PR #1, opened 2016-02-16
Fix vulkaninfo on win32 to use existing console

hlandau:
The vulkaninfo utility is compiled as /SUBSYSTEM:APPLICATION but
attempts to allocate its own console and print to it. Not only
does this appear to be unreliable (sometimes no output is printed),
it is unnecessary if the utility is compiled as /SUBSYSTEM:CONSOLE.

This removes unnecessary console manipulation code. By using
/SUBSYSTEM:CONSOLE, the ordinary main() entrypoint function can be
used.

The console is still enlarged if it was opened exclusively for the
process by detecting whether it was so opened. Likewise, the process
still sleeps forever if the console is exclusive to the process.
The process behaves like a normal command line process when executed
from the command line.

IMPORTANT: I don't know how you generate the .vcxproj files included
in the SDK, as I can't find them here, but you need to change them
from /SUBSYSTEM:APPLICATION to /SUBSYSTEM:CONSOLE in line with this
change.

Fixes lunarg issue 350.

### DIFF
--- a/demos/vulkaninfo.c
+++ b/demos/vulkaninfo.c
@@ -48,11 +48,17 @@
 
 #define snprintf _snprintf
 
-bool consoleCreated = false;
+// Returns nonzero if the console is used only for this process. Will return
+// zero if another process (such as cmd.exe) is also attached.
+static int ConsoleIsExclusive(void) {
+    DWORD pids[2];
+    DWORD num_pids = GetConsoleProcessList(pids, ARRAYSIZE(pids));
+    return num_pids <= 1;
+}
 
 #define WAIT_FOR_CONSOLE_DESTROY                                               \
     do {                                                                       \
-        if (consoleCreated)                                                    \
+        if (ConsoleIsExclusive())                                                    \
             Sleep(INFINITE);                                                   \
     } while (0)
 #else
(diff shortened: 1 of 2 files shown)


demos/vulkaninfo.c
```diff
@@ -1133,7 +1137,7 @@ static void app_gpu_dump(const struct app_gpu *gpu) {
     app_dev_dump(&gpu->dev);
 }
 
-int main(int argc, char **argv) {
+int main_actual(int argc, char **argv) {
```
dustin-lunarg: Renaming the original main to main_actual will lead to an undefined reference to main error on Linux.  Go ahead and merge your main and main_actual into a single main, using the appropriate _WIN32 #ifdef directives to only enable the Windows specific console code for the Windows build.

dustin-lunarg:
Thanks Hugo!  This is a nice improvement.  The change looks good overall.  There is just one issue with the Linux build to deal with.  I have posted a comment in the diff regarding the issue, if you want to fix it yourself.  Otherwise, I can check out your change locally and add the necessary adjustments.

The Visual Studio project files are generated by CMake.  In the demos/CMakeLists.txt file you should find an 'add_executable' macro for vulkaninfo, which contains a WIN32 argument.  The WIN32 is what triggers CMake to generate the project with /SUBSYSTEM:WINDOWS.  Removing it should generate a project file that uses /SUBSYTEM:CONSOLE.

hlandau:
Whoops, how did I miss that. Fixed, and cmake updated.

demos/vulkaninfo.c
```diff
@@ -1133,6 +1139,30 @@ static void app_gpu_dump(const struct app_gpu *gpu) {
     app_dev_dump(&gpu->dev);
 }
 
+// Enlarges the console window to have a large scrollback size.
+static void ConsoleEnlarge() {
```
dustin-lunarg: An #ifdef _WIN32 is needed for this function.  After that is added, this change should be good to merge.

hlandau:
Fixed. Sorry for the repeated broken code on Linux, I don't have this codebase prepared for building on Linux.

dustin-lunarg:
This fix has been merged to the master branch.

PR #2, opened 2016-02-17
[WIP] Replace private repo references

corngood:
Here's what I needed to do for `update_external_sources.sh` to work on linux without private repo access.

WIP because I didn't bother with the .cmd yet, or look into the implications of removing the svn stuff.  Still I got a working build out of it, so here it is for reference.


karl-lunarg:
Hi, thanks for the fix and the pull request!   We missed this during the roll-out, but it would have been hard to pull off because of things switching to public all at the same time.   There's another pull request to address the same issue, so I will be pulling that one in, test it, and commit it back to this branch and to master.  Thanks!

PR #4, opened 2016-02-18
There were some old private svn references left

clintar:


karl-lunarg:
We're in the process of updating the public LunarGLASS repo with the correct code.  The portions of the script that use svn to fetch and build LunarGLASS are going to be removed from the script in the LoaderAndValidationLayers repo and moved to a new script in the VulkanTools repo, which is really the repo that needs LunarGLASS.  This new script will fetch and build from the updated public LunarGLASS repo and so will not need a password.

See https://github.com/LunarG/VulkanTools/issues/4 for more info.
Thank you for providing the pull request.  We won't pull your code due to the changes mentioned above.  I'll also close this request now.  Please track the issue at https://github.com/LunarG/VulkanTools/issues/4.

PR #10, opened 2016-02-20
Fix wayland support

jekstrand:
This little series gets Wayland support working with the loader.  The first commit makes all of the different WSI platforms configurable.  We should, perhaps, consider autodetection on Linux via pkg-config, but that can come later.


jon-lunarg:
Thanks for fixes. cherry picked to master branch,

PR #13, opened 2016-02-21
Check boolean values in vkCreateSampler

rdb:
This change adds checks for the boolean values in vkCreateSampler.  The anisotropyEnable field is erroneously missing from the 1.0.3 spec, which makes it easy to overlook setting it.  This causes a crash in some drivers.

I wasn't sure against which branch to make this PR.  Let me know if I should re-make it against the sdk-1.0.3 branch instead of master.


dustin-lunarg:
Change was rebased with master locally and pushed directly to Vulkan-LoaderAndValidationLayers.

PR #14, opened 2016-02-21
Remove unnecessary null pointer checks

### LINKED ISSUE #6: Remove unnecessary null pointer checks
[An extra null pointer check is not needed in functions](http://dietmar-kuehl.de/mirror/c++-faq/freestore-mgmt.html#faq-16.8 "Do I need to check for NULL before delete p?") like the following.
* [VkRenderFramework::ShutdownFramework](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/1affe90f0ec7f9bccb6841a56a2a5b66861efe6a/tests/vkrenderframework.cpp#L163 "VkRenderFramework::ShutdownFramework function")
* [_DESCRIPTOR_POOL_NODE](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/1affe90f0ec7f9bccb6841a56a2a5b66861efe6a/layers/draw_state.h#L448 "Destructor for _DESCRIPTOR_POOL_NODE")

1ace:
Fixes #6


layers/draw_state.cpp
```diff
@@ -6929,9 +6916,7 @@ static void deleteRenderPasses(layer_data* my_data)
             }
```
1ace: (Lines 6907­–6914) I didn't touch them because I wasn't sure, but I think those `if` can also be removed.
It will obviously not result in a double-free, but it might free things that are not meant to be freed yet?
I'm about 80% confident there :P

tobine:
Looks good to me. Verified that it builds and cube --validate is still running. Pushed to upstream master with slight modification to commit message.
Thanks for the clean-up!

PR #15, opened 2016-02-21
Use `spirv/spirv.hpp` as canonical path

1ace:
The (maybe better) alternative would be to add `-I/usr/include/spirv/` in the build system, but I don't know enough about cmake to know where/how to do that :]


tobine:
The CMakeLists.txt file in the layers dir currently contains the following block


    include_directories(
	    ${CMAKE_CURRENT_SOURCE_DIR}
	    ${CMAKE_CURRENT_SOURCE_DIR}/../loader
	    ${CMAKE_CURRENT_SOURCE_DIR}/../include/vulkan
	    ${CMAKE_CURRENT_BINARY_DIR}
	    ${PROJECT_SOURCE_DIR}/../glslang/SPIRV
    )

If the user has run "update_external_sources.[bat|sh]" from the root of the repo, the last line above will cause the correct spirv.hpp to be picked up from the glslang dir. Have you run the update_external_sources script for your platform, and is the glslang  spirv.hpp then correctly included in your build?

The change as you've requested breaks the Linux build. If the last line above is changed to "${PROJECT_SOURCE_DIR}/../glslang" and the #include line is changed to:

    #include <SPIRV/spirv.hpp>

the linux build still works. I'm not sure that's the right change, though, so let me know what you think of my suggestion above.

jon-lunarg:
Thanks, pushed fix (plus appropriate CMake file change) to master branch.

PR #16, opened 2016-02-21
Remove extra `layers/` from path

### LINKED ISSUE #9: Link to vk_layer_settings.txt broken
A link to the _vk_layer_settings.txt_ is broken in this [README](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/sdk-1.0.3/layers/README.md).
It seems to include an unnecessary extra "layers/"

1ace:
Fixes #9


karl-lunarg:
Thanks for the fix!  Applied manually to master branch.

1ace:
Thanks :)

I have a few other patches, but I was wondering: what's your preferred way of receiving patches?
- Mails (`git send-email`)?
- The Khronos GitLab instance *(I noticed there's more discussion activity over there, although there hasn't been any merges since the release)*?
- PRs here?
- Something else?

I guess you could use a [CONTRIBUTING.md](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) :)

Oh, and why is `sdk-1.0.3` a branch instead of a tag on `master`?
And do you want all changes applied to both branches (when it makes sense)?

karl-lunarg:
It is best to submit PRs here (GitHub).   Although GitLab had a lot of traffic, we expect / want most of it over here on GitHub.  NDA stuff should go to GitLab of course.

It would also be easiest to submit the changes from a branch based on master.

The intent of the sdk-1.0.3 branch was to support the release of the 1.0.3.1 SDK and any other 1.0.3.x SDKs that we'd have to release while still on the Vulkan 1.0.3 header.  Yes, we could have made it a tag and then created the branch off that tag only when we needed it.  But folks executing our process are used to branches and I think that any further discussion on tag vs branch would probably generate a long discussion.

Thanks for the suggestion for a CONTRIBUTING.md.  We'll look into it.

PR #21, opened 2016-02-23
Fix MSVS warnings in loader code.

null77:
This fixes a few occurrences of the following warnings:

 - Unreachable code
 - Cast from unsigned int to char
 - Variable shadowing
 - Possibly uninitialized value used.


karl-lunarg:
Pushed to master.  Thanks!

PR #23, opened 2016-02-23
layers: Allow draw_state layout mismatch with UNDEFINED

stroyan:
Also allow CB initialLayout of VK_IMAGE_LAYOUT_UNDEFINED to match
any layout.


tobine:
Pushed commit to master

PR #24, opened 2016-02-23
layers: draw_state VK_WHOLE_SIZE in barriers

stroyan:
Allow use of VK_WHOLE_SIZE in barriers validation.


tobine:
pushed commit to master

PR #26, opened 2016-02-23
Fixed markStoreImagesAndBuffersAsWritten

szdarkhack:
It now correctly treats images as images instead of buffers.


mark-lunarg:
Fix was pushed to Master, thanks for the PR.  Assigned to Tobin in case he wanted to include this fix as well.

PR #28, opened 2016-02-24
layers: Make layer option names more consistent and update docs

mark-lunarg:
There was an inconsistent mix of old- and new-style names, and the documentation for the layer settings did not match the layers themselves.  First shot at adding come consistency.  (There is a corresponding sibling change for VulkanTools).


tobine:
Merged this to master

PR #29, opened 2016-02-24
Add spirv.hpp include file to fix the build.

lentinem:


courtney-g:
We've been getting this from the glslang repository as a peer of this one, setup via update_external_sources.sh. I didn't think we were changing that relationship.

lentinem:
This seems kind of overkill for a single header file so that the layers can build (I've ignored external sources for layer builds for quite a while). Wouldn't this also create issues such that if that header changed and the layers were not updated yet it would break? I'm open to being overruled here but I would at least like to understand why.

courtney-g:
The build process for the LoaderAndValidationLayers is to run update-external-sources any time you pull in new code from master. update-external-sources will make sure the peer repositories have their heads set at the commit they need to build properly. There are multiple users of this SPIRV header. I think we'd prefer that all the users will get it from the same place.

jon-lunarg:
I think this can be closed with no action.  Continue doing what Courtney says using update_external_sources.sh to get spirv header.

Michael are you okay with this?

PR #34, opened 2016-02-24
Fixes for building on Windows

cnorthrop:
These are tested on Android, built from Windows and Linux

### DIFF
--- a/buildAndroid/jni/Android.mk
+++ b/buildAndroid/jni/Android.mk
@@ -143,19 +143,4 @@ LOCAL_CPPFLAGS += -DVK_USE_PLATFORM_ANDROID_KHR
 LOCAL_LDLIBS    := -llog
 include $(BUILD_SHARED_LIBRARY)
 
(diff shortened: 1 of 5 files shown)


ianelliottus:
I cherry-picked and pushed the changes to the target branch.

buildAndroid/jni/Android.mk
```diff
@@ -143,19 +143,4 @@ LOCAL_CPPFLAGS += -DVK_USE_PLATFORM_ANDROID_KHR
 LOCAL_LDLIBS    := -llog
 include $(BUILD_SHARED_LIBRARY)
 
-include $(CLEAR_VARS)
```
lentinem: Can we comment this out instead of removing it. Eventually it should be enabled.

cnorthrop: Planning to turn on the validation tests very soon, so we won't lose this recipe.  I've got it working on a branch that I'll send to you.

PR #35, opened 2016-02-25
layers: Add component awareness to SC interface matching

chrisforbes:
- Adds more tests for mismatch by component and by location
- Removes dead handling of builtins (we only care about collecting builtin /decorations/ to aid in the interface block handling;
    we don't really care about producing the set of builtins used overall.)

Some SC error messages change slightly, as interstage locations now include the component index.

Fixes internal bug 26609754

V2: Remove dead code

(This patch is already on the android-* branch, but the issue it addresses is present on master too)


tobine:
merged to master

PR #37, opened 2016-02-25
layers: LX322, Track semaphore state across QueueBindSparse calls

mark-lunarg:
Hey Tobin, can you take a look and see if this is reasonable?  I have not found any way to test this API, but if it is used it causes invalid MemTracker errors.  This is mostly straight from QueueSubmit.


mark-lunarg:
Pushed to Master.

PR #53, opened 2016-02-29
WIP: Count attachments as the same if they have overlapping memory.

lentinem:


lentinem:
I can't assign but it would be helpful if someone could do a cr on this. It is untested atm which is why it's marked WIP.

chrisforbes:
This looks like it does the right thing, but it would be nice to have some tests.

I think there's big improvements that could be made in the subpass/attachment tracking code in general, but that's way out of scope for this PR.

layers/draw_state.h
```diff
@@ -325,9 +333,17 @@ class BUFFER_NODE : public BASE_NODE {
     unique_ptr<VkBufferCreateInfo> create_info;
 };
 
+// Store the DAG.
+struct DAGNode {
```
tobine: Would prefer DAG_NODE as name just to match naming of all the other "*_NODE" types

tobine:
Nothing obviously incorrect to me. Can you rebase on latest and repush and I'll test it out and walk through the code a bit?

layers/draw_state.h
lentinem: This isn't new and was just moved from the cpp to the h. We definitely have some naming inconsistencies throughout the code. I'm ok with changing this but it might be worth actually having a set of rules for struct, function, var naming.

lentinem:
Rebased.

tobine:
Ran this through tri/cube/smoke --validate and it's clean there. Pushed to master.

PR #57, opened 2016-03-01
WIP: Add checking for may alias bit.

lentinem:
Not sure how to make this only diff against the prior pull request.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,4 +2,5 @@
 # We'll use defaults from the LLVM style, but with 4 columns indentation.
 BasedOnStyle: LLVM
 IndentWidth: 4
+LineLength: 132
 ...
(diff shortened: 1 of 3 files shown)


lentinem:
Rebased: Note there are some formatting changes that somehow didn't get merged from the change this depended on. I would be in favor of doing them anyway.

tobine:
I think that format was updated due to modified clang-format for layers with larger line length. I'd like to just skip the format CL and only pull the alias check changes from second CL.

tobine:
On second thought, alias commit doesn't cherry pick cleanly without the format changes so I'll just take them both. Having the shortly line length isn't an issue.

lentinem:
I just pushed the new line length

tobine:
Thanks. Validated and pushed to master.

PR #59, opened 2016-03-02
Miscellaneous fixes for DrawState layer

chrisforbes:


layers/draw_state.cpp
```diff
@@ -3619,8 +3618,7 @@ VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkCreateDevice(VkPhysicalDevice g
         my_device_data->physDevProperties.features =
             *pCreateInfo->pEnabledFeatures;
     } else {
-        my_instance_data->instance_dispatch_table->GetPhysicalDeviceFeatures(
-            gpu, &my_device_data->physDevProperties.features);
+        memset(&my_device_data->physDevProperties.features, 0, sizeof(VkPhysicalDeviceFeatures));
```
lentinem: Does features never get used?

chrisforbes: It's used by a bunch of checks, and will be used by more shortly. The issue
here is that setting pEnabledFeatures=NULL was being treated as "enable all
supported features", whereas the spec says it means "disable all features".

(There's an open spec issue for whether it should be allowed at all, since
the valid usage block contradicts other guidance in the Features chapter--
but this just gets the layer in line with the existing spec)

lentinem: Bleh. That does seem to be true; I would have hoped NULL didn't mean disable but it seems to be what it is.

lentinem:
+1

PR #61, opened 2016-03-02
Allow use of queue family ignored.

lentinem:


mark-lunarg:
Thanks Michael, found this yesterday on the smoke test.  Tested on Windows and pushed to Master.

PR #65, opened 2016-03-02
Fix extension support in unique objects

lentinem:


mark-lunarg:
Hey Michael --  Running this on Linux and Windows required that I add DestroySurfaceKHR to createInstanceRegisterExtensions() in the generated layer's header files.   I'm doing this now for VulkanTools and will push both when that's working.

PR #74, opened 2016-03-03
layers: Validate pipeline derivative requirements in DrawState

chrisforbes:
While looking at what if anything SC needed to do for derivatives (which is nothing!) I figured I'd get the DS base pipeline checks out of the way.


courtney-g:
LGTM

tobine:

Issue #75, opened 2016-03-03
Subpass index not being validated during pipeline creation


jhprince:
When calling vkCreateGraphicsPipelines, the subpass index represented in VkGraphicsPipelineCreateInfo is not validated against the amount of subpasses created with the vkCreateRenderPass function call.

chrisforbes:
Fixed in master.

PR #76, opened 2016-03-03
layers: Validate subpass index in CreateGraphicsPipelines

chrisforbes:
Not validating this upfront meant that users got really weird errors
about mismatches between their FS outputs and color attachments. Now
they should hit the real problem first.

Fixes GH #75

### DIFF
--- a/layers/draw_state.cpp
+++ b/layers/draw_state.cpp
@@ -1999,6 +1999,23 @@ static VkBool32 verifyPipelineCreateState(layer_data* my_data, const VkDevice de
         }
     }
 
+    // Ensure the subpass index is valid. If not, then validate_pipeline_shaders
+    // produces nonsense errors that confuse users. Other layers should already
+    // emit errors for renderpass being invalid.
+    auto rp_data =
+        my_data->renderPassMap.find(pPipeline->graphicsPipelineCI.renderPass);
+    if (rp_data != my_data->renderPassMap.end() &&
+        pPipeline->graphicsPipelineCI.subpass >=
+            rp_data->second->pCreateInfo->subpassCount) {
+      skipCall |= log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                          (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
+                          DRAWSTATE_INVALID_PIPELINE_CREATE_STATE, "DS",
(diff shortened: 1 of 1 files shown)


layers/draw_state.cpp
```diff
@@ -1999,6 +1999,16 @@ static VkBool32 verifyPipelineCreateState(layer_data* my_data, const VkDevice de
         }
     }
 
+    // Ensure the subpass index is valid. If not, then validate_pipeline_shaders
```
lentinem: please run git clang-format on this

chrisforbes: Fixed.

layers/draw_state.cpp
```diff
@@ -1999,6 +1999,23 @@ static VkBool32 verifyPipelineCreateState(layer_data* my_data, const VkDevice de
...
+      skipCall |= log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                          (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
+                          DRAWSTATE_INVALID_PIPELINE_CREATE_STATE, "DS",
+                          "Invalid Pipeline CreateInfo State: Subpass index %u "
+                          "is out of range for this renderpass (0..%u)",
+                          pPipeline->graphicsPipelineCI.subpass,
+                          rp_data->second->pCreateInfo->subpassCount - 1);
+    }
+
     if (!validate_pipeline_shaders(my_data, device, pPipeline)) {
```
tobine: Did you want to move this to an "else" block or at least prevent the validate_pipeline_shaders() call if skipCall is true? Otherwise we'll still end up running validate_pipeline_shaders even when the subpass is invalid, hence still getting the weird errors as well. If that's intended and the only goal here is for the pre-emptive error before the weird errors, then this change LGTM.

lentinem: For the most part we do this everywhere. The fundamental assumption is that once an error happens; it could trigger other incorrect errors anywhere (say a call to draw with an invalid renderpass which will lead to all sorts of things). IMO the overhead of making it so that once something fails nothing else checks anything related to that failure is quite large. That being said I'm fine with doing something like that locally here but I don't see much of a benefit as lots of errors would be thrown elsewhere in this case.

chrisforbes: Yeah, I just wanted a good error to be first.

lentinem:
+1

tobine:

PR #77, opened 2016-03-04
Fix assert for layouts.

lentinem:
Currently we assume that the layout must have been set for a subresource.
While this is true it may be set globally instead of on that specific
subresource directly causing the old assert to throw.


tobine:

PR #78, opened 2016-03-04
Enforce that unique objects if enabled happens before draw_state

lentinem:


layers/draw_state.cpp
```diff
@@ -135,6 +142,29 @@ struct layer_data {
...
+};
+
+template<class TCreateInfo>
+void ValidateLayerOrdering(const TCreateInfo& createInfo) {
+    bool foundLayer = false;
+    for (uint32_t i = 0; i < createInfo.enabledLayerCount; ++i) {
+        if (!strcmp(createInfo.ppEnabledLayerNames[i], ds_global_layers[0].layerName)) {
+            foundLayer = true;
+        }
+        // This has to be logged to console as we don't have a callback at this point.
```
courtney-g: Not necessarily. The VK_EXT_debug_report extension defines a VkDebugReportCallbackCreateInfoEXT structure that can be linked into the pNext parameter of the CreateInstance structure and provide a callback for use during CreateInstance.

layers/draw_state.cpp
```diff
@@ -3600,6 +3629,8 @@ VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkCreateInstance(const VkInstance
 
     init_draw_state(my_data, pAllocator);
 
+    ValidateLayerOrdering(*pCreateInfo);
```
courtney-g: If this discovers an issue, should we just fail the CreateInstance?

lentinem: I would prefer not to. IMO the application should always have the option of completely ignoring the layers even when they are enabled.

courtney-g: Hmm. We'll generate an error if the app asks for a layer that doesn't exist. Seems that if the layer order is invalid that should generate an error. App can do another CreateInstance without layers if it wants. If this is only generated via layers enabled via the property mechanism then I agree, app shouldn't be bothered as it was not the one that specified the layers.

lentinem: This particular error is required by the spec though. If we modify the spec to have ordering requirements then I would agree but as is they can have completely valid spec code and then still fail this function. This is also why this is a warning instead of an error as it's technically not wrong; it can just expose issues with validation.

mark-lunarg:
I agree with making it a Warning, the user should be able to do whatever they want layer-wise.  Seems like the same check should go in object_tracker as well.

jon-lunarg:
I agree that it should just be a warning reported via debug_report  and the CreateInstance shouldn't fail if the order is wrong.

lentinem:
This check should go in almost every layer (haven't done a complete analysis of which ones) but I think doing 1 first makes sense so that we can agree on how this is done.

jon-lunarg:
Should this check happen in the loader instead of draw_state  object_tracker, etc.  Seems best to add the check to the loader only once instead of having it duplicated in several layers.

tobine:
I've been sitting on this PR as there doesn't seem to be a consensus on best way to handle layer ordering enforcement. Let's close on this in our meeting today.

KarenGhavam:
Tobin. We discussed this in our sync meeting on wednesday and are fine with this merge request. I don't think you want me to do the merge..... Giving it back to you.

tobine:
merged to master

PR #80, opened 2016-03-04
Remove unused var.

lentinem:


chrisforbes:
:+1: This will need to come back eventually, when accurate member sizes are added, but is dead currently.

PR #81, opened 2016-03-04
Register event on creation and delete on deletion.

lentinem:
Needs to be rebased after unused var is removed.


tobine:

PR #86, opened 2016-03-07
layers: GH41 Flag error if cmd buffer references destroyed framebuffer

tobine:
Create framebuffer node and include set of cmd buffers that reference that FB.
At the time when RenderPass is bound to cmd buffer, or when secondary cmd buffer
is created with framebuffer inheritence, store a reference from FB to cmd buffer.

When FB destroyed, set any cmd buffers referencing it as INVALID and store FB ref.

At submit time, if CB is INVALID, check for any destroyed FBs and report them.

### DIFF
--- a/layers/draw_state.cpp
+++ b/layers/draw_state.cpp
@@ -115,7 +115,7 @@ struct layer_data {
     unordered_map<VkQueryPool, QUERY_POOL_NODE> queryPoolMap;
     unordered_map<VkSemaphore, SEMAPHORE_NODE> semaphoreMap;
     unordered_map<void*,                 GLOBAL_CB_NODE*>                    commandBufferMap;
-    unordered_map<VkFramebuffer,         VkFramebufferCreateInfo*>           frameBufferMap;
+    unordered_map<VkFramebuffer, FRAMEBUFFER_NODE> frameBufferMap;
     unordered_map<VkImage, vector<ImageSubresourcePair>> imageSubresourceMap;
     unordered_map<ImageSubresourcePair, IMAGE_NODE> imageLayoutMap;
     unordered_map<VkRenderPass,          RENDER_PASS_NODE*>                  renderPassMap;
@@ -3391,6 +3391,7 @@ static void resetCB(layer_data* my_data, const VkCommandBuffer cb)
         pCB->uniqueBoundSets.clear();
         pCB->destroyedSets.clear();
         pCB->updatedSets.clear();
+        pCB->destroyedFramebuffers.clear();
         pCB->boundDescriptorSets.clear();
         pCB->waitedEvents.clear();
         pCB->semaphores.clear();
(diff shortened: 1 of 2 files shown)


layers/draw_state.cpp
```diff
@@ -4104,21 +4105,48 @@ static bool validateCommandBufferState(layer_data *dev_data,
...
+                          fb_string.str().c_str());
+              causeReported = true;
+            } else if (!causeReported) {
+              skipCall |= log_msg(
+                  dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                  VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
+                  reinterpret_cast<uint64_t&>(pCB->commandBuffer), __LINE__,
+                  DRAWSTATE_INVALID_COMMAND_BUFFER, "DS",
+                  "You are submitting command buffer %#" PRIxLEAST64
+                  " that is invalid due to an unknown cause. Validation should "
```
lentinem: Why is this unknown? seems like it would be easy enough to make casueReported a enum that we can switch over and combine this with above.

layers/draw_state.cpp
```diff
@@ -7040,8 +7105,12 @@ VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkCreateFramebuffer(VkDevice devi
             localFBCI->pAttachments = new VkImageView[localFBCI->attachmentCount];
             memcpy((void*)localFBCI->pAttachments, pCreateInfo->pAttachments, localFBCI->attachmentCount*sizeof(VkImageView));
         }
+        FRAMEBUFFER_NODE fbNode = {};
+        fbNode.createInfo = *localFBCI;
+        std::pair<VkFramebuffer, FRAMEBUFFER_NODE> fbPair(*pFramebuffer,
```
lentinem: I would inline this with std::make_pair but am fine leaving it if you prefer this way.

layers/draw_state.cpp
```diff
@@ -7399,34 +7468,47 @@ VkBool32 VerifyFramebufferAndRenderPassLayouts(VkCommandBuffer cmdBuffer, const
...
-                if (!FindLayout(pCB, image, sub, node)) {
-                    SetLayout(pCB, image, sub, newNode);
-                    continue;
-                }
-                if (newNode.layout != node.layout) {
-                    skip_call |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__, DRAWSTATE_INVALID_RENDERPASS, "DS",
-                                 "You cannot start a render pass using attachment %i where the intial layout differs from the starting layout.", i);
-                }
-            }
+      const VkImageView& image_view = framebufferInfo.pAttachments[i];
```
lentinem: What was changed in this block?

layers/draw_state.cpp
```diff
@@ -7485,11 +7570,12 @@ void TransitionFinalSubpassLayouts(VkCommandBuffer cmdBuffer, const VkRenderPass
...
         return;
     }
-    const VkFramebufferCreateInfo* pFramebufferInfo = framebuffer_data->second;
+    const VkFramebufferCreateInfo framebufferInfo =
+        framebuffer_data->second.createInfo;
     for (uint32_t i = 0; i < pRenderPassInfo->attachmentCount; ++i) {
-        const VkImageView& image_view = pFramebufferInfo->pAttachments[i];
-        SetLayout(dev_data, pCB, image_view,
-                  pRenderPassInfo->pAttachments[i].finalLayout);
+      const VkImageView& image_view = framebufferInfo.pAttachments[i];
```
lentinem: This looks wrong to me. I thought the indentation for clang-format was 4. Are you running with the correct file?

layers/draw_state.h
```diff
@@ -669,10 +675,12 @@ typedef struct _GLOBAL_CB_NODE {
     VkFramebuffer                framebuffer;
     // Capture unique std::set of descriptorSets that are bound to this CB.
     std::set<VkDescriptorSet>    uniqueBoundSets;
-    // Keep running track of which sets are bound to which set# at any given time
     // Track descriptor sets that are destroyed or updated while bound to CB
     std::set<VkDescriptorSet>    destroyedSets;
     std::set<VkDescriptorSet>    updatedSets;
+    unordered_set<VkFramebuffer> destroyedFramebuffers;
```
lentinem: Does it make sense to combine this with destroyedSets and just have destroyedObjects or does knowing what type help?

layers/draw_state.cpp
tobine: Currently there should be no cases where causeReported is false. I stewed over this code quite a bit and added this final block as a defensive mechanism in the event that we were missing any case where an INVALID cmd buffer was submitted, or if someone in the future causes such a case, I didn't want this code to be cleared without *some* error being flagged so I put this in both to make sure an error is flagged, but also to signal that validation should be updated to explicitly handle this case.

Using an enum here isn't ideal as there may be multiple cases that cause INVALID cmd buffer to be submitted so the separate "if" blocks are meant to report each of these cases.

I'm inclined to stick with the defensive code for now, although I don't like it a ton, and put in a TODO here to improve the INVALID cmd buffer reporting in general. We still have various holes that need to be filled here so it probably calls for a rework in the near future.

layers/draw_state.cpp
tobine: Interesting. I didn't change anything here and code looks the same so not sure why it was reformatted.

tobine: Ah, now I see it. I changed "pFramebufferInfo->" to "framebufferInfo." git clang-format just reformatted a larger section than may have been expected.

layers/draw_state.cpp
tobine: I just ran "git clang-format" which I thought I had set up correctly but it's possible my config is not correct. Mark just said that he's planning to run all layers through clang-format today so if I get this in first, his change will override any errors and I can confirm separately that I have correct format.

layers/draw_state.h
tobine: It makes sense to simplify this in some fashion, as there are still more cases of different resource types that can be deleted and cause INVALID cmd buffers, so I don't think we want separate sets for each. There's also a hole in this current fix where it doesn't handle FBs deleted out from under an in-flight cmd buffer.

I'd like to just consider this a spot-fix for a specific bug that was reported, and go ahead with the core code and perhaps a few added "TODOs". Then I'll file a separate bug to implement a more complete solution for handling deletion any resource that a cmd_buffer depends on. We do handle many of these cases already w/ the BASE_NODE and in_use tracking, but we need to build that out to handle missing cases. I'd like to get the mem_tracker merge done first, and then can address this issue fully.

layers/draw_state.cpp
lentinem: I would probably at least remove the else in this case (to be more consistent so that we have a bunch of if (...) {output} if (!output) ...) and add a comment to the affect that this should never happen. Did you see this happen in practice? I'm a bit of a fan of putting an assert here but realize that that might cause issues for users - depends how confident you are that it will never happen.

layers/draw_state.cpp
lentinem: Try git clang-format --style=file

layers/draw_state.h
lentinem: I view the deleted out from under as part of fb lifetime which is on the list of valid usage tasks. The lifetime checks all have a common implementation (in_use) which I like and I would want something similar for the invalid cmd buffer submission checks like these so we can just add new objects into the checks more easily. I think this framework can be used but doing something like merging the sets will probably make it easier.

tobine:
I changed "else if" to "if", added a few "TODO" comments, correctly ran git clang-format and went ahead and pushed this to make SDK freeze. I'll open separate GH issue on myself to unify INVALID cmd_buffer handling.

PR #89, opened 2016-03-08
Use size_t types for output size_t.

lentinem:


tobine:
cherry-picked this fix and merged to master

PR #107, opened 2016-03-10
layers: GH96 Fix error messages for bad layout

tobine:
Instead of printing attachements as #s in the error messages, print the layout string.


lentinem:
+1

tobine:
merged to master

PR #108, opened 2016-03-10
layers: GH98 Only check fence in-use once per QueueSubmit

tobine:
In vkQueueSubmit, perform the fence in-use check first.
Then validate each submit, which includes marking the current fence in-use.


lentinem:
+1

tobine:
merged to master

Issue #67, opened 2016-03-03
Memory Layer complaining about invalid memory being read by vkCmdCopyImageToBuffer(), even though it is correctly filled earlier.


KarenGhavam:
This issue came from LunarXchange (issue number 303).

Consider the following command buffers:

Command Buffer 1: Blit image A to image B, possibly performing format conversion along the way. Command Buffer 2: Copy whole image B's contents to buffer X with a vkCmdCopyImageToBuffer() call.

Now, let's build a single submission consisting of the two command buffers. Let Command Buffer 1 signal a semaphore, that Command Buffer 2 waits upon before starting execution.

Once the command buffers finish executing, we try to map the buffer X's memory buffer into process space and read from it.

With this chain of action, I'm getting a call-back from the Memory layer, telling that "VkCmdCopyImageToBuffer(): Cannot read invalid memory <image B's memory buffer>, please fill the memory before using.", which I believe is incorrect.

lentinem:
What order are the command buffers being executed in? The spec requires 1 is submitted before 2 and this could trigger if thats not the case (section 5.5).

lentinem:
Jon; I'm not sure how to find out more from the original submitter. Can you ask them for more info on the ordering and if possible a repro?

KarenGhavam:
The original submitter was Dominik Witczak. I have sent him email to come to this issue and answer your questions.

KarenGhavam:
Michael,

Dominick Witczak was original submitter. I will ask him to begin watching
the Github issue and to answer your question.

Karen Ghavam
CEO
LunarG, Inc. - 3D Graphics Software Innovations
karen@lunarg.com
970-988-9043

kbiElude:
I'll need to re-check this, this issue has been reported months ago so I can't tell whether the problem still reproduces.

jon-lunarg:
Dominik:

Maybe I don't understand your case,  But  it seems if both command buffers  are in the same QueueSubmit, then  a semaphore can't be used to synchronize between command buffer 1 and command buffer 2 access to image B.  How about using a memory barrier for synchronize access to image B?

lentinem:
Assigning to Karen as we need a repro for this but I'm not sure Dominik's user id and Karen created it.

kbiElude:
[...]

PR #112, opened 2016-03-11
#67: Add missing interception for vkCmdBlitImage() calls.

DominikWitczakAMD:
Mem Tracker did not intercept the vkCmdBlitImage() call, which caused the problem described in #67 due to a domino effect.


tobine:
+1

mark-lunarg:
Pushed to Master.

PR #131, opened 2016-03-14
layers: draw state image/layout cleanup

dustin-lunarg:
Remove tracked device memory, subresource, and layout data associated
with an image when that image is destroyed.  Eliminates spurious
draw state messages that report invalid layout issues when the
destroyed image's handle is recylced and the new image state is
validated with the old image's data.


lentinem:
+1

tobine:
cherry-picked and pushed to master. Thanks!

Issue #132, opened 2016-03-15
Draw State: "Cannot delete event X which is in use by a command buffer" error prints the pointer as a uint64


DominikWitczakAMD:
This isn't too helpful. The event handle should be printed out as a pointer instead. MR coming.

PR #133, opened 2016-03-15
#132: Event handle should be printed as a hexadecimal value

DominikWitczakAMD:
MR for #132


tobine:
cherry-picked and pushed to master. Thanks!

PR #135, opened 2016-03-15
layers: GH130 Handle index mapping for sparse bindings

tobine:
If a binding has descriptorCount == 0, account for this when setting up bindingToIndexMapping.
This initial solution will keep a mapping for every binding, but we still lose the Type and
other state associated with empty bindings. Included a "TODO" to implement more complete fix.

### DIFF
--- a/layers/draw_state.cpp
+++ b/layers/draw_state.cpp
@@ -231,6 +231,7 @@ static unordered_map<void *, layer_data *> layer_data_map;
 static int globalLockInitialized = 0;
 static loader_platform_thread_mutex globalLock;
 #define MAX_TID 513
+#define MAX_DESCRIPTOR_INDEX 0xFFFFFFFF
 static loader_platform_thread_id g_tidMapping[MAX_TID] = {0};
 static uint32_t g_maxTID = 0;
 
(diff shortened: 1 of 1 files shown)


tobine:
This is bad fix, closing

layers/draw_state.cpp
```diff
@@ -5130,8 +5132,14 @@ vkCreateDescriptorSetLayout(VkDevice device, const VkDescriptorSetLayoutCreateIn
                             DRAWSTATE_INVALID_LAYOUT, "DS", "duplicated binding number in "
                                                             "VkDescriptorSetLayoutBinding"))
                     return VK_ERROR_VALIDATION_FAILED_EXT;
+            } else if (pCreateInfo->bindingCount == 0) {
+                // TODO : Handling empty bindings as a special case and adjusting the index mapping
+                //  below to skip empty bindings. This is not ideal, and would be better to track
+                //  the state of the empty binding & have mapping point to index for that state.
+                pNewNode->bindingToIndexMap[pCreateInfo->pBindings[i].binding] = MAX_DESCRIPTOR_INDEX;
```
lentinem: I agree with the comment here but it's ok for now. Probably worth submitting a bug for this as P2 enhancement so it doesn't get lost.

PR #137, opened 2016-03-15
layers: GH130 Fix to correctly get type/flags from set binding

tobine:
We were using bindingToIndexMap to do look-ups into type and stage vectors which is incorrect.
bindingToIndexMap is to index into the original pBindings map that was passed in when the
layout was created. This fix does this to correctly pull type and stage of a binding in draw_state.

### DIFF
--- a/layers/draw_state.cpp
+++ b/layers/draw_state.cpp
@@ -1341,11 +1341,13 @@ static bool has_descriptor_binding(layer_data *my_data, vector<VkDescriptorSetLa
     auto const layout_node = my_data->descriptorSetLayoutMap[(*pipelineLayout)[slot.first]];
 
     auto bindingIt = layout_node->bindingToIndexMap.find(slot.second);
-    if (bindingIt == layout_node->bindingToIndexMap.end())
+    if ((bindingIt == layout_node->bindingToIndexMap.end()) || (layout_node->createInfo.pBindings == NULL))
         return false;
 
-    type = layout_node->descriptorTypes[bindingIt->second];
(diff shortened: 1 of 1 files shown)


layers/draw_state.cpp
```diff
@@ -1344,8 +1344,9 @@ static bool has_descriptor_binding(layer_data *my_data, vector<VkDescriptorSetLa
     if (bindingIt == layout_node->bindingToIndexMap.end())
         return false;
 
-    type = layout_node->descriptorTypes[bindingIt->second];
-    stage_flags = layout_node->stageFlags[bindingIt->second];
+    VkDescriptorSetLayoutBinding binding = layout_node->createInfo.pBindings[bindingIt->second];
```
lentinem: Should check for existence.

tobine: Do you mean check for existence of pBindings?
The layout node is constructed in vkCreateDescriptorSetLayout() and if we have a node we're guaranteed to have a pBindings.
I could add an assert that bindingIt->second is < createInfo.bindingCount. That should always be true and will ensure that we don't overstep the bounds of pBindings array.
Does that capture your intent?

lentinem: Yea, I understand that it will always be true in theory but I've tried to make it a habit to double check existence just in case as it can cause crashes if we find a case where theory does not always line up with practice.

tobine:
Updated with some improved defense mechanisms

tobine:
pushed to master

Issue #101, opened 2016-03-10
Mem tracker layer: "Invalid usage flag for buffer X used by vkCmdCopyQueryPoolResults()" error incorrectly reported


DominikWitczakAMD:
In the debug call-back, the layer informs the target buffer should have VK_BUFFER_USAGE_TRANSFER_DST_BIT usage set during creation. I believe this used to be a requirement back in the days, but has eventually been lifted. Checked the latest spec language, but I wasn't able to find any language backing this error.

Proposal is to remove the check:

![image](https://cloud.githubusercontent.com/assets/17761066/13666864/feb47a64-e6b3-11e5-957f-c9be13340f28.png)

lentinem:
Can you post the error that was output? It would be easier to trace down if it's correct or not with that information.

DominikWitczakAMD:
Sorry, of course:

![image](https://cloud.githubusercontent.com/assets/17761066/13698802/a719b6d4-e776-11e5-9a0f-92090918940f.png)

lentinem:
Spec MR has been submitted so closing.

PR #141, opened 2016-03-16
#101: Fixes invalid call-back for buffers targetted by vkCmdCopyQuery…

DominikWitczakAMD:
…PoolResults()

This addresses #101


lentinem:
-1. It's still not clear to me that this isn't required. The spec says: "VK_BUFFER_USAGE_TRANSFER_DST_BIT indicates that the buffer can be used as the destination of a transfer command." as well as "vkCmdCopyQueryPoolResults is considered to be a transfer operation" which implies (thought its not explicitly stated) that this should be used.

janharaldfredriksen-arm:
This got to be required - it shouldn't be any different than other copy (or transfer) commands in this respect.

This condition isn't stated in the Valid Usage section for vkCmdCopyQueryPoolResults, but I suspect we just forgot and that we should add "dstBuffer must have been created with VK_BUFFER_USAGE_TRANSFER_DST_BIT usage flag" there.

Do you agree? If so, shall we create a spec bug instead?

DominikWitczakAMD:
Don't mind either way! If that's what your driver needs to work fine, I'm all for getting this requirement added to the valid usage section.

The only reason I put up this MR is that I wanted to pinch someone for a reaction ;) Apparently, it worked!

lentinem:
As probably obvious from my comment I agree with Jan Harald and can propose that change unless you want to do so.

DominikWitczakAMD:
Yes, please, that would be very helpful.

lentinem:
https://gitlab.khronos.org/vulkan/vulkan/merge_requests/1213

tobine:
closing

PR #144, opened 2016-03-16
layers: validate capabilities against device features

chrisforbes:


layers/core_validation.cpp
```diff
@@ -2519,6 +2519,149 @@ static bool descriptor_type_match(layer_data *my_data, shader_module const *modu
...
 }
 
+
+static VkBool32 validate_shader_capabilities(layer_data *my_data, VkDevice dev, shader_module const *src)
+{
+    VkBool32 pass = VK_TRUE;
+
+    auto enabledFeatures = &my_data->physDevProperties.features;
+
+#define REQUIRE_FEATURE(feature) \
```
lentinem: Can we make this a function instead of a macro? It looks like it could take in parameter feature and the string name that you want to print.

chrisforbes: Well, the whole point was to avoid having to specify all the feature names twice, but sure.

lentinem: Generally I prefer functions to macros for readability , debugability, and a few other smaller reasons so if at all possible I would really prefer that. Could also make a function that took in the var name and returned the string name like we do in a number of other places if you only want to pass in the one arg.

chrisforbes:
@lentinem updated. Do you prefer that?

layers/core_validation.cpp
```diff
@@ -2519,6 +2519,170 @@ static bool descriptor_type_match(layer_data *my_data, shader_module const *modu
...
+                if (log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_EXT, /* dev */0,
+                            __LINE__, SHADER_CHECKER_BAD_CAPABILITY, "SC",
+                            "Shader declares capability %u, not supported in Vulkan.",
+                            insn.word(1)))
+                    pass = VK_FALSE;
+                break;
+            }
+        }
+    }
+#undef REQUIRE_FEATURE
```
lentinem: Not needed

layers/core_validation.cpp
```diff
@@ -2519,6 +2519,170 @@ static bool descriptor_type_match(layer_data *my_data, shader_module const *modu
...
+                    "enabled on the device",
+                    feature_name)) {
+            return false;
+        }
+    }
+}
+
+static VkBool32 validate_shader_capabilities(layer_data *my_data, VkDevice dev, shader_module const *src)
+{
+    VkBool32 pass = VK_TRUE;
```
lentinem: might make sense to invert pass to fail as most other functions work that way but I don't have a strong preference.

layers/core_validation.cpp
```diff
@@ -2519,6 +2519,170 @@ static bool descriptor_type_match(layer_data *my_data, shader_module const *modu
...
+            case spv::CapabilityImage1D:
+            case spv::CapabilitySampledBuffer:
+            case spv::CapabilityImageBuffer:
+            case spv::CapabilityImageQuery:
+            case spv::CapabilityDerivativeControl:
+                // Always supported by a Vulkan 1.0 implementation -- no feature bits.
+                break;
+
+            case spv::CapabilityGeometry:
+                pass = require_feature(my_data, enabledFeatures->geometryShader, "geometryShader") && pass;
```
lentinem: would probably use pass &=.

lentinem:
Yes! Thanks!

layers/core_validation.cpp
chrisforbes: Thanks, will fix.

layers/core_validation.cpp
chrisforbes: All of SC has the logic this way. I'm happy to invert it all when there's nothing more exciting happening, but it's a lot of churn.

lentinem: Ah ok. probably better to leave it for now then.

chrisforbes:
Feedback incorporated.

layers/core_validation.cpp
```diff
@@ -2519,6 +2519,161 @@ static bool descriptor_type_match(layer_data *my_data, shader_module const *modu
     }
 }
 
+static VkBool32 require_feature(layer_data *my_data, bool feature, char const *feature_name) {
```
lentinem: Sorry; missed this before but this will break windows as we are implicitly casting between VkBool32 and bool. I would in general prefer bool as VkBool32 was really for c code that didn't necessarily have bool but most of the code uses VkBool32 so I'm ok either way.

chrisforbes: Good point. Fixed.

tobine:
pushed to master

PR #148, opened 2016-03-17
layers: Fix validation of vkCmdBindDescriptorSets() during renderpass

tobine:
Spec allows vkCmdBindDescriptorSets() to be called inside or outside of a
renderpass so killing two core_validation checks that violate this behavior.

NOTE : This GitHub diff on this is ridiculous. Just removed two checks and reduced indentation for a big block.


lentinem:
+1 as far as a I can seems fine but it is hard to parse.

tobine:
fix pushed to master

PR #150, opened 2016-03-17
Descriptor counting

chrisforbes:


layers/core_validation.cpp
```diff
@@ -154,7 +154,7 @@ struct layer_data {
     unordered_map<VkImage, vector<ImageSubresourcePair>> imageSubresourceMap;
     unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE> imageLayoutMap;
     unordered_map<VkRenderPass, RENDER_PASS_NODE *> renderPassMap;
-    unordered_map<VkShaderModule, shader_module *> shaderModuleMap;
+    unordered_map<VkShaderModule, unique_ptr<shader_module>> shaderModuleMap;
```
lentinem: Whats the motivation for using unique_ptr instead of the object?

layers/core_validation.cpp
```diff
@@ -1299,51 +1299,73 @@ unsigned get_constant_value(shader_module const *src, unsigned id) {
...
+        describe_type_inner(ss, src, insn.word(3));
+        break;
     case spv::OpTypeStruct: {
-        dst += sprintf(dst, "struct of (");
+        ss << "struct of (";
         for (unsigned i = 2; i < insn.len(); i++) {
-            dst = describe_type(dst, src, insn.word(i));
-            dst += sprintf(dst, i == insn.len() - 1 ? ")" : ", ");
+            describe_type_inner(ss, src, insn.word(i));
+            if (i == insn.len() - 1) {
```
lentinem: I kind of prefer the ternary in this case but don't have too much of a preference

lentinem:
+1. Seems good just a few minor comments which can be addressed later if that makes more sense.

layers/core_validation.cpp
```diff
@@ -1299,51 +1299,73 @@ unsigned get_constant_value(shader_module const *src, unsigned id) {
...
+        break;
     default:
-        return dst + sprintf(dst, "oddtype");
+        ss << "oddtype";
+        break;
     }
 }
 
+
+static std::string describe_type(shader_module const *src, unsigned type) {
```
tobine: Does it make sense to change this guy to return "const char*" and just "return ss.str().c_str()"? That would avoid all of the repeated c_str() calls when calling log_msg().

lentinem: That would break scope as the lifetime of the string would be terminated when this function exited and you would be returning an invalid pointer.

tobine: Good point. Probably wise to favor valid pointers over reduced calls to c_str()...

layers/core_validation.cpp
chrisforbes: I've not done the work to prove that `shader_module` behaves well as `unordered_map::mapped_type`. It's not a trivial type. In particular, `spirv_inst_iter` is a pair of iterators into `shader_module::words`.

tobine:
pushed to master

PR #151, opened 2016-03-17
Track lifetime when using cross-queue semaphores.

lentinem:


layers/core_validation.cpp
```diff
@@ -5315,23 +5343,41 @@ vkQueueSubmit(VkQueue queue, uint32_t submitCount, const VkSubmitInfo *pSubmits,
...
+            const VkSemaphore &semaphore = submit->pWaitSemaphores[i];
+            semaphoreList.push_back(semaphore);
+            if (dev_data->semaphoreMap[semaphore].signaled) {
+                dev_data->semaphoreMap[semaphore].signaled = 0;
             } else {
                 skipCall |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT,
                                     VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT, 0, __LINE__, DRAWSTATE_QUEUE_FORWARD_PROGRESS,
                                     "DS", "Queue %#" PRIx64 " is waiting on semaphore %#" PRIx64 " that has no way to be signaled.",
-                                    (uint64_t)(queue), (uint64_t)(submit->pWaitSemaphores[i]));
+                                    reinterpret_cast<uint64_t &>(queue), reinterpret_cast<const uint64_t &>(semaphore));
```
tobine: should first cast also be "const"?

lentinem: Generally casting to a const won't hurt anything but it's not really a benefit in this case either. The semaphore had to be const because the variable itself was const whereas the variable queue is not const. We could make it a policy to cast everything that isn't being written to const which I'd be ok with but would need a number of other changes throughout the code.

layers/core_validation.cpp
```diff
@@ -5315,23 +5343,41 @@ vkQueueSubmit(VkQueue queue, uint32_t submitCount, const VkSubmitInfo *pSubmits,
...
+            }
+            const VkQueue &other_queue = dev_data->semaphoreMap[semaphore].queue;
```
tobine: Do we want special handling for NULL queue case here?
Also, should vkCreateSemaphore() be updated to explicitly set initial SEMAPHORE_NODE queue to VK_NULL_HANDLE?

lentinem: In theory that shouldn't be possible but I've said that before so yea probably makes sense to be safe and add that case. Will upload shortly.

lentinem: Done

tobine:
Hi Mark,
If you have time today can you give this PR a quick test and merge to master? Code LGTM.
I can merge Monday if you can't get to it.

mark-lunarg:
Hey Tobin, I'm out today but if I get a chance I'll try it before Monday...

tobine:
Pushed to master

PR #152, opened 2016-03-18
layers: Fix signed/unsigned comparison warnings in thread_check.h

MichaelMcDonnell:
Hi,

The file generator.py generates thread_check.h and emits signed/unsigned comparisons. The size or limits are passed as uint32_t, so the index in the for loop just needs to be the same type. It is very similar to the common problem of comparing an int with a size_t while iterating through a for loop. The change fixes 36 warnings.

Michael


lentinem:
+1

jon-lunarg:
Thanks  for the fixes.
Cheery-picked this to master branch.

MichaelMcDonnell:
Thanks Jon!

PR #153, opened 2016-03-18
WIP: Update event tracking

lentinem:
Account for sequential cross command buffer and cross queue tracking.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5115,6 +5115,9 @@ void updateTrackedCommandBuffers(layer_data *dev_data, VkQueue queue, VkQueue ot
         }
         other_queue_data->second.untrackedCmdBuffers.clear();
     }
+    for (auto eventStagePair : other_queue_data->second.eventToStageMap) {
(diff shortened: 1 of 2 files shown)


tobine:
Note that I'm seeing seg faults attempting to test this CL. The line I'm faulting on for cube --validate is in vkQueueSubmit:

    for (auto &function : pCB->eventUpdates) {

lentinem:
Thats odd; should be fixed now though.

tobine:
Thanks, Michael. It's working now. I still need to go through the code a bit more but will merge if I don't see any issues.

lentinem:
Thanks. If you can take a look at not just functionality but the design. As I mentioned there are a number of other state tracking fixes that need to happen and would use the same basic design as this but I would like to get feedback on if others would be comfortable using this framework for those types of tasks going forward.

stroyan:
sourceStageMask should be logged as 0x%x instead of %d format.

lentinem:
Done

tobine:
With a bit of reading and re-reading I think I like this approach. Cherry-picked, updated resetCB to clear eventUpdates and pushed to master.

PR #158, opened 2016-03-21
layers: GH157 Update error messages to be consistent

tobine:
Print dynamic offset in error messages as hex to be consistent.
Also fix typos of offet->offset


lentinem:
+1. Hard to make sure from the diffs but it looks like this is pretty much just using reinterpret_cast?

tobine:
Updates to reinterpret_cast and changes one parameter to be printed as hex instead of decimal.

tobine:
pushed to master

PR #161, opened 2016-03-21
demos: Only use shaderClipDistance feature in tri if the device suppo…

jon-lunarg:
…rts it

Allows tri to still run with Ilo driver.

Change-Id: Icb0e1094677372be946faffcc0defb0f94f07ef2

### DIFF
--- a/demos/tri.c
+++ b/demos/tri.c
@@ -244,6 +244,7 @@ struct demo {
     VkPhysicalDeviceMemoryProperties memory_properties;
 
     bool validate;
+    bool use_shaderClipDistance;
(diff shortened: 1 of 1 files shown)


chrisforbes:
You also need to actually /not use/ the feature if it's not present. This will require a variant of the shader.

I think a better fix would be to just expose the feature bit in ilo -- it clearly works.

chrisforbes:
(See LunarG/VulkanTools#14 for the similar nulldrv change)

jon-lunarg:
Okay, abandoning theis PR.  I made this fix to Ilo driver to report shaderclipdistance as a supported feature. Thanks.

Issue #117, opened 2016-03-12
Swapchain Layer: Unnecessary error when selecting VK_PRESENT_MODE_FIFO_KHR without enumeration


szdarkhack:
The swapchain extension spec notes that VK_PRESENT_MODE_FIFO_KHR is *required* (extended spec section E.3.6 issue 11), and thus no enumeration of presentation modes should be needed to select it. Currently the layer indiscriminately complains that vkGetPhysicalDeviceSurfacePresentModesKHR() has not been called. There should probably be a special check to silence the error in this particular situation.

szdarkhack:
Fixed here #118

szdarkhack:
The PR has been accepted, closing.

PR #167, opened 2016-03-23
layers: GH117: swapchain layer warns if app uses neither semaphore no…

ianelliottus:
This addresses github issue #117 (for the Vulkan-Doc repository).  Applications should call
vkAcquireNextImageKHR with either a valid semaphore and/or fence.  Issue a
warning if both are set to VK_NULL_HANDLE.

### DIFF
--- a/layers/swapchain.cpp
+++ b/layers/swapchain.cpp
@@ -1744,9 +1744,9 @@ VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkAcquireNextImageKHR(VkDevice de
                               VK_KHR_SWAPCHAIN_EXTENSION_NAME);
     }
     if ((semaphore == VK_NULL_HANDLE) && (fence == VK_NULL_HANDLE)) {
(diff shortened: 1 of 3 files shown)


tobine:
:+1:

layers/swapchain.cpp
```diff
@@ -1743,6 +1743,11 @@ VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkAcquireNextImageKHR(VkDevice de
                               "%s() called even though the %s extension was not enabled for this VkDevice.", __FUNCTION__,
                               VK_KHR_SWAPCHAIN_EXTENSION_NAME);
     }
+    if ((semaphore == VK_NULL_HANDLE) && (fence == VK_NULL_HANDLE)) {
```
lentinem: This seems like this should be an error. With no way to wait how would the app tell when it could display?

jjulianoatnv: I agree with ERROR not WARNING. Otherwise it looks good. Thanks.

ianelliottus: Agreed.  I will change it to an error.

ianelliottus: I changed it to an error.  I left the LOG_WARNING macro, in case I need it later.  I can take it out if you prefer.

lentinem:
+1

ianelliottus:
Thanks for the review!  I cherry-picked and pushed the fix to the "master" branch.

PR #170, opened 2016-03-23
layers: LX450, Tighten up queueFamilyIndex validation, fix crash

mark-lunarg:
For CreateBuffer and CreateImage the QFIs should be ignored unless
sharingMode is set to CONCURRENT.  Also added QFI validation to
CreateCommandPool and added a special case for when an index is
incorrectly set to QUEUE_FAMILY_IGNORED.

Change-Id: I17639230ea7aa58ab89b9b0dc33e55927e9f1c84


tobine:
merged to master

PR #175, opened 2016-03-23
layers: Get rid of 1K limit on debug message length

chrisforbes:
Someone on windows: please verify that this behaves correctly.

### DIFF
--- a/layers/vk_layer_logging.h
+++ b/layers/vk_layer_logging.h
@@ -212,6 +212,21 @@ static inline VkBool32 will_log_msg(debug_report_data *debug_data, VkFlags msgFl
     return true;
 }
 
+#ifdef WIN32
+static inline int vasprintf(char **strp, char const *fmt, va_list ap) {
+    *strp = nullptr;
+    int size = _vscprintf(fmt, ap);
+    if (size >= 0) {
+        *strp = (char *)malloc(size+1);
+        if (!*strp) {
+            return -1;
+        }
+        _vsnprintf(*strp, size+1, fmt, ap);
+    }
+    return size;
+}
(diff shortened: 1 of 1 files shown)


layers/vk_layer_logging.h
```diff
@@ -212,6 +212,18 @@ static inline VkBool32 will_log_msg(debug_report_data *debug_data, VkFlags msgFl
     return true;
 }
 
+#ifdef WIN32
```
courtney-g: Why only Windows? Or is this not provided by Windows?

layers/vk_layer_logging.h
```diff
@@ -230,12 +242,14 @@ static inline VkBool32 log_msg(debug_report_data *debug_data, VkFlags msgFlags,
         return false;
     }
 
-    char str[1024];
     va_list argptr;
     va_start(argptr, format);
-    vsnprintf(str, 1024, format, argptr);
+    char *str;
+    vasprintf(&str, format, argptr);
```
courtney-g: May need ifdef WIN32 here as well depending on answer above.

layers/vk_layer_logging.h
chrisforbes: This is not provided by windows.

courtney-g:
LGTM

layers/vk_layer_logging.h
stroyan: If malloc fails and vasprintf returns -1, then it would be better to pass a fixed string like "out of memory" rather than passing on a NULL to debug_report_log_msg.

jon-lunarg:
cherry picked to master

PR #176, opened 2016-03-24
layers: Relax descriptor type match for texel buffers slightly

chrisforbes:
The descriptor doesn't really provide a sampler, but the consuming code
doesn't really look at one either.

Allow VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER to match against
OpTypeSampledImage with an image type having Dim=Buffer and Sampled=1.


tobine:
merged to master

PR #179, opened 2016-03-24
tests: Fix InvalidQueueFamilyIndex test

tobine:
Need to enable parameter_validation layer for layer_validation_tests.cpp.
Also format of error message changed when the check migrated.


mark-lunarg:
Pushed to Master.  Thanks Tobin.

PR #182, opened 2016-03-24
layers: GH177 Fix deadlocks

tobine:
Fix some early return cases where lock wasn't correctly released.


lentinem:
+1

mark-lunarg:
+1, pushed to Master.  Thanks Tobin.

PR #185, opened 2016-03-24
layers: WSI functions should not work unless enabled.

ianelliottus:
The WSI extension functions are special.  They are statically exported from the
loader.  This means that an application can directly link against them, and not
have to call a vkGet*ProcAddr function.  However, the application is still
reponsible for for enabling the appropriate WSI extension.

An application was observed that called the WSI functions without enabling the
appropriate extensions.  It worked.  When that application was run with the
validation layers, it crashed.  A change was made to the layer to make sure
that an error was always logged.  It was decided that the loader should change
so that the terminator functions not work (and print a brief message) if an
application calls the function without enabling its extension.  This is done in
the terminator so that the validation layer can be called, which should log a
good error message; and yet still done here in case the application is never
tried with validation turned on.

### DIFF
--- a/loader/wsi.c
+++ b/loader/wsi.c
@@ -257,8 +257,9 @@ terminator_GetPhysicalDeviceSurfaceSupportKHR(VkPhysicalDevice physicalDevice,
     struct loader_instance *ptr_instance =
         (struct loader_instance *)phys_dev->this_icd->this_instance;
     if (!ptr_instance->wsi_surface_enabled) {
-        fprintf(stderr, "VK_KHR_surface extension not enabled.  "
-                "vkGetPhysicalDeviceSurfaceSupportKHR not executed!\n");
+        loader_log(ptr_instance, VK_DEBUG_REPORT_ERROR_BIT_EXT, 0,
+                   "VK_KHR_VK_KHR_surface extension not enabled.  "
+                   "vkGetPhysicalDeviceSurfaceSupportKHR not executed!\n");
         return VK_SUCCESS;
     }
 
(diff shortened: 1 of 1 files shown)


loader/wsi.c
```diff
@@ -250,8 +251,18 @@ terminator_GetPhysicalDeviceSurfaceSupportKHR(VkPhysicalDevice physicalDevice,
                                               uint32_t queueFamilyIndex,
                                               VkSurfaceKHR surface,
                                               VkBool32 *pSupported) {
+    // First, check to ensure the appropriate extension was enabled:
     struct loader_physical_device *phys_dev =
         (struct loader_physical_device *)physicalDevice;
+    struct loader_instance *ptr_instance =
+        (struct loader_instance *)phys_dev->this_icd->this_instance;
+    if (!ptr_instance->wsi_surface_enabled) {
+        fprintf(stderr, "VK_KHR_surface extension not enabled.  "
```
courtney-g: This would be good to report via debug_report as well.

courtney-g:
+1

jon-lunarg:
The rest of the loader code uses loader_log() rather than fprintf() to log errors. This code should use loader_log()  also since loader_log() will log as appropriately controlled by EXT_debug_report  and also the VK_LOADER_DEBUG env var.

Otherwise , LGTM.

ianelliottus:
I converted the fprintf() calls to loader_log() calls.  Ready to go now?

jon-lunarg:
Thanks Ian! cherry picked the changes into master branch

PR #186, opened 2016-03-24
Fix layout transitions for combined aspect masks.

lentinem:


tobine:
LGTM, merged to master

PR #187, opened 2016-03-24
layers: LX250 Validate attachment layouts in render pass setup

MarkY-LunarG:
Compare render pass VkAttachmentReference with
VkAttachmentDescription to verify compatiblity.

If attachment description was different that attachment format, we never logged an error. 

Change-Id: I014ac0029b520e8c4c7e57ce0ccb5dcbcc03ff0f

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -9235,6 +9235,17 @@ VkBool32 ValidateLayouts(const layer_data *my_data, VkDevice device, const VkRen
                                     "Layout for color attachment is %s but can only be COLOR_ATTACHMENT_OPTIMAL or GENERAL.",
                                     string_VkImageLayout(subpass.pColorAttachments[j].layout));
                 }
+            // Verify that the render pass color attachment initial/final layout is compatible with the sub-pass attachment layout.
+            } else if (!(pCreateInfo->pAttachments[subpass.pColorAttachments[j].attachment].initialLayout == VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL ||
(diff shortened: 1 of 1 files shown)


layers/core_validation.cpp
```diff
@@ -9235,6 +9235,17 @@ VkBool32 ValidateLayouts(const layer_data *my_data, VkDevice device, const VkRen
                                     "Layout for color attachment is %s but can only be COLOR_ATTACHMENT_OPTIMAL or GENERAL.",
                                     string_VkImageLayout(subpass.pColorAttachments[j].layout));
                 }
+            // Verify that the render pass color attachment initial/final layout is compatible with the sub-pass attachment layout.
```
lentinem: I don't think this is correct. Can you paste the spec wording that lead to this? The intial/final layouts can be anything as long as they match the actual resource layout when the pass beings/ends (there are checks for those elsewhere).

MarkY-LunarG: I don't see a spec section regarding this.  But, at the least, this seems to be a reasonable Warning.  Should I convert it from an Error to a Warning?  The result of someone doing this is not being able to see their resulting image, and I believe, the driver could crash.

lentinem: I'm not sure this is a warning either. In many cases the transition automatically happens from initial to the attachment's layout and from the attachment's layout to final. In those cases I would expect the application to set initial to whatever it is before the pass is started (often preinitialized, undefined, or present) as then they don't need to do an explicit barrier.

It does seem to be worth making this a warning if those conditions are not met (see spec section on subpass dependency) though as in that case the initial and (not or which seems to be what this is checking) final should definitely match.

PR #188, opened 2016-03-24
layers: Fix issue when sub-passes have diff attachment count.

MarkY-LunarG:
In core_validation, if the attachment count between a primary
and secondary sub-pass was different, then it should treat
the shorter one as UNUSED.  So, if the longer one has the
attachment flags set to UNUSED, it shouldn't be an error.

Change-Id: I3647403551962ef522e7e2ba7ab10a5cb7466c52


layers/core_validation.cpp
```diff
@@ -2246,21 +2246,22 @@ static bool attachment_references_compatible(const uint32_t index, const VkAttac
...
         if ((pPrimaryAttachments[pPrimary[index].attachment].format ==
              pSecondaryAttachments[pSecondary[index].attachment].format) &&
             (pPrimaryAttachments[pPrimary[index].attachment].samples ==
              pSecondaryAttachments[pSecondary[index].attachment].samples))
-            return true;
+            success = true;
     }
     // Format and sample counts didn't match
-    return false;
+    return success;
```
mark-lunarg: Might it be a little clearer to use 'result' instead of 'success'?  Or maybe follow what other CV functions do I guess.

layers/core_validation.cpp
```diff
@@ -2246,21 +2246,22 @@ static bool attachment_references_compatible(const uint32_t index, const VkAttac
                                              const uint32_t primaryCount, const VkAttachmentDescription *pPrimaryAttachments,
                                              const VkAttachmentReference *pSecondary, const uint32_t secondaryCount,
                                              const VkAttachmentDescription *pSecondaryAttachments) {
+    bool success = false;
     if (index >= primaryCount) { // Check secondary as if primary is VK_ATTACHMENT_UNUSED
-        if (VK_ATTACHMENT_UNUSED != pSecondary[index].attachment)
-            return false;
+        if (VK_ATTACHMENT_UNUSED == pSecondary[index].attachment)
+            success = true;
```
lentinem: return true seems slightly preferable as there is no additional logic that happens after this point if this condition is met.

lentinem:
+1. Only minor comments.

layers/core_validation.cpp
MarkY-LunarG: I'm switching to immediate return per Mike's request.

MarkY-LunarG:
Somehow fixes and pull/push process grabbed a bunch more changes.  Will try to resolve.

PR #189, opened 2016-03-25
Add check for renderArea being within framebuffer bounds.

lentinem:


layers/core_validation.h
```diff
@@ -197,6 +197,7 @@ typedef enum _DRAW_STATE_ERROR {
     DRAWSTATE_NO_PIPELINE_BOUND,             // Unable to identify a bound pipeline
     DRAWSTATE_INVALID_POOL,                  // Invalid DS pool
     DRAWSTATE_INVALID_SET,                   // Invalid DS
+    CORE_INVALID_RENDER_AREA,                // Invalid DS layout
```
tobine: Comment is incorrect.
For now just skip w/ "DRAWSTATE_" as the prefix in this enum. If we want to migrate the enums to new names we should do it wholesale. Right now doc validator still checks for specific prefixes so that also needs to be updated if/when we modify enum naming scheme.

lentinem: Bleh. I guess mass changing is ok as it's really just find/replace. Will upload shortly.

tobine:
merged to master

PR #194, opened 2016-03-25
layers: Replace ptr-to-ptr descriptor data struct with vector

tobine:
We were using GENERIC_HEADER** ppDescriptors to track updates to slots.
This was pretty messy so make it a bit cleaner with a vector instead.


lentinem:
+1

tobine:
merged to master

PR #196, opened 2016-03-25
layers: Fix issue when sub-passes have diff attachment count.

MarkY-LunarG:
In core_validation, if the attachment count between a primary
and secondary sub-pass was different, then it should treat
the shorter one as UNUSED.  So, if the longer one has the
attachment flags set to UNUSED, it shouldn't be an error.

Change-Id: I3647403551962ef522e7e2ba7ab10a5cb7466c52


MarkY-LunarG:
This was MR 188, but I had issues trying to update it.  I have incorporated the commetns from that merge request.

lentinem:
+1

tobine:
merged to master

PR #197, opened 2016-03-25
layers: Update descriptor tracking in core_validation

tobine:
NOTE: No rush on this one. I'll give it at least the weekend to stew on it.

Old mem_tracker code to record updated storage image/buffers was naive in
that is gathered any image/buffer updates to descriptor sets and marked as
updated every such image/buffer that was referenced by currently bound sets.

In reality sets may be bound but not used. This is captured in the
"active_slots" construct of the pipeline.

The code here updates tracking for the updated buffers by waiting until
a draw cmd for gfx pipe or dispatch cmd for compute pipe, and then gathering
the updated images/buffers based on which slots are active.

I've multi-purposed validate_draw_state() to now be
validate_and_update_draw_state() as it includes recording of the images/buffers
from the active_slots, which is then used to mark those as read.
Also, that same function is now being called for compute dispatch cmds so
"draw" state is not ideal.
There is still some rework that needs to be done to improve this situation.


layers/core_validation.cpp
```diff
@@ -2903,10 +2907,23 @@ static VkBool32 validate_dynamic_offsets(layer_data *my_data, const GLOBAL_CB_NO
...
+                        for (j = 0; j < pWDS->descriptorCount; ++j) {
+                            pCB->updateBuffers.insert(dev_data->bufferViewMap[pWDS->pTexelBufferView[j]].buffer);
+                        }
+                    } else if (pWDS->descriptorType == VK_DESCRIPTOR_TYPE_STORAGE_BUFFER ||
+                               pWDS->descriptorType == VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC) {
+                        for (j = 0; j < pWDS->descriptorCount; ++j) {
+                            pCB->updateBuffers.insert(pWDS->pBufferInfo[j].buffer);
                         }
                     }
+                    i += j; // Advance i to end of this set of descriptors (++i at end of for loop will move 1 index past
```
chrisforbes: I'd do `i += pWDS->descriptorCount;` and narrow the scope of `j` back to each of its uses.

layers/core_validation.cpp
```diff
@@ -2968,54 +2989,55 @@ static VkBool32 validate_draw_state(layer_data *my_data, GLOBAL_CB_NODE *pCB, Vk
...
-            for (size_t i = 0; i < pPipe->vertexBindingDescriptions.size(); i++) {
-                if ((pCB->currentDrawData.buffers.size() < (i + 1)) || (pCB->currentDrawData.buffers[i] == VK_NULL_HANDLE)) {
-                    result |= log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0,
-                                      __LINE__, DRAWSTATE_VTX_INDEX_OUT_OF_BOUNDS, "DS",
-                                      "The Pipeline State Object (%#" PRIxLEAST64
-                                      ") expects that this Command Buffer's vertex binding Index " PRINTF_SIZE_T_SPECIFIER
-                                      " should be set via vkCmdBindVertexBuffers.",
-                                      (uint64_t)pCB->lastBound[VK_PIPELINE_BIND_POINT_GRAPHICS].pipeline, i);
+            // For given active slots, verify any dynamic descriptors and record updated images & buffers
+            result |= validate_and_update_drawtime_descriptor_state(my_data, pCB, activeSetBindingsPairs);
```
chrisforbes: activeBindingSetPairs isn't going to be sensible for non-graphics pipelines yet -- the SC side needs support for compute added.

tobine: Hi Chris,
Yeah, I now see that we're not even calling verifyPipelineCreateState for the compute case.
What's really needed for these changes is to be able to call validate_and_capture_pipeline_shader_state() for a compute pipeline. Any guestimate on how far away we are from being able to do that?

As an interim solution I can either skip the compute-related checks momentarily, or I can keep some version of the old solution alive. If we could be parsing compute shaders correctly this week, I would think it's ok to just skip the compute-related for now. If it's longer I should probably keep some form of the compute checks alive.

chrisforbes: The changes to `validate_and_capture_pipeline_shader_state()` are pretty straightforward -- is just a matter of where they fall in the list of things to build.

tobine:
Just force pushed an update here that provides a temporary workaround for compute pipelines/shaders that matches the functionality we previously had in these cases. Included comments to update the code when shader_checker can parse compute shaders.

chrisforbes:
Needs a rebase, but otherwise looks OK now.

layers/core_validation.cpp
```diff
@@ -2834,17 +2832,22 @@ static uint32_t getBindingEndIndex(const LAYOUT_NODE *pLayout, const uint32_t bi
...
+//  3. Grow updateBuffers for pCB to include buffers from STORAGE*_BUFFER descriptor buffers
+static VkBool32 validate_and_update_drawtime_descriptor_state(
+    layer_data *dev_data, GLOBAL_CB_NODE *pCB,
+    const vector<std::pair<SET_NODE *, unordered_set<uint32_t>>> &activeSetBindingsPairs) {
     VkBool32 result = VK_FALSE;
 
     VkWriteDescriptorSet *pWDS = NULL;
     uint32_t dynOffsetIndex = 0;
     VkDeviceSize bufferSize = 0;
+    uint32_t j = 0;
```
lentinem: Is there a particular reason why j is broken out now"?

layers/core_validation.cpp
```diff
@@ -2903,10 +2906,23 @@ static VkBool32 validate_dynamic_offsets(layer_data *my_data, const GLOBAL_CB_NO
...
-                            i += j; // Advance i to end of this set of descriptors (++i at end of for loop will move 1 index past
-                                    // last of these descriptors)
+                        }
+                    } else if (pWDS->descriptorType == VK_DESCRIPTOR_TYPE_STORAGE_IMAGE) {
+                        for (j = 0; j < pWDS->descriptorCount; ++j) {
+                            pCB->updateImages.insert(pWDS->pImageInfo[j].imageView);
+                        }
+                    } else if (pWDS->descriptorType == VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER) {
+                        for (j = 0; j < pWDS->descriptorCount; ++j) {
+                            pCB->updateBuffers.insert(dev_data->bufferViewMap[pWDS->pTexelBufferView[j]].buffer);
```
lentinem: Might be worth adding an assert to make sure the map has that element.

layers/core_validation.cpp
```diff
@@ -2917,18 +2933,57 @@ static VkBool32 validate_dynamic_offsets(layer_data *my_data, const GLOBAL_CB_NO
...
 
 // Validate overall state at the time of a draw call
-static VkBool32 validate_draw_state(layer_data *my_data, GLOBAL_CB_NODE *pCB, VkBool32 indexedDraw) {
+static VkBool32 validate_and_update_draw_state(layer_data *my_data, GLOBAL_CB_NODE *pCB, const VkBool32 indexedDraw,
+                                               const VkPipelineBindPoint bindPoint) {
+    VkBool32 result = VK_FALSE;
     // First check flag states
-    VkBool32 result = validate_draw_state_flags(my_data, pCB, indexedDraw);
-    PIPELINE_NODE *pPipe = getPipeline(my_data, pCB->lastBound[VK_PIPELINE_BIND_POINT_GRAPHICS].pipeline);
+    if (VK_PIPELINE_BIND_POINT_GRAPHICS == bindPoint)
```
lentinem: Is this always true? Doesn't this only get called from draw functions?

lentinem:
+1. Mostly looks good.

layers/core_validation.cpp
tobine: I think I was attempting to be "smart" by avoiding the re-decls in each loop. Looking at the code now I don't like that at all and I'm going to move the decls back to the loops where j is properly scoped to each loop.

layers/core_validation.cpp
tobine: yep

layers/core_validation.cpp
tobine: Currently this is always true, but the plan is to soon multi-purpose this function to handle compute as well when compute descriptors are properly tracked.

tobine:
merged to master

PR #198, opened 2016-03-25
layers: Initialize IMAGE_CMD_BUF_LAYOUT_NODE.initialLayout

mark-lunarg:
Uninitialized initalLayout in core_validation caused scads of bad layout-
transition errors in DOTA2.

Change-Id: I3d8b913222e8562ef02333a4f804d78c28269949

### DIFF
--- a/layers/core_validation.h
+++ b/layers/core_validation.h
@@ -458,7 +458,7 @@ typedef struct _IMAGE_LAYOUT_NODE {
 
 class IMAGE_CMD_BUF_LAYOUT_NODE {
   public:
-    IMAGE_CMD_BUF_LAYOUT_NODE() {}
+    IMAGE_CMD_BUF_LAYOUT_NODE() : initialLayout(VK_IMAGE_LAYOUT_UNDEFINED) {}
     IMAGE_CMD_BUF_LAYOUT_NODE(VkImageLayout initialLayoutInput, VkImageLayout layoutInput)
         : initialLayout(initialLayoutInput), layout(layoutInput) {}
 


mark-lunarg:
Michael, this seems to be clean from the testing I've done -- on Windows when uninitialized, the enum was unrecognized and caused errors such as "457615 DS(ERROR): object: 0 type: 6 location: 4875 msgCode: 6: Cannot submit cmd buffer using image with layout VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL when first use is Unhandled VkImageLayout."  Was zero-initialization (LAYOUT_UNDEFINED) intended?

layers/core_validation.h
```diff
@@ -458,7 +458,7 @@ typedef struct _IMAGE_LAYOUT_NODE {
 
 class IMAGE_CMD_BUF_LAYOUT_NODE {
   public:
-    IMAGE_CMD_BUF_LAYOUT_NODE() {}
+    IMAGE_CMD_BUF_LAYOUT_NODE() : initialLayout(VK_IMAGE_LAYOUT_UNDEFINED) {}
```
lentinem: If this is causing issues we should really find the root cause here. The problem with this is IMAGE_CMD_BUF_LAYOUT_NODE() should only be used as a temporary holder and should never be accessed unless initialLayout and layout are set. If it is there is a bug somewhere. If we set initialLayout like this we are masking that bug and potentially hiding other issues as the image this class represents will likely not have layout UNDEFINED.

lentinem: Can you share other info (e.g. line numbers, var values) about what you are seeing with this?

lentinem:
I ran my samples by adding an assert on line 3676 where FindLayout is as that should always be true and if not for some reason could lead to use of uninit vars. I also changed the default constructor to use IMAGE_CMD_BUF_LAYOUT_NODE() : initialLayout(VK_IMAGE_LAYOUT_MAX_ENUM), layout(VK_IMAGE_LAYOUT_MAX_ENUM) {} in the hope that would cause a throw but didn't see any errors. Can you try these and see if the assert throws and if you see MAX_ENUM being used?

mark-lunarg:
Sure Michael.

SetLayout (below) creates 'node', which is uninitialized.  FindLayout is called (returns False) and then SetLayout(pCB, imgpair, {node.initialLayout, layout}); is executed, which adds a Node to the map in which .initialLayout is still uninitialized while .layout is fine.  It never matches any IMAGE_LAYOUT enums, resulting in false positives.

void SetLayout(GLOBAL_CB_NODE *pCB, ImageSubresourcePair imgpair, const VkImageLayout &layout) {
    // TODO (mlentine): Maybe make vector a set?
    if (std::find(pCB->imageSubresourceMap[imgpair.image].begin(), pCB->imageSubresourceMap[imgpair.image].end(), imgpair) !=
        pCB->imageSubresourceMap[imgpair.image].end()) {
        pCB->imageLayoutMap[imgpair].layout = layout;
    } else {
        // TODO (mlentine): Could be expensive and might need to be removed.
        assert(imgpair.hasSubresource);
        IMAGE_CMD_BUF_LAYOUT_NODE node;
        FindLayout(pCB, imgpair.image, imgpair.subresource, node);
        SetLayout(pCB, imgpair, {node.initialLayout, layout});
    }
}

lentinem:
Thanks Mark. Thats what the assert I was refering to above. This should really always return true but until we can figure that out can you try changing to if (!FindLayout(pCB, imgpair.image, imgpair.subresource, node)) {node.initialLayout = layout;} and see if that fixes it for you. I'm ok with that being checked it as it's still correct but would eventually like to try to narrow down why that returns false if either we can get a sample that causes that or if someone on your side has some time to look into that.

mark-lunarg:
Michael, dota2 on Windows produces these errors, but it also has other layout-related errors and I haven't narrowed it down to either the app or the layers yet.  I pushed your fix, and will look into this more this coming week.

PR #201, opened 2016-03-26
vkjson: Fix include paths and clang errors

critsec:


courtney-g:
+1

libs/vkjson/vkjson.cc
```diff
@@ -456,7 +456,7 @@ inline void VisitForWrite(Visitor* visitor, const T& t) {
   Iterate(visitor, const_cast<T*>(&t));
 }
 
-template <typename T, typename = EnableForStruct<T>, typename = void>
+template <typename T, typename /*= EnableForStruct<T>*/, typename /*= void*/>
```
pimanttr: EnableForStruct is needed for SFINAE - to make sure this overload is only used if T is a struct. I'm not sure how this compiles without it - nor what problem it causes on clang?

libs/vkjson/vkjson.cc
```diff
@@ -634,7 +634,7 @@ class JsonReaderVisitor {
   std::string* errors_;
 };
 
-template <typename T, typename = EnableForStruct<T>>
+template <typename T, typename /*= EnableForStruct<T>*/>
```
pimanttr: (same here)

critsec:
The problem is that the default value is stated twice, clang is issuing an error about redefining template default values. For line 459, see line 361.

pimanttr:
Ah, ok, got it, the forward declarations already have it so it's fine to remove in the definition.

+1 then

PR #202, opened 2016-03-28
layers: intercept device functions in vkGetInstanceProcAddr

olvaffe:
Each layer should be a conformant Vulkan implementation first, and that
requires a correctly implemented vkGetInstanceProcAddr.

Practically, not all loaders implement dynamic dispatch for unknown device
functions.  And not all loaders intercept functions they do not care, such
as vkCreateimage.  A loader might choose to call down to layers/drivers
when it encounters

  vkGetInstanceProcAddr(instance, "vkCreateImage");

and a layer intercepting vkCreateImage should respond to that.


layers/core_validation.cpp
```diff
@@ -10887,25 +10905,23 @@ VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vkGetDeviceProcAddr(VkD
     if (!strcmp(funcName, "vkCreateEvent"))
         return (PFN_vkVoidFunction)vkCreateEvent;
 
+    if (!strcmp(funcName, "vkCreateSwapchainKHR"))
```
courtney-g: Why does core_validation need to return it's WSI functions when WSI is not enabled?
We needed it for swapchain so that it could verify the extensions were enabled, but it doesn't seem like it's needed here. You don't need the tests above, if the app cannot get to those functions.

lentinem: +1 to courtney's comment.

vk-layer-generate.py
```diff
@@ -1194,6 +1193,13 @@ def generate_intercept(self, proto, qual):
             "DestroySwapchainKHR",
             "GetSwapchainImagesKHR"
         ]
+        khr_swapchain_functions= [
```
lentinem: Similarly, why is this needed?

jon-lunarg:
If layers are really going to handle GetInstanceProcAddr() from the app  (not just for instance layer bootstrapping as they currently do)  as Olv is advocating.  Then that implies they need to return a valid pointer prior to device extensions being enabled.  This is what Olvs code does. The question then is what should a layer do if the extension is not enabled and the entry points get called?  It looks like Olvs proposal is they call down the chain and return without doing any processing in the layer itself.

My questions/comments are twofold:
1) Prior to this change if the wsi swapchain extension wasn't enabled then the layer would return NULL for GDPA. With this new pattern, the layer always returns it's device extension entry points regardless of if the extension is enabled or not.  I think this would allow for successful app GIPA from the LunarG loader of device extensions that are implemented by a layer but disabled. In programming the dynamic device extension dispatch table, I believe the loader relies on layers and
ICDs to return NULL for disabled extensions.

2) The specification of GetInstanceProcAddr (including device level entry points including extensions) was done with the assumption that the loader would implement dynamic dispatch.  I think if it was envisioned a loader would not implement dynamic dispatch then a different specification fro GetInstanceProcAddr would have been arrived at.

courtney-g:
On Mar 28, 2016 12:37 PM, "Jon Ashburn" <notifications@github.com> wrote:
>
> If layers are really going to handle GetInstanceProcAddr() from the app
(not just for instance layer bootstrapping as they currently do) as Olv is
advocating.

I don't see why that's necessary, the layers can only live in between the
loader and the icd, it does not need to have the exact same API interface.
Olvs example is a good way to discuss pros and cons. What I see is a one
time initialization check has now become a runtime check just because the
layer returned it's function pointer.
That doesn't seem very Vulkan like.

Then that implies they need to return a valid pointer prior to device
extensions being enabled. This is what Olvs code does. The question then is
what should a layer do if the extension is not enabled and the entry points
get called? It looks like Olvs proposal is they call down the chain and
return without doing any processing in the layer itself.
>
> My questions/comments are twofold:
> 1) Prior to this change if the wsi swapchain extension wasn't enabled
then the layer would return NULL for GDPA. With this new pattern, the layer
always returns it's device extension entry points regardless of if the
extension is enabled or not. I think this would allow for successful app
GIPA from the LunarG loader of device extensions that are implemented by a
layer but disabled. In programming the dynamic device extension dispatch
table, I believe the loader relies on layers and
> ICDs to return NULL for disabled extensions.
>
> 2) The specification of GetInstanceProcAddr (including device level entry
points including extensions) was done with the assumption that the loader
would implement dynamic dispatch. I think if it was envisioned a loader
would not implement dynamic dispatch then a different specification fro
GetInstanceProcAddr would have been arrived at.

jon-lunarg:
Seems like the fundamental question is should the layers try to be a close as possible to the spec or not in their GIPA and GDPA functions?

If the answer is no then we should just document better the existing behaviours.  If the answer is yes, then this patch is desirable to have (probably with some  modifications). My concern is the loader probably relies on GDPA (not GIPA) of a device extension returning  NULL if the extension is not enabled.  I don't think this patch does that.  I would want GIPA of device extension would return valid entrypoint regardless of if it is enabled or not. GDPA of device extension would return NULL if the extension is not enabled.

A secondary concern is if we start changing the interface between layers and loader to make the layers more spec compliant, we may need to version the interface;  so layers in the wild can work with old and new interface depending on if the system has an old or new loader.

lentinem:
I'm still unclear here as to why layers must intercept extensions to be as close as possible to the spec. All layers will return the next level in the chain when the layer does not intercept the function which seems more correct to me than intercepting and just calling down. Also, going forward it will have to work this way because there will be extensions that the layers are not aware of.

jon-lunarg:
Hi Olv:
What is status of this Pull Request?  Is it still desired or has it become obsolete?

olvaffe:
Ahh, I set my mail filers incorrectly and missed all github notifications.  I am sorry about that.

layers/core_validation.cpp
olvaffe: It is to piggyback GIPA on GDPA.  Then in vkCreateSwapchainKHR, for example, the extension is checked again.

As for spec compliant, this fixes GIPA("vkCreateSwapchainKHR") but regresses GDPA("vkCreateSwapchainKHR").

vk-layer-generate.py
olvaffe: To make GIPA intercept swapchain functions as GDPA does.  Otherwise, a loader must dispatch GIPA("vkCreateSwapchainKHR") to itself.  This is fine because it imposes a neglectable overhead.

But it will be an issue when a layer intercepts in GDPA("vkSomeDeviceFunctionVENDOR") but not not in GIPA("vkSomeDeviceFunctionVENDOR").  If the device function is unknown to a loader (Android loader), the loader could not dispatch the function to itself but has to generate a dynamic dispatch function.  That is a lot of works for a loader to work around an incompliance of a layer.

olvaffe:
@lentinem The motivation of this pull request is to allow the loader to do less by asking:

  If a layer intercepts a function in its GDPA, it should do so in its GIPA.

It is also more correct this way IMO.  For extensions a layer is unaware of, since it does not intercept the extension in GDPA, it should not intercept the extension in GIPA either.

olvaffe:
@jon-lunarg I agree with all your points.

I believe the deviation of layer behavior from the spec has to do with the desktop loader envisioning dynamic dispatches.

It is also true that this pull request makes GIPA more compliant (by handling device commands correctly) and GDPA less compliant (by not returning NULL for un-enabled extensions).

I do want to formalize the current loader-layer interface and call it v0

  https://gist.github.com/olvaffe/bbab7607528dd7107e268f5d0eec84f6

No loader depends on this pull request AFAIK.  We may incorporate this pull request (when it is fixed) into v0 interface, or delay it for v1.

But yes, I do want GIPA and GDPA to be compliant at some point because our loader would like to avoid dynamic dispatches when possible.

olvaffe:
@courtney-g A layer can return a function pointer from GDPA that does no runtime check.  Yes, it will be more works for layers if they want to avoid the runtime checks.

When the GIPA-returned device function pointer is called without enabling the associated device extension, the behavior is undefined.  I chose to call down, but a layer can do whatever it likes.

The spec is clear GDPA should return NULL for un-enabled extensions.  This pull request makes some layers behave worse in that regard.  But the swapchain layer also behaves that way.  IMO, all of them need to be fixed.

ianelliottus:
> The spec is clear GDPA should return NULL for un-enabled extensions. This pull request makes some layers behave worse in that regard. But the swapchain layer also behaves that way. IMO, all of them need to be fixed.

I've been wondering about this w.r.t. layers.  Can the purpose and functionality of the layer legitimately affect what it does for GIPA and GDPA?  I think the answer is yes.  It seem appropriate (given the purpose of a given layer) for them to return NULL depending on whether they NEED to intercept a given function, and that this need can be influenced by whether the extension is enabled.

For example, the swapchain layer is trying to validate proper usage of some of the WSI extensions.  It wants to return valid function pointers for all of the device functions, so that it can issue an error if the app calls the functions without the extension being enabled.  It wants to keep track of whether the device extension is enabled, to aid it in its validation purpose.   Other layers don't need to perform such a check--they only need to do anything w.r.t. to the WSI device functions if the appropriate extension is enabled.

Just seems like another angle to consider.

jon-lunarg:
Regarding this:
It is also true that this pull request makes GIPA more compliant (by handling device commands correctly) and GDPA less compliant (by not returning NULL for un-enabled extensions).

I think you could change the patch a little  so the GDPA could still return NULL for un-enabled extensions.  And at the same time GIPA could handle the device commands.  It would just mean a little less sharing of code I believe.

Then I think we could merge this sort of change of in if GDPA wasn't regressing.

olvaffe:
@ianelliottus Calling `vkGetDeviceProcAddr(device, "vkCreateSwapchainKHR")` without enabling `VK_KHR_swapchain` is an invalid usage.  A validation layer could validate that.  The spec also states that the call must return `NULL`, and I think that should be honored as well.  Instead of returning a valid function pointer that checks whether the extension is enabled, the validation layer should log an error via debug_report and return `NULL` when its GDPA is called.

olvaffe:
@jon-lunarg Yes, I can fix.

layers/core_validation.cpp
lentinem: So now, if I call GIPA or GDPA when the extension is not enabled I get back a function pointer to the function defined here. IMO it should instead pass through and return the next layer or driver (loader)'s wsi function.

vk-layer-generate.py
lentinem: But this won't intercept VENDOR anyway as this only knows about the wsi functions.

layers/core_validation.cpp
olvaffe: A layer intercepting vkCreateSwapchainKHR does not know if VK_KHR_swapchain is going to be enabled when its GIPA is called to query the function.  It cannot just pass through and call down the chain.

It works for the desktop loader because the desktop loader guarantees it never queries device functions with GIPA.  The desktop loader makes that guarantee through dynamic dispatches and JSON files, which we do not support.

vk-layer-generate.py
olvaffe: It does not need to.  This change essentially says, if GDPA intercepts a function, GIPA should too.  If a vendor function is unknown to a layer, it cannot be intercepted in the layer's GDPA.  Then it does not need to be intercepted in the layer's GIPA.

jon-lunarg:
Regarding this  GDPA behavior for swapchain, I agree with Olv's statement cited below as this seems a better approach.  Although since the Android and desktop loader's export the  WSI swapchain extension entrypoints, the swapchain also needs to check for the extension being enabled within the entrypoints themselves (eg CreateSwapchainKHR).

Calling vkGetDeviceProcAddr(device, "vkCreateSwapchainKHR") without enabling VK_KHR_swapchain is an invalid usage. A validation layer could validate that. The spec also states that the call must return NULL, and I think that should be honored as well. Instead of returning a valid function pointer that checks whether the extension is enabled, the validation layer should log an error via debug_report and return NULL when its GDPA is called.

olvaffe:
Closing this and I will update layers one by one.  Starting with https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/443

ianelliottus:
I just noticed that I was included in this discussion, that I hadn't been following (mail filters are difficult for github/gitlab) ...

> Regarding this GDPA behavior for swapchain, I agree with Olv's statement cited below as this seems a better approach. Although since the Android and desktop loader's export the WSI swapchain extension entrypoints, the swapchain also needs to check for the extension being enabled within the entrypoints themselves (eg CreateSwapchainKHR).

This seems fine with me.

PR #205, opened 2016-03-28
layers: GH195 Fix core_validation dynamic state checks

tobine:
Overhauled validation of dynamic state based on latest details in the
VkPipelineDynamicStateCreateInfo section of the spec.
Because some of the state checks need to be predicated on different pipeline
state, removed the previous flag-based system for check predication and
made it more clear and explicit exactly which pipeline states were gating
each dynamic state check with "if" clauses in validate_draw_state_flags().

The biggest change here is for blend constants, which was the focus of the
referenced GH195. For this case, added blendConstantsEnabled bool to
PIPELINE_NODE and enable it by loop through all attachments to identify if
any have blendEnable set and use the blend constant in any of their
blendFactors.

Also updated all the tests that were affected by this change.


mark-lunarg:
Tobin, this looks good.  Passes Dota2 on Windows, and I also ran it through our build server so it's clean there.  Shall I push it?

tobine:
If you have time today, please push it, otherwise I'll push it tomorrow morning. Thanks!

mark-lunarg:
Pushed to Master.

PR #206, opened 2016-03-29
Strip array level from both per-vertex TCS output and TES input

### LINKED ISSUE #120: SC layer: Tessellation evaluation shader input interface mismatch errors
This reports errors similar to those reported in issue #106.

A tess control shader with output defined as 
```
layout(vertices=4) out;
layout(location = 0) out vec3 ctrlPosition[];
patch out float ctrlTess;
```
matched with a tess eval shader with input defined as
```
layout(location = 0) in vec3 ctrlPosition[];
patch in float ctrlTess;
layout(quads, equal_spacing, ccw) in;
```
produces the following errors:

> [SC] Code 2: Type mismatch on location 0.0: 'ptr to output arr[4] of vec3 of float32' vs 'ptr to input arr[32] of vec3 of float32'

(repeats for each of location 0.0, 1.0, 2.0, 3.0)

followed by
> [SC] Code 4: tessellation evaluation shader consumes input location 4.0 which is not written by tessellation control shader

(repeats for each location 4.0 - 31.0)

The tess eval shader explicitly consumes only indices [0]-[3] of ctrlPosition.  It seems like there should ideally be some awareness of the tess control shader's `layout(vertices=N) out;` setting when validating the interfaces between these stages.

chrisforbes:
The array sizes declared needn't match -- and often don't. Also, track whether the `patch` decoration has been applied to the decoration, and require it to match across this interface.

Fixes #120.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -1363,6 +1363,20 @@ struct interface_var {
     /* TODO: collect the name, too? Isn't required to be present. */
 };
 
+struct shader_stage_attributes {
+    char const *const name;
+    bool arrayed_input;
+    bool arrayed_output;
+};
+
+static shader_stage_attributes shader_stage_attribs[] = {
+    {"vertex shader", false, false},
+    {"tessellation control shader", true, true},
+    {"tessellation evaluation shader", true, false},
+    {"geometry shader", true, false},
+    {"fragment shader", false, false},
+};
+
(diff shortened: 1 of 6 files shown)


chrisforbes:
Feel free to review, but don't merge this quite yet. I have tests that I need to clean up and add.

tobine:
So far so good. Assign back to me when you're done and I'll complete review and run some tests prior to merge.

chrisforbes:
Should be good to go. There was a fair bit of test framework unmangling required to make this testable.

tobine:

PR #207, opened 2016-03-29
layers: LX446, Allow GetDeviceQueue to be called multiple times

mark-lunarg:
Calling this API more than once on the same queue hoses the fence
validation.

Change-Id: I524347fd11d3b30a66004cf62937d0446e70df22


mark-lunarg:
For review.

layers/core_validation.cpp
```diff
@@ -5609,18 +5609,23 @@ VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkGetFenceStatus(VkDevice device,
...
+                                                            VkQueue *pQueue) {
     layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
     dev_data->device_dispatch_table->GetDeviceQueue(device, queueFamilyIndex, queueIndex, pQueue);
     loader_platform_thread_lock_mutex(&globalLock);
-    dev_data->queues.push_back(*pQueue);
-    QUEUE_NODE *pQNode = &dev_data->queueMap[*pQueue];
-    pQNode->device = device;
+
+    // Add queue to tracking list if it is new
+    if (find(dev_data->queues.begin(), dev_data->queues.end(), *pQueue) == dev_data->queues.end()) {
```
tobine: If we're using find would like to see std::find for clarity

tobine:
Looking at the code I don't see any valid reason that "queues" in layer_data is a "vector<VkQueue>" vs. just an unordered_set<VkQueue>. I propose changing it to an unordered_set which would simplify the code change here as you could do the "if" check on "if (queues.count() == 0)" instead of using the "find" function.

PR #211, opened 2016-03-30
Adding template for gradle wrapper build

ggfan:
Windows bat file is not updated yet (android-generated.bat), ndk-build and mac build works.


build-android/gradle-templates/.gradle/2.10/taskArtifacts/cache.properties
```diff
@@ -0,0 +1 @@
+#Tue Mar 29 11:04:00 PDT 2016
```
courtney-g: Do we need this file?

build-android/gradle-templates/common.gradle
```diff
@@ -0,0 +1,72 @@
...
+def ndkDir = properties.getProperty('ndk.dir')
+
+model {
+    android {
+        compileSdkVersion = 23
+        buildToolsVersion = "23.0.2"
+
+        defaultConfig.with {
+            minSdkVersion.apiLevel = 22
+            targetSdkVersion.apiLevel = 24
```
courtney-g: Is it possible to define multiple configurations that a user can select?
I.e. default needs to be apiLevel: N-Preview
And that will pull in the Vulkan headers and system library.
But for folks developing on Marshmellow on the handful of devices that support Vulkan, we'll want to use our wrapper library and Vulkan header since apiLevel: 23 does not have them.

build-android/gradle-templates/local.properties
```diff
@@ -0,0 +1,12 @@
...
+#
+# This file must *NOT* be checked into Version Control Systems,
+# as it contains information specific to your local configuration.
+#
+# Location of the SDK. This is only used by Gradle.
+# For customization when using a Version Control System, please read the
+# header note.
+#Mon Mar 28 09:29:17 PDT 2016
+sdk.dir=/Users/gfan/dev/sdk_current
+ndk.dir=/Users/gfan/dev/ndk_current
```
courtney-g: Do we need a new line here?

courtney-g: Do we need this file at all?
It contains references for a specific person, not very useful as a default. Will Android Studio figure this out on it's own.

lentinem:
Many of these files are auto-generated and for now I would prefer if those do not get committed so we are consistent. Fetching them/creating them via the generate script is more consistent with everything else. Also, why are we renaming the directory? It makes it harder to diff and would be nice to have that in a separate change if needed.

build-android/jni/Android.mk
```diff
@@ -16,23 +16,26 @@
...
 MY_PATH := $(LOCAL_PATH)
 SRC_DIR := $(LOCAL_PATH)/../../
+LAYER_DIR := $(LOCAL_PATH)/../generated
 
 include $(CLEAR_VARS)
 LOCAL_MODULE := layer_utils
-LOCAL_SRC_FILES += $(SRC_DIR)/layers/vk_layer_config.cpp
-LOCAL_SRC_FILES += $(SRC_DIR)/layers/vk_layer_extension_utils.cpp
-LOCAL_SRC_FILES += $(SRC_DIR)/layers/vk_layer_utils.cpp
+LOCAL_SRC_FILES += $(LAYER_DIR)/common/vk_layer_config.cpp
```
lentinem: Why are we copying files around? This seems kind of wasted work.

build-android/jni/Android.mk
```diff
@@ -105,13 +110,13 @@ include $(BUILD_SHARED_LIBRARY)
...
-LOCAL_SRC_FILES += $(MY_PATH)/../generated/vk_safe_struct.cpp
-LOCAL_SRC_FILES += $(SRC_DIR)/layers/vk_layer_table.cpp
+LOCAL_SRC_FILES += $(LAYER_DIR)/layer-src/unique_objects/unique_objects.cpp
+LOCAL_SRC_FILES += $(LAYER_DIR)/layer-src/unique_objects/vk_safe_struct.cpp
+LOCAL_SRC_FILES += $(LAYER_DIR)/common/vk_layer_table.cpp
 LOCAL_C_INCLUDES += $(SRC_DIR)/include \
                     $(SRC_DIR)/layers \
-		    $(MY_PATH)/../generated \
-		    $(SRC_DIR)/loader
+		                $(MY_PATH)/../generated/include \
```
lentinem: Indentation?

courtney-g:
BTW, this should be applied to the master branch. I'll merge it from there into the android_layers branch.

build-android/gradle-templates/.gradle/2.10/taskArtifacts/cache.properties
ggfan: need to delete it, thanks

build-android/gradle-templates/common.gradle
ggfan: yes, we would make "android-N" as default. I will fix them all next checkin

build-android/gradle-templates/local.properties
ggfan: this file will not be checked in at all in the end.  I will make sure it is deleted

build-android/jni/Android.mk
ggfan: since android studio could only handle files at the dir level, so only file related to that project need to be sit in that source directory.  so we copy files into separate directory for each proj, so people could browse code inside android studio.  it is old story.

Since android studio is using the its own source file tree, might just keep ndk-build to point to the same source tree.

the structure name change was required by Courtney, not needed anymore? confirm it I am happy to make it back

build-android/jni/Android.mk
ggfan: yes, will do it next checkin

cnorthrop:
Branch is working for me both with ndk-build and gradlew (after I modified local.properties to point to my SDK/NDK).

jon-lunarg:
What is the status of this Pull Request?

cnorthrop:
These changes made it in via PR #294.

Issue #215, opened 2016-03-30
CoreValidation: vkCmdBindDescriptorSets() has an invalid per-DS check


DominikWitczakAMD:
In core_validation.cpp, vkCmdBindDescriptorSets() issues the following check in a loop which iterates over all defined descriptor sets:

![image](https://cloud.githubusercontent.com/assets/17761066/14142934/d9d29412-f688-11e5-8abf-2f5175827826.png)

The block I marked with red color seems to be a copy-paste of the green block. The check on its own definitely makes sense, but only if the red block is removed.

PR coming in a bit..

tobine:
Dominik fixed this with #217 
Sorry I hadn't seen this issue when I first closed that PR.

PR #216, opened 2016-03-30
#215: Removes an invalid check in vkCmdBindDescriptorSets()

DominikWitczakAMD:


tobine:
The comment on that check is a quote from spec section 13.2.5 Descriptor Set Binding:

    dynamicOffsetCount must equal the total number of dynamic descriptors in the sets being bound.

It's the last sentence of the 5th paragraph.
Therefore the check itself should be valid. If you think there's an error in how the check is being executed, that's a separate issue, so please provide details in new issue if that's the case.

DominikWitczakAMD:
That check *is* valid, however its current location is not. If you look closely in the diff, you'll notice the same check outside the loop has been left intact.

It doesn't make any sense to compare the number of dynamic descriptors defined for set N with the total number of dynamic descriptors passed by the user, if the user had specified more than one descriptor set.

tobine:
Thanks, Dominik, you're right. All I saw in diff was killed check so a comment in the PR would have helped.

I missed this repeated code when I was fixing a rebase conflict.
There is similar repeated code in the same place for storing dynamicOffsets. I'll clean that up as well.

DominikWitczakAMD:
Sounds good. Thanks, Michael!

tobine:
merged to master and created #217 to remove similar duplicate code related to dynamicOffsets

PR #221, opened 2016-03-30
WIP: Fix find layout to search on each aspect bit.

lentinem:


Tony-LunarG:
Looks good to me - fixes the sample issue and still catches problems in CTS

Tony-LunarG:
Do you think this is good to go?  We're about to freeze/branch for another SDK release, and having this issue fixed would be desirable.

lentinem:
It's marked WIP because I haven't had a chance to do any testing but if you have tested it it should be fine. Might be too late now though. Sorry I didn't see this Friday.

Tony-LunarG:
Ok - I committed it to master. Thanks!

PR #222, opened 2016-03-30
layers: LX265 Error if color blend attch count != subpass attch.

MarkY-LunarG:
Spec states that the VkPilelineColorBlendAttachmentState
attachmentCount must be equal to the subpass'
colorAttachmentCount.

Change-Id: I2b281cda2c12f2003d8ee68b326d78a9aaadf573

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -3409,6 +3409,14 @@ static VkBool32 validatePipelineState(layer_data *my_data, const GLOBAL_CB_NODE 
                 VkSampleCountFlagBits subpassNumSamples = (VkSampleCountFlagBits)0;
                 uint32_t i;
 
+                if (pPipeline->cbStateCI.attachmentCount != pSD->colorAttachmentCount) {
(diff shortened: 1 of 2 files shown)


tobine:
Check looks good but the change breaks the NumSamplesMismatch VkLayerValidationTest due to the early return causing that error to not be hit. Please check that out and fix it. Does it need to return early? If not it could store result in skipCall like many other functions.
Otherwise may have to update the test.
Also, since the NumSamplesMismatch test apparently hits this new check, I recommend that you repurpose that test for hitting this case, and add another case in that same repurposed test to also hit the NumSamplesMismatch case. That way we get an extra test. Make sense?
You can refer to a number of tests, such as CopyDescriptorUpdateErrors, for examples on checking multiple error cases in a single test.

MarkY-LunarG:
There were actually multiple tests that showed this problem.  In all cases, there was no blend state attachments added (bad).  The delayed return would have fixed this case, but not the other cases.  I did make your suggested change regarding not returning early because I think we should return as much information as possible. However, I also added the appropriate number of blend attachments to all the existing tests so they wouldn't hit this new error check.  Then I added a new test to catch the case that was triggering my new error.

tobine:
Great update. pushed to master.

PR #223, opened 2016-03-30
layers: LX250 Verify color and depth/stencil are different.

MarkY-LunarG:
Validate that the color and depth/stencil attachments in a
subpass are different.

Change-Id: I0010270e50334cd1dc5b275c29c76265f570b186


MarkY-LunarG:
This is replacing the previous work I had done for LX250.  This simply checks to make sure the same attachment isn't being used as a depth and color render target at the same time.  An attachment can only be used as an input and 1 output at most.

layers/core_validation.cpp
```diff
@@ -9185,6 +9185,17 @@ VkBool32 ValidateDependencies(const layer_data *my_data, const VkRenderPassBegin
             for (auto overlapping_attachment : overlapping_attachments[attachment]) {
                 output_attachment_to_subpass[overlapping_attachment].push_back(i);
             }
+
+            for (uint32_t j = 0; j < subpass.colorAttachmentCount; ++j) {
```
tobine: Since we already loop through color attachments above I'd like to avoid loop-within a loop here. One alternative would be to have declare a local "unordered_set<> color_att" above that you insert the color attachments into when you process them, and then check if (color_att.count(attachment)) in the depth stencil case.

tobine:
One fix-up recommended and please rebase as Michael just updated this code.

layers/core_validation.cpp
MarkY-LunarG: Isn't that still going to be a linear search worse case.  Perhaps I should just move the check inside the color loop and that way we only loop once total.    Of course, we have the extra overhead of one more compare instruction "if (depth...attachment != UNUSED) {}"  Not that it's going to be a huge impact either way.  The most color surfaces I've seen output to at once were 6 (for a render to cubemap scenario).

MarkY-LunarG:
Made your suggested changes, rebased, and it pulled in a lot of items into this.  Not sure how to clean it up without restarting a new one.  If that's what you want, let me know.

PR #224, opened 2016-03-30
layers: Don't stash pointer to caller memory in bound function

### LINKED ISSUE #212: CoreValidation: vkCmdWaitEvents() caches a pointer to an array which is likely to become invalid by the time validation occurs
In core_validation.cpp::vkCmdWaitEvents(), there's an interesting piece of code:

image

This code assumes **pEvents** will continue to be valid infinitely after vkCmdWaitEvents() call leaves, which is not necessarily true. In my case, I'm passing an array, whose data comes from stack memory of the main thread, so the core_validation.cpp::vkCmdWaitEvents() never crashes. However, the data that function later tries to access is noise, which results in an incorrect debug call-back occuring at queue submission time.

Also, the debug message is invalid (the reported pointer is obviously not a fence handle):

image

chrisforbes:
We were already accumulating all the events the CB touches -- so just
pass an index into that.

Fixes #212

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -8716,24 +8716,25 @@ VkBool32 ValidateBarriers(const char *funcName, VkCommandBuffer cmdBuffer, uint3
     return skip_call;
 }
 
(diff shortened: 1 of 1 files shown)


lentinem:
+1

chrisforbes:
Added a handful of other fixes.

tobine:
merged to master

PR #225, opened 2016-03-31
WIP: Mem ref cleanup

chrisforbes:
Gets rid of some crazy std::list usage in the GLOBAL_CB_NODE & DEVICE_MEM_INFO structs.

Ignore the first 4 commits -- they are in another PR and just provided a good starting point for this work.

### DIFF
--- a/layers/core_validation.h
+++ b/layers/core_validation.h
@@ -891,7 +891,7 @@ struct GLOBAL_CB_NODE {
     std::unordered_set<VkCommandBuffer> secondaryCommandBuffers;
     // MTMTODO : Scrub these data fields and merge active sets w/ lastBound as appropriate
     vector<std::function<VkBool32()>> validate_functions;
-    list<VkDeviceMemory> pMemObjList; // List container of Mem objs referenced by this CB
+    std::unordered_set<VkDeviceMemory> memObjs;
     vector<std::function<bool(VkQueue)>> eventUpdates;
 };
 
--- a/layers/core_validation.h
+++ b/layers/core_validation.h
@@ -123,7 +123,6 @@ struct MEMORY_RANGE {
 // Data struct for tracking memory object
 struct DEVICE_MEM_INFO {
     void *object;      // Dispatchable object used to create this memory (device of swapchain)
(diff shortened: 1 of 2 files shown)


layers/core_validation.h
```diff
@@ -123,12 +139,11 @@ struct MEMORY_RANGE {
...
 struct DEVICE_MEM_INFO {
     void *object;      // Dispatchable object used to create this memory (device of swapchain)
-    uint32_t refCount; // Count of references (obj bindings or CB use)
     bool valid;        // Stores if the memory has valid data or not
     VkDeviceMemory mem;
     VkMemoryAllocateInfo allocInfo;
-    list<MT_OBJ_HANDLE_TYPE> pObjBindings;        // list container of objects bound to this memory
-    list<VkCommandBuffer> pCommandBufferBindings; // list container of cmd buffers that reference this mem object
+    unordered_set<MT_OBJ_HANDLE_TYPE> objBindings; // objects bound to this memory
+    unordered_set<VkCommandBuffer> commandBufferBindings; // cmd buffers referencing this memory
```
lentinem: It seems like set is a win if there are a lot of draw calls otherwise list might be better. I would be curious what larger apps do in practice but I'm ok with set as I'm inclined to believe the number of draw calls can get large for various apps.

lentinem:
+1

tobine:
With these changes when I run tri or cube w/ --validate, I get a list of 3 errors when I close the app. Here's one of the 3, the other 2 errors just have different mem obj addresses:

    ERROR: [MEM] Code 9 : While trying to clear mem binding for buffer obj 0x1728450, unable to find that object referenced by mem obj 0x1728670

Please check those out and fix accordingly.

chrisforbes:
@tobine I'd accidentally inverted the condition that produces that error. Fixed.

tobine:
Did you push updates for tri --validate to work? I'm still seeing the same 3 error messages when it shuts down with the latest code.

Can you check that out again?
Also, can you do me another favor and if/when you think this is ok to merge, update the title of this PR or just create a new one that no longer has "WIP" in the title.

tobine:
merged to master

PR #227, opened 2016-03-31
loader: vk{Create|Destroy}Instance can have multiple tmp callbacks

ianelliottus:
During code review of the Android version of the temporary debug_report
callbacks code, it was decided to allow an array of
VkDebugReportCallbackCreateInfoEXT structs to be passed to vkCreateInstance().
This code implements that, using some new utility functions in order to help
keep the code clean.

### DIFF
--- a/loader/debug_report.c
+++ b/loader/debug_report.c
@@ -168,7 +168,7 @@ util_CopyDebugReportCreateInfos(const void *pChain,
                                 VkDebugReportCallbackCreateInfoEXT **infos,
                                 VkDebugReportCallbackEXT **callbacks)
 {
-    *num_callbacks = 0;
+    uint32_t n = *num_callbacks = 0;
 
     // NOTE: The loader is not using pAllocator, and so this function doesn't
     // either.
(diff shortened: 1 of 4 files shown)


loader/debug_report.c
```diff
@@ -156,6 +156,113 @@ void util_DestroyDebugReportCallback(struct loader_instance *inst,
...
+
+    // NOTE: The loader is not using pAllocator, and so this function doesn't
+    // either.
+
+    const void *pNext = pChain;
+    while (pNext) {
+        // 1st, count the number VkDebugReportCallbackCreateInfoEXT:
+        if (((VkDebugReportCallbackCreateInfoEXT *)pNext)->sType ==
+            VK_STRUCTURE_TYPE_DEBUG_REPORT_CREATE_INFO_EXT) {
+            (*num_callbacks)++;
```
courtney-g: We should count in a local variable and only update num_callbacks when everything is done. We don't want to change anything if we hit some error.

loader/debug_report.c
```diff
@@ -156,6 +156,113 @@ void util_DestroyDebugReportCallback(struct loader_instance *inst,
...
+        }
+        pNext = (void *)((VkDebugReportCallbackCreateInfoEXT *)pNext)->pNext;
+    }
+    if (*num_callbacks) {
```
courtney-g: Can return early here if no callbacks. Saves an indentation level. That's 124 bytes smaller source :-)

loader/loader.h
```diff
@@ -293,7 +293,9 @@ struct loader_instance {
 
     bool debug_report_enabled;
     VkLayerDbgFunctionNode *DbgFunctionHead;
-    VkDebugReportCallbackCreateInfoEXT debugReportCreateInfo;
+    uint32_t num_tmp_callbacks;
+    VkDebugReportCallbackCreateInfoEXT *tmp_dbg_create_infos;
```
courtney-g: Need to be sure to initialize these pointers to NULL

loader/trampoline.c
```diff
@@ -473,7 +505,10 @@ vkDestroyInstance(VkInstance instance,
     if (ptr_instance->phys_devs)
         loader_heap_free(ptr_instance, ptr_instance->phys_devs);
     if (callback_setup) {
-        util_DestroyDebugReportCallback(ptr_instance, instance_callback, NULL);
+        util_DestroyDebugReportCallbacks(ptr_instance,
+                                         pAllocator,
+                                         ptr_instance->num_tmp_callbacks,
+                                         ptr_instance->tmp_callbacks);
```
courtney-g: util_FreeDebucReportCallbacks here?

loader/loader.h
ianelliottus: Yes, I saw this.  However, it bit me in Android, where the struct wasn't initialized.  I think I'll do it here as well.

ianelliottus:
I incorporated all of your feedback.

courtney-g:
LGTM

jon-lunarg:
Thanks Ian.  cherry picked squashed commit to master

PR #228, opened 2016-03-31
[For Review Only] LX458 + CmdCopyImage Valid Usage

mark-lunarg:


layers/image.cpp
```diff
@@ -649,66 +653,223 @@ vkCmdClearDepthStencilImage(VkCommandBuffer commandBuffer, VkImage image, VkImag
...
-               VkImageLayout dstImageLayout, uint32_t regionCount, const VkImageCopy *pRegions) {
-    VkBool32 skipCall = VK_FALSE;
-    layer_data *device_data = get_my_data_ptr(get_dispatch_key(commandBuffer), layer_data_map);
-    auto srcImageEntry = device_data->imageMap.find(srcImage);
-    auto dstImageEntry = device_data->imageMap.find(dstImage);
+// Returns true if [x, xoffset] and [y, yoffset] overlap
+static bool ranges_intersect(int32_t start, uint32_t start_offset, int32_t end, uint32_t end_offset) {
+    bool result = false;
+    int32_t intersection_min = std::max(start, end);
+    int32_t intersection_max = std::min(start + static_cast<int32_t>(start_offset), end + static_cast<int32_t>(end_offset));
```
tobine: Do we know that the offset will always safely cast to int32_t?
Otherwise if offset is huge seems like casting behavior is implementation dependent according to a quick search.

I know offsets w/ highest bit set are unlikely, but if it's not restricted by spec, need to consider that case and make sure we behave correctly.

layers/image.cpp
```diff
@@ -649,66 +653,223 @@ vkCmdClearDepthStencilImage(VkCommandBuffer commandBuffer, VkImage image, VkImag
...
-                log_msg(device_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
-                        (uint64_t)commandBuffer, __LINE__, IMAGE_INVALID_IMAGE_ASPECT, "IMAGE", str);
+    // Intentionally fall through to 1D case
+    case VK_IMAGE_TYPE_1D: // Validate x and Width
+        if ((offset->x + extent->width > image->extent.width) || (offset->x < 0) ||
+            ((offset->x + static_cast<int32_t>(extent->width)) < 0)) {
+            result = true;
         }
+        break;
+    default:
```
tobine: Would be nice to send error message here in the event that there is a future extension that adds a new image type that we don't handle.

layers/image.cpp
```diff
@@ -649,66 +653,223 @@ vkCmdClearDepthStencilImage(VkCommandBuffer commandBuffer, VkImage image, VkImag
...
+            }
+
+            // For each region the layerCount member of srcSubresource and dstSubresource must match
+            if (pRegions[i].srcSubresource.layerCount != pRegions[i].dstSubresource.layerCount) {
+                std::stringstream ss;
+                ss << "vkCmdCopyImage: number of layers in source and destination subresources for pRegions[" << i
+                   << "] do not match";
+                skipCall |=
+                    log_msg(device_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
+                            (uint64_t)commandBuffer, __LINE__, IMAGE_INVALID_EXTENTS, "IMAGE", "%s", ss.str().c_str());
```
tobine: All of the c-style casts (12 total) should be changed to reinterpret_casts.

tobine:
LTGM, just a few nit-picks that I noted

layers/image.cpp
mark-lunarg: Hmmm.  If the extent is negative, an error would have already been thrown in the exceeds_bounds() checks.  I could instead cast the ints to uints, which might give bad results, but the results are already undefined due to the previous error.

layers/image.cpp
mark-lunarg: Parameter_validation would have already caught such an issue.  I'll add a comment saying as much.

mark-lunarg:
Updated and pushed to master.

layers/image.cpp
tobine: I was thinking of a case where upper-most bit of unit32_t is "1", such as start_offset 0x80000000. Is there a spec and/or device limit that would prevent such an offset? If not, then the way that it's cast to an int is implementation-dependent.

I see that scissor test, for example has restriction:

    Evaluation of (offset.x + extent.width) must not cause a signed integer addition overflow

I don't see such a restriction on CmdCopyImage values, but there are restrictions on extents in relation to actual image size. I just wanted to flag this in case there's a hole here if we see such huge offsets in extents.

mark-lunarg: I switched the casts from the uints to the ints.  If they were negative an error would already would have been output.  If the values were so large (in either case) we'd overflow as you point out.  There's nothing in the spec as there is for the scissor test -- this seems like an oversight.

PR #229, opened 2016-03-31
Don't validate memory if used for both input and output.

lentinem:


tobine:
Change looks good, any further context for educational purposes? Seems like if its an input attachment, it would still have "first_read" of "true" even if used as an output, so why not validate memory in this case?

lentinem:
The only case this changes is when an attachment is both input and output. It still sets first_read to true if its an input and validates memory is valid but not if it's also an output. In that case we assume it doesn't need to be validated as we can't add the logic necessary to try to find out if it's actually being written or read first in that case.

tobine:
Got it, thanks. Looks like #230 includes and improves upon this change. Both look good to me so I'll push shortly.

tobine:
merged to master

PR #231, opened 2016-03-31
Fix loader_unwrap_physical_device() to cast to correct type

pdaniell-nv:
The trampoline loader_unwrap_physical_device() utility function was
casting the VkPhysicalDevice parameter to (loader_physical_device*)
instead of (loader_physical_device_tramp*). It worked previously because
it just so happened that the phys_dev member was in the same location in
both structs.


jon-lunarg:
thanks, cherry-picked to master

PR #232, opened 2016-03-31
Trampoline vkEnumeratePhysicalDevices should use layer-returned physical device count

pdaniell-nv:
Modify the trampoline vkEnumeratePhysicalDevices() implementation to
enumerate the number of physical devices based on what the top-most
layer reports and not what the terminator_EnumeratePhysicalDevices
counted. This allows intermediate layers to modify the physical device
count.


jon-lunarg:
thanks, cherry picked to master

Issue #233, opened 2016-03-31
vkEnumeratePhysicalDevices trampoline<->terminator dependency


pdaniell-nv:
There is a dependency between the trampoline vkEnumeratePhysicalDevices implementation in the loader and the terminator_EnumeratePhysicalDevices. The dependency is that the trampoline code assumes that the pPhysicalDeviceCount and pPhysicalDevices outputs from the top-most layer match up with those output from the terminator_EnumeratePhysicalDevices. It uses this assumption in order to figure out the "this_icd" pointer associated with each VkPhysicalDevice handle it wraps.

This dependency breaks the layering model and causes the loader to fail to operate correctly if a layer manipulates the physical device list between vkEnumeratePhysicalDevices and terminator_EnumeratePhysicalDevices. In theory the trampoline code should not need to know which "icd" belongs to each VkPhysicalDevice. It's up to layers to wrap and unwrap VkPhysicalDevice handles and do things like inject their own or filter out physical devices if they want.

I have a pending MR to modify the loader so the trampoline code no longer needs to know about the "this_icd" pointer of the VkPhysicalDevice handle.

PR #234, opened 2016-03-31
Fix vkEnumeratePhysicalDevices trampoline/terminator dependency

pdaniell-nv:
Fixes issue #233 

There is a dependency between the trampoline vkEnumeratePhysicalDevices implementation in the loader and the terminator_EnumeratePhysicalDevices. The dependency is that the trampoline code assumes that the pPhysicalDeviceCount and pPhysicalDevices outputs from the top-most layer match up with those output from the terminator_EnumeratePhysicalDevices. It uses this assumption in order to figure out the "this_icd" pointer associated with each VkPhysicalDevice handle it wraps.

This dependency breaks the layering model and causes the loader to fail to operate correctly if a layer manipulates the physical device list between vkEnumeratePhysicalDevices and terminator_EnumeratePhysicalDevices. In theory the trampoline code should not need to know which "icd" belongs to each VkPhysicalDevice. It's up to layers to wrap and unwrap VkPhysicalDevice handles and do things like inject their own or filter out physical devices if they want.

This MR modifies the trampoline wrapped VkPhysicalDevice handle struct loader_physical_device_tramp to no longer need the "struct loader_icd *this_icd" pointer passed up from the terminator function. This removes the dependency and allows layers to mess with the physical device list.

### DIFF
--- a/loader/loader.c
+++ b/loader/loader.c
@@ -636,7 +636,7 @@ loader_init_device_extensions(const struct loader_instance *inst,
 }
 
 VkResult loader_add_device_extensions(const struct loader_instance *inst,
-                                      struct loader_icd *icd,
+                                      PFN_vkEnumerateDeviceExtensionProperties fpEnumerateDeviceExtensionProperties,
                                       VkPhysicalDevice physical_device,
                                       const char *lib_name,
                                       struct loader_extension_list *ext_list) {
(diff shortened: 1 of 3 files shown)


jon-lunarg:
Thanks, cherry picked to master branch.

PR #235, opened 2016-03-31
Enable gradle build

ggfan:
Added gradle build for layers:
create one gradle project for each layer
  all layers templated from the same build.gradle
with the change, application could add dependencies in its
build.gradle file, 
android.sources { 
  main {
     jniLibs {
        dependencies {
            project ":/dir/to/core_validation"
            project ":/dir/to/device_limits"
            ...other layers...
        }
    } //jniLibs
 } //main
} //android.sources
layers will be automatically build and packaged into apk files


cnorthrop:
These changes got cherry-picked in via PR #294.  Thanks!

PR #236, opened 2016-04-01
layers: GH226 Add check for descriptor slot to make sure it's updated

tobine:
We have an existing check to make sure that a set has been updated, but there was
still a hole where individual slots may not have an update. This adds an additional
check to make sure that we flag any slots that are used by not updated at draw time.


lentinem:
+1

tobine:

PR #237, opened 2016-04-01
layers: LX414: Fix device_limits alignment check

dustin-lunarg:
Fix buffer/descriptor offset alignment checks in device_limits:
 - vkUpdateDescriptorSets offset check was not performed with
   the correct physical device alignments.
 - vkBindBufferMemory offset check was always performed with the
   UBO alignment.  This check has been moved to core_validation,
   where the offset is checked against the alignment associated
   with its usage type and the alignment retrieved from
   vkGetBufferMemoryRequirements.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5954,6 +5954,48 @@ vkBindBufferMemory(VkDevice device, VkBuffer buffer, VkDeviceMemory mem, VkDevic
         skipCall |= validate_buffer_image_aliasing(dev_data, buffer_handle, mem, memoryOffset, memRequirements,
                                                    dev_data->memObjMap[mem].bufferRanges, dev_data->memObjMap[mem].imageRanges,
                                                    VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT);
+        // Validate memory requirements alignment
+        if (vk_safe_modulo(memoryOffset, memRequirements.alignment) != 0) {
+            skipCall |=
+                log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_PHYSICAL_DEVICE_EXT, 0,
+                        __LINE__, DRAWSTATE_INVALID_BUFFER_MEMORY_OFFSET, "DS",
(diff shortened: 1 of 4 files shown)


layers/core_validation.cpp
```diff
@@ -6022,6 +6022,58 @@ vkBindBufferMemory(VkDevice device, VkBuffer buffer, VkDeviceMemory mem, VkDevic
...
+                skipCall |=
+                    log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_PHYSICAL_DEVICE_EXT,
+                            0, __LINE__, DRAWSTATE_INVALID_UNIFORM_TEXEL_BUFFER_OFFSET, "DS",
+                            "vkBindBufferMemory(): memoryOffset is %#" PRIxLEAST64 " but must be a multiple of "
+                            "device limit minTexelBufferOffsetAlignment %#" PRIxLEAST64,
+                            memoryOffset, dev_data->physDevProperties.properties.limits.minTexelBufferOffsetAlignment);
+            }
+        }
+        if (usage & VK_BUFFER_USAGE_STORAGE_TEXEL_BUFFER_BIT) {
+            if (vk_safe_modulo(memoryOffset, dev_data->physDevProperties.properties.limits.minTexelBufferOffsetAlignment) != 0) {
```
tobine: The checks for UNIFORM & STORAGE texel buffers appear exactly the same. Can we merge them into a single error enum case and change the top level "if" check to:

    if (usage & (VK_BUFFER_USAGE_UNIFORM_TEXEL_BUFFER_BIT | VK_BUFFER_USAGE_STORAGE_TEXEL_BUFFER_BIT)) {

If that makes sense it will save about 10 lines of code which is a big win in my book.

dustin-lunarg: The only difference between the two texel buffer checks is the DRAWSTATE_INVALID_UNIFORM_TEXEL_BUFFER_OFFSET vs DRAWSTATE_INVALID_STORAGE_TEXEL_BUFFER_OFFSET error code.  Do you think those should be combined into a single DRAWSTATE_INVALID_TEXEL_BUFFER_OFFSET case?  Seems like that would make sense.

dustin-lunarg: Done

tobine:
merged to master

Issue #238, opened 2016-04-01
CoreValidation: "Num samples mismatch" error reported for subpasses which do not use any color/DS attachments


DominikWitczakAMD:
Consider a GFX pipeline, whose renderpass defines a subpass without any color & DS attachments. In my case, I'm using image store ops fired from within a FS to produce output.

For this scenario, GFX pipeline has the VkPipelineMultisampleStateCreateInfo::rasterizationSamples set to SAMPLE_COUNT_1_BIT, since that's the granularity I need. However, at vkCmdBindPipeline() call time, I'm getting the following error debug call-back:

![image](https://cloud.githubusercontent.com/assets/17761066/14205461/944774e2-f80b-11e5-92f1-637fc9ff2cc2.png)

![image](https://cloud.githubusercontent.com/assets/17761066/14205473/ad6ab18c-f80b-11e5-9166-c31c2f26b751.png)

If I'm correct, the check should be skipped for subpass which do not have any attachments assigned.

tobine:
Merged your #239 that fixes this issue so closing. Thanks!

PR #239, opened 2016-04-01
#238: Fixes the "num samples mismatch" error being shown incorrectly …

DominikWitczakAMD:
…for subpasses with 0 attachments


tobine:
merged to master

PR #244, opened 2016-04-01
layers: Clean up PIPELINE_NODE using safe_struct

tobine:
PIPELINE_NODE was storing ptrs from the app and has some ugly initialization code
for the gfx pipeline create struct. This commit switches pipeline node to use
safe_struct wrappers for Gfx and Compute create info.
There were also some bugs found in the safe_struct wrappers making this migration
so those are cleaned up in this commit as well.


layers/core_validation.cpp
```diff
@@ -2723,7 +2724,8 @@ static VkBool32 validate_pipeline_shader_stage(layer_data *dev_data,
 // Validate that the shaders used by the given pipeline and store the active_slots
 //  that are actually used by the pipeline into pPipeline->active_slots
 static VkBool32 validate_and_capture_pipeline_shader_state(layer_data *my_data, PIPELINE_NODE *pPipeline) {
-    VkGraphicsPipelineCreateInfo const *pCreateInfo = &pPipeline->graphicsPipelineCI;
+    VkGraphicsPipelineCreateInfo const *pCreateInfo =
```
chrisforbes: `auto` and lose the cast

layers/core_validation.cpp
```diff
@@ -2737,7 +2739,8 @@ static VkBool32 validate_and_capture_pipeline_shader_state(layer_data *my_data,
     auto pipelineLayout = pCreateInfo->layout != VK_NULL_HANDLE ? &my_data->pipelineLayoutMap[pCreateInfo->layout] : nullptr;
 
     for (uint32_t i = 0; i < pCreateInfo->stageCount; i++) {
-        VkPipelineShaderStageCreateInfo const *pStage = &pCreateInfo->pStages[i];
+        VkPipelineShaderStageCreateInfo const *pStage =
```
chrisforbes: ditto, although you'll also need to change the type taken by `validate_pipeline_shader_stage` to be your wrapped version.

layers/core_validation.cpp
```diff
@@ -2784,7 +2787,8 @@ static VkBool32 validate_and_capture_pipeline_shader_state(layer_data *my_data,
 }
 
 static VkBool32 validate_compute_pipeline(layer_data *my_data, PIPELINE_NODE *pPipeline) {
-    VkComputePipelineCreateInfo const *pCreateInfo = &pPipeline->computePipelineCI;
+    VkComputePipelineCreateInfo const *pCreateInfo =
```
chrisforbes: ditto

chrisforbes:
My bigger question it - do we actually need to keep the CI around here? We've got various other copies of bits of it, in more helpful forms.

layers/core_validation.cpp
```diff
@@ -3187,26 +3191,36 @@ static VkBool32 verifyPipelineCreateState(layer_data *my_data, const VkDevice de
...
+    if (pPipeline->graphicsPipelineCI.pInputAssemblyState &&
+        pPipeline->graphicsPipelineCI.pInputAssemblyState->topology == VK_PRIMITIVE_TOPOLOGY_PATCH_LIST) {
         if (~pPipeline->active_shaders & VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT) {
             skipCall |= log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
                                 DRAWSTATE_INVALID_PIPELINE_CREATE_STATE, "DS", "Invalid Pipeline CreateInfo State: "
                                                                                "VK_PRIMITIVE_TOPOLOGY_PATCH_LIST primitive "
                                                                                "topology is only valid for tessellation pipelines");
         }
-        if (!pPipeline->tessStateCI.patchControlPoints || (pPipeline->tessStateCI.patchControlPoints > 32)) {
+        if (!pPipeline->graphicsPipelineCI.pTessellationState) {
```
lentinem: Why is this here? Seems like not having state should be ok in this case.

layers/core_validation.cpp
```diff
@@ -3410,12 +3308,24 @@ static VkBool32 validatePipelineState(layer_data *my_data, const GLOBAL_CB_NODE
...
                 uint32_t i;
 
-                if (pPipeline->cbStateCI.attachmentCount != pSD->colorAttachmentCount) {
-                    skipCall |= log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_PIPELINE_EXT,
-                                        reinterpret_cast<const uint64_t &>(pipeline), __LINE__, DRAWSTATE_INVALID_RENDERPASS, "DS",
-                                        "Mismatch between blend state attachment count %u and subpass %u color attachment "
-                                        "count %u!  These must be the same.",
-                                        pPipeline->cbStateCI.attachmentCount, pCB->activeSubpass, pSD->colorAttachmentCount);
+                if (!pPipeline->graphicsPipelineCI.pColorBlendState) {
+                    skipCall |= log_msg(
```
lentinem: Ditto. Also the message would be wrong.

layers/core_validation.cpp
tobine: If topology is VK_PRIMITIVE_TOPOLOGY_PATCH_LIST tessellation shader stages must be enabled which means that pTessellationState must not be NULL. Relevant spec quotes:

    If pStages includes tessellation shader stages, the topology member of pInputAssembly must be VK_PRIMITIVE_TOPOLOGY_PATCH_LIST

    If the tessellation shaders feature is not enabled, topology must not be VK_PRIMITIVE_TOPOLOGY_PATCH_LIST

    If pStages includes tessellation shader stages, the topology member of pInputAssembly must be VK_PRIMITIVE_TOPOLOGY_PATCH_LIST

lentinem: Ah missed the indent; should still change the message though.

layers/core_validation.cpp
tobine: Yes, this is the check that's just wrong in general. Just filed #256 for @MarkY-LunarG to look into and I'm going to kill this existing, incorrect check from this PR.

layers/core_validation.cpp
tobine: done

layers/core_validation.cpp
tobine: done

layers/core_validation.cpp
tobine: I don't like this one as it requires changing 2 function signatures. A benefit of the safe_* structs is that they can exactly cast to the types they wrap so I prefer to just keep the cast here and not change the function signatures.

layers/core_validation.cpp
tobine: The message for this case is customized. It notes that no pTessellationState is provided and says that one needs to be provided w/ patchControlPoints value >0 and <=32.

tobine:
@chrisforbes Are you referring to storing graphicsPipelineCI? That's used like 70 times in the code so not easy to get rid of.

layers/core_validation.cpp
lentinem: But this change has nothing to do with checking that patchControlPoints is >0 and <=32. IMO this is confusing for the user as I would go and check to see why my points were invalid. I would probably realize at this point that I didn't have points and fix it but what we are checking is that tessellation state exists. I realize that technically this does say that no state exists but could be less confusing.

tobine:
Updated based on comments, rebased to master and force pushed update

layers/core_validation.cpp
tobine: updated and force pushed

chrisforbes:
LGTM

tobine:

Issue #100, opened 2016-03-10
DrawState layer: "Cannot delete semaphore X which is in use" error incorrectly reported


DominikWitczakAMD:
In principle, this issue builds upon #99.

The two command buffers we submit with a single vkQueueSubmit() call are synchronized as per my last comment in #67. tl;dr: the first cmd buffer signals a semaphore A, and the second cmd buffer waits on it before executing. We then submit both command buffers, wait on a fence passed with the vkQueueSubmit() call and proceed with releasing the objects.

At this point, when I call vkDeleteSemaphore(), I'm getting a misleading debug call-back informing me the semaphore is still in use. Here's the obligatory call stack:

![image](https://cloud.githubusercontent.com/assets/17761066/13666257/e2cd6fac-e6b0-11e5-9cf3-4a157d3be6a6.png)

lentinem:
Will try to find something obvious but will likely need a more solid repro for this.

lentinem:
This is likely fixed by 110 as well.

DominikWitczakAMD:
Fixed, closing.

DominikWitczakAMD:
Managed to reproduce the error in a different scenario:

1. Create a semaphore A.
2. Submit a command buffer with the semaphore used as a signal semaphore.
3. Submit another command buffer, with the semaphore used this time as a wait semaphore.
4. Destroy the semaphore.

This results in the same "Cannot delete semaphore X which is in use" error being reported.

DominikWitczakAMD:
Potentially related to #134

lentinem:
Were these the same queues or different queues? If they were the same brief testing did not repro. If they were different then it should be the same issue as #99.

DominikWitczakAMD:
The issue continues to reproduce, I'm afraid. Just to ensure we're talking about the same scenario, here's how I'm getting that call-back:

1. Submit a command buffer 1 to a compute queue family. Use semaphore A as a signal semaphore here.
2. Submit a command buffer 2 to a universal queue family. Use the semaphore A as a wait signal in this step (along with a relevant dst stage mask). Also use a fence X here.
3. Wait for fence X to be set.
4. Destroy command buffer 1

lentinem:
Just to clarify; are you still using draw state or have you switched to core_validation?

DominikWitczakAMD:
Switched to core_validation. I always sync to the latest master version before trying to reproduce any of the reported issues.

[...]

PR #245, opened 2016-04-01
Issue 99, 100, LX459: layers: LX459,GH#99,#100, Fix semaphore reference count

mark-lunarg:
QueueSubmit waitSemaphore refcounts were getting incremented instead of
decremented, resulting in invalid 'semaphore still in use' errors.

Change-Id: I8ac224115b8ee43637b8de4b377750b277cfd22b


mark-lunarg:
Pushed to Master.

PR #252, opened 2016-04-04
layers: Fix 24 char truncation warnings on VS2015.

MichaelMcDonnell:
Casts the values to signed char to silence 24 warning C4309 on
VS2015. The warning happens because the values are larger than
what can be represented in a signed char, but the expected
overflow works just fine with VS2015.

### DIFF
--- a/layers/vk_layer_utils.cpp
+++ b/layers/vk_layer_utils.cpp
@@ -568,14 +568,14 @@ VkDeviceSize vk_safe_modulo(VkDeviceSize dividend, VkDeviceSize divisor) {
     return result;
 }
 
-static const char UTF8_ONE_BYTE_CODE = 0xC0;
-static const char UTF8_ONE_BYTE_MASK = 0xE0;
-static const char UTF8_TWO_BYTE_CODE = 0xE0;
-static const char UTF8_TWO_BYTE_MASK = 0xF0;
-static const char UTF8_THREE_BYTE_CODE = 0xF0;
-static const char UTF8_THREE_BYTE_MASK = 0xF8;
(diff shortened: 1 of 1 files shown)


layers/vk_layer_utils.cpp
```diff
@@ -568,14 +568,14 @@ VkDeviceSize vk_safe_modulo(VkDeviceSize dividend, VkDeviceSize divisor) {
...
 
-static const char UTF8_ONE_BYTE_CODE = 0xC0;
-static const char UTF8_ONE_BYTE_MASK = 0xE0;
-static const char UTF8_TWO_BYTE_CODE = 0xE0;
-static const char UTF8_TWO_BYTE_MASK = 0xF0;
-static const char UTF8_THREE_BYTE_CODE = 0xF0;
-static const char UTF8_THREE_BYTE_MASK = 0xF8;
-static const char UTF8_DATA_BYTE_CODE = 0x80;
-static const char UTF8_DATA_BYTE_MASK = 0xC0;
+static const char UTF8_ONE_BYTE_CODE = (char)0xC0;
```
dustin-lunarg: Although you'll find C-style casts in the existing code, we're trying to limit new code to the appropriate C++ casts.  Can you change this to use static_cast?

MichaelMcDonnell:
Sure I can change it to C++ static_casts. I'm away from my computer until Friday.

dustin-lunarg:
I changed the type declarations from char to uint8_t to resolve this.  Thanks for the submissions to cleanup the warnings!  With VS2013, we're now down to one warning from glslang and one debug link warning.

MichaelMcDonnell:
Ok sounds great! I hate those noisy warnings :-)

PR #253, opened 2016-04-04
tests: Initialize fields so test does not crash.

MichaelMcDonnell:
I had forgotten to define the LD_LIBRARY_PATH and VK_LAYER_PATH
environment variables in VS2015 before running the tests. The
test crashed when it tried to delete the un-initialized pointers.
I have added a default initialization to VK_NULL_HANDLE so you
get an error message instead of a crash.


dustin-lunarg:
Cherry picked and pushed to master with the following edits:
* Moved inst and m_device initialization to start of initialization list.  The original order was an issue for the Android (clang) builds due to -Werror,-Wreorder.
* Changed VkDeviceObj* m_device initialization value to NULL.
* Added explicit initialization values for m_commandPool and m_commandBuffer to your commit.
* Ran: git clang-format --style=file

The original author info and commit message were preserved, with the addition of a Gerrit Change-Id for our internal review/build system.  Thanks!

MichaelMcDonnell:
Thank you for the detailed explanation. I learned something new :-)

PR #254, opened 2016-04-04
Fix def file generation to depend on vulkan.py.

null77:
I noticed this when writing alternate build scripts for the vulkan layers. Perhaps vk.py was the old name, but I only see a vulkan.py now.


tobine:

PR #255, opened 2016-04-04
Allow vk-generate.py to work in Python 2.

null77:
There were two places where this script called super, one with
Python 2 and 3 compatible syntax and one with Python 3 only. Fix
it to use the compatible syntax in both locations.

This is pretty important for ANGLE since Google tools only use
Python 2, and will be for the forseeable future.


tobine:

PR #262, opened 2016-04-05
Fix 3 MSVS warnings in loader and layers.

null77:
warning C4389: signed/unsigned mismatch
warning C4701: potentially uninitialized local variable used
warning C4702: unreachable code


layers/core_validation.cpp
```diff
@@ -1499,7 +1499,7 @@ static void collect_interface_by_location(layer_data *my_data, shader_module con
         assert(insn != src->end());
         assert(insn.opcode() == spv::OpVariable);
 
-        if (insn.word(3) == sinterface) {
+        if (insn.word(3) == (uint32_t)sinterface) {
```
lentinem: Please use the appropriate c++ style cast instead of c style casts.

layers/core_validation.cpp
```diff
@@ -8978,7 +8977,7 @@ VkBool32 CheckDependencyExists(const layer_data *my_data, const int subpass, con
     VkBool32 result = VK_TRUE;
     // Loop through all subpasses that share the same attachment and make sure a dependency exists
     for (uint32_t k = 0; k < dependent_subpasses.size(); ++k) {
-        if (subpass == dependent_subpasses[k])
+        if ((uint32_t)subpass == dependent_subpasses[k])
```
lentinem: Ditto.

null77:
Fixed to static_cast, uploaded new commit. Is there a style guide for the loader or layers?

Also I wasn't able to get clang-format to work correctly, the diff I got seemed to change the formatting in a number of places that weren't the lines I had changed.

lentinem:
Thanks. Unfortunately there isn't a unified style guide at this point outside of running clang-format using the file in the repo. I've been an advocate of https://google.github.io/styleguide/cppguide.html with exceptions for things like line length which we agreed would be longer and tab width which is 4 but a good chunk of the layers don't follow this yet.

As far as clang format goes you should be able to run git clang-format --style=file which should pickup the .clang-format file but I have admittedly not used windows and thus haven't tried it there.

null77:
Naturally I'm a fan of the google guide as well. This modified .clang-format is what ANGLE does as well, it might be that the file is either out of date or my clang-format binary is different from the one you use internally. I'll let you guys handle the formatting side, thanks!

dustin-lunarg:
Cherry picked to master and pushed.

PR #263, opened 2016-04-05
layers: Change include order of loader platform.

null77:
This fixes a bug where on Windows, windows.h would include
winsock.h before winsock2.h, causing numerous warnings.


courtney-g:
+1

dustin-lunarg:
Cherry picked and pushed to master.

PR #264, opened 2016-04-05
layers: Fix return values of bool functions.

null77:
These functions were returning VK_ERROR_VALIDATION_FAILED_EXT,
which was giving a warning in Visual Studio. The app shouldn't be
checking for non-bool return values of these functions.


tobine:
:+1:

dustin-lunarg:
Cherry picked to master and pushed

PR #265, opened 2016-04-05
layers: Fix VkBool32/bool unsafe mix warnings

dustin-lunarg:
Fixes the MSVS warnings regarding unsafe mix of VkBool32 and bool for the
validation layers.  Converts all Boolean values internal to the layer
to bool/true/false.  The VK_TRUE and VK_FALSE values continue to be used
with the Vulkan API components.


tobine:

PR #266, opened 2016-04-05
Validate attachment index is in range.

lentinem:


layers/core_validation.cpp
```diff
@@ -9379,8 +9379,25 @@ VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkCreateRenderPass(VkDevice devic
...
+                if (attachment >= pCreateInfo->attachmentCount) {
+                    skip_call |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0,
+                                         __LINE__, DRAWSTATE_INVALID_RENDERPASS, "DS",
+                                         "Preserve attachment %d cannot be greater than the total number of attachments %d.",
+                                         attachment, pCreateInfo->attachmentCount);
+                }
+            }
             for (uint32_t j = 0; j < subpass.colorAttachmentCount; ++j) {
-                uint32_t attachment = subpass.pColorAttachments[j].attachment;
+                uint32_t attachment = subpass.pResolveAttachments[j].attachment;
```
tobine: Need null ptr check here for pResolveAttachments

layers/core_validation.cpp
```diff
@@ -9379,8 +9379,25 @@ VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkCreateRenderPass(VkDevice devic
             dev_data->renderPassMap[*pRenderPass]->attachment_first_layout;
         for (uint32_t i = 0; i < pCreateInfo->subpassCount; ++i) {
             const VkSubpassDescription &subpass = pCreateInfo->pSubpasses[i];
+            for (uint32_t j = 0; j < subpass.preserveAttachmentCount; ++j) {
+                uint32_t attachment = subpass.pPreserveAttachments[j];
+                if (attachment >= pCreateInfo->attachmentCount) {
```
tobine: This would be a good place to add in this additional valid usage check:

    The value of any element of the pPreserveAttachments member in any given element of pSubpasses must not be VK_ATTACHMENT_UNUSED

layers/core_validation.cpp
```diff
@@ -9379,8 +9379,25 @@ VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkCreateRenderPass(VkDevice devic
...
+                                         __LINE__, DRAWSTATE_INVALID_RENDERPASS, "DS",
+                                         "Preserve attachment %d cannot be greater than the total number of attachments %d.",
+                                         attachment, pCreateInfo->attachmentCount);
+                }
+            }
             for (uint32_t j = 0; j < subpass.colorAttachmentCount; ++j) {
-                uint32_t attachment = subpass.pColorAttachments[j].attachment;
+                uint32_t attachment = subpass.pResolveAttachments[j].attachment;
+                if (attachment >= pCreateInfo->attachmentCount && attachment != VK_ATTACHMENT_UNUSED) {
+                    skip_call |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0,
```
tobine: This is weird. Seems like you hit a GitHub diff bug. I'm seeing a bunch of blank space where the "skip_call = log_msg..." txt should be. I pulled the CL and the code is there so not sure what's up with this diff. Can anyone else confirm they're seeing the same behavior in the GitHub online diff?

layers/core_validation.cpp
lentinem: This covers that because ATTACHMENT_UNUSED will always be greater than the count.

layers/core_validation.cpp
chrisforbes: Looks broken online to me too. Github wasn't having a good day today, this may be an artifact of that.

tobine:

PR #268, opened 2016-04-05
layers: Merge binding maps into image and buffer maps

tobine:
mem_tracker layer had specific maps to track memory binding for images
and buffers. This change merges the data from those maps into IMAGE_NODE
and BUFFER_NODE respectively and eliminates the binding maps.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -37,19 +37,19 @@
 // Turn on mem_tracker merged code
 #define MTMERGESOURCE 1
 
+#include <SPIRV/spirv.hpp>
+#include <algorithm>
+#include <assert.h>
+#include <iostream>
+#include <list>
+#include <map>
+#include <set>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <assert.h>
+#include <string>
 #include <unordered_map>
 #include <unordered_set>
-#include <map>
-#include <string>
-#include <iostream>
-#include <algorithm>
-#include <list>
-#include <SPIRV/spirv.hpp>
-#include <set>
 
 #include "vk_loader_platform.h"
 #include "vk_dispatch_table_helper.h"
@@ -105,16 +105,9 @@ struct layer_data {
     std::vector<VkDebugReportCallbackEXT> logging_callback;
     VkLayerDispatchTable *device_dispatch_table;
     VkLayerInstanceDispatchTable *instance_dispatch_table;
-#if MTMERGESOURCE
-// MTMERGESOURCE - stuff pulled directly from MT
-    uint64_t currentFenceId;
-    // Maps for tracking key structs related to mem_tracker state
-    // Images and Buffers are 2 objects that can have memory bound to them so they get special treatment
-    unordered_map<uint64_t, MT_OBJ_BINDING_INFO> imageBindingMap;
-    unordered_map<uint64_t, MT_OBJ_BINDING_INFO> bufferBindingMap;
-// MTMERGESOURCE - End of MT stuff
-#endif
+
     devExts device_extensions;
+    uint64_t currentFenceId;
     unordered_set<VkQueue> queues;  // all queues under given device
     // Global set of all cmdBuffers that are inFlight on this device
     unordered_set<VkCommandBuffer> globalInFlightCmdBuffers;
@@ -152,11 +145,8 @@ struct layer_data {
     PHYS_DEV_PROPERTIES_NODE physDevProperties;
 // MTMERGESOURCE - added a couple of fields to constructor initializer
     layer_data()
-        : report_data(nullptr), device_dispatch_table(nullptr), instance_dispatch_table(nullptr),
-#if MTMERGESOURCE
-        currentFenceId(1),
-#endif
-        device_extensions(){};
+        : report_data(nullptr), device_dispatch_table(nullptr), instance_dispatch_table(nullptr), device_extensions(),
+          currentFenceId(1){};
 };
 
 static const VkLayerProperties cv_global_layers[] = {{
@@ -260,25 +250,24 @@ static void clear_cmd_buf_and_mem_references(layer_data *my_data, const VkComman
 
 #define MAX_BINDING 0xFFFFFFFF
 
-static MT_OBJ_BINDING_INFO *get_object_binding_info(layer_data *my_data, uint64_t handle, VkDebugReportObjectTypeEXT type) {
-    MT_OBJ_BINDING_INFO *retValue = NULL;
+static VkDeviceMemory *get_object_mem_binding(layer_data *my_data, uint64_t handle, VkDebugReportObjectTypeEXT type) {
     switch (type) {
     case VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT: {
-        auto it = my_data->imageBindingMap.find(handle);
-        if (it != my_data->imageBindingMap.end())
-            return &(*it).second;
+        auto it = my_data->imageMap.find(VkImage(handle));
+        if (it != my_data->imageMap.end())
+            return &(*it).second.mem;
         break;
     }
     case VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT: {
-        auto it = my_data->bufferBindingMap.find(handle);
-        if (it != my_data->bufferBindingMap.end())
-            return &(*it).second;
+        auto it = my_data->bufferMap.find(VkBuffer(handle));
+        if (it != my_data->bufferMap.end())
+            return &(*it).second.mem;
         break;
     }
     default:
         break;
     }
-    return retValue;
+    return nullptr;
 }
 // MTMERGESOURCE - end section
 #endif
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -445,12 +383,12 @@ static VkBool32 validate_usage_flags(layer_data *my_data, void *disp_obj, VkFlag
...
 // Pulls image info and then sends actual vs. desired usage off to helper above where
 //  an error will be flagged if usage is not correct
-static VkBool32 validate_image_usage_flags(layer_data *my_data, void *disp_obj, VkImage image, VkFlags desired, VkBool32 strict,
+static VkBool32 validate_image_usage_flags(layer_data *dev_data, void *disp_obj, VkImage image, VkFlags desired, VkBool32 strict,
                                            char const *func_name, char const *usage_string) {
     VkBool32 skipCall = VK_FALSE;
-    MT_OBJ_BINDING_INFO *pBindInfo = get_object_binding_info(my_data, (uint64_t)image, VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT);
-    if (pBindInfo) {
-        skipCall = validate_usage_flags(my_data, disp_obj, pBindInfo->create_info.image.usage, desired, strict, (uint64_t)image,
+    const auto &image_node = dev_data->imageMap.find(image);
```
chrisforbes: You don't want the ref here. This type is `unordered_map<VkImage, IMAGE_NODE>::iterator`. Just `auto` (or `auto const`).

layers/core_validation.cpp
```diff
@@ -37,19 +37,20 @@
 // Turn on mem_tracker merged code
 #define MTMERGESOURCE 1
 
+#include <SPIRV/spirv.hpp>
+#include <algorithm>
+#include <assert.h>
+#include <bits/unordered_map.h>
```
chrisforbes: Why are you including this? This is an internal library header, and including it here breaks the build.

layers/core_validation.cpp
```diff
@@ -459,12 +397,12 @@ static VkBool32 validate_image_usage_flags(layer_data *my_data, void *disp_obj,
...
 // Pulls buffer info and then sends actual vs. desired usage off to helper above where
 //  an error will be flagged if usage is not correct
-static VkBool32 validate_buffer_usage_flags(layer_data *my_data, void *disp_obj, VkBuffer buffer, VkFlags desired, VkBool32 strict,
+static VkBool32 validate_buffer_usage_flags(layer_data *dev_data, void *disp_obj, VkBuffer buffer, VkFlags desired, VkBool32 strict,
                                             char const *func_name, char const *usage_string) {
     VkBool32 skipCall = VK_FALSE;
-    MT_OBJ_BINDING_INFO *pBindInfo = get_object_binding_info(my_data, (uint64_t)buffer, VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT);
-    if (pBindInfo) {
-        skipCall = validate_usage_flags(my_data, disp_obj, pBindInfo->create_info.buffer.usage, desired, strict, (uint64_t)buffer,
+    const auto &buffer_node = dev_data->bufferMap.find(buffer);
```
chrisforbes: Ditto

layers/core_validation.cpp
```diff
@@ -10497,23 +10405,18 @@ vkGetSwapchainImagesKHR(VkDevice device, VkSwapchainKHR swapchain, uint32_t *pCo
             IMAGE_LAYOUT_NODE image_layout_node;
             image_layout_node.layout = VK_IMAGE_LAYOUT_UNDEFINED;
             image_layout_node.format = swapchain_node->createInfo.imageFormat;
-            dev_data->imageMap[pSwapchainImages[i]].createInfo.mipLevels = 1;
-            dev_data->imageMap[pSwapchainImages[i]].createInfo.arrayLayers = swapchain_node->createInfo.imageArrayLayers;
+            auto &image_node = dev_data->imageMap[pSwapchainImages[i]];
```
chrisforbes: :+1: nice.

layers/core_validation.h
```diff
@@ -564,7 +555,8 @@ class IMAGE_CMD_BUF_LAYOUT_NODE {
 class BUFFER_NODE : public BASE_NODE {
   public:
     using BASE_NODE::in_use;
-    unique_ptr<VkBufferCreateInfo> create_info;
```
chrisforbes: This seems OK; container's mapped type is `BUFFER_NODE` rather than ptr.. -- `unordered_map` provides fairly incredible guarantees about validity of references even across `rehash`, so there's no internal moving going on; possibly able to narrow lock window by holding `BUFFER_NODE` by `unique_ptr` though, and `move` into the map. Out of scope for this, though.

layers/core_validation.h
```diff
@@ -542,6 +532,7 @@ class IMAGE_NODE : public BASE_NODE {
   public:
     VkImageCreateInfo createInfo;
     VkDeviceMemory mem;
+    bool valid; // If this is a swapchain image backing memory track valid here as it doesn't have DEVICE_MEM_INFO
```
chrisforbes: Odd that this is dropped in the middle of the mem-related fields, but OK.

layers/core_validation.cpp
```diff
@@ -8641,8 +8550,9 @@ VkBool32 ValidateBarriers(const char *funcName, VkCommandBuffer cmdBuffer, uint3
         }
 
         auto buffer_data = dev_data->bufferMap.find(mem_barrier->buffer);
-        uint64_t buffer_size =
-            buffer_data->second.create_info ? reinterpret_cast<uint64_t &>(buffer_data->second.create_info->size) : 0;
+        uint64_t buffer_size = (buffer_data->second.createInfo.sType == VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO)
```
chrisforbes: Not new, but a size being `uint64_t` rather than `size_t` suggests not quite right.

layers/core_validation.cpp
```diff
@@ -260,19 +251,19 @@ static void clear_cmd_buf_and_mem_references(layer_data *my_data, const VkComman
 
 #define MAX_BINDING 0xFFFFFFFF
 
-static MT_OBJ_BINDING_INFO *get_object_binding_info(layer_data *my_data, uint64_t handle, VkDebugReportObjectTypeEXT type) {
-    MT_OBJ_BINDING_INFO *retValue = NULL;
+static VkDeviceMemory *get_object_mem_binding(layer_data *my_data, uint64_t handle, VkDebugReportObjectTypeEXT type) {
+    VkDeviceMemory *retValue = nullptr;
```
chrisforbes: Not new, but this is weird code. There's one use of `retValue` at the end of the function; would be clearer to drop this and just return `nullptr` there.

layers/core_validation.cpp
```diff
@@ -301,59 +292,6 @@ static void delete_cmd_buf_info(layer_data *my_data, VkCommandPool commandPool,
...
-    // Swap Chain is very unique, use my_data->imageBindingMap, but copy in
-    // SwapChainCreatInfo's usage flags and set the mem value to a unique key. These is used by
-    // vkCreateImageView and internal mem_tracker routines to distinguish swap chain images
-    case VK_DEBUG_REPORT_OBJECT_TYPE_SWAPCHAIN_KHR_EXT: {
-        auto pCI = &my_data->imageBindingMap[handle];
-        memset(pCI, 0, sizeof(MT_OBJ_BINDING_INFO));
-        pCI->mem = MEMTRACKER_SWAP_CHAIN_IMAGE_KEY;
-        pCI->valid = false;
-        pCI->create_info.image.usage =
-            const_cast<VkSwapchainCreateInfoKHR *>(static_cast<const VkSwapchainCreateInfoKHR *>(pCreateInfo))->imageUsage;
```
chrisforbes: Good riddance to /that/...

layers/core_validation.cpp
```diff
@@ -152,11 +146,8 @@ struct layer_data {
     PHYS_DEV_PROPERTIES_NODE physDevProperties;
 // MTMERGESOURCE - added a couple of fields to constructor initializer
     layer_data()
-        : report_data(nullptr), device_dispatch_table(nullptr), instance_dispatch_table(nullptr),
-#if MTMERGESOURCE
-        currentFenceId(1),
-#endif
-        device_extensions(){};
+        : report_data(nullptr), device_dispatch_table(nullptr), instance_dispatch_table(nullptr), device_extensions(),
```
lentinem: why do you have device_extensions() here?

layers/core_validation.cpp
```diff
@@ -152,11 +146,8 @@ struct layer_data {
...
+          currentFenceId(1){};
```
lentinem: I would prefer a space here: "(1) {}". Strange that clang doesn't require that.

layers/core_validation.cpp
lentinem: +1

layers/core_validation.cpp
lentinem: Why? IMO it doesn't really matter here.

layers/core_validation.cpp
```diff
@@ -803,18 +739,19 @@ static VkBool32 set_sparse_mem_binding(layer_data *dev_data, void *dispObject, V
     if (mem == VK_NULL_HANDLE) {
         skipCall = clear_object_binding(dev_data, dispObject, handle, type);
     } else {
-        MT_OBJ_BINDING_INFO *pObjBindInfo = get_object_binding_info(dev_data, handle, type);
-        if (!pObjBindInfo) {
+        VkDeviceMemory *pMemBinding = get_object_mem_binding(dev_data, handle, type);
```
lentinem: IMO we are way too pointer happy. It's out of the scope of this change but it's worth looking at when we really need pointers and when references/copies of handles will suffice.

layers/core_validation.cpp
```diff
@@ -5914,7 +5829,7 @@ vkBindBufferMemory(VkDevice device, VkBuffer buffer, VkDeviceMemory mem, VkDevic
     uint64_t buffer_handle = (uint64_t)(buffer);
     VkBool32 skipCall =
         set_mem_binding(dev_data, device, mem, buffer_handle, VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT, "vkBindBufferMemory");
-    add_object_binding_info(dev_data, buffer_handle, VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT, mem);
+    dev_data->bufferMap[buffer].mem = mem;
```
lentinem: Would be good to check for existence in the map. We should not be creating one if buffer is not valid.

layers/core_validation.cpp
lentinem: This should really be VkDeviceSize.

layers/core_validation.cpp
```diff
@@ -10237,33 +10147,33 @@ VKAPI_ATTR VkResult VKAPI_CALL vkBindImageMemory(VkDevice device, VkImage image,
...
-    // Track objects tied to memory
-    uint64_t image_handle = (uint64_t)(image);
-    skipCall =
-        set_mem_binding(dev_data, device, mem, image_handle, VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT, "vkBindImageMemory");
-    add_object_binding_info(dev_data, image_handle, VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT, mem);
-    {
+    const auto &image_node = dev_data->imageMap.find(image);
+    if (image_node != dev_data->imageMap.end()) {
+        // Track objects tied to memory
+        uint64_t image_handle = (uint64_t)(image);
```
lentinem: c++ cast

layers/core_validation.cpp
```diff
@@ -10237,33 +10147,33 @@ VKAPI_ATTR VkResult VKAPI_CALL vkBindImageMemory(VkDevice device, VkImage image,
...
-        dev_data->imageMap[image].mem = mem;
-        dev_data->imageMap[image].memOffset = memoryOffset;
-        dev_data->imageMap[image].memSize = memRequirements.size;
+        print_mem_list(dev_data, device);
         loader_platform_thread_unlock_mutex(&globalLock);
+        if (VK_FALSE == skipCall) {
+            result = dev_data->device_dispatch_table->BindImageMemory(device, image, mem, memoryOffset);
+            loader_platform_thread_lock_mutex(&globalLock);
+            dev_data->memObjMap[mem].image = image;
+            image_node->second.memOffset = memoryOffset;
```
lentinem: Kind of odd that mem is set on image_node always but memOffset and memSize only if the call succeeds. I would move the setting of mem to here.

layers/core_validation.cpp
tobine: Bizzarre... I have no idea how that include got in there.

layers/core_validation.cpp
tobine: Yeah this code sucks returning the ptr into the struct. I've updated the return as suggested. Still need to find a way to kill or improve it further, though.

layers/core_validation.cpp
tobine: Default initializer for device_extensions struct. Same code as before, just moved its location relative to currentFenceId.

layers/core_validation.cpp
tobine: will do

layers/core_validation.cpp
tobine: :+1:

layers/core_validation.cpp
tobine: yes. I had hoisted it up thinking there was a dependency prior to calling down the chain but double-checking I see that i was incorrect. Moving mem assignment down.

layers/core_validation.cpp
tobine: I tried to add the space and clang-format kills is

tobine:

PR #271, opened 2016-04-05
Fix regression with dEQP-VK.api.object_management.alloc_callback_fail.device.

pdaniell-nv:
When the loader vkCreateDevice trampoline code calls down to
vkCreateDevice and it fails don't update the dev->device field with a bogus device
handle.

### DIFF
--- a/loader/loader.c
+++ b/loader/loader.c
@@ -3559,7 +3559,9 @@ VkResult loader_create_device_chain(const struct loader_physical_device_tramp *p
         loader_create_info.pNext = &create_info_disp;
         res = fpCreateDevice(pd->phys_dev, &loader_create_info, pAllocator,
                              &created_device);
-        dev->device = created_device;
+        if (res == VK_SUCCESS) {
+            dev->device = created_device;
+        }
     } else {
         // Couldn't find CreateDevice function!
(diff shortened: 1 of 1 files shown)


loader/loader.c
```diff
@@ -3559,7 +3559,9 @@ VkResult loader_create_device_chain(const struct loader_physical_device_tramp *p
         loader_create_info.pNext = &create_info_disp;
         res = fpCreateDevice(pd->phys_dev, &loader_create_info, pAllocator,
                              &created_device);
-        dev->device = created_device;
+        if (res == VK_SUCCESS) {
```
courtney-g: This should be:
 if (res != VK_SUCCESS) {
   return res;
}
dev->device = created_device;

I don't think we want to call loader_init_device_dispatch_table if the CreateDevice failed.

pdaniell-nv: Before the regression was introduced in 1d41bc502102f0ae112c775489b3bd2b1d7996ff it used to call the loader_init_device_dispatch_table() function anyway. But we can change it.

jon-lunarg:
Squashed the two commits and pushed to master branch; this included Courtney's suggested change to the commit to return early and sklp the init_dispatch.  Thanks for the fix Piers.

PR #274, opened 2016-04-06
Fix loader bad behavior scribbling through ptr to const caller memory

chrisforbes:
See commit messages for detailed explanation

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -90,6 +90,11 @@ endif()
 add_executable(vulkaninfo vulkaninfo.c)
 target_link_libraries(vulkaninfo ${LIBRARIES})
 
+if (NOT WIN32)
+    add_executable(const-lies const-lies.cc)
+    target_link_libraries(const-lies ${LIBRARIES})
+endif()
+
 if(UNIX)
     add_executable(tri tri.c ${CMAKE_BINARY_DIR}/demos/tri-vert.spv ${CMAKE_BINARY_DIR}/demos/tri-frag.spv)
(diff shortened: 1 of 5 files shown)


courtney-g:
Why is a commit with "DO NOT MERGE: demos: linux demo for loader" on here?

jon-lunarg:
Chris:

This all looks good to me for the loader patch except for one thing.  The patch  always puts the expanded layer names at the beginning of the list.  This is bad since it doesn't preserve the order of layers from the app.  For example in your test program do:
auto names = alloc_protectable<char const *>();
    names[0] = "VK_LAYER_LUNARG_basic";
    names[1] = "VK_LAYER_LUNARG_standard_validation";
    names[2] = "VK_LAYER_LUNARG_generic";

Instead of having the expanded order (closest to the app first) be basic, 8 validation layers, generic. Your patch makes the order basic, generic, 8 validation layers

I think with a some minor modifications your patch could be altered to preserve the order during expansion.

chrisforbes:
@jon-lunarg I agree that would be a better thing to do -- I just did what loader_expand_layer_names was specified to do in its description, placing all the expanded names at the end of the list.

@courtney-g The other commit is a toy demo program. I don't consider it production quality- if it were to be merged, it should be a test case somewhere, and should be adapted to work on all platforms.

chrisforbes:
@jon-lunarg Added one more commit which preserves order during expansion.

jon-lunarg:
Thanks Chris for the patches!

Cherry-picked to master branch.

PR #288, opened 2016-04-06
Ensure bind point for subpass is graphics.

lentinem:


tobine:

PR #289, opened 2016-04-06
layers: Change the test about "too many images".

ianelliottus:
Given new spec language coming out in early April (clarifying various aspects
of acquiring presentable images), a change was needed to the WSI-swapchain
validation layer.  The new validate-layer test will work for "BlitModel"
drivers that advertise a minImageCount of 1.  Failing the test now results in
an error instead of a performance warning.

### DIFF
--- a/layers/swapchain.cpp
+++ b/layers/swapchain.cpp
@@ -1614,8 +1614,8 @@ vkDestroySwapchainKHR(VkDevice device, VkSwapchainKHR swapchain, const VkAllocat
     // TODOs:
     //
     // - Implement a check for validity language that reads: All uses of
-    //   presentable images acquired from pname:swapchain and owned by the
-    //   application must: have completed execution
+    //   presentable images acquired from pname:swapchain must: have completed
+    //   execution
     bool skipCall = false;
     layer_data *my_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
     loader_platform_thread_lock_mutex(&globalLock);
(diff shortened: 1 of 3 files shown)


layers/swapchain.cpp
```diff
@@ -1745,31 +1745,36 @@ VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkAcquireNextImageKHR(VkDevice de
...
-        uint32_t imagesOwnedByApp = 0;
+    SwpPhysicalDevice *pPhysicalDevice = pDevice->pPhysicalDevice;
+    if (pSwapchain && pPhysicalDevice && pPhysicalDevice->gotSurfaceCapabilities) {
+        // Look to see if the application has already acquired the maximum
+        // number of images, and this will push it past the spec-defined
+        // limits:
+        uint32_t minImageCount = pPhysicalDevice->surfaceCapabilities.minImageCount;
+        uint32_t imagesAcquiredByApp = 0;
         for (uint32_t i = 0; i < pSwapchain->imageCount; i++) {
             if (pSwapchain->images[i].ownedByApp) {
```
courtney-g: Want to change the structure field as well? ownedByApp => acquiredByApp

courtney-g:
One more owned => acquired change, otherwise looks good.

layers/swapchain.cpp
ianelliottus: Thanks, I will!

ianelliottus:
Cherry-picked and pushed to master

PR #290, opened 2016-04-06
Don't exceed max allowed color attachments and fix get proc addr names.

lentinem:


layers/device_limits.cpp
```diff
@@ -676,11 +699,13 @@ VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vkGetDeviceProcAddr(VkD
         return (PFN_vkVoidFunction)vkDestroyDevice;
     if (!strcmp(funcName, "vkGetDeviceQueue"))
         return (PFN_vkVoidFunction)vkGetDeviceQueue;
-    if (!strcmp(funcName, "CreateCommandPool"))
+    if (!strcmp(funcName, "vkCreateRenderPass"))
+        return (PFN_vkVoidFunction)vkCreateRenderPass;
+    if (!strcmp(funcName, "vkCreateCommandPool"))
```
chrisforbes: This looks like a fix for another issue. At least call it out in the commit message?

layers/device_limits.cpp
```diff
@@ -468,6 +468,29 @@ VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL vkDestroyDevice(VkDevice device, cons
     layer_data_map.erase(key);
 }
 
+VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkCreateRenderPass(VkDevice device, const VkRenderPassCreateInfo *pCreateInfo,
+                                                                  const VkAllocationCallbacks *pAllocator,
+                                                                  VkRenderPass *pRenderPass) {
+    layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
+    bool skip_call = false;
+    if (dev_data) {
```
chrisforbes: Do you need to test for this? Things are /deeply/ wrong if you don't have layer data for the device; elsewhere we don't check.

layers/device_limits.cpp
```diff
@@ -468,6 +468,29 @@ VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL vkDestroyDevice(VkDevice device, cons
...
+                                     reinterpret_cast<uint64_t>(device), __LINE__, DEVLIMITS_INVALID_ATTACHMENT_COUNT, "DL",
+                                     "Cannot create a render pass with %d color attachments. Max is %d.",
+                                     pCreateInfo->pSubpasses[i].colorAttachmentCount, max_color_attachments);
+            }
+        }
+    }
+    if (skip_call) {
+        return VK_ERROR_VALIDATION_FAILED_EXT;
+    }
+    return get_my_data_ptr(get_dispatch_key(device), layer_data_map)
```
chrisforbes: `return dev_data->device_dispatch_table->CreateRenderPass(...`

layers/device_limits.cpp
lentinem: I could remove it; I'm not too picky on this point so whichever is fine with me.

chrisforbes: Well, despite the check you're about to segfault on :490 in this case so I'd just remove it.

layers/device_limits.cpp
```diff
@@ -468,6 +468,28 @@ VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL vkDestroyDevice(VkDevice device, cons
...
 }
 
+VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkCreateRenderPass(VkDevice device, const VkRenderPassCreateInfo *pCreateInfo,
+                                                                  const VkAllocationCallbacks *pAllocator,
+                                                                  VkRenderPass *pRenderPass) {
+    layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
+    bool skip_call = false;
+    if (dev_data) {
+        uint32_t max_color_attachments = dev_data->physicalDeviceProperties.limits.maxColorAttachments;
+        for (uint32_t i = 0; i <= pCreateInfo->subpassCount; ++i) {
```
chrisforbes: `<`. You're reading off the end.

chrisforbes:
:+1: with those issues fixed.

tobine:

PR #293, opened 2016-04-07
layers: GH241 vkSetEvent should affect queue stageMasks

tobine:
vkSetEvent may occur asynchronously from cmd buffer execution
and the results of vkSetEvent should be immediately visible
across all queues of a device.

This fix will update any event that a queue has seen so that
its VK_PIPELINE_STAGE_HOST_BIT is set at vkSetEvent() time.


lentinem:
-1 to prevent checking it in until we are in agreement.

tobine:
cherry-picked and pushed to master and filed #297 for some follow-up work

PR #296, opened 2016-04-07
Lxml removal and consistent source code generation

stroyan:
These patches change from using lxml to python's xml.etree.
That removes a dependency on lxml.
There is also a patch that changes the generated code to have a consistent order.
Before that patch the generated code would vary between builds.
Finally, there is a removal of a unicode header that crept into vk-layer-generate.py.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -88,14 +88,6 @@ Windows 7+ with additional required software packages:
   - Ensure the pip module is installed (it should be by default)
   - Need python3.3 or later to get the Windows py.exe launcher that is used to get python3 rather than python2 if both are installed on Windows
   - 32 bit python works
-- Python lxml package must be installed
-  - Download the lxml package from
-        http://www.lfd.uci.edu/~gohlke/pythonlibs/#lxml
-        32-bit latest for Python 3.5 is: lxml-3.5.0-cp35-none-win32.whl
-        64-bit latest for Python 3.5 is: lxml-3.5.0-cp35-none-win_amd64.whl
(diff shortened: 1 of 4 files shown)


tobine:
LGTM :+1:

dustin-lunarg:
Cherry picked to master and pushed.

null77:
I believe this has a bug on Windows - with VK_STRUCTURE_TYPE_WIN32_SURFACE_CREATE_INFO_KHR. Please see a more robust fix that I've tested with Python 3 and 2 here: https://github.com/null77/Vulkan-LoaderAndValidationLayers/commit/36152c614fb1e480b09b832bf58521c965804467. Sorry was just in the process of testing that code more before I uploaded it - stroyan, I had seen your patch on the mailing list.

stroyan:
The problem you saw with VK_STRUCTURE_TYPE_WIN32_SURFACE_CREATE_INFO_KHR was found and fixed before this merge request was pushed.  I just expanded the patterns that can be mapped to VK_STRUCTURE_ names.  Your comment parsing fix is interesting.  It could be more flexible for future additional structure enum names.  Hopefully a resolution to https://github.com/KhronosGroup/Vulkan-Docs/issues/34 will move that information into new xml fields instead of comments so the parsing will be straight-forward and robust.

null77:
Ah, you're correct, just ran it through my code and compiles fine. Thanks for pointing out that issue, I agree the best fix would be to fix vk.xml. Even if we can do a more robust parser, the best solution is to include the info in the XML.

Thanks for fixing this!

PR #300, opened 2016-04-08
[for review]layers: use std mutex

jeremy-lunarg:
Mark and I would like to move to std::mutex. Is this okay? If so, I'll move all the layers to this method.

### DIFF
--- a/layers/image.cpp
+++ b/layers/image.cpp
@@ -31,15 +31,16 @@
 // Allow use of STL min and max functions in Windows
 #define NOMINMAX
 
+#include <algorithm>
+#include <assert.h>
 #include <inttypes.h>
+#include <memory>
+#include <mutex>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <assert.h>
-#include <vector>
-#include <algorithm>
 #include <unordered_map>
-#include <memory>
+#include <vector>
 using namespace std;
 
 #include "vk_loader_platform.h"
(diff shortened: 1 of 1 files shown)


layers/image.cpp
```diff
@@ -357,18 +349,18 @@ vkCreateImage(VkDevice device, const VkImageCreateInfo *pCreateInfo, const VkAll
         result = device_data->device_dispatch_table->CreateImage(device, pCreateInfo, pAllocator, pImage);
     }
     if (result == VK_SUCCESS) {
-        loader_platform_thread_lock_mutex(&globalLock);
+        std::lock_guard<std::mutex> lock(::mutex);
```
lentinem: Why do you need to global namespace mutex? I'd also prefer using globalLock as the mutex as it's more descriptive although in the interest of variable naming consistency i would vote for using global_lock.

lentinem:
As I've said before, I'm very in favor of doing this in general and would like to see this used everywhere.

layers/image.cpp
jeremy-lunarg: The std namespace is sucked in at the top. Actually, it had a redundant using-directive and I removed it, but I'd be in favor of removing both and going with explicit std scope resolution throughout.

jeremy-lunarg:
Sounds good to me. I'm fine with global_lock. I'll work on updating all the
layers.

layers/image.cpp
lentinem: Ah ok. +1 to removing both and using explicit scope resolution.

tobine:
I agree with previous comments:

1. Kill "using std"
2. rename "mutex"->"global_lock"

Looks good to push with those changes. Just assign back to me if you want me to merge it.

PR #301, opened 2016-04-08
Fix for GH issues 99 and 100.

mark-lunarg:
This corner case occurs because command buffers and semaphores are cleaned up through fence references, but QueueSubmit calls does not require fences so the refs were left hanging.  Dominik, can you take a look?  It passes your app here, but I'm interested in more complicated QueueSubmit patterns you may have access to.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -4864,8 +4864,7 @@ static void decrementResources(layer_data *my_data, VkCommandBuffer cmdBuffer) {
         auto semaphoreNode = my_data->semaphoreMap.find(semaphore);
         if (semaphoreNode != my_data->semaphoreMap.end()) {
             // Each CB in semaphore's list may have reference to this semaphore
-            for (auto cb : semaphoreNode->second.linkedCmdBuffers) {
-                GLOBAL_CB_NODE *refCB = getCBNode(my_data, cb);
+            for (auto refCB : semaphoreNode->second.linkedCmdBuffers) {
                 for (auto semaphoreRef : refCB->semaphores) {
                     auto semaphoreRefNode = my_data->semaphoreMap.find(semaphoreRef);
                     if (semaphoreRefNode != my_data->semaphoreMap.end()) {
@@ -5236,7 +5235,7 @@ vkQueueSubmit(VkQueue queue, uint32_t submitCount, const VkSubmitInfo *pSubmits,
             // For each signal/wait semaphore, add ref to this cmdbuffer's list and vice-versa
             for (auto semaphore : semaphoreList) {
                 pCBNode->semaphores.push_back(semaphore);
-                dev_data->semaphoreMap[semaphore].linkedCmdBuffers.push_back(submit->pCommandBuffers[i]);
+                dev_data->semaphoreMap[semaphore].linkedCmdBuffers.push_back(pCBNode);
             }
             pCBNode->submitCount++;
             skipCall |= validatePrimaryCommandBufferState(dev_data, pCBNode);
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -5406,7 +5420,19 @@ static inline void removeInFlightCmdBuffer(layer_data *dev_data, VkCommandBuffer
             if ((q != queue) &&
                 (dev_data->queueMap[q].inFlightCmdBuffers.find(cmd_buffer) != dev_data->queueMap[q].inFlightCmdBuffers.end())) {
                 dev_data->globalInFlightCmdBuffers.insert(cmd_buffer);
-                break;
+                return;
```
tobine: Can't just early return here. If there were 2+ other queues that had the given cmd buffer inFlight then early return would cause cmd buffer to only be added back for first one.

layers/core_validation.h
```diff
@@ -607,6 +607,8 @@ class SEMAPHORE_NODE : public BASE_NODE {
     uint32_t signaled;
     SemaphoreState state;
     VkQueue queue;
+    // Track command buffers linked by semaphores
+    vector<VkCommandBuffer> linkedCmdBuffers;
```
tobine: any reason to have this as vector vs. unordered_set?

layers/core_validation.cpp
```diff
@@ -5223,8 +5233,12 @@ vkQueueSubmit(VkQueue queue, uint32_t submitCount, const VkSubmitInfo *pSubmits,
         for (uint32_t i = 0; i < submit->commandBufferCount; i++) {
             skipCall |= ValidateCmdBufImageLayouts(submit->pCommandBuffers[i]);
             pCBNode = getCBNode(dev_data, submit->pCommandBuffers[i]);
-            pCBNode->semaphores = semaphoreList;
-            pCBNode->submitCount++; // increment submit count
+            // For each signal/wait semaphore, add ref to this cmdbuffer's list and vice-versa
+            for (auto semaphore : semaphoreList) {
+                pCBNode->semaphores.push_back(semaphore);
+                dev_data->semaphoreMap[semaphore].linkedCmdBuffers.push_back(submit->pCommandBuffers[i]);
```
tobine: save pCBNode here to avoid re-look-up in the loop above

layers/core_validation.cpp
```diff
@@ -4863,7 +4863,17 @@ static void decrementResources(layer_data *my_data, VkCommandBuffer cmdBuffer) {
     for (auto semaphore : pCB->semaphores) {
         auto semaphoreNode = my_data->semaphoreMap.find(semaphore);
         if (semaphoreNode != my_data->semaphoreMap.end()) {
-            semaphoreNode->second.in_use.fetch_sub(1);
+            // Each CB in semaphore's list may have reference to this semaphore
+            for (auto cb : semaphoreNode->second.linkedCmdBuffers) {
+                GLOBAL_CB_NODE *refCB = getCBNode(my_data, cb);
```
tobine: If we save GLOBAL_CB_NODE* in "linkedCmdBuffers" data struct instead of raw VkCommandBuffers, we can avoid re-look-up of refCB here.

layers/core_validation.cpp
```diff
@@ -4863,7 +4863,17 @@ static void decrementResources(layer_data *my_data, VkCommandBuffer cmdBuffer) {
...
+                for (auto semaphoreRef : refCB->semaphores) {
```
tobine: Trying to wrap my head around this triple "for" loop. In the "dumb" case where a single CB has a single semaphore, we're adding two extra loop cases for no benefit. Here's one thought:

1. Make linkedCmdBuffers an unordered_set
2. Guard the new double-for loop check with if (linkedCmdBuffers.count() > 1)
3. In the "else" case, we can just fetch_sub and then clear linkedCmdBuffers without descending extra double-for loop

layers/core_validation.cpp
mark-lunarg: Thanks, fixed.

layers/core_validation.h
mark-lunarg: Unordered_sets are for unique references.  This dataset must support multiple references to the same command buffer.

layers/core_validation.cpp
mark-lunarg: Done.

layers/core_validation.cpp
mark-lunarg: Done.

layers/core_validation.cpp
mark-lunarg: See note above about unordered_set.

layers/core_validation.cpp
```diff
@@ -5391,26 +5404,41 @@ static bool cleanInFlightCmdBuffer(layer_data *my_data, VkCommandBuffer cmdBuffe
...
+
+// If given queue is valid, remove the cmd_buffer from that queues inFlightCmdBuffer set. Finally, check all other
+// queues and if given cmd_buffer is not in flight on any other queue, Remove given cmd_buffer and all linked
+// cmd_buffers from the global inFlight set.
+// Note: This function assumes that the global lock is held by the calling thread.
 static inline void removeInFlightCmdBuffer(layer_data *dev_data, VkCommandBuffer cmd_buffer, VkQueue queue) {
-    // Pull it off of global list initially, but if we find it in any other queue list, add it back in
-    dev_data->globalInFlightCmdBuffers.erase(cmd_buffer);
-    if (dev_data->queueMap.find(queue) != dev_data->queueMap.end()) {
+    bool notFoundInAnyQueue = TRUE;
```
lentinem: use lowercase true

layers/core_validation.h
```diff
@@ -601,12 +601,17 @@ class FENCE_NODE : public BASE_NODE {
...
+struct GLOBAL_CB_NODE;
+
 class SEMAPHORE_NODE : public BASE_NODE {
   public:
     using BASE_NODE::in_use;
     uint32_t signaled;
     SemaphoreState state;
     VkQueue queue;
+    // Track command buffers linked by semaphores
+    std::vector<GLOBAL_CB_NODE *> linkedCmdBuffers;
```
lentinem: I'm still a little confused as to how this fixes the problem. Specifically the current structure has fence->cmdbuffers and if there is no fence it uses untracked cmdbuffers. It also has the semaphore->queue map such that cmdbuffers from one queue get mapped into the other queue when the semaphore is waited on. When the fence is waited on (or wait idle) it then loops through each cmdbuffer include those transferred via later submits with seamphores or fences. I really want to understand whats going on here so we don't run into more issues down the road.

layers/core_validation.cpp
```diff
@@ -5391,26 +5404,41 @@ static bool cleanInFlightCmdBuffer(layer_data *my_data, VkCommandBuffer cmdBuffe
...
+    if (notFoundInAnyQueue && (dev_data->queueMap.find(queue) != dev_data->queueMap.end())) {
         dev_data->queueMap[queue].inFlightCmdBuffers.erase(cmd_buffer);
         for (auto q : dev_data->queues) {
             if ((q != queue) &&
                 (dev_data->queueMap[q].inFlightCmdBuffers.find(cmd_buffer) != dev_data->queueMap[q].inFlightCmdBuffers.end())) {
-                dev_data->globalInFlightCmdBuffers.insert(cmd_buffer);
-                break;
+                notFoundInAnyQueue = FALSE;
```
lentinem: false

layers/core_validation.h
mark-lunarg: Ah, I had no idea what untrackedCmdBuffers was meant to be used for.  I'll see about getting that to work instead.

mark-lunarg:
Replaced by pull-request #309.

PR #304, opened 2016-04-11
layers: renderpass compatibility: there is only one d/s attachment

chrisforbes:
Previously this function assumed there were as many d/s attachments as
color attachments, which is invalid. Besides not making sense, we'd end
up dereferencing memory beyond that pointed to by
VkSubpassDescription::pDepthStencilAttachment if we had multiple color
attachments.


tobine:

PR #307, opened 2016-04-11
WIP: Remove duplicate count decrement.

lentinem:


tobine:
LGTM :+1:

tobine:
Tested this out and all 3 demos now flag "DS: Cannot delete semaphore <addr> which is in use."

Switching to :-1: until errors are fixed.

lentinem:
Yea I found that. It's more complicated than this so I labeled it WIP. Will remove if I think I have something but it might take a bit.

mark-lunarg:
Michael is right -- this is the correct thing to do but only part of it, and it'll be part of MR301 when I get it reworked.  Most of the problems fixed at this point (the right way) except for one cross-queue case.

mark-lunarg:
See pull request #309.

PR #308, opened 2016-04-11
layers: Cleanup core_validation physical device properties structs

tobine:
Pull physical device memory properties into layer_data and make sure it's
correctly populated at vkCreateDevice() time.
Clarify a few ambiguous "my_data" variables as "dev" or "instance".
Rename physical device properties to be lower case. This is the start of
a slow move to unify code on common standard with lowercase variable names.


layers/core_validation.cpp
```diff
@@ -4714,15 +4713,13 @@ VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL vkDestroyDevice(VkDevice device, cons
...
-#if MTMERGESOURCE
 VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL
 vkGetPhysicalDeviceMemoryProperties(VkPhysicalDevice physicalDevice, VkPhysicalDeviceMemoryProperties *pMemoryProperties) {
-    layer_data *my_data = get_my_data_ptr(get_dispatch_key(physicalDevice), layer_data_map);
-    VkLayerInstanceDispatchTable *pInstanceTable = my_data->instance_dispatch_table;
+    layer_data *instance_data = get_my_data_ptr(get_dispatch_key(physicalDevice), layer_data_map);
+    VkLayerInstanceDispatchTable *pInstanceTable = instance_data->instance_dispatch_table;
     pInstanceTable->GetPhysicalDeviceMemoryProperties(physicalDevice, pMemoryProperties);
-    memcpy(&memProps, pMemoryProperties, sizeof(VkPhysicalDeviceMemoryProperties));
+    memcpy(&instance_data->phys_dev_mem_props, pMemoryProperties, sizeof(VkPhysicalDeviceMemoryProperties));
```
chrisforbes: This isn't really correct (it assumes this is per-instance state, but it's actually per-physical-device state).

It's also unused (only device copy is used). Possibly just don't hook vkGetPhysicalDeviceMemoryProperties at all now.

tobine: Yes we just had one global copy previously which was bad. I'm now storing it per device which should be sufficient and I'll kill this code.

tobine:
Force pushed an update that addressed Chris's comment and also added an additional CL with a bunch more clean-up.

chrisforbes:
:+1:

tobine:

PR #309, opened 2016-04-12
layers: GH99-100, Fix core_validation CmdBuffer/Semaphore corner cases

mark-lunarg:
Michael and Tobin -- This MR should replace 301.  Using tracked command buffers simplified things.  There were a few corner cases in Dominik's app -- the aforementioned ones, along with calling waitforfences twice in a row on the same fence, and making two separate queueSubmit calls sharing a signal/wait semaphore but on different queues.  This caused removeInFlightCommandBuffer to add the CB back into the global inflight list as the CB was on a different queue.  Please take a look at that part of the patch, you guys might know of some pre-existing way to make that connection and prevent the dangling reference.  With these changes, our test apps and Dominik's run without errors and I'll continue testing to look for regressions.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -4975,6 +4975,7 @@ static void trackCommandBuffers(layer_data *my_data, VkQueue queue, uint32_t sub
         if (fence_data == my_data->fenceMap.end()) {
             return;
         }
+
         if (queue_data != my_data->queueMap.end()) {
             prior_fences = queue_data->second.lastFences;
             queue_data->second.lastFences.clear();
@@ -4984,7 +4985,7 @@ static void trackCommandBuffers(layer_data *my_data, VkQueue queue, uint32_t sub
             }
             queue_data->second.untrackedCmdBuffers.clear();
         }
-        fence_data->second.cmdBuffers.clear();
+
         fence_data->second.priorFences = prior_fences;
         fence_data->second.needsSignaled = true;
         fence_data->second.queue = queue;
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -5391,26 +5394,39 @@ static bool cleanInFlightCmdBuffer(layer_data *my_data, VkCommandBuffer cmdBuffe
...
     dev_data->globalInFlightCmdBuffers.erase(cmd_buffer);
     if (dev_data->queueMap.find(queue) != dev_data->queueMap.end()) {
         dev_data->queueMap[queue].inFlightCmdBuffers.erase(cmd_buffer);
+        bool cbOnDifferentQueue = true;
         for (auto q : dev_data->queues) {
             if ((q != queue) &&
                 (dev_data->queueMap[q].inFlightCmdBuffers.find(cmd_buffer) != dev_data->queueMap[q].inFlightCmdBuffers.end())) {
-                dev_data->globalInFlightCmdBuffers.insert(cmd_buffer);
+                // Do not add CB back in if it is on any of this queue's fence's cmdBuffers list
+                for (auto qFence : dev_data->queueMap[queue].lastFences) {
```
tobine: I see what you're doing here but not sure why it's valid?
If the CB is on that queue's lastFences, then isn't the CB still in-flight globally? I thought lastFences were the last set of fences submitted on queue that haven't yet been verified to be complete.
I think just some further comments or clarification may be what's needed here.

layers/core_validation.cpp
```diff
@@ -5190,9 +5191,10 @@ vkQueueSubmit(VkQueue queue, uint32_t submitCount, const VkSubmitInfo *pSubmits,
         vector<VkSemaphore> semaphoreList;
         for (uint32_t i = 0; i < submit->waitSemaphoreCount; ++i) {
             const VkSemaphore &semaphore = submit->pWaitSemaphores[i];
+            semaphoreList.push_back(semaphore);
             if (dev_data->semaphoreMap[semaphore].signaled) {
                 dev_data->semaphoreMap[semaphore].signaled = 0;
-                dev_data->semaphoreMap[semaphore].in_use.fetch_sub(1);
+                //dev_data->semaphoreMap[semaphore].in_use.fetch_sub(1);
```
tobine: can kill this line

tobine:
I just have the one comment that I don't fully understand why updated "removeInFlightCmdBuffer()" works.

Assuming that's correct, though, my general view is that this is an excellent fix for a tricky bug. Nice work!

That said, there are a few further updates I'd like to see to this code in the near future, although if there's a pressing need for this fix, I'm fine if these are follow-on clean-up changes.
At the time I wrote "removeInFlightCmdBuffer()" I didn't like the little trick to add the CB back into the globalInFlightCmdBuffers data struct, but I lived with it as there was just one extra check. Now that this logic is more complicated, I'd like the function to be re-factored to have the general flow:

1. remove the cmd_buffer from the given queue (if any)
2. Do all the logic to determine if the cmd_buffer should be removed from globalInFlightCmdBuffers
3. Only remove from globalInFlightCmdBuffers at the end of function if logic dictates it

And update the comment to reflect the new general flow.

vkWaitForFences now has:

1. Loop over fenceCount and do stuff
2. decrementResources
3. Loop over fenceCount and do stuff

I'd like the "stuff" in 1 & 3 to be cleaned up in such a way that we can either decrementResources before or after the loop and then have a single loop. I see that currently there are some dependencies between the 3 separate steps noted above, but I think with a bit of refactoring that code can be simplified and only a single loop will be needed.

layers/core_validation.cpp
mark-lunarg: It is unclear what the original use for lastFences was -- in these cases it does have the last fence for this queue, but that might just be because the use cases are simplistic.  What about plumbing fence[i] through from WaitForFences -- I think that might be more targeted and safe.  I'll give that a shot.

layers/core_validation.cpp
mark-lunarg: Thanks.

mark-lunarg:
I will rework RIFCB to ditch the remove/re-add hack.  The other stuff might need to come later -- I'll add a TODO.

mark-lunarg:
There's a new failure case for consecutive QueueSubmits to the same queue w/o Semaphores and only a fence on the second.  This should be legal but fails currently.

tobine:
Any new validation layer test cases that you can implement to capture some of these special cases would be fantastic.

mark-lunarg:
That's a great idea, I have about half-a-dozen scenarios to contribute.  Also, I'm now passing the fence into remvoeInFlightCmdBuffers explicitly, and for QueueWaitIdle pass NULL and then use all the fences in lastFences.  This seems like the correct thing to do, what do you think?

tobine:
Yes, all of this sounds good to me

lentinem:
-1. I'm still unclear as to what this is accomplishing and how that fixes the semaphore not being decremented issue. I think we should probably have a f2f discussion about this maybe on Friday.

mark-lunarg:
There isn't just a semaphore issue, there are a handful of intertwined issues.  Dominik originally had a couple of issues -- 99 and 100.  This addresses both of those and a few other problems that turned up.      We can discuss on Friday if it's in the morning, I'll be unavailable in the afternoon.

lentinem:
Right. However, I don't understand what about this fixes that particular issue which in theory it does do right?

mark-lunarg:
Had a fresh look this morning and ran across this.  I'm guessing this was the fix Michael was looking for.  Do y'all still want to have the meeting?

lentinem:
Yes!!! However, I think that line needs to be moved up before the if instead of removed completely (fence is reused).

mark-lunarg:
Moving the line before the 'if' appears to work fine.  If it's OK with y'all, I'll finish up testing and push this change.

lentinem:
SGTM.

mark-lunarg:
Subsumed by PR 399 & 400.

PR #310, opened 2016-04-12
layers: Clean up semaphore state in core_validation

tobine:
No need to have 3 semaphore states. Killed SemaphoreState struct and
updated all the code to just use the "signaled" bool.

Also minor clean-up of queueMap destroy code.


mark-lunarg:
+1.  I've wanted to do this all week!

tobine:

PR #311, opened 2016-04-12
layers: LX468, Prevent bad core_validation error message

mark-lunarg:
Tobin, can you give this a quick check?  If it's OK, I can push it after you have a look.

At BeginRenderpass time, a reference is made between the FB and CB but
it was not removed in EndRenderpass resulting in an INVALID CB.

Change-Id: Ibbc078b582b204bee26cdc510575c57023cf9d6b

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -4369,6 +4369,7 @@ static void resetCB(layer_data *dev_data, const VkCommandBuffer cb) {
         pCB->status = 0;
         pCB->viewports.clear();
         pCB->scissors.clear();
+
         for (uint32_t i = 0; i < VK_PIPELINE_BIND_POINT_RANGE_SIZE; ++i) {
             // Before clearing lastBoundState, remove any CB bindings from all uniqueBoundSets
             for (auto set : pCB->lastBound[i].uniqueBoundSets) {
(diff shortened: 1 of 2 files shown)


tobine:
:-1: The FrameBuffer continues to reference the cmd buffer until the FrameBuffer is destroyed (resulting in invalidation of the CB) or the CB is reset (looks like we don't correctly handle this case).

I think the bug here is that at resetCB() time,  we need to make sure that for each FB referenced by the CB being reset, that FB needs to have the CB removed from its referencingCmdBuffers. Does that make sense?

It looks like we currently don't have a complete list of FBs referenced by a CB, but only a single FB in pCB->framebuffer.

The slow and ugly way to fix would be at resetCB time, iterate over all FBs and remove and CB matches from their respective referencingCmdBuffers. Probably the right way to fix it, though, is to just replace the framebuffer member in GLOBAL_CB_NODE w/ an unordered_set that tracks all FBs. Then, at resetCB time, for each FB in that set, clear the given CB from that FB's referencingCmdBuffers.

mark-lunarg:
Thanks Tobin. We'll try to work it that way then.

layers/core_validation.cpp
```diff
@@ -9855,27 +9867,28 @@ static bool validateFramebuffer(layer_data *dev_data, VkCommandBuffer primaryBuf
...
-        }
-        auto fb_data = dev_data->frameBufferMap.find(secondary_fb);
-        if (fb_data == dev_data->frameBufferMap.end()) {
-            skip_call |=
-                log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
-                        DRAWSTATE_INVALID_SECONDARY_COMMAND_BUFFER, "DS", "vkCmdExecuteCommands() called w/ invalid Cmd Buffer %p "
-                                                                          "which has invalid framebuffer %" PRIx64 ".",
-                        (void *)secondaryBuffer, (uint64_t)(secondary_fb));
-            return skip_call;
+    for (auto primary_fb : pCB->framebuffers) {
```
mark-lunarg: Tobin, is this the right thing to do for 'primary_fb'?  Seems odd to have multiple primaries.

tobine: No, good catch. Here we only want to validate against the FB from the primary CB's current activeRenderPass. So I believe you can get that from dev_data->renderPassMap[pCB->activeRenderPass]->fb.

mark-lunarg:
Take another look?

tobine:

PR #312, opened 2016-04-12
WIP : make unique_objects handles always unique

tobine:
This is simple way to make sure all unique_objects handles are unique.
Created a simple allocator that grabs an initial batch of unique objects
to pass out, and then just keeps growing the batch as needed, not freeing
any of the unique objects until destroy device time.

Putting this out for early review to get feedback/input.


dustin-lunarg:
Would substituting a sequence number for the non-dispatchable handles be a viable alternative to the current wrapper based approach?  This would avoid the need to deal with the allocation pools, and would have the added benefit that any validation messages logged for these objects would show a simple integer value instead of an address. Something like:

CreateFence(device, pFence)
    DispatchTable(device)->CreateFence(device, pFence)
    uint64 unique_handle = g_sequence_number++
    g_object_map[unique_handle] = *pFence
    *pFence = unique_handle

GetFenceStatus(device, fence)
    DispatchTable(device)->GetFenceStatus(device, g_object_map[fence])

tobine:
Dustin, yes, Chris had same sequence number suggestion last night so I've been working on that alternative this morning and should have a PR today. Overall I like it better and seems cleaner.

lentinem:
This one is obsolete correct? Close?

tobine:
replaced with #325

PR #313, opened 2016-04-12
Fix issue when physical devices are queried multiple time

jbreton-nv:
In vkEnumeratePhysicalDevices, inst->phys_devs is freed if non-null and realloated just after.  

There could be some issue if it's not reallocated at the exact same memory location.  

Physical devices received from previous vkEnumeratePhysicalDevices might point at invalid memory

For example an application that does:  

    vkEnumeratePhysicalDevices(instance, &gpuCount, m_gpu);
    [...]
    vkEnumeratePhysicalDevices(instance, &gpuCount, dummy);

might have issue later when using m_gpu.

I think we should find a way to avoid releasing physical devices memory in vkEnumeratePhysicalDevices

### DIFF
--- a/loader/trampoline.c
+++ b/loader/trampoline.c
@@ -563,14 +563,14 @@ vkEnumeratePhysicalDevices(VkInstance instance, uint32_t *pPhysicalDeviceCount,
         loader_platform_thread_unlock_mutex(&loader_lock);
         return VK_ERROR_INITIALIZATION_FAILED;
     }
-    if (inst->phys_devs)
-        loader_heap_free(inst, inst->phys_devs);
     count = (inst->total_gpu_count < *pPhysicalDeviceCount) ?
 	    inst->total_gpu_count : *pPhysicalDeviceCount;
     *pPhysicalDeviceCount = count;
-    inst->phys_devs = (struct loader_physical_device_tramp *)loader_heap_alloc(
-        inst, count * sizeof(struct loader_physical_device_tramp),
-        VK_SYSTEM_ALLOCATION_SCOPE_INSTANCE);
+    if (!inst->phys_devs) {
+        inst->phys_devs = (struct loader_physical_device_tramp *)loader_heap_alloc(
+            inst, inst->total_gpu_count * sizeof(struct loader_physical_device_tramp),
+            VK_SYSTEM_ALLOCATION_SCOPE_INSTANCE);
+    }
     if (!inst->phys_devs) {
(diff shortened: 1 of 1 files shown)


loader/trampoline.c
```diff
@@ -563,14 +563,14 @@ vkEnumeratePhysicalDevices(VkInstance instance, uint32_t *pPhysicalDeviceCount,
...
-        loader_heap_free(inst, inst->phys_devs);
     count = (inst->total_gpu_count < *pPhysicalDeviceCount) ?
 	    inst->total_gpu_count : *pPhysicalDeviceCount;
     *pPhysicalDeviceCount = count;
-    inst->phys_devs = (struct loader_physical_device_tramp *)loader_heap_alloc(
-        inst, count * sizeof(struct loader_physical_device_tramp),
-        VK_SYSTEM_ALLOCATION_SCOPE_INSTANCE);
+    if (!inst->phys_devs) {
+        inst->phys_devs = (struct loader_physical_device_tramp *)loader_heap_alloc(
+            inst, inst->total_gpu_count * sizeof(struct loader_physical_device_tramp),
```
jon-lunarg: inst->total_gpu_count is wrong here (line 571).  It should be count instead to allow for layers returning a different count than the loader terminator code enumerated.

Otherwise the patch looks  good to me.

jon-lunarg: I think count would always be the same from one enumeratePD to another enumeratePD  within the same instance with the same set of ICDs and same set of layers enabled.  Unless we allow for an ICD to update it's driver installed that enumerates a different number of physicalDevices, while another process has an active CreateInstance.

jbreton-nv: 'count' could be less than 'inst->total_gpu_count' since it's computed based on '(inst->total_gpu_count < *pPhysicalDeviceCount)'  

If the app does:  

    int one = 1;
    int gpuCount = <actual number of GPU>;
    vkEnumeratePhysicalDevices(instance, &one , m_firstGpu);
    vkEnumeratePhysicalDevices(instance, &gpuCount , m_gpu);

You might have an issue when processing the second vkEnumeratePhysicalDevices, since inst->phys_devs will have a size of one physical device and won't be reallocated

jon-lunarg: Okay you convinced me.  The allocation should be based on total_gpu_count but the returnedphysicalDeviceCount and physicalDevice list will be based on count.

jon-lunarg:
cherry picked to master. Thanks!

PR #314, opened 2016-04-12
 Fixed total count of VkDescriptorType in vkCreateDescriptorPool

McNopper:
VkDescriptorPoolSize array can contain more than one of same VkDescriptorType.


McNopper:
The specification does not restrict, that of each VkDescriptorType only one, unique element has to be present in VkDescriptorPoolSize.

PR #315, opened 2016-04-12
Fixed total count of VkDescriptorType in vkCreateDescriptorPool

McNopper:
VkDescriptorPoolSize array can contain more than one of same VkDescriptorType.

The specification does not restrict, that of each VkDescriptorType only one, unique element has to be present in VkDescriptorPoolSize.


tobine:

Thanks for the fix!

PR #317, opened 2016-04-13
layers: cv: get rid of device-scoped current renderpass tracking

chrisforbes:
This was completely redundant to the (correct) CB-scoped renderpass
tracking, and only served to break things when different threads
simultaneously recorded command buffers.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -136,9 +136,6 @@ struct layer_data {
     unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE> imageLayoutMap;
     unordered_map<VkRenderPass, RENDER_PASS_NODE *> renderPassMap;
     unordered_map<VkShaderModule, unique_ptr<shader_module>> shaderModuleMap;
-    // Current render pass
-    VkRenderPassBeginInfo renderPassBeginInfo;
-    uint32_t currentSubpass;
     VkDevice device;
 
     // Device specific data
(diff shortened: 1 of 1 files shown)


layers/core_validation.cpp
```diff
@@ -9477,7 +9469,6 @@ VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL vkCmdNextSubpass(VkCommandBuffer comm
     layer_data *dev_data = get_my_data_ptr(get_dispatch_key(commandBuffer), layer_data_map);
     loader_platform_thread_lock_mutex(&globalLock);
     GLOBAL_CB_NODE *pCB = getCBNode(dev_data, commandBuffer);
-    TransitionSubpassLayouts(commandBuffer, &dev_data->renderPassBeginInfo, ++dev_data->currentSubpass);
```
lentinem: It's hard to tell from the diff but I assume this is still called elsewhere?

layers/core_validation.cpp
```diff
@@ -9529,7 +9520,6 @@ VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL vkCmdEndRenderPass(VkCommandBuffer co
     }
 #endif
     GLOBAL_CB_NODE *pCB = getCBNode(dev_data, commandBuffer);
-    TransitionFinalSubpassLayouts(commandBuffer, &dev_data->renderPassBeginInfo);
```
lentinem: Same here.

layers/core_validation.cpp
chrisforbes: Yes -- just a few lines away, with the correct per-CB state.

lentinem:
+1

tobine:

PR #318, opened 2016-04-13
Fix for ValidateCmdBufImageLayouts

FslNopper:
Filter out the transition from old layouts VK_IMAGE_LAYOUT_UNDEFINED or VK_IMAGE_LAYOUT_PREINITIALIZED

This case only happens, when the above layouts are "left". Can not happen later,
as a transition back to these layouts is forbidden and is checked at another place.


lentinem:
-1. Please explain why you think this is valid. If the image is in an undefined or preinitialized layout it must be transitioned in some form before use.

McNopper:
Agree, an image has to be transitioned before use and this can be done with either vkCmdWaitEvents or vkCmdPipelineBarrier, having one or more VkImageMemoryBarrier.
Problem is, that ValidateCmdBufImageLayouts is reporting wrong e.g. when vkCmdPipelineBarrier is called doing a first transition. The validation is complaining about current and expected layout but actually this function is doing the transition between old and new layout.

McNopper:
So from my side still +1

McNopper:
The ValidateCmdBufImageLayouts function is so far okay checking other commands, but for the command, actually changing the layout, an exception is needed.

The fix does this.

McNopper:
But I am currently thinking about the use case, when the image has not been transitioned and the image is used by another function. Let me revist the code.

lentinem:
Can you send a repro? It sounds like this is a bug somewhere but I don't think the correct solution is to exempt UNDEFINED and PREINITIALIZED. I'm not sure what this has to do with it currently changing the layout as that would apply regardless of format.

McNopper:
I'll try to find the bug and agree, this does not solve it.

McNopper:
Grrr ... wrong number of level count in subresource range. My code was buggy, validation is correct.

PR #325, opened 2016-04-13
Tobin unique objs uint map

tobine:
overhaul unique objects to prevent ever having repeat handles


demos/tri.c
```diff
@@ -1258,7 +1258,7 @@ char *demo_read_spv(const char *filename, size_t *psize) {
 
 static VkShaderModule demo_prepare_vs(struct demo *demo) {
     void *vertShaderCode;
-    size_t size;
+    size_t size = 0;
```
lentinem: Why is this needed?

chrisforbes:
You're going to need to lock around accesses to the mapping. I see you tried to avoid it by using atomics, but that's not enough.

demos/tri.c
tobine: Google coding style for local variables recommends initializing :) : http://google.github.io/styleguide/cppguide.html#Local_Variables

Also while I was testing, due to running in wrong dir, the demo_read_spv() was failing and uninitialized size had ridiculously high value that was causing goofy crash that was immediately clear that demo_read_spv() was failing.

lentinem:
Agree with chris. However, we could just use write locks with unlimited reads which I think is worth it in this case due to the frequency of locking and the fact that it's autogenerated so it shouldn't be too hard.

chrisforbes:
@lentinem do we have a preferred rwlock primitive?

lentinem:
I would use std::shared_mutex. However, after a brief search it appears that windows may not support that so we may need a wrapper for windows ... or use boost ... *shudders*.

tobine:
Updated this pull request.
Using a std::mutex to protect all map accesses and unique_id increments so switched unique_id to a unit64_t instead of atomic uint.

This change results in a measurable improvement in smoketest --validate performance. Over first 10 5s segments with master vs. this change avg framerate goes from 3.83fps to 4.39fps for ~15% improvement. A teeny tiny drop in the bucket, but still very nice.

demos/tri.c
lentinem: A note on the style: the point is that you shouldn't declare size and then init before use but initing twice is not recommended (the zero value would never be used). That being said if read_spv doesn't set to 0 on failure then this is still good.

tobine:
Tony, once you verify that this fixes CTS bugs, please push or assign back to me and I'll push. Thanks.

Tony-LunarG:
This passes the tests in CTS that once consistently got the threading errors - assigning back to Tobin.

tobine:

PR #332, opened 2016-04-13
layers: Validate correct vkSetEvent() use

tobine:
Make sure vkSetEvent() is not called on an event that's in use
by and in-flight cmd buffer. Such use violates queue forward progress.


lentinem:
+1

tobine:

PR #333, opened 2016-04-13
Small perf things

chrisforbes:


lentinem:
+1 although I'd like to know why those sets where indeed sets to being with and if there was a reason for that before changing them.

tobine:
My initial thought in making those sets was that it would maintain the order of when the error conditions were introduced. That's not a big deal, though, and I'm planning to change that delayed CB_INVALID tracking code to have a single, unified data struct that stores object and type so any such errors can be reported uniformly. I'll update again at that point.

chrisforbes:
@tobine That's not how `std::set` works, though -- the ordering is on the keys (by default, by `std::less`), not insertion order. So, nothing lost :).

PR #343, opened 2016-04-14
Remove error when only implicit dependencies are used.

lentinem:


layers/core_validation.cpp
```diff
@@ -8768,14 +8768,8 @@ static bool CheckDependencyExists(const layer_data *my_data, const int subpass,
         if (prev_elem == node.prev.end() && next_elem == node.next.end()) {
             // If no dependency exits an implicit dependency still might. If so, warn and if not throw an error.
```
tobine: It looks like this comment should be updated as well, yes?

tobine:

PR #345, opened 2016-04-14
layers: Correctly remove secondary cmd buffers from in-flight

tobine:
Secondary command buffers are considered executing, or "in-flight"
at the time they're bound into a primary command buffer. This leads
to a couple of corner case bugs if the primary command buffer is
never submitted on a queue. This change fixes those bugs:

1. If a secondary cmd buffer is being freed and is in the global
in-flight set, remove it if associated primary is not in-flight
2. When a primary cmd buffer is reset, remove its secondary
cmd buffers from in-flight


layers/core_validation.cpp
```diff
@@ -4354,6 +4354,10 @@ static void resetCB(layer_data *dev_data, const VkCommandBuffer cb) {
         pCB->drawData.clear();
         pCB->currentDrawData.buffers.clear();
         pCB->primaryCommandBuffer = VK_NULL_HANDLE;
+        // Make sure any secondaryCommandBuffers are removed from globalInFlight
```
lentinem: In theory this should never happen. When a command buffer is done all secondaries are done if it is reset without being completed that will throw an error. Might be good to somehow flag this either via an error message or with an assert.

layers/core_validation.cpp
```diff
@@ -5900,6 +5904,40 @@ vkDestroyDescriptorPool(VkDevice device, VkDescriptorPool descriptorPool, const
...
+        // Primary CB or secondary where primary is also in-flight is an error
+        if ((cb_node->createInfo.level != VK_COMMAND_BUFFER_LEVEL_SECONDARY) ||
+            (dev_data->globalInFlightCmdBuffers.count(cb_node->primaryCommandBuffer))) {
+            skip_call |= log_msg(
+                dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
+                reinterpret_cast<const uint64_t &>(cb_node->commandBuffer), __LINE__, DRAWSTATE_INVALID_COMMAND_BUFFER_RESET, "DS",
+                "Attempt to %s command buffer (%#" PRIxLEAST64 ") which is in use.", action,
+                reinterpret_cast<const uint64_t &>(cb_node->commandBuffer));
+        } else { // Secondary CB w/o primary in-flight, remove from in-flight
+            dev_data->globalInFlightCmdBuffers.erase(cb_node->commandBuffer);
```
lentinem: This could also be primary not in flight? Seems like another case for an assert as well. Secondary should be cleared ideally in decrement resources with the primary.

layers/core_validation.cpp
```diff
@@ -5900,6 +5904,40 @@ vkDestroyDescriptorPool(VkDevice device, VkDescriptorPool descriptorPool, const
...
+    return skip_call;
+}
+// Iterate over all cmdBuffers in given commandPool and verify that each is not in use
+static bool validateCommandBuffersNotInUse(layer_data *dev_data, const VkCommandPool commandPool, const char *action) {
+    bool skip_call = false;
+    auto pool_data = dev_data->commandPoolMap.find(commandPool);
+    if (pool_data != dev_data->commandPoolMap.end()) {
+        for (auto cmd_buffer : pool_data->second.commandBuffers) {
+            if (dev_data->globalInFlightCmdBuffers.count(cmd_buffer)) {
+                skip_call |= validateCommandBufferNotInUse(dev_data, getCBNode(dev_data, cmd_buffer), action);
```
lentinem: If in use validation not in use is a bit weird. I think this is mainly to clear the secondary cmd buffers but I would at least name something else.

lentinem:
+1. This is mostly ok. Minor name change.

yavn:
Hi, can you check this against `dEQP-VK.api.command_buffers.secondary_execute_twice` from CTS dev branch? I think it still complaints when a secondary buffer is reused in another primary buffer (this should be allowed as long as `VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT` is set):

> DS(ERROR): object: 0 type: 6 location: 5068 msgCode: 32: CB 0x6049e40 was submitted with secondary buffer 0x608c3d0 but that buffer has subsequently been bound to primary cmd buffer 0x608b4c0.

tobine:

@yavn Regarding VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT validation error that you note, that's a separate issue and I've filed Vulkan-Docs issue [182](https://github.com/KhronosGroup/Vulkan-Docs/issues/182) to clarify.

PR #346, opened 2016-04-14
layers: lx465 update core_validation image layout warning.

MarkY-LunarG:
If sub-resource data is present, print out that info as part of
the error message when the image layout is incorrect.

Change-Id: Ibd4d95e22d846b823b5c76986e8646624c109bc6


lentinem:
+1

tobine:

PR #348, opened 2016-04-14
layers: Fix lx474 by adding image extent = 0 validation check.

MarkY-LunarG:
Check only the valid dimensions for the imageType provided in
the create info.

Change-Id: I11d484caf1085b7b0ae4a402453b5fdcadd7b67d


layers/image.cpp
```diff
@@ -299,6 +299,36 @@ vkCreateImage(VkDevice device, const VkImageCreateInfo *pCreateInfo, const VkAll
...
+    // Intentional fall-through
+    case VK_IMAGE_TYPE_1D:
+        if (pCreateInfo->extent.width == 0) {
+            failedMinSize = true;
+        }
+        break;
+    default:
+        break;
+    }
+    if (failedMinSize) {
```
lentinem: I'm ok with this but you could get rid of the need for failedMinSize by having the cases break if != 0 and then the final one print the error.

tobine:

PR #349, opened 2016-04-14
layers: Add fix for lx172 to catch invalid line widths.

MarkY-LunarG:
Validate line width in both vkCreateGraphicsPipelines and in
vkCmdSetLineWidth.  Also, add a warning in vkCmdSetLineWidth
if the user calls it but doesn't enable dynamic line width.

Change-Id: I62118da9cb5282fcc22b1506e9be2db82b5f4a40

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -2936,6 +2936,31 @@ static bool validate_and_update_draw_state(layer_data *my_data, GLOBAL_CB_NODE *
     return result;
 }
 
+// Validate HW line width capabilities prior to setting requested line width.
+static bool verifyLineWidth(layer_data *my_data, DRAW_STATE_ERROR dsError, const uint64_t &target, float lineWidth) {
+    bool skip_call = false;
+
+    // First check to see if the physical device supports wide lines.
+    if ((VK_FALSE == my_data->phys_dev_properties.features.wideLines) && (1.0f != lineWidth)) {
+        skip_call |= log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, target, __LINE__,
+                             dsError, "DS", "Attempt to set lineWidth to %f but physical device wideLines feature "
+                                            "not supported/enabled so lineWidth must be 1.0f!",
+                             lineWidth);
+    } else {
+        // Otherwise, make sure the width falls in the valid range.
+        if ((my_data->phys_dev_properties.properties.limits.lineWidthRange[0] > lineWidth) ||
+            (my_data->phys_dev_properties.properties.limits.lineWidthRange[1] < lineWidth)) {
+            skip_call |= log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, target,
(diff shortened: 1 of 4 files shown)


layers/core_validation.cpp
```diff
@@ -2945,6 +2945,31 @@ static bool validate_and_update_draw_state(layer_data *my_data, GLOBAL_CB_NODE *
     return result;
 }
 
+// Validate HW line width capabilities prior to setting requested line width.
+static VkBool32 verifyLineWidth(layer_data *my_data, DRAW_STATE_ERROR dsError, uint64_t &target, float lineWidth) {
```
lentinem: I would use const uint64_t &target.

lentinem:
+1

layers/core_validation.cpp
tobine: bool return type

layers/core_validation.cpp
```diff
@@ -2945,6 +2945,31 @@ static bool validate_and_update_draw_state(layer_data *my_data, GLOBAL_CB_NODE *
...
+    VkBool32 skipCall = VK_FALSE;
```
tobine: bool here and use "skip_call" I'm trying to standardize variable names on lower-case going fwd

layers/core_validation.cpp
```diff
@@ -3102,6 +3127,14 @@ static bool verifyPipelineCreateState(layer_data *my_data, const VkDevice device
                                 pPipeline->graphicsPipelineCI.pTessellationState->patchControlPoints);
         }
     }
+    // If a rasterization state is provided, make sure that the line width conforms to the HW.
+    if (pPipeline->graphicsPipelineCI.pRasterizationState) {
+        VkBool32 dynLineWidth = isDynamic(pPipeline, VK_DYNAMIC_STATE_LINE_WIDTH);
```
tobine: switch to "bool" type
"dyn_line_width" preferred

layers/core_validation.cpp
```diff
@@ -6991,6 +7024,17 @@ VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL vkCmdSetLineWidth(VkCommandBuffer com
     if (pCB) {
         skipCall |= addCmd(dev_data, pCB, CMD_SETLINEWIDTHSTATE, "vkCmdSetLineWidth()");
         pCB->status |= CBSTATUS_LINE_WIDTH_SET;
+
+        PIPELINE_NODE *pPipeTrav = getPipeline(dev_data, pCB->lastBound[VK_PIPELINE_BIND_POINT_GRAPHICS].pipeline);
+        VkBool32 dynLineWidth = isDynamic(pPipeTrav, VK_DYNAMIC_STATE_LINE_WIDTH);
```
tobine: switch to "bool" type
"dyn_line_width" preferred

tobine:
Cube and Tri are now hitting errors based on this check:

`ERROR: [DS] Code 10 : Attempt to set lineWidth to 0.000000 but physical device wideLines feature not supported/enabled so lineWidth must be 1.0f!`

Can you update those demos so they are no longer flagging these errors?

MarkY-LunarG:
Updated with fixes to demos.  Also, I already updated and committed fixes to VulkanSamples which had several of the same issues.

tobine:

PR #350, opened 2016-04-14
layers: Update core_validation to fix lx265.

MarkY-LunarG:
Tobin discovered an issue with my previous change.  This resolves
the issue with the blend state targeting only a specific subpass
and verifying the subpass attachment count matches the blend
attachment count.

Change-Id: Iacdd880dde32267da095a3e9baa9e8eb66fdc760


lentinem:
+1

tobine:

PR #351, opened 2016-04-14
layers: Validate ONE_TIME_SUBMIT for secondary cmd buffers

### LINKED ISSUE #341: Update validation of secondary cmd buffers to handle ONE_TIME_SUBMIT_BIT
Spec language was updated a while back so make sure validation matches it:

If any given element of pCommandBuffers contains commands that execute secondary command buffers created with VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT, each such secondary command buffer must not have been previously submitted without re-recording that command buffer

tobine:
Previously were only validating *_ONE_TIME_SUBMIT_BIT for primary cmd buffers.
After clarification on spec, moving the check so it will also hit secondary
command buffers.

Fixes #341


lentinem:
+1

tobine:

PR #352, opened 2016-04-14
Remove the ICD and Layers search path for Windows

pdaniell-nv:
This makes it more like Linux which leaves it up to the system
LoadLibrary() call to find the .dll as specified in the .json files.
Hard-coding "C:" or any other search path creates the possibility of
picking up the wrong .dll. The LoadLibrary() implementation does the
right thing without having to search in its behalf and give it a full path. If the .json wanted to use a full path it would have specified that in the first place. No need for the loader to do any searching.

This fixes a bug with the loader where a system may have more than 1 OS installation and the in use installation is installed on say E:, but there is also an install in C: too, which we're not booted into. By hard-coding "C:\" in the search path the loader picks up the other OS installation's Vulkan ICD, which is not what the .json of the current OS expected.

### DIFF
--- a/loader/vk_loader_platform.h
+++ b/loader/vk_loader_platform.h
@@ -254,10 +254,10 @@ using namespace std;
 #define DEFAULT_VK_REGISTRY_HIVE HKEY_LOCAL_MACHINE
 #define DEFAULT_VK_DRIVERS_INFO "SOFTWARE\\Khronos\\Vulkan\\Drivers"
 // TODO: Are these the correct paths
-#define DEFAULT_VK_DRIVERS_PATH "C:\\Windows\\System32;C:\\Windows\\SysWow64"
+#define DEFAULT_VK_DRIVERS_PATH ""
 #define DEFAULT_VK_ELAYERS_INFO "SOFTWARE\\Khronos\\Vulkan\\ExplicitLayers"
 #define DEFAULT_VK_ILAYERS_INFO "SOFTWARE\\Khronos\\Vulkan\\ImplicitLayers"
-#define DEFAULT_VK_LAYERS_PATH "C:\\Windows\\System32;C:\\Windows\\SysWow64"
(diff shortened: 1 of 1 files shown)


jon-lunarg:
The exsiting code was obviously wrong! Thanks, fix is on master branch.

loader/vk_loader_platform.h
```diff
@@ -254,10 +254,10 @@ using namespace std;
 #define DEFAULT_VK_REGISTRY_HIVE HKEY_LOCAL_MACHINE
 #define DEFAULT_VK_DRIVERS_INFO "SOFTWARE\\Khronos\\Vulkan\\Drivers"
 // TODO: Are these the correct paths
-#define DEFAULT_VK_DRIVERS_PATH "C:\\Windows\\System32;C:\\Windows\\SysWow64"
+#define DEFAULT_VK_DRIVERS_PATH ""
```
jjulianoatnv: Line 2553 of loader.c reads 

>                     // a filename which is assumed in a system directory
>                     loader_get_fullpath(library_path, DEFAULT_VK_DRIVERS_PATH,
>                                         sizeof(fullpath), fullpath);

The comment is no longer accurate, misleading.
The new reality is that the .json file decides whether:

- the file is loaded from an absolute pathname, or
- the file is found by applying the platform's normal search path algorithm

Is there value in retaining the DEFAULT_VK_DRIVERS_PATH macro anymore? Maybe it should be removed altogether.

jon-lunarg: Agreed that DEFAULT_VK_DRIVERS_PATH should just be removed (and the layers default path also).  I think the comment is half correct.  The library_path is from the JSON  file and this else clause  is the case where the library_path was just a filename. I think the code  and comment can be cleaned up. I don't think it is necessary anymore to call loader_get_fullpath() in this case (the else clause).  Similar thing applies to layers.

PR #353, opened 2016-04-14
layers: Fix 32-bit VS2013 build

dustin-lunarg:
Change reinterpret_cast<uint64_t> to reinterpret_cast<uint64_t&> for
unique_objects generated and static code.


tobine:

Verified android build and basic tests. I'm getting lazy with my casts. Sorry, I'll be better in the future.

PR #381, opened 2016-04-18
Update cube and tri to run on android.

lentinem:


tobine:
I was re-basing this onto master and there are a few merge conflicts in cube.c that I'm not sure of the best way to merge. Michael, can you take a look at rebase and re-push if you're able to get the conflicts fixed. I'll try to then get it pushed in a timely manner so we don't hit conflicts again.

lentinem:
Ok done (I think). Please test this on linux as with all the x11/xcb changes I had to merge in I may have broken something.

demos/cube.c
```diff
@@ -1993,6 +2061,7 @@ static void demo_create_window(struct demo *demo) {
         exit(1);
     }
 }
+<<<<<<< 0d3f96c9766e309ab81b36fc908cdfde2ac6915a
```
ianelliottus: Looks like this line was left over from a merge conflict

ianelliottus:
Question: Can you resize the window for these demos, and if so, is there a problem when the swapchain is recreated?

lentinem:
I'm mostly running on android which makes that difficult to test for me. Tobin may be able to quickly check.

tobine:
@ianelliottus window resizing on linux appears to be working fine for both cube and tri.

@lentinem there were a number of issues on Linux but I massaged the code and verified that it's working on both linux and android. I created new #445 and assigned to Tony to get his final ok since all of the merge conflicts were with his recent changes.

ianelliottus:
@tobine I was wondering about resizing on Android.

lentinem:
@tobine I made some comments on that CL.

PR #382, opened 2016-04-18
layers: Fix object_tracker bugs

tobine:
A couple of bugs in object_tracker hand-coded sections that MarkY tracked down.
In DestroyInstance case we were attempting to delete wrong map key.
Then, for vkDestroyCommandPool we were using wrong function to clean-up the
command buffers in that pool which would result in occassional crashes.


mark-lunarg:
:+1:

tobine:

PR #383, opened 2016-04-18
layers: Cleanup core_validation fence tracking

tobine:
There were two separate mechanisms for tracking fence state in core_validation.
This change removes the last[Retired|Submitted]Id per queue method.
We also no longer over-write the fence createInfo to track fence state, but
just use the needsSignaled bool for all fence state tracking.
Finally, there was a validation test/check flagging an error if an
unsignaled fence was reset, but this is valid based on the spec so removed
the check and changed test to be a positive test expected to pass.


layers/core_validation.cpp
```diff
@@ -6121,29 +6037,16 @@ VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkResetFences(VkDevice device, ui
...
-            // Validate fences in SIGNALED state
-            if (!(fence_item->second.createInfo.flags & VK_FENCE_CREATE_SIGNALED_BIT)) {
-                // TODO: I don't see a Valid Usage section for ResetFences. This behavior should be documented there.
-                skipCall = log_msg(dev_data->report_data, VK_DEBUG_REPORT_WARNING_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_FENCE_EXT,
-                                   (uint64_t)pFences[i], __LINE__, MEMTRACK_INVALID_FENCE_STATE, "MEM",
-                                   "Fence %#" PRIxLEAST64 " submitted to VkResetFences in UNSIGNALED STATE", (uint64_t)pFences[i]);
-            } else {
-                fence_item->second.createInfo.flags =
-                    static_cast<VkFenceCreateFlags>(fence_item->second.createInfo.flags & ~VK_FENCE_CREATE_SIGNALED_BIT);
+            fence_item->second.needsSignaled = true;
```
lentinem: Shouldn't this be setting it to false?

lentinem:
There is now a redundant line in trackCommandBuffers that sets needsSignaled to true that can be removed.

tobine:
Submitted #387 to replace this PR

PR #387, opened 2016-04-19
layers: Cleanup core_validation fence tracking

tobine:
There were two separate mechanisms for tracking fence state in core_validation.
This change removes the last[Retired|Submitted]Id per queue method.
We also no longer over-write the fence createInfo to track fence state, but
just use the needsSignaled bool for all fence state tracking.
Finally, there was a validation test/check flagging an error if an
unsignaled fence was reset, but this is valid based on the spec so removed
the check and changed test to be a positive test expected to pass.

Note : This replaces #383 with a few more fix-ups


tobine:

Issue #377, opened 2016-04-18
SC layer: Tessellation evaluation to geometry shader interface error


yavn:
It seems there's one more interface mismatch case when dealing with user defined IO block:

```
// tessellation evaluation

layout(location = 0) out VertexData {
    vec4 in_gs_tessCoord;
    int  in_gs_primitiveID;
} ib_out;
```

```
// geometry

layout(location = 0) in VertexData {
    vec4 in_gs_tessCoord;
    int  in_gs_primitiveID;
} ib_in[];
```

I'm getting the following errors:

> SC(ERROR): object: 0 type: 0 location: 1545 msgCode: 1: Type mismatch on location 0.0: 'vec4 of float32' vs 'vec4 of float32'
> SC(ERROR): object: 0 type: 0 location: 1545 msgCode: 1: Type mismatch on location 1.0: 'sint32' vs 'sint32'

Here TESE output is not arrayed, but GEOM input is, and the arrayness is at the block level rather than block member. It seems this case is not covered by `types_match()` function.

Example test is available in CTS branch `302-tessellation-shader-tests` test group `dEQP-VK.tessellation.invariance`. More complex IO blocks are in `dEQP-VK.tessellation.user_defined_io`.

FslNopper:
Did you test with the latest master branch? I had the same issue with 1.0.8, but it is seems to be fixed now.

yavn:
Just tested with the latest master, and yes it still happens. Perhaps some other mismatch situation was fixed. I remember getting TESC into TESE errors before and these are fixed I think, but now it's TESE into GEOM.

chrisforbes:
Looking into this.

chrisforbes:
Right, the problem is the additional array level is not expressed in the member type -- it's in the block type. types_match is still looking for one to throw away, though.

chrisforbes:
Fix in #389 will solve this for reasonable cases. We'll miss producing errors for cases around arrays of blocks which aren't arrays of vertices -- but that's a crazy edge case.

chrisforbes:
Fixed in master.

yavn:
Works for me, thanks!

PR #389, opened 2016-04-20
layers: block members do not express extra array level in member type

chrisforbes:
Fixes GH #377.


tobine:

PR #393, opened 2016-04-20
Fix for clearing layouts and cmdbuffers.

lentinem:


tobine:

PR #395, opened 2016-04-20
doc: formalize the current loader-layer interface

olvaffe:
Expand the section "Discovery of layer entry points" into "Layer Library
Interface".  In general, the new section defines an interface that allows
a loader to discover and insert layers while staying compliant.  Other
sections have additional guidelines for layers to follow.


courtney-g:
+1

loader/LoaderAndLayerInterface.md
```diff
@@ -1030,45 +1030,147 @@ VkCommandBuffer as the first parameter and also include vkCreateDevice. Future
...
+#### Layer Libraries and Manifest Files
+
+The layer libraries and the manifest files must be kept in sync.  On Windows
+and Linux, the loader uses manifest files to discover layer libraries and
+layers.  On Android, the loader inspects the layer libraries directly.
+
+The manifest file specifies in its "functions", "vkGetInstanceProcAddr" node
+the function name the loader will use to discover
+<layerName>GetInstanceProcAddr.  When the layer library exports
+<layerName>GetInstanceProcAddr, the JSON node must explicitly specify that
```
jon-lunarg: When using a manifest file the name of GIPA and GDPA can be arbitrary.

loader/LoaderAndLayerInterface.md
```diff
@@ -1030,45 +1030,147 @@ VkCommandBuffer as the first parameter and also include vkCreateDevice. Future
...
+
+   When a layer library contains only one layer, this function may
+   alternatively be named `vkGetInstanceProcAddr`.
+
+ - `<layerName>GetDeviceProcAddr` behaves as if `<layerName>`'s
+   `vkGetDeviceProcAddr` is called.
+
+   When a layer library contains only one layer, this function may
+   alternatively be named `vkGetDeviceProcAddr`.
+
```
jon-lunarg: On WIndows and Linux the layer properties and extensions are queried from manifest file.  Thus the layer library need not export or even implement  vkEnumerate*Properties.

loader/LoaderAndLayerInterface.md
```diff
@@ -1030,45 +1030,147 @@ VkCommandBuffer as the first parameter and also include vkCreateDevice. Future
...
+   function never fails.
+
+ - `<layerName>GetInstanceProcAddr` behaves as if `<layerName>`'s
+   `vkGetInstanceProcAddr` is called, except
+
+   - when `pName` is `vkEnumerateInstanceLayerProperties`,
+     `vkEnumerateInstanceExtensionProperties`, or
+     `vkEnumerateDeviceLayerProperties` (but _not_
+     `vkEnumerateDeviceExtensionProperties`), it returns a function pointer to
+     the function of the same name defined by this interface.
```
jon-lunarg: why but _not_ EnumerateDeviceExtensionsProperties

loader/LoaderAndLayerInterface.md
```diff
@@ -1030,45 +1030,147 @@ VkCommandBuffer as the first parameter and also include vkCreateDevice. Future
...
+`VK_ERROR_LAYER_NOT_PRESENT`, including when `pLayerName` is `NULL`.
+
+`vkEnumerateDeviceLayerProperties` must always fail with
+`VK_ERROR_LAYER_NOT_PRESENT`.
+
+`vkEnumerateDeviceExtensionProperties` must always fail with
+`VK_ERROR_LAYER_NOT_PRESENT`, except when `pLayerName` is `NULL`.  It must
+handle the case where `pLayerName` is `NULL`, usually by chaining to other
+layers.
+
```
jon-lunarg: The above lines 1060 to 1072 are very confusing to me;  Can we just remove them? What purpose do they serve?  The layer must always return a failure code for these for but the layer library must always return a success code based on the statements below. How do you plan on implementing these different requirements?  I would assume by having two different functions that behave differently.  But they are documented as the same names, how do you handle the name collision?

olvaffe: I will remove them.

A layer and a layer library containing the layer are two different entities.  They both have a function named, say, vkEnumerateDeviceLayerProperties, but they are unrelated and serve different purposes.

A layer's vkEnumerateDeviceLayerProperties is required by the spec.  It is not called nor queryable.  It does not need to be implemented at all.  (Or if you like, you can have a static dummyEnumerateDeviceLayerProperties and a static dummyGetInstanceProcAddr that returns it).  Name collision is not an issue.

A layer library's vkEnumerateDeviceLayerProperties behaves similar to a layer's vkEnumerateDeviceLayerProperties.  A subtle difference is the layer library's function must ignore physicalDevice.  It must never fail either.

olvaffe: Actually, I need to keep

    vkEnumerateDeviceExtensionProperties` must handle the case where `pLayerName` is `NULL`, usually by chaining to other layers.

loader/LoaderAndLayerInterface.md
olvaffe: Because we want

  `<layerName>GetInstanceProcAddr(instance, "vkEnumerateDeviceExtensionProperties");`

to return `<layerName>`'s `vkEnumerateDeviceExtensionProperties`, which is defined by line 1069 to 1072, instead of the layer library's function defined by line 1109 to 1112.

In practice, all layers will define a single vkEnumerateDeviceExtensionProperties that satisfies both line-1069-to-1072 requirements and line-1109-to-1112 requirements.

loader/LoaderAndLayerInterface.md
olvaffe: I want to require all layers to implement them.  All validation layers in this repository already implement them.  It is not a lot of work, and it allows layers to be truly cross-platform.

loader/LoaderAndLayerInterface.md
olvaffe: Yes, that is described in line 1156 to 1157.

loader/LoaderAndLayerInterface.md
jon-lunarg: okay

jon-lunarg:
Cherry picked to master. Thanks!

PR #399, opened 2016-04-21
Fix cmd buffer tracking

lentinem:
Also included mark's tests which can be committed separately but this makes them all pass. There were two issues here:

1) Semaphore updating was done too early relative to cmd buffer tracking so that had to be moved otherwise the update would happen before anything was tracked and it would miss some updates.

2) When on multiple queues the inflightcmdbuffers list would be on a different queue than the one we were waiting on. To address this I removed that concept and replaced it with the in_use concept that we use elsewhere.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -4378,6 +4378,7 @@ static bool addCmd(const layer_data *my_data, GLOBAL_CB_NODE *pCB, const CMD_TYP
 static void resetCB(layer_data *dev_data, const VkCommandBuffer cb) {
     GLOBAL_CB_NODE *pCB = dev_data->commandBufferMap[cb];
     if (pCB) {
+        pCB->in_use.store(0);
         pCB->cmds.clear();
         // Reset CB state (note that createInfo is not cleared)
         pCB->commandBuffer = cb;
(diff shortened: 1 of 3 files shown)


layers/core_validation.cpp
```diff
@@ -4897,6 +4898,41 @@ static bool validateAndIncrementResources(layer_data *my_data, GLOBAL_CB_NODE *p
...
+                                         "Cannot get query results on queryPool %" PRIu64
+                                         " with index %d which was guarded by unsignaled event %" PRIu64 ".",
+                                         (uint64_t)(queryEventsPair.first.pool), queryEventsPair.first.index, (uint64_t)(event));
+                }
+            }
+        }
+    }
+    return skip_call;
+}
+// Remove given cmd_buffer from the global inFlight set.
```
tobine: comment needs to be updated

tobine:
I rebased this onto my fence tracking update fix. Filed new #400. Please give that a look to make sure everything looks ok.

mark-lunarg:
The other day, you ran through an impressive high-level description of what was going in in the tracking within and across the queues, etc.  As we go along there will be more and more folks going through this source -- could you add an overall description of how the tracking is implemented, or maybe Tobin-like comments stating what the routines/data structures are being used for?

PR #405, opened 2016-04-21
layers: Break out VkDescriptorSetLayout code into its own class

tobine:
This is the start of a refactor to pull code out of core_validation.cpp
into its own separate classes. I'm starting with descriptor set code as
it isolates reasonably well and it's old and could use some updating anyway.

For this first piece I've broken VkDescriptorSetLayout into its own
class currently called DescriptorSetLayout. I don't know if that's a great
name as it's close to VkDescriptorSetLayout, so I'm open to changing it.
Provided a brief class description in comment in new file descriptor_sets.h.

I made the class interfaces based on what other code is currently using.
I'm planning to pull more descriptor set code into its own classes and I
anticipate that will cause some flux in the class interfaces until most
of the work is done.

Follow-on changes will pull more descriptor code into separate classes
and spin out the class implementations into a separate descriptor_sets.cpp
file.

### DIFF
--- a/layers/descriptor_sets.h
+++ b/layers/descriptor_sets.h
@@ -1,0 +1,262 @@
+/* Copyright (c) 2015-2016 The Khronos Group Inc.
+ * Copyright (c) 2015-2016 Valve Corporation
+ * Copyright (c) 2015-2016 LunarG, Inc.
+ * Copyright (C) 2015-2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


layers/descriptor_sets.h
```diff
@@ -0,0 +1,250 @@
...
+    VkDescriptorSetLayoutCreateFlags flags_;
+    uint32_t binding_count_; // # of bindings in this layout
+    vector<safe_VkDescriptorSetLayoutBinding> bindings_;
+    uint32_t descriptor_count_; // total # descriptors in this layout
+    uint32_t dynamic_descriptor_count_;
+};
+DescriptorSetLayout::DescriptorSetLayout()
+    : layout_(VK_NULL_HANDLE), flags_(0), binding_count_(0), descriptor_count_(0), dynamic_descriptor_count_(0) {}
+// Construct DescriptorSetLayout instance from given create info
+DescriptorSetLayout::DescriptorSetLayout(debug_report_data *report_data, const VkDescriptorSetLayoutCreateInfo *p_create_info,
```
lentinem: Just started looking at this but I would strongly prefer the implementations to be in a cpp file.

tobine: Yep, planning to do that with next set of changes but I can do now if required.
I'm going to have all of the descriptor_set related classes defined in the header and all of their implementations in .cpp.

tobine:
Replaced w/ #413

PR #408, opened 2016-04-22
layers: GH384 Fix to correctly recognize immutable samplers

tobine:
Actual fix this time. If a binding is immutable samplers
skip over it when verifying descriptor bindings.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -2668,121 +2668,124 @@ static bool validate_and_update_drawtime_descriptor_state(
         SET_NODE *set_node = set_bindings_pair.first;
         LAYOUT_NODE *layout_node = set_node->pLayout;
         for (auto binding : set_bindings_pair.second) {
-            uint32_t startIdx = getBindingStartIndex(layout_node, binding);
-            uint32_t endIdx = getBindingEndIndex(layout_node, binding);
-            for (uint32_t i = startIdx; i <= endIdx; ++i) {
-                // We did check earlier to verify that set was updated, but now make sure given slot was updated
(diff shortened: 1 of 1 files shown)


layers/core_validation.cpp
```diff
@@ -2668,16 +2668,17 @@ static bool validate_and_update_drawtime_descriptor_state(
...
                 // TODO : Would be better to store set# that set is bound to so we can report set.binding[index] not updated
                 // For immutable sampler w/o combined image, don't need to update
-                if ((set_node->pLayout->createInfo.pBindings[i].descriptorType == VK_DESCRIPTOR_TYPE_SAMPLER) &&
-                    (set_node->pLayout->createInfo.pBindings[i].descriptorCount != 0) &&
-                    (set_node->pLayout->createInfo.pBindings[i].pImmutableSamplers)) {
-                    // Nothing to do here
+                if ((set_node->pLayout->createInfo.pBindings[binding_index].descriptorType == VK_DESCRIPTOR_TYPE_SAMPLER) &&
+                    (set_node->pLayout->createInfo.pBindings[binding_index].descriptorCount != 0) &&
+                    (set_node->pLayout->createInfo.pBindings[binding_index].pImmutableSamplers)) {
+                    i = endIdx; // Advance to end of this binding, no more work for immutable samplers
```
lentinem: This is the same as a break which would be a bit cleaner but this check also doesn't depend on i so can't we do this outside the loop and only do the loop if false?

tobine:
Yeah, I like that better, Michael. Pushed updated.

lentinem:
+1

tobine:

PR #409, opened 2016-04-22
tests: Add WaitForUnsubmittedFence to layer tests

Tony-LunarG:
This is from the CTS api.buffer.createBuffer_* tests which get validation errors when creating a fence, resetting it, waiting for it, and then destroying it.


mark-lunarg:
This test fails with current validation -- don't push until fix is completed.

lentinem:
PR #461 should fix this but I haven't been able to test yet.

mark-lunarg:
This appears to be a driver bug.  Driver should wait for fence to timeout even though it has not been submitted (could happen on a separate thread).  Some drivers wait, some don't.

PR #413, opened 2016-04-25
layers: Break out VkDescriptorSetLayout code into its own class

tobine:
NOTE : Rebased this from #405. Planning to push this afternoon if I don't hear any objections.

This is the start of a refactor to pull code out of core_validation.cpp
into its own separate classes. I'm starting with descriptor set code as
it isolates reasonably well and it's old and could use some updating anyway.

For this first piece I've broken VkDescriptorSetLayout into its own
class currently called DescriptorSetLayout. I don't know if that's a great
name as it's close to VkDescriptorSetLayout, so I'm open to changing it.
Provided a brief class description in comment in new file descriptor_sets.h.

I made the class interfaces based on what other code is currently using.
I'm planning to pull more descriptor set code into its own classes and I
anticipate that will cause some flux in the class interfaces until most
of the work is done.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -56,10 +56,8 @@
 #endif
 #include "vk_struct_size_helper.h"
 #include "core_validation.h"
-#include "vk_layer_config.h"
 #include "vk_layer_table.h"
 #include "vk_layer_data.h"
-#include "vk_layer_logging.h"
 #include "vk_layer_extension_utils.h"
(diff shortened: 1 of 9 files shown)


mark-lunarg:
Hey Tobin, we're gonna code-freeze this afternoon, can you please hold off until we're locked down?

tobine:
Sure thing. Gimme the heads-up when you're locked.

tobine:

PR #419, opened 2016-04-26
layers: Add copy constructors and initializers for safe_* structs

tobine:
Any safe struct for which default copy constructor was incorrect was
broken (anything with embedded ptrs). This change fixes that by adding
correct copy constructors. Also had to add initializers that can take
object ptrs in order for copy constructor code to work.


lentinem:
+1

tobine:

PR #420, opened 2016-04-27
layers: Emit safe_*::ptr() members to eliminate some cast noise

chrisforbes:
These casts were annoying noise, and uncheckable. Move them into the
generator where we know the one type that makes sense.

Drop spurious reinterpret_cast of stage create info to itself -- it's
all raw types once the root pCreateInfo is unwrapped.


tobine:

PR #442, opened 2016-04-28
layers: Fix -Wunused-result warning for vasprintf

chrisforbes:
In case of allocation failure, glibc's vasprintf doesn't make any
guarantees about the ptr -- the return value is the only way to handle
this safely.


courtney-g:
+1

tobine:

PR #443, opened 2016-04-28
Update VK_LAYER_device_limits

olvaffe:
This pull request attempts to clean up device_limits.  The highlights are:

 - layer commands are put in device_limits namespace with the vk prefix dropped
 - GIPA and GDPA are refactored to be compliant with the spec
 - avoid invalid GIPA(NULL, "vkCreateDevice") call

My goal is to use device_limits to set up a basic layer guideline.  I can then update other layers using the same guideline.  Suggestions are welcomed.


layers/device_limits.cpp
```diff
@@ -781,3 +736,136 @@ VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vkGetInstanceProcAddr(V
...
     }
 }
+
+static PFN_vkVoidFunction
+intercept_core_instance_command(const char *name) {
+    static const struct {
+        const char *name;
+        PFN_vkVoidFunction proc;
+    } core_instance_commands[] = {
+#define ENTRY(cmd) { "vk" # cmd, reinterpret_cast<PFN_vkVoidFunction>(cmd) }
```
mark-lunarg: OLV, this macro doesn't save us all that much (and macros give me the shivers) -- could we just eliminate it and do the typing here?

olvaffe: This actually saves us from issues such as the one fixed by commit 40db7ffcf75097e778d0a43fdd0d17b4615a4d2f.

With that weighted in, do you still prefer the macro be eliminated?  I can see the reason and I am fine eliminating it.

mark-lunarg: Yes, please.  Though this fits in well with the style guidelines and is not at all objectionable in itself, for the layers we'd like to discourage the use of macros in general.

olvaffe: Pull request force-updated.  It should be macro-free now.

dustin-lunarg:
:+1:

olvaffe:
Am I supposed to merge this and #451 myself? :smiley:

I looks like I have the permissions.  I will do the merges tomorrow unless someone objects.

tobine:
Yes, in general if you get a thumbs up and have run through the tests in contributing.md then you can go ahead and push if you have permissions.
I'll run both the PRs in the morning and if they look good I may just push them for you.

olvaffe:
Thanks, Tobin.  I did ran vk_layer_validation_tests manually and there was no regression.  I am following CONTRIBUTING.md now.  I will update here only when I find new issues.

layers/device_limits.cpp
```diff
@@ -749,12 +747,13 @@ intercept_core_instance_command(const char *name) {
         { "vkGetPhysicalDeviceQueueFamilyProperties", reinterpret_cast<PFN_vkVoidFunction>(GetPhysicalDeviceQueueFamilyProperties) },
         { "vkGetPhysicalDeviceMemoryProperties", reinterpret_cast<PFN_vkVoidFunction>(GetPhysicalDeviceMemoryProperties) },
         { "vkGetPhysicalDeviceSparseImageFormatProperties", reinterpret_cast<PFN_vkVoidFunction>(GetPhysicalDeviceSparseImageFormatProperties) },
-        { "vkEnumerateInstanceLayerProperties", reinterpret_cast<PFN_vkVoidFunction>(vkEnumerateInstanceLayerProperties) },
-        { "vkEnumerateDeviceLayerProperties", reinterpret_cast<PFN_vkVoidFunction>(vkEnumerateDeviceLayerProperties) },
-        { "vkEnumerateInstanceExtensionProperties", reinterpret_cast<PFN_vkVoidFunction>(vkEnumerateInstanceExtensionProperties) },
-        { "vkEnumerateDeviceExtensionProperties", reinterpret_cast<PFN_vkVoidFunction>(EnumerateDeviceExtensionProperties) },
```
olvaffe: vkEnumerateDeviceExtensionProperties is incorrectly removed here.

tobine:

PR #445, opened 2016-04-28
Android cube tri

tobine:
Tony, please give this a look and verify that it runs for you on Windows. This is a series of changes to get cube and tri working for android. There were some conflicts with your recent updates that I had to resolve. I verified that cube/tri and running on linux/nvidia and android so if Windows looks good to you these changes should be ready to push.


demos/cube.c
```diff
@@ -336,16 +336,16 @@ struct demo {
     HINSTANCE connection;        // hInstance - Windows Instance
     char name[APP_NAME_STR_LEN]; // Name to put on the window/icon
     HWND window;                 // hWnd - window handle
-#elif defined(VK_USE_PLATFORM_XLIB_KHR)
+#elif !defined(VK_USE_PLATFORM_ANDROID_KHR)
```
lentinem: I see why this wasn't working but can we use if defined(XLIB) || defined(XCB). I think this breaks windows if not and additionally I really would like everything to actually check for the appropriate flags.

demos/cube.c
```diff
@@ -2064,8 +2064,7 @@ static void demo_create_window(struct demo *demo) {
         exit(1);
     }
 }
-<<<<<<< 0d3f96c9766e309ab81b36fc908cdfde2ac6915a
-#else
+#elif !defined(VK_USE_PLATFORM_ANDROID_KHR)
```
lentinem: Ditto here. We should not need XLIB functions if XLIB is not defined.

demos/cube.c
```diff
@@ -2422,22 +2422,22 @@ static void demo_init_vk(struct demo *demo) {
                 demo->extension_names[demo->enabled_extension_count++] =
                     VK_KHR_WIN32_SURFACE_EXTENSION_NAME;
             }
-#if defined(VK_USE_PLATFORM_XLIB_KHR)
+#elif !defined(VK_USE_PLATFORM_ANDROID_KHR)
```
lentinem: Same.

tobine:
force pushed update w/ fixed "defined()" blocks

demos/cube.c
```diff
@@ -336,11 +336,11 @@ struct demo {
     HINSTANCE connection;        // hInstance - Windows Instance
     char name[APP_NAME_STR_LEN]; // Name to put on the window/icon
     HWND window;                 // hWnd - window handle
-#elif defined(VK_USE_PLATFORM_XLIB_KHR)
+#elif defined(VK_USE_PLATFORM_XLIB_KHR) | defined(VK_USE_PLATFORM_XCB_KHR)
```
Tony-LunarG: So if you define either XCB or XLIB you get both?

lentinem: Yes. This isn't a change from what currently exists. While I would really like to have everything using these defines the way it currently is is that if WIN32 is not defined you get both. This doesn't work on android but is equivalent to this statement on windows/linux. I had originally split them out but I think there was code in the XCB path that used the types defined here.

Tony-LunarG:
Ok - both demos run and resize without issues on Nvidia Windows 7

tobine:

PR #447, opened 2016-04-28
Add a compile-time option for layer search paths.

null77:
This allows me to embed compile-time search paths into my layer binaries to deploy on the Chromium bots for testing, as well as on developer machines. This is much easier for me than using environment varaibles. Was discussing this with jon@lunarg


jon-lunarg:
fixed build issue on Linux and pushed the commits to master branch

courtney-g:
Does this need an update to loader/LoaderAndLayerInterface.md as well?

jon-lunarg:
Perhaps a section could be added on compile time config options  for paths. Since the loader built and shipped would have an empty default path  this feature wouldn't normally be available on either Windows or Linux.

PR #448, opened 2016-04-28
Add an output directory option to genvk.py.

null77:
Currently for the threading utils and parameter checker generator,
the script is hard-coded to write to the current directory. Add
an option to add a custom output directory.

This allows me to use these scripts within the ANGLE build.


generator.py
```diff
@@ -561,7 +565,10 @@ def beginFile(self, genOpts):
         # Open specified output file. Not done in constructor since a
         # Generator can be used without writing to a file.
         if (self.genOpts.filename != None):
-            self.outFile = open(self.genOpts.filename, 'w')
+            if (self.genOpts.genDirectory != None):
+                self.outFile = open(self.genOpts.genDirectory + '/' + self.genOpts.filename, 'w')
```
dustin-lunarg: Can you change this to use os.path.join()?

null77: Done, good catch.

null77:
Uploaded an updated CL with os.path.join instead of string concatenation.

dustin-lunarg:
Cherry-picked to master and pushed.

null77:
Thanks!

PR #451, opened 2016-04-29
Update VK_LAYER_image

olvaffe:
Similar to pull request #443, but for VK_LAYER_image.


layers/image.cpp
```diff
@@ -1318,38 +1282,144 @@ VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vkGetDeviceProcAddr(VkD
...
+        PFN_vkVoidFunction proc;
+    } core_instance_commands[] = {
+        { "vkGetInstanceProcAddr", reinterpret_cast<PFN_vkVoidFunction>(GetInstanceProcAddr) },
+        { "vkCreateInstance", reinterpret_cast<PFN_vkVoidFunction>(CreateInstance) },
+        { "vkDestroyInstance", reinterpret_cast<PFN_vkVoidFunction>(DestroyInstance) },
+        { "vkCreateDevice", reinterpret_cast<PFN_vkVoidFunction>(CreateDevice) },
+        { "vkGetPhysicalDeviceProperties", reinterpret_cast<PFN_vkVoidFunction>(GetPhysicalDeviceProperties) },
+    };
+
+    // we should never be queried for these commands
```
courtney-g: Reference please?
Why won't these be called?

layers/image.cpp
```diff
@@ -1318,38 +1282,144 @@ VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vkGetDeviceProcAddr(VkD
...
+    return nullptr;
+}
+
+static PFN_vkVoidFunction
+intercept_core_device_command(const char *name) {
+    static const struct {
+        const char *name;
+        PFN_vkVoidFunction proc;
+    } core_device_commands[] = {
+        { "vkGetDeviceProcAddr", reinterpret_cast<PFN_vkVoidFunction>(GetDeviceProcAddr) },
```
courtney-g: One of the most common issues I've seen in Get*ProcAddr in layers is forgetting to intercept a function.
How do we know this is the complete list?
Is there something we could do to validate the list?

courtney-g:
Code looks good, just a few comments/questions.
+1

layers/image.cpp
olvaffe: In `vkGetInstanceProcAddr` at the bottom of the file, these three functions are intercepted.

Another way to look at this is, in the [interface definition](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/loader/LoaderAndLayerInterface.md#layer-library-interface-version-0), the three functions are special-cased in `<layerName>GetInstanceProcAddr`.

layers/image.cpp
olvaffe: We could enable -Wunused-function and that should catch the issue fixed by this commit https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/443/commits/673031c493dd0ddcb0bdaa6985a028575e9549d7

There is another issue fixed by this commit https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/443/commits/4a0095eb7e0b78540559d5dfe1e0c29e7e3545db.  I think it can be fixed by using macros, but the current coding style is to avoid macros in layers.

layers/image.cpp
```diff
@@ -1310,13 +1309,14 @@ intercept_core_instance_command(const char *name) {
         { "vkCreateInstance", reinterpret_cast<PFN_vkVoidFunction>(CreateInstance) },
         { "vkDestroyInstance", reinterpret_cast<PFN_vkVoidFunction>(DestroyInstance) },
         { "vkCreateDevice", reinterpret_cast<PFN_vkVoidFunction>(CreateDevice) },
-        { "vkEnumerateInstanceLayerProperties", reinterpret_cast<PFN_vkVoidFunction>(vkEnumerateInstanceLayerProperties) },
-        { "vkEnumerateInstanceExtensionProperties", reinterpret_cast<PFN_vkVoidFunction>(vkEnumerateInstanceExtensionProperties) },
-        { "vkEnumerateDeviceLayerProperties", reinterpret_cast<PFN_vkVoidFunction>(vkEnumerateDeviceLayerProperties) },
-        { "vkEnumerateDeviceExtensionProperties", reinterpret_cast<PFN_vkVoidFunction>(EnumerateDeviceExtensionProperties) },
```
olvaffe: vkEnumerateDeviceExtensionProperties is incorrectly removed here.

tobine:

PR #454, opened 2016-04-29
tests: Initialize reserved flags as 0

dustin-lunarg:
Initialize the flags members of the pipeline state CreateInfo structs
that are reserved for future use to 0.


karl-lunarg:
+1
I don't see any reason to not push this into master

tobine:

All tests passing again, thanks, Dustin

PR #456, opened 2016-04-29
layers: Fix access violation

dustin-lunarg:
Move the get_dispatch_key(instance/device) call before calling the
driver's vkDestroyDevice/vkDestroyInstance functions in
unique_objects.  This addresses an access violation error reported
by Application Verifier on Windows due to the previously destroyed
instance/device handle being dereferenced.


mark-lunarg:
+1

tobine:

Issue #437, opened 2016-04-27
Add tests for swapchain validation layer


tobine:
The currest list of validation errors as well as any associated tests for the swapchain validation layer are documented in [vk_validation_layer_details.md](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/vk_validation_layer_details.md).

This documentation should be examined according to the existing validation code and the Vulkan specification and additional test(s) should be added as appropriate.

ianelliottus:
Just noticed this tracking issue.  I dealt with this a while back, but didn't close this issue.  There are some time-consuming tests that should be added sometime (I left TODO statements for what was needed).

Jon Ashburn also added some incomplete support for the VK_KHR_display extension.  It would be good for somebody to finish that support sometime.

PR #457, opened 2016-04-29
tests: #437 Start testing the WSI swapchain validation layer

ianelliottus:
This is the initial version of the EnableWsiBeforeUse test case.  It verifies
    that the swapchain validation layer issues an error if any of the WSI functions
    are called when the corresponding extension isn't enabled.  This Initial
    version tests the functions for the following extensions:
    
    - VK_KHR_xcb_surface
    - VK_KHR_surface
    - VK_KHR_swapchain
    
    There is curently one exception: vkQueuePresentKHR().  This function dispatches
    from a VkQueue.  The layer uses the swapchain parameter in order to look up the
    VkDevice that the extension should have been enabled for.  Since we didn't
    create a real swapchain, the lookup doesn't occur (the layer doesn't chase NULL
    pointers), and so the error doesn't occur.
    
    This test case still needs to test the following extensions:
    
    - VK_KHR_android_surface
    - VK_KHR_mir_surface
    - VK_KHR_wayland_surface
    - VK_KHR_win32_surface
    - VK_KHR_xlib_surface
    
    When the swapchain layer supports the following extensions, this test case
    should test them:
    
    - VK_KHR_display
    - VK_KHR_display_swapchain

Note: The test code caught that one WSI function (vkDestroySurfaceKHR()) wasn't checking whether the VK_KHR_surface extension was enabled.  That is fixed with this pull request.

### DIFF
--- a/layers/swapchain.cpp
+++ b/layers/swapchain.cpp
@@ -859,6 +859,14 @@ vkDestroySurfaceKHR(VkInstance instance, VkSurfaceKHR surface, const VkAllocatio
     layer_data *my_data = get_my_data_ptr(get_dispatch_key(instance), layer_data_map);
     std::unique_lock<std::mutex> lock(global_lock);
     SwpSurface *pSurface = &my_data->surfaceMap[surface];
+    SwpInstance *pInstance = &(my_data->instanceMap[instance]);
+
+    // Validate that the platform extension was enabled:
+    if (pInstance && !pInstance->surfaceExtensionEnabled) {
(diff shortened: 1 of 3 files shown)


courtney-g:
LGTM

tobine:
These changes look good but when I run on Nvidia Linux I see the output message "Error: Undefined DisplayServer". Is this expected? Is there any way to suppress the message? I don't like the fact that one of the first things that a user will see when they start running tests is an "Error:" message.

tobine:
That's my bad, the "Error: Undefined DisplayServer" is not coming from these changes.

PR #459, opened 2016-04-29
Add fence in use for bind sparse info.

lentinem:


tobine:

PR #461, opened 2016-04-29
Remove fences once they have been waited on.

lentinem:


tobine:

Issue #463, opened 2016-05-02
CoreValidation: A weird "num of descriptor count in a binding must be equal to the total number of descriptors defined in DS layout, or else!" check results in invalid debug call-backs at vkCmdBindDescriptorSets() call time.


DominikWitczakAMD:
Consider the following descriptor set layout:

![image](https://cloud.githubusercontent.com/assets/17761066/14956991/123686f6-1083-11e6-95d0-cfc3c841d94b.png)

There is a very weird check in the same function which results in a debug call-back I'm worried might be invalid:

![image](https://cloud.githubusercontent.com/assets/17761066/14957030/4f63c28c-1083-11e6-936a-e9b5b5bc2cae.png)

I'm not sure what the intent behind this check was, but I don't think we should be assuming that the number of items under a specific binding should always be equal to the total number of descriptors defined for a descriptor set.

I'm seeing this call-back at vkCmdBindDescriptorSets() call time, when I take a vile attempt to bind the discussed descriptor set :)

The relevant change must have been submitted recently. I don't recall seeing this callback a few weeks ago.

tobine:
Yep dumb bug, it's just supposed to be checking against the specific binding count, not the entire layout count. Fix coming shortly.

tobine:
Just submitted #466 to fix this.

tobine:
Fix was cherry-picked and pushed to master

PR #466, opened 2016-05-02
layers: GH463 Fix layout compatibility check

tobine:
Fix for #463 
After lh and rh layout descriptorCounts were verified to be the same, the
next check was testing the descriptor count for each binding in the lh binding
against the entire descriptor count for the rh layout. The check should be made
just against the descriptor count for the specific binding we're checking in
the rh layout.


lentinem:
+1

tobine:

Issue #465, opened 2016-05-02
No debug call-back reported if vkCmdDispatch() is recorded without a pipeline being bound to the compute bind point


DominikWitczakAMD:
I've just encountered an application which tried to dispatch compute invocations without having bound a compute pipeline first. This should have resulted in a debug call-back, but apparently existing validation layers do not capture such cases.

We need this fixed asap. It's very easy to fall for this trap.

tobine:
Submitted #467 to add this check

tobine:
Fix was cherry-picked and pushed to master

PR #467, opened 2016-05-02
layers: GH465 Add validation for now bound VkPipeline

tobine:
NOTE: Diff is pretty manged due to indentation change, but at the core this just adds a check for both Gfx and Compute cases to make sure that a pipeline is bound ad Draw/Dispatch time. Addresses #465 

At the time of a Draw or Dispatch, a pipeline must be bound to the
command buffer. This adds a check to verify that.
Added tests for both draw and compute cases to verify the check.


layers/core_validation.cpp
```diff
@@ -2757,108 +2757,112 @@ static bool validate_and_update_draw_state(layer_data *my_data, GLOBAL_CB_NODE *
...
-                                result |=
-                                    log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT,
-                                            VK_DEBUG_REPORT_OBJECT_TYPE_DESCRIPTOR_SET_EXT, (uint64_t)pSet->set, __LINE__,
-                                            DRAWSTATE_DESCRIPTOR_SET_NOT_UPDATED, "DS",
-                                            "DS %#" PRIxLEAST64 " bound but it was never updated. It is now being used to draw so "
-                                            "this will result in undefined behavior.",
-                                            (uint64_t)pSet->set);
-                            }
+    // TODO : When Compute shaders are properly parsed, fix this section to validate them as well
+    if (state.pipelineLayout && (VK_PIPELINE_BIND_POINT_GRAPHICS == bindPoint)) {
```
lentinem: It's a little hard to tell but I think this makes the entire set of logic below GRAPHICS so it might be simpler to check for !GRAPHICS and return early.

lentinem:
+1. Only minor comments.

tobine:

PR #468, opened 2016-05-02
tests: Re-enable and flesh out InvalidUsageBits

Tony-LunarG:
Re-enable ifdef'ed usage test - check 3 paths, 2 image paths and a buffer path.

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -1127,55 +1127,66 @@ TEST_F(VkLayerTest, ResetUnsignaledFence) {
     m_errorMonitor->VerifyNotFound();
 }
 
-/* TODO: Update for changes due to bug-14075 tiling across render passes */
-#if 0
 TEST_F(VkLayerTest, InvalidUsageBits)
 {
-    // Initiate Draw w/o a PSO bound
-
     m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT,
         "Invalid usage flag for image ");
 
     ASSERT_NO_FATAL_FAILURE(InitState());
-    VkCommandBufferObj commandBuffer(m_device);
-    BeginCommandBuffer();
+    VkImageObj image(m_device);
+    // Initialize image with USAGE_SAMPLED
+    image.init(128, 128, VK_FORMAT_D32_SFLOAT_S8_UINT,
+               VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT, VK_IMAGE_TILING_LINEAR, 0);
 
-    const VkExtent3D e3d = {
-        .width = 128,
-        .height = 128,
-        .depth = 1,
-    };
-    const VkImageCreateInfo ici = {
-        .sType = VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO,
-        .pNext = NULL,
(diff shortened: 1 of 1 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -153,6 +153,18 @@ class ErrorMonitor {
         return result;
     }
 
+    VkBool32 CheckForOtherMessages(const char *msgString) {
+        VkBool32 result = VK_FALSE;
+        test_platform_thread_lock_mutex(&m_mutex);
```
chrisforbes: Should unlock this somewhere

tests/layer_validation_tests.cpp
```diff
@@ -153,6 +153,18 @@ class ErrorMonitor {
         return result;
     }
 
+    VkBool32 CheckForOtherMessages(const char *msgString) {
```
chrisforbes: Why VK types here?

tests/layer_validation_tests.cpp
```diff
@@ -1127,55 +1139,65 @@ TEST_F(VkLayerTest, ResetUnsignaledFence) {
...
+    // Buffer usage not set to TRANSFER_SRC and image usage not set to
+    // TRANSFER_DST
+    BeginCommandBuffer();
+    vkCmdCopyBufferToImage(m_commandBuffer->GetBufferHandle(), buffer.handle(),
+                           image.handle(), VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL,
+                           1, &region);
 
     m_errorMonitor->VerifyFound();
+    if (!m_errorMonitor->CheckForOtherMessages(
+            "Invalid usage flag for image ")) {
```
chrisforbes: This is kindof ugly -- but I see why you've done it.

As an alternative, did you consider just calling `vkCmdCopyBufferToImage`twice, once for each of the errors you're looking for?

Otherwise, we possibly need some stronger support for specifying (multiple) expectations.

Thoughts?

Tony-LunarG:
I've added the unlock.  The VkBool32 was staying consistent with other routines.  I can certainly call vkCmdCopyBufferTo Image a second time if that's cleaner.

tests/layer_validation_tests.cpp
```diff
@@ -1127,55 +1127,65 @@ TEST_F(VkLayerTest, ResetUnsignaledFence) {
...
+                           image.handle(), VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL,
+                           1, &region);
+
+    m_errorMonitor->VerifyFound();
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                                         "Invalid usage flag for image ");
+    vkCmdCopyBufferToImage(m_commandBuffer->GetBufferHandle(), buffer.handle(),
+                           image.handle(), VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL,
+                           1, &region);
+
```
chrisforbes: You need another `VerifyFound` here

chrisforbes: Yep, I like this /much/ better than adding plumbing to hunt through other messages.

mark-lunarg:
Hey Tony, please add a TEST_DESCRIPTION to the test and go ahead and push it...

Tony-LunarG:
Done

PR #475, opened 2016-05-03
Breaking Descriptor Sets into separate class

tobine:
I think this is ready to go pending a bit more testing from LunarG and a final rebase. Updated commit message below.

layers: Rearchitect Descriptor Set validation code
This change pulls all of the DescriptorSet code out of core_validation.cpp and into
its own files/classes in descriptor_set.h/cpp.

See header file for complete class documentation.

These changes pass tri/cube/smoketest --validate.
All related layer validation tests are also updated and passing.
Finally, I ran it through mustpass CTS and did not hit any issues related to these changes.

Going forward this is a general outline for how we would like to evolve core_validation.
That is, we'd like to move the functionality of the checks into reasonable classes and
just have core_validation call into those classes to do the majority of the work.


dustin-lunarg:
There seems to be a file missing from this change: core_validation_types.h

tobine:
@dustin-lunarg force pushed update with missing file.

layers/core_validation.cpp
```diff
@@ -113,7 +111,7 @@ struct layer_data {
     unordered_map<VkCommandPool, CMD_POOL_INFO> commandPoolMap;
     unordered_map<VkDescriptorPool, DESCRIPTOR_POOL_NODE *> descriptorPoolMap;
     unordered_map<VkDescriptorSet, SET_NODE *> setMap;
-    unordered_map<VkDescriptorSetLayout, DescriptorSetLayout *> descriptorSetLayoutMap;
+    unordered_map<VkDescriptorSetLayout, cvdescriptorset::DescriptorSetLayout *> descriptorSetLayoutMap;
```
lentinem: Why cv?

layers/core_validation.cpp
```diff
@@ -2141,7 +2139,7 @@ static bool verify_set_layout_compatibility(layer_data *my_data, const SET_NODE
         return false;
     }
     auto layout_node = my_data->descriptorSetLayoutMap[pipeline_layout_it->second.descriptorSetLayouts[layoutIndex]];
-    return layout_node->IsCompatible(pSet->p_layout, &errorMsg);
+    return pSet->p_ds->IsCompatible(layout_node, &errorMsg);
```
lentinem: I would prefer more descriptive names.

layers/descriptor_sets.cpp
```diff
@@ -0,0 +1,947 @@
...
+ *
+ * Author: Tobin Ehlis <tobine@google.com>
+ */
+
+#include "descriptor_sets.h"
+#include "vk_enum_string_helper.h"
+#include "vk_safe_struct.h"
+#include <sstream>
+
+cvdescriptorset::DescriptorSetLayout::DescriptorSetLayout()
```
lentinem: Do we use this function? I would prefer not having it if we can get away with it.

layers/core_validation.cpp
tobine: "cv" -> "core validation"
DescriptorSetLayout was a bit generic and I anticipate high probably for name collision so the namespace is meant to identify this as core validation's implementation of specialized DescriptorSet* classes.

layers/descriptor_sets.cpp
tobine: Yeah, looks like we can kill it so will do.

layers/core_validation.cpp
tobine: Such as?
"Pipeline Layout Compatibility" is a pervasive spec topic so I based the name off of that.
"Set->IsCompatible(layout..." seems clear. "Set->IsSetCompatibleWithLayout(layout..." seems redundant. I may just be too ingrained in the terms to be objective, though.

lentinem: I mean instead of p_ds i would just say descriptor_set

tobine: Ahh, ok. That makes more sense then. Will do.

mark-lunarg:
Hey Tobin, when I ran this through Dota2 it worked fine but I did notice that before this patch the app would generate (correctly, I believe) thousands of the messages below which are not showing up now.  Might be nothing but I though it was worth a note.

The app is counting on the ICD returning OUT_OF_MEMORY to know when to allocate new descriptors -- this is against the spec but we don't have an app-side fix for it yet.

`DS(ERROR): object: 0x8c86 type: 20 location: 3988 msgCode: 25: Unable to allocate 1 descriptors of type VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER from pool 0x8c88. This pool only has 0 descriptors of this type remaining.`

layers/core_validation.h
```diff
@@ -401,18 +328,21 @@ struct PIPELINE_LAYOUT_NODE {
...
-    GENERIC_HEADER *pUpdateStructs;
-    uint32_t descriptorCount;                   // Total num of descriptors in this set
-    vector<GENERIC_HEADER*> pDescriptorUpdates; // Vector where each index points to update node for its slot
-    DescriptorSetLayout *p_layout;              // DescriptorSetLayout for this set
-    SET_NODE *pNext;
-    unordered_set<VkCommandBuffer> boundCmdBuffers; // Cmd buffers that this set has been bound to
-    SET_NODE()
-        : set(VK_NULL_HANDLE), pool(VK_NULL_HANDLE), pUpdateStructs(nullptr), descriptorCount(0), p_layout(nullptr),
-          pNext(nullptr){};
+    cvdescriptorset::DescriptorSet *descriptor_set;
```
chrisforbes: Do we really need another level of indirection here, or can `cvdescriptorset::DescriptorSet` replace `SET_NODE` entirely? If the indirection is necessary, hold this by `unique_ptr` and remove the destructor.

layers/core_validation_types.h
```diff
@@ -0,0 +1,136 @@
...
+ * Author: Courtney Goeltzenleuchter <courtneygo@google.com>
+ * Author: Tobin Ehlis <tobine@google.com>
+ * Author: Chris Forbes <chrisf@ijw.co.nz>
+ * Author: Mark Lobodzinski <mark@lunarg.com>
+ */
+#ifndef CORE_VALIDATION_TYPES_H_
+#define CORE_VALIDATION_TYPES_H_
+
+// Check for noexcept support
+#if defined(__clang__)
```
chrisforbes: We're accumulating a bunch of copies of this noise now...

layers/descriptor_sets.cpp
```diff
@@ -0,0 +1,942 @@
...
+                    if (descriptors_.at(start_idx + i)->updated) {
+                        image_set->insert(static_cast<ImageDescriptor *>(descriptors_.at(start_idx + i).get())->GetImageView());
+                        num_updates++;
+                    }
+                }
+            } else if (TexelBuffer == descriptors_.at(start_idx)->descriptor_class) {
+                for (uint32_t i = 0; i < p_layout_->GetDescriptorCountFromBinding(binding); ++i) {
+                    if (descriptors_.at(start_idx + i)->updated) {
+                        auto bufferview = static_cast<TexelDescriptor *>(descriptors_.at(start_idx + i).get())->GetBufferView();
+                        auto buffer = buffer_view_map_->at(bufferview).buffer;
```
chrisforbes: What do you do if one of these `at` throws?

layers/descriptor_sets.cpp
```diff
@@ -0,0 +1,942 @@
...
+    descriptor_class = PlainSampler;
+    if (immut) {
+        sampler_ = *immut;
+        immutable_ = true;
+        updated = true;
+    }
+}
+bool cvdescriptorset::ValidateSampler(const VkSampler sampler,
+                                      const std::unordered_map<VkSampler, std::unique_ptr<SAMPLER_NODE>> *sampler_map) {
+    return (sampler_map->count(sampler) != 0) ? true : false;
```
chrisforbes: `return sampler_map->count(sampler) != 0;` please

layers/descriptor_sets.cpp
```diff
@@ -0,0 +1,942 @@
...
+            break;
+        case VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER:
+        case VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC:
+        case VK_DESCRIPTOR_TYPE_STORAGE_BUFFER:
+        case VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC:
+            for (uint32_t di = 0; di < p_layout_->GetDescriptorCountFromIndex(i); ++di)
+                descriptors_.push_back(std::unique_ptr<Descriptor>(new BufferDescriptor(type, buffer_map_)));
+            break;
+        default:
+            break;
```
chrisforbes: Silent default case here is worrying -- if it were ever hit, all subsequent indices get quietly misaligned.

layers/descriptor_sets.h
```diff
@@ -78,195 +81,287 @@
...
+                           const std::unordered_map<VkSampler, std::unique_ptr<SAMPLER_NODE>> *);
+    bool WriteUpdate(const VkWriteDescriptorSet *, const uint32_t, std::string *) override;
+    bool CopyUpdate(const Descriptor *, std::string *) override;
+
+  private:
+    VkSampler sampler_;
+    bool immutable_;
+    VkImageView image_view_;
+    VkImageLayout image_layout_;
+    const std::unordered_map<VkSampler, std::unique_ptr<SAMPLER_NODE>> *sampler_map_;
```
chrisforbes: Do you really need to store ptrs to all these maps inside every descriptor?

tests/layer_validation_tests.cpp
```diff
@@ -2942,6 +2955,9 @@ TEST_F(VkLayerTest, InvalidDynamicOffsetCases) {
     Draw(1, 0, 0, 0);
     m_errorMonitor->VerifyFound();
 
+    vkFreeMemory(m_device->device(), mem, NULL);
```
chrisforbes: These should really be the other way around.

layers/descriptor_sets.cpp
```diff
@@ -0,0 +1,942 @@
...
+                                                              std::unordered_set<VkImageView> *image_set) const {
+    std::unordered_set<uint32_t> binding_set;
+    p_layout_->FillBindingSet(&binding_set);
+    return GetStorageUpdates(binding_set, buffer_set, image_set);
+}
+// Starting at given current_binding with binding_remaining descriptors, parse over update_count
+//  descriptor updates and verify that for any binding boundaries that are crossed, the next binding(s) are all consistent
+//  Consistency means that their type, stage flags, and whether or not they use immutable samplers matches
+//  If so, return true. If not, fill in error_msg and return false
+bool cvdescriptorset::DescriptorSet::VerifyUpdateConsistency(uint32_t current_binding, uint32_t binding_remaining,
```
chrisforbes: This is really awkward. All the callers must call `GetDescriptorCountFromBinding(current_binding)` /and/ you must call it here when you step through to the next. Perhaps it would be better to pass the initial offset in instead, and only call `GetDescriptorCountFromBinding()` here?

chrisforbes: This function might also live better in the descriptor set layout rather than the descriptor set instance -- it doesn't care about any instance state.

layers/descriptor_sets.h
```diff
@@ -78,195 +81,287 @@
...
+    // Return true if any part of set has ever been updated
+    bool IsUpdated() const { return some_update_; };
+    // Return true if the binding at the given global index has been updated
+    bool IsUpdated(const uint32_t global_index) const;
+
+  private:
+    bool ValidateUpdate(const VkWriteDescriptorSet *, const uint32_t, std::string *) const;
+    bool VerifyUpdateConsistency(uint32_t, uint32_t, uint32_t, const char *, std::string *) const;
+    bool some_update_; // has any part of the set ever been updated?
+    bool full_update_; // has every descriptor in the set been updated?
```
chrisforbes: This isn't quite what it says on the tin -- it's `has every descriptor in the set been updated in a single update`. It also appears to be unused.

layers/descriptor_sets.cpp
```diff
@@ -0,0 +1,942 @@
...
+            for (uint32_t di = 0; di < p_layout_->GetDescriptorCountFromIndex(i); ++di)
+                descriptors_.push_back(std::unique_ptr<Descriptor>(new BufferDescriptor(type, buffer_map_)));
+            break;
+        default:
+            break;
+        }
+    }
+}
+// For given global_index return bool of whether or not the underlying descriptor has been updated
+bool cvdescriptorset::DescriptorSet::IsUpdated(const uint32_t global_index) const {
```
chrisforbes: Is this overload actually used anywhere?

layers/descriptor_sets.cpp
```diff
@@ -0,0 +1,942 @@
...
+    if (global_index >= descriptors_.size() || !descriptors_.at(global_index))
```
chrisforbes: When is `descriptors_.at(global_index) == nullptr` ? The above function to add descriptors /never/ inserts `nullptr`.

layers/descriptor_sets.cpp
```diff
@@ -0,0 +1,942 @@
...
+                      << rh_ds_layout->GetStageFlagsFromBinding(binding->binding);
+            *error_msg = error_str.str();
+            return false;
+        }
+    }
+    return true;
+}
+
+bool cvdescriptorset::DescriptorSetLayout::IsNextBindingConsistent(const uint32_t binding) const {
+    if (!binding_to_index_map_.count(binding + 1))
```
chrisforbes: This function would be a lot clearer if you grabbed refs to the two binding structures upfront, rather than having `at` calls spread throughout the logic.

layers/descriptor_sets.cpp
```diff
@@ -0,0 +1,942 @@
...
+      image_to_swapchain_map_(image_to_swapchain_map), swapchain_map_(swapchain_map) {
+    // Foreach binding, create default descriptors of given type
+    for (uint32_t i = 0; i < p_layout_->GetBindingCount(); ++i) {
+        auto type = p_layout_->GetTypeFromIndex(i);
+        switch (type) {
+        case VK_DESCRIPTOR_TYPE_SAMPLER: {
+            auto immut_sampler = p_layout_->GetImmutableSamplerPtrFromIndex(i);
+            for (uint32_t di = 0; di < p_layout_->GetDescriptorCountFromIndex(i); ++di) {
+                if (immut_sampler)
+                    descriptors_.push_back(std::unique_ptr<Descriptor>(new SamplerDescriptor(immut_sampler + di, sampler_map_)));
```
chrisforbes: You could use `emplace_back` here to reduce the noise a little.

layers/core_validation.cpp
```diff
@@ -6447,8 +5904,22 @@ vkAllocateDescriptorSets(VkDevice device, const VkDescriptorSetAllocateInfo *pAl
...
+                    if (log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                                VK_DEBUG_REPORT_OBJECT_TYPE_DESCRIPTOR_SET_LAYOUT_EXT, (uint64_t)pAllocateInfo->pSetLayouts[i],
+                                __LINE__, DRAWSTATE_INVALID_LAYOUT, "DS", "Unable to find set layout node for layout %#" PRIxLEAST64
+                                                                          " specified in vkAllocateDescriptorSets() call",
+                                (uint64_t)pAllocateInfo->pSetLayouts[i])) {
+                        lock.unlock();
+                        return VK_ERROR_VALIDATION_FAILED_EXT;
+                    }
+                }
                 // Create new set node and add to head of pool nodes
```
chrisforbes: This is now misleading.

layers/descriptor_sets.cpp
```diff
@@ -0,0 +1,942 @@
...
+            log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DESCRIPTOR_SET_LAYOUT_EXT,
+                    reinterpret_cast<uint64_t &>(layout_), __LINE__, DRAWSTATE_INVALID_LAYOUT, "DS",
+                    "duplicated binding number in "
+                    "VkDescriptorSetLayoutBinding");
+        }
+        binding_to_global_start_index_map_[p_create_info->pBindings[i].binding] = global_index;
+        global_index += p_create_info->pBindings[i].descriptorCount ? p_create_info->pBindings[i].descriptorCount - 1 : 0;
+        binding_to_global_end_index_map_[p_create_info->pBindings[i].binding] = global_index;
+        global_index++;
+        bindings_.push_back(new safe_VkDescriptorSetLayoutBinding(&p_create_info->pBindings[i]));
```
chrisforbes: Your recent changes made safe_* correctly copy-constructible -- is there a reason these aren't held by value? If you did that, you wouldn't need the destructor.

layers/core_validation.cpp
```diff
@@ -2808,15 +2669,15 @@ static bool validate_and_update_draw_state(layer_data *my_data, GLOBAL_CB_NODE *
                 activeSetBindingsPairs.push_back(std::make_pair(pSet, setBindingPair.second));
                 // Make sure set has been updated if it has no immutable samplers
                 //  If it has immutable samplers, we'll flag error later as needed depending on binding
-                if (!pSet->pUpdateStructs) {
```
chrisforbes: Does this special case really pay for itself still?

layers/core_validation.cpp
```diff
@@ -3803,146 +3438,36 @@ static bool dsUpdate(layer_data *my_data, VkDevice device, uint32_t descriptorWr
...
-                // Layout bindings match w/ update, now verify that update type & stageFlags are the same for entire update
-                if ((skipCall = validateUpdateConsistency(my_data, device, layout_binding->descriptorType, pUpdate, startIndex,
-                                                          endIndex)) == false) {
-                    // The update is within bounds and consistent, but need to
-                    // make sure contents make sense as well
-                    if ((skipCall = validateUpdateContents(my_data, &pWDS[i], layout_binding->pImmutableSamplers)) == false) {
-                        // Update is good. Save the update info
-                        // Create new update struct for this set's shadow copy
-                        GENERIC_HEADER *pNewNode = NULL;
-                        skipCall |= shadowUpdateNode(my_data, device, pUpdate, &pNewNode);
```
chrisforbes: Can the `shadowUpdateNode` function now be removed too?

layers/descriptor_sets.cpp
chrisforbes: This applies everywhere -- if we're going to start using throwing constructs, we'd better have some strategy for containing any exceptions within the layer. We shouldn't expect the loader, other layers, or the client app to be tolerant to random exceptions escaping.

chrisforbes:
With the various inline comments fixed, I think this is more or less OK.

Will note that a single patch +1493/-952 was pretty obnoxious to review, but is probably "too big to reject". This splits things up, but also dramatically changes the data structures for backing descriptor set instances, and fixes a number of bugs that were inherent in the old design (updates overflowing into the next binding; updates that write a subset of a binding; etc). These should be called out in the commit message, and ideally should have tests.

I have general concerns about the size of descriptor instances, and impact on future locking improvements -- there's a /lot/ of non-owning pointers to bits of `layer_data` everywhere now; it may not be obvious in future what the locking requirements are.

layers/descriptor_sets.cpp
chrisforbes: In fact, all subsequent indices do get silently misaligned after a binding with `descriptorCount == 0`, since it consumes one global index, but no ptr is pushed.

layers/descriptor_sets.cpp
```diff
@@ -0,0 +1,942 @@
...
+    auto num_updates = 0;
+    if (!p_layout_->HasBinding(update->dstBinding)) {
+        std::stringstream error_str;
+        error_str << "DescriptorSet " << set_ << " does not have copy update dest binding of " << update->dstBinding << ".";
+        *error = error_str.str();
+        return false;
+    }
+    if (!src_set->HasBinding(update->srcBinding)) {
+        std::stringstream error_str;
+        error_str << "DescriptorSet " << set_ << " does not have copy update src binding of " << update->srcBinding << ".";
```
chrisforbes: `set_` refers to the wrong descriptor set here.

tobine:
@mark-lunarg Interesting note on Dota2 not flagging that error. The code that handles that check is in validate_descriptor_availability_in_pool() function which is not changed with this commit. However, that code did change a couple weeks ago with my DescriptorLayout class change (SHA fce842878e9ddcc7f37e1c457a4b018d52358087).
If possible, can you go back to one commit before that change (SHA b71037a2f223bc54f1fe878d597950fbbc43ea29), test Dota2, and then test with that change? If the messages disappear with that change then I may have introduced a bug there. If it would be possible to capture this behavior of Dota2 in a test, that would be ideal.

mark-lunarg:
Spent a lot more time moving around in the app and I did get those error messages, sorry for the false alarm.  I did see a dota2 crash in `resetCB` (stack pasted in at the bottom) when doing some UI stuff I don't normally mess with.  In `resetCB`,  `pCB->lastBound` contained two entries one of which was all zeros.  This might be uninitialized data or a spot where [] was used for a set lookup with a bad object, but the crash was at this line:
`set_node->second->descriptor_set->RemoveBoundCommandBuffer(pCB->commandBuffer);
`

At this point setMap has about ~13000 entries and about a third look to be invalid, either all zeros or like so:
image

`0xddd...ddd` signifies freed heap memory.  Looks like old/invalid entries are not getting removed from the map?

Stack Trace:
image

layers/core_validation.h
tobine: Yes, you're right, it's easy to just kill SET_NODE so I did.

layers/descriptor_sets.cpp
tobine: Killed all of the at() calls in these classes.
I originally introduced them out of ignorance when I realized I couldn't use "[]" operator or class members of unordered_map type in functions that were declared "const" for the class. Switched the "at()" calls to get a const iter ref w/ "find()", check against "end()" and then reference values off of the iter.
Thanks for helping remove this bit of my ignorance.

layers/descriptor_sets.cpp
tobine: done

layers/descriptor_sets.h
tobine: Probably not. Added a "TODO" comment related to this for time being and will address in future update. I originally had a "Validate()" then "Update()" pattern for calling into descriptors but realized the actions were closely linked and pushed the Validate() down into Update() for each descriptor which requires the map ptrs.
I'll look to improve this in the future, though, as we should just be able to contain maps at the DescriptorSet level.

tests/layer_validation_tests.cpp
tobine: done

layers/descriptor_sets.cpp
tobine: I like this. I'll update function and push down to Layout level.

layers/descriptor_sets.h
tobine: full_update_ killed. I added it early thinking of optimizations but don't actually use it and it's incorrect as you point out.

layers/descriptor_sets.cpp
tobine: Nope, killed. All the functionality that this was intended to expose is contained within the ValidateDrawState() function.

layers/descriptor_sets.cpp
tobine: never, killed

layers/descriptor_sets.cpp
tobine: killed "at()" calls

layers/descriptor_sets.cpp
tobine: done

layers/core_validation.cpp
tobine: good catch. preventing code rot, one comment at a time!

layers/core_validation.cpp
tobine: Yes, I think so. I can probably be convinced otherwise but plan to keep it for now.
At this point in the code we know the set is valid and have to pull it anyway to construct the set/bindings pair. It's only one simple call to check for updated in which case I think an error is most likely as immutable sampler case will be minority. This flags an error as soon as we possibly can.

layers/core_validation.cpp
tobine: yep, killed. Getting core_validation.cpp oh-so-close to 10k lines.

layers/descriptor_sets.cpp
tobine: Only reason is that I didn't have copy-constructible types when this was first implemented and those were added for a different purpose so I hadn't gotten to using them here.
I'll do that now.
Changed to held by value and killed destructor.

tobine:
Pushed a bunch of new CLs on base change as a result of Chris's input. Thanks @chrisforbes 
Doing another round of testing and if all looks good I plan on pushing this in another hour.

chrisforbes:
:+1:

tobine:
@mark-lunarg good find. We weren't removing entries from setMap as we deleted the DescriptorSets ptrs. Just pushed another CL to fix that issue. I'm pretty comfortable with this change now and will try to get it pushed asap.

mark-lunarg:
Hey Tobin, just tried this on Dota2 and am still getting the crash I described this morning, above.  I'd rather not push until it's working -- want me to start looking at it from this side?

tobine:
Yes, please do. Did you definitely have "layers: Fix descriptorSet cleanup bugs" change that I just pushed?

mark-lunarg:
Pulled the new stuff.  That particular issue seems to be OK now but after two runs I've seen a crash in CmdBindPipeline (object was trashed) and now in the destructor for DescriptorSet. I don't think it's quite ready for Master, and will keep at it from here -- how would you like to proceed?

tobine:
I need repros if possible. We can use this as an opportunity to construct or at least plan some tests that will hit these cases. Please do your best to debug and summarize fail cases when you get a good understanding.

Note that I do see one other hole I'm patching now. We invalidate command buffers bound to descriptor sets when vkFreeDescriptorSets() is called, but not when descriptor sets are freed due to their pool being destroyed or reset.

mark-lunarg:
It is croaking (pretty consistently I think) in ResetDescriptorPool, for at least one of the issues.

tobine:
Alright, there was a big bug where I was failing to remove freed sets from the Pool's unordered_set of descriptorSets. That should definitely make things better. Just pushed fix. Please try again.

mark-lunarg:
LGTM :+1:   This has been fun and all, but can I buy y'all a Windows box?

tobine:

Issue #470, opened 2016-05-03
VkLayerTest.EnableWsiBeforeUse triggering SEH exception on Windows


dustin-lunarg:
The EnableWsiBeforeUse test is crashing on Windows due to a combination of an incomplete test and a bug in the swapchain layer.  

* Commit 832f5c4 contains some WSI surface creation code that is #ifdef'd off for some platforms, including Windows, until it can be tested.  The problem is that the other WSI calls are being made with a VK_NULL_HANDLE surface handle, which triggers a crash in the swapchain layer.  The entire test needs to be disabled for the untested platforms.
* See #471 for a description of the swapchain layer bug that produces the crash.

ianelliottus:
Sorry about that!

I have a Windows laptop now, and brought it into work, but I haven't been able to figure out how to run the tests on Windows.  There's a run_all_tests.ps1 file that doesn't run from my Cygwin-bash shell.  I did Start->Run "powershell", and tried to run the script, but I get an error about scripts being turned off.  How can I run the tests on Windows?

dustin-lunarg:
For the PowerShell scripts, you probably need to use Set-ExecutionPolicy to allow scripts to be run: https://technet.microsoft.com/en-us/library/ee176961.aspx

You can also just run vk_layer_validation_tests.exe directly.  I either run the _run_all_tests.ps1 from the build/test directory, or run vk_layer_validation_tests.exe from the VIsualStudio debugger.

ianelliottus:
Hmm.  I don't see any *.exe files under my dbuilt/tests directory.

ianelliottus:
@dustin-lunarg, please try the pull request I just pushed (https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/476) and tell me if that solves the problem for now.

dustin-lunarg:
PR #470 shifts the issue to vkDestroySwapChainKHR.  Similar problem: 
1. vkCreateswapChainKHR skips swap chain creation due to validation error
2. vkGetSwapchainImagesKHR is called with VK_NULL_HANDLE, which adds an invalid entry to the map.
3.  vkDestroySwapChainKHR tries to dereference a pointer obtained from the invalid map entry.

ianelliottus:
[...]

PR #476, opened 2016-05-03
tests: #470 Call WSI surface functions on supported platforms.

ianelliottus:
Windows and Android are currently not supported by the EnableWsiBeforeUse test.
This was previously addressed with #ifdef's for the platform-specific code.
However, this means that the surface variable is VK_NULL_HANDLE for the
unsupported platforms, which is causing problems with Windows debug builds, run
from VisualStudio.  This should ensure those functions aren't called on Windows
and Android.


ianelliottus:
This short-term work-around doesn't fix the problem.  Per discussion in the issue #470, I'm going to fix this correctly.

PR #477, opened 2016-05-03
tests: GH426 Cleanup of dynamic state binding validation tests

### LINKED ISSUE #426: Add tests for dynamic state binding validation cases
This is a top-level issue to capture adding validation tests for dynamic state validation cases. The comprehensive list of validation layer checks is contained in [vk_validation_layer_details.md](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/vk_validation_layer_details.md). Checks that don't currently have tests identified are below. Tests should be added where missing and the document should be updated.

LINE_WIDTH_NOT_BOUND
DEPTH_BIAS_NOT_BOUND
BLEND_NOT_BOUND
DEPTH_BOUNDS_NOT_BOUND
STENCIL_NOT_BOUND
VIEWPORT_SCISSOR_MISMATCH

Note that a number of dynamic state binding tests already exist so the owner of this issue should scrub existing tests in addition to checks above to identify possibility for reuse and any areas where tests already exist but are not correctly documented.

tobine:
All dynamic state binding cases had tests, but were not correctly documented.
This changes merges them all into a single test, DynamicStatesNotBound, to speed
test time and documentation was updated to record that test for the 9 related tests.

Fixes #426


lentinem:
I'm ok with this but in general I prefer smaller tests if they can be separated as when something breaks it's easier to reproduce and trace especially when you are running that test repeatedly.

tobine:

PR #479, opened 2016-05-03
Core_Validation:  BindImageMemory and memory aliasing bug fixes

mark-lunarg:
This was generating errors in some advanced apps we have been running.  The included test is for the memory aliasing problem, though any validation error raised by the old code will cause a crash as the error codes were not passed back up the chain.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5613,65 +5613,15 @@ static bool validateIdleBuffer(const layer_data *my_data, VkBuffer buffer) {
     return skip_call;
 }
 
-VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL
-vkDestroyBuffer(VkDevice device, VkBuffer buffer, const VkAllocationCallbacks *pAllocator) {
-    layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
-    bool skipCall = false;
-    std::unique_lock<std::mutex> lock(global_lock);
-    if (!validateIdleBuffer(dev_data, buffer) && !skipCall) {
-        lock.unlock();
-        dev_data->device_dispatch_table->DestroyBuffer(device, buffer, pAllocator);
-        lock.lock();
-    }
-    dev_data->bufferMap.erase(buffer);
-}
-
-VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL
-vkDestroyBufferView(VkDevice device, VkBufferView bufferView, const VkAllocationCallbacks *pAllocator) {
-    layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
-    dev_data->device_dispatch_table->DestroyBufferView(device, bufferView, pAllocator);
-    std::lock_guard<std::mutex> lock(global_lock);
-    auto item = dev_data->bufferViewMap.find(bufferView);
-    if (item != dev_data->bufferViewMap.end()) {
-        dev_data->bufferViewMap.erase(item);
-    }
-}
-
-VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL vkDestroyImage(VkDevice device, VkImage image, const VkAllocationCallbacks *pAllocator) {
-    layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
-    bool skipCall = false;
-    if (!skipCall)
-        dev_data->device_dispatch_table->DestroyImage(device, image, pAllocator);
-
-    std::lock_guard<std::mutex> lock(global_lock);
-    const auto& entry = dev_data->imageMap.find(image);
-    if (entry != dev_data->imageMap.end()) {
-        // Clear any memory mapping for this image
-        auto mem_entry = dev_data->memObjMap.find(entry->second.mem);
-        if (mem_entry != dev_data->memObjMap.end())
-            mem_entry->second.image = VK_NULL_HANDLE;
-
-        // Remove image from imageMap
-        dev_data->imageMap.erase(entry);
-    }
-    const auto& subEntry = dev_data->imageSubresourceMap.find(image);
-    if (subEntry != dev_data->imageSubresourceMap.end()) {
-        for (const auto& pair : subEntry->second) {
-            dev_data->imageLayoutMap.erase(pair);
-        }
-        dev_data->imageSubresourceMap.erase(subEntry);
-    }
-}
-#if MTMERGESOURCE
 static bool print_memory_range_error(layer_data *dev_data, const uint64_t object_handle, const uint64_t other_handle,
                                      VkDebugReportObjectTypeEXT object_type) {
     if (object_type == VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT) {
         return log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, object_type, object_handle, 0,
-                       MEMTRACK_INVALID_ALIASING, "MEM", "Buffer %" PRIx64 " is alised with image %" PRIx64, object_handle,
+                       MEMTRACK_INVALID_ALIASING, "MEM", "Buffer %" PRIx64 " is aliased with image %" PRIx64, object_handle,
                        other_handle);
     } else {
         return log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, object_type, object_handle, 0,
-                       MEMTRACK_INVALID_ALIASING, "MEM", "Image %" PRIx64 " is alised with buffer %" PRIx64, object_handle,
+                       MEMTRACK_INVALID_ALIASING, "MEM", "Image %" PRIx64 " is aliased with buffer %" PRIx64, object_handle,
                        other_handle);
     }
 }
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -5611,6 +5660,10 @@ vkDestroyBuffer(VkDevice device, VkBuffer buffer, const VkAllocationCallbacks *p
         dev_data->device_dispatch_table->DestroyBuffer(device, buffer, pAllocator);
         lock.lock();
     }
+
+    VkDeviceMemory mem = dev_data->bufferMap[buffer].mem;
+    remove_memory_ranges(reinterpret_cast<uint64_t &>(buffer), mem, dev_data->memObjMap[mem].bufferRanges);
```
tobine: Is it safe to use this buffer object after call down chain to DestroyBuffer?

Also, for map look-ups where we're unsure if key is in map, we should move to at() function such as "dev_data->bufferMap.at(buffer).mem" as [] operator can insert new elements. Applies to both these lines, I believe. If we know that a key is in map, should use [] operator as it's faster.

layers/core_validation.cpp
```diff
@@ -5632,7 +5685,11 @@ VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL vkDestroyImage(VkDevice device, VkIma
         dev_data->device_dispatch_table->DestroyImage(device, image, pAllocator);
 
     std::lock_guard<std::mutex> lock(global_lock);
-    const auto& entry = dev_data->imageMap.find(image);
+    VkDeviceMemory mem = dev_data->imageMap[image].mem;
+    remove_memory_ranges(reinterpret_cast<uint64_t &>(image), mem, dev_data->memObjMap[mem].imageRanges);
+    clear_object_binding(dev_data, reinterpret_cast<uint64_t &>(image), VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT);
+
```
tobine: Should move these 3 lines below the "if" check below when we know that "image" is in imageMap

layers/core_validation.cpp
```diff
@@ -9911,15 +9931,18 @@ VKAPI_ATTR VkResult VKAPI_CALL vkBindImageMemory(VkDevice device, VkImage image,
...
         lock.lock();
-        skipCall |= validate_buffer_image_aliasing(dev_data, image_handle, mem, memoryOffset, memRequirements,
-                                                   dev_data->memObjMap[mem].imageRanges, dev_data->memObjMap[mem].bufferRanges,
-                                                   VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT);
+
+        const MEMORY_RANGE range =
+            insert_memory_ranges(image_handle, mem, memoryOffset, memRequirements, dev_data->memObjMap[mem].imageRanges);
+        skipCall |=
+            validate_memory_range(dev_data, dev_data->memObjMap[mem].bufferRanges, range, VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT);
+
```
tobine: Potentially guard these calls to make sure mem is in memObjMap such as:

    if (dev_data->memObjMap.find(mem) != dev_data->memObjMap.end())

layers/core_validation.cpp
```diff
@@ -5650,48 +5707,8 @@ VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL vkDestroyImage(VkDevice device, VkIma
...
-    MEMORY_RANGE range;
-    range.handle = handle;
-    range.memory = mem;
-    range.start = memoryOffset;
-    range.end = memoryOffset + memRequirements.size - 1;
-    ranges.push_back(range);
-    return validate_memory_range(dev_data, other_ranges, range, object_type);
-}
 
+#if MTMERGESOURCE
```
tobine: Can this either be removed now or moved down below your updates? I'd like to get rid of all the MTMERGESOURCE references (code from original mem_tracker) as we feel that the code within them has been updated and there is not any duplicated code in those blocks vs the other code that was in draw_state before the merge.

layers/core_validation.cpp
mark-lunarg: Yes, thanks!

layers/core_validation.cpp
mark-lunarg: Happy to remove them.  I'll do it as a separate commit though.

layers/core_validation.cpp
mark-lunarg: Buffer is safe to use after calling DestroyBuffer as it is only used as a key or index in these routines.  I think we do know that the buffer is in the map -- we'd never have made it this far in this layer, or OT should have complained previously.

tobine:
I'm sorry, Mark, I was unclear in my comment regarding MTMERGESOURCE. I use those to track where merged code has been cut-and-pasted vs. thoroughly reviewed and intelligently merged. If you've made any changes inside/around such MTMERGESOURCE blocks and you feel that there is not redundancy in that code area, please remove the MTMERGESOURCE block from just that code.

I want all of the other MTMERGESOURCE blocks to remain for now to make sure those areas get proper attention before removing the block. There are definitely some merged areas that still need to be cleaned up.

layers/core_validation.cpp
```diff
@@ -5623,7 +5660,16 @@ vkDestroyBuffer(VkDevice device, VkBuffer buffer, const VkAllocationCallbacks *p
...
+    // Clean up memory binding and range information for buffer
+    auto bufferEntry = dev_data->bufferMap.find(buffer);
+    if (bufferEntry != dev_data->bufferMap.end()) {
+        auto memEntry = dev_data->memObjMap.find(bufferEntry->second.mem);
+        if (memEntry != dev_data->memObjMap.end()) {
+            remove_memory_ranges(reinterpret_cast<uint64_t &>(buffer), bufferEntry->second.mem, memEntry->second.bufferRanges);
+        }
+        clear_object_binding(dev_data, reinterpret_cast<uint64_t &>(buffer), VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT);
+        dev_data->bufferMap.erase(bufferEntry);
+    }
```
tobine: This is nice. Only single map look-ups are made where needed and then the result of those look-ups are used when valid.

layers/core_validation.cpp
```diff
@@ -5640,19 +5686,22 @@ vkDestroyBufferView(VkDevice device, VkBufferView bufferView, const VkAllocation
...
-
+    const auto &imageEntry = dev_data->imageMap.find(image);
+    if (imageEntry != dev_data->imageMap.end()) {
+        // Clean up memory mapping, bindings and range references for image
+        auto memEntry = dev_data->memObjMap.find(imageEntry->second.mem);
+        if (memEntry != dev_data->memObjMap.end()) {
+            remove_memory_ranges(reinterpret_cast<uint64_t &>(image), imageEntry->second.mem, memEntry->second.imageRanges);
+            clear_object_binding(dev_data, reinterpret_cast<uint64_t &>(image), VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT);
+            memEntry->second.image = VK_NULL_HANDLE;
+        }
```
tobine: same here, looks good

layers/core_validation.cpp
```diff
@@ -5718,9 +5726,14 @@ vkBindBufferMemory(VkDevice device, VkBuffer buffer, VkDeviceMemory mem, VkDevic
...
-                                                   dev_data->memObjMap[mem].bufferRanges, dev_data->memObjMap[mem].imageRanges,
-                                                   VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT);
+
+        // Track and validate bound memory range information
+        if (dev_data->memObjMap.find(mem) != dev_data->memObjMap.end()) {
+            const MEMORY_RANGE range =
+                insert_memory_ranges(buffer_handle, mem, memoryOffset, memRequirements, dev_data->memObjMap[mem].bufferRanges);
+            skipCall |= validate_memory_range(dev_data, dev_data->memObjMap[mem].imageRanges, range,
+                                              VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT);
+        }
```
tobine: I'd like to see this code change to match code I commented on above. find() outside the if check, then use results of the find (if valid) inside the if check to prevent any further map look-ups.

layers/core_validation.cpp
tobine: Would like to see this block of code similar to case from 5694-5702 above.

layers/core_validation.cpp
```diff
@@ -5623,7 +5672,16 @@ vkDestroyBuffer(VkDevice device, VkBuffer buffer, const VkAllocationCallbacks *p
         dev_data->device_dispatch_table->DestroyBuffer(device, buffer, pAllocator);
         lock.lock();
     }
-    dev_data->bufferMap.erase(buffer);
+    // Clean up memory binding and range information for buffer
+    auto &bufferEntry = dev_data->bufferMap.find(buffer);
```
tobine: missing const

layers/core_validation.cpp
```diff
@@ -5623,7 +5672,16 @@ vkDestroyBuffer(VkDevice device, VkBuffer buffer, const VkAllocationCallbacks *p
...
+    if (bufferEntry != dev_data->bufferMap.end()) {
+        auto &memEntry = dev_data->memObjMap.find(bufferEntry->second.mem);
```
tobine: missing const

layers/core_validation.cpp
```diff
@@ -5718,9 +5739,16 @@ vkBindBufferMemory(VkDevice device, VkBuffer buffer, VkDeviceMemory mem, VkDevic
         buffer_node->second.mem = mem;
         VkMemoryRequirements memRequirements;
         dev_data->device_dispatch_table->GetBufferMemoryRequirements(device, buffer, &memRequirements);
-        skipCall |= validate_buffer_image_aliasing(dev_data, buffer_handle, mem, memoryOffset, memRequirements,
-                                                   dev_data->memObjMap[mem].bufferRanges, dev_data->memObjMap[mem].imageRanges,
-                                                   VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT);
+
+        // Track and validate bound memory range information
+        auto &memEntry = dev_data->memObjMap.find(mem);
```
tobine: missing const

layers/core_validation.cpp
```diff
@@ -9921,15 +9949,22 @@ VKAPI_ATTR VkResult VKAPI_CALL vkBindImageMemory(VkDevice device, VkImage image,
...
         VkMemoryRequirements memRequirements;
         lock.unlock();
         dev_data->device_dispatch_table->GetImageMemoryRequirements(device, image, &memRequirements);
         lock.lock();
-        skipCall |= validate_buffer_image_aliasing(dev_data, image_handle, mem, memoryOffset, memRequirements,
-                                                   dev_data->memObjMap[mem].imageRanges, dev_data->memObjMap[mem].bufferRanges,
-                                                   VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT);
+
+        // Track and validate bound memory range information
+        auto &memEntry = dev_data->memObjMap.find(mem);
```
tobine: missing const

tobine:
Your changes didn't compile due to missing const on 4 "auto &" decls that have been noted for educational purposes. I went ahead and applied a fix-up for those issues.

Issue #471, opened 2016-05-03
Swapchain layer inserting invalid std::unordered_map entries


dustin-lunarg:
The swapchain layer calls std::unordered_map::operator[] to retrieve records associated with handles without first checking to see if an entry for the handle exists in the map. When no entry is present for the handle, operator[] inserts a new entry with a value-initialized record. This can lead to an invalid pointer dereference with Windows Debug builds, where the memory is initialized with the 0xcdcdcdcd debug pattern. The operator[] calls need to replaced with find() to avoid adding unwanted entries to the map, and the value returned by find() needs to be checked against end() to determine if it was present before it is used.

ianelliottus:
Can you give an example of how I can reproduce this, please?

dustin-lunarg:
You can reproduce it by providing any unrecognized or invalid handle, such as VK_NULL_HANDLE, to the swapchain layer's vkDestroySrufaceKHR function.  The VkLayerTest.EnableWsiBeforeUse test will currently reproduce this error on Windows.

To see what is happening you can call vkDestroySurfaceKHR(instance, 0 , 0) without calling any other WSI functions and set a break point on line [861](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/swapchain.cpp#L861) of the current code.  At break, you will see that my_data->surfaceMap has a size of 0.  After stepping over this line, you will see the size increase to 1.

Basically, you don't want to use operator[] to search for an item in a map, as it has the side effect of inserting new entries for values it does not find.

ianelliottus:
I just created pull request #482 (see: https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/482).  It addresses this and other issues.

jon-lunarg:
This issue has been fixed AFAIK, closing

PR #482, opened 2016-05-04
Swapchain validation layer fixes, and turn on EnableWsiBeforeUse test for Windows

ianelliottus:
This addresses 3 sets of changes:

1) Fixes the way "count" variables are tested in the swapchain layer.

2) Fix the swapchain layer for issue #471.  The layer now uses find()/end() for looking up values that should be in an unordered map, and gets NULL if the value isn't there.

3) Fully turns on EnableWsiBeforeUse test for Windows (which couldn't be done until #471 was fixed).

### DIFF
--- a/layers/swapchain.cpp
+++ b/layers/swapchain.cpp
@@ -85,7 +85,11 @@ static void createDeviceRegisterExtensions(VkPhysicalDevice physicalDevice, cons
     pDisp->QueuePresentKHR = (PFN_vkQueuePresentKHR)gpa(device, "vkQueuePresentKHR");
     pDisp->GetDeviceQueue = (PFN_vkGetDeviceQueue)gpa(device, "vkGetDeviceQueue");
 
-    SwpPhysicalDevice *pPhysicalDevice = &my_instance_data->physicalDeviceMap[physicalDevice];
+    SwpPhysicalDevice *pPhysicalDevice = NULL;
+    {
+        auto it = my_instance_data->physicalDeviceMap.find(physicalDevice);
+        pPhysicalDevice = (it == my_instance_data->physicalDeviceMap.end()) ? NULL : &it->second;
+    }
     if (pPhysicalDevice) {
         my_device_data->deviceMap[device].pPhysicalDevice = pPhysicalDevice;
         pPhysicalDevice->pDevice = &my_device_data->deviceMap[device];
@@ -290,7 +294,11 @@ vkCreateInstance(const VkInstanceCreateInfo *pCreateInfo, const VkAllocationCall
 VK_LAYER_EXPORT VKAPI_ATTR void VKAPI_CALL vkDestroyInstance(VkInstance instance, const VkAllocationCallbacks *pAllocator) {
     dispatch_key key = get_dispatch_key(instance);
     layer_data *my_data = get_my_data_ptr(key, layer_data_map);
-    SwpInstance *pInstance = &(my_data->instanceMap[instance]);
+    SwpInstance *pInstance = NULL;
+    {
+        auto it = my_data->instanceMap.find(instance);
+        pInstance = (it == my_data->instanceMap.end()) ? NULL : &it->second;
+    }
 
     // Call down the call chain:
     my_data->instance_dispatch_table->DestroyInstance(instance, pAllocator);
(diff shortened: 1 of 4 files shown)


layers/swapchain.cpp
```diff
@@ -85,7 +85,9 @@ static void createDeviceRegisterExtensions(VkPhysicalDevice physicalDevice, cons
     pDisp->QueuePresentKHR = (PFN_vkQueuePresentKHR)gpa(device, "vkQueuePresentKHR");
     pDisp->GetDeviceQueue = (PFN_vkGetDeviceQueue)gpa(device, "vkGetDeviceQueue");
 
-    SwpPhysicalDevice *pPhysicalDevice = &my_instance_data->physicalDeviceMap[physicalDevice];
+    DECLARE_AND_FIND_IN_MAP(SwpPhysicalDevice, pPhysicalDevice,
```
mark-lunarg: We're trying to cut down/eliminate macros in the layers and had plans to remove the existing ones in swapchain.  The layers are getting redesigned to be consistent and streamlined and this will make that harder -- are there special circumstances that should set swapchain apart in this effort?

ianelliottus: Hi @mark-lunarg.  Sorry I didn't see this last night.

I used a macro because it seemed that I went from a relatively-smaller amount of simple code, to a larger amount of more-complex code, that was repeated all over the place.  It just seemed like what a macro is good for.  I can replace the macros with the logic if that's what you folks want.  Please advise, and I'll get to it this morning.

mark-lunarg: Thanks Ian.  We had previously agreed to stick to the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) as much as possible and it's our long-term goal to have all the layers looking clean and consistent, so that's what we'd prefer to shoot for.

lentinem: The general reason is that most things can be done without macros and avoid common pitfalls with things like scoping. Ian, is there a reason a function can't be used for this?

ianelliottus: @lentinem, I started off thinking that all cases were going to involve converting cases of 1 line of code that both declared and initialized a variable, to multiple lines of code that would do essentially the same thing.  The declaration has to be in the scope of the line that I was replacing.  I could create a templatized function to set the variable.  Is that what you prefer?

I'm sorry that I didn't pick-up earlier on what's considered clean in the layers:-).  I'm just an occasional layer writer:-).

ianelliottus: @lentinem and @mark-lunarg, I'll aim to please, but I'm not sure whether you'd both prefer that I create a templatized function, or just replace the macros with in-line code.  Please advise.  Thanks!

mark-lunarg: In lieu of replacing all of these calls in all of the layers I'd say just use the in-line code so that they are more consistent.

ianelliottus: Okay, I'll do that.  Thanks!

ianelliottus: @mark-lunarg, another preference question: do you want me to add a commit to this pull request, or replace the previous request with one that in-lines the previous-macro code?

mark-lunarg:
Hey Ian, you can just make the changes to your existing branch and force push to ian_160504a_FixSwapchainLayerAndTest.  Github will understand

ianelliottus:
Will do.

ianelliottus:
All squashed and pushed

ianelliottus:
Sorry @mark-lunarg.  I meant I force-pushed my new commit to the ian_160504a_FixSwapchainLayerAndTest branch.  I haven't pushed it to master yet.  Does this version look okay to you?  Does it check out okay?

layers/swapchain.cpp
```diff
@@ -1528,7 +1674,11 @@ static bool validateCreateSwapchainKHR(VkDevice device, const VkSwapchainCreateI
 
     // Validate pCreateInfo->oldSwapchain:
     if (pCreateInfo && pCreateInfo->oldSwapchain) {
-        SwpSwapchain *pOldSwapchain = &my_data->swapchainMap[pCreateInfo->oldSwapchain];
+        SwpSwapchain *pOldSwapchain = NULL;
+        {
+          auto it = my_data->swapchainMap.find(pCreateInfo->oldSwapchain);
```
mark-lunarg: Has this stuff taken a trip through clang-format?

mark-lunarg:
Looks OK...

ianelliottus:
Thanks.  I cherry-picked and pushed the commits to master.

PR #483, opened 2016-05-04
Sparse color attachments

chrisforbes:
Previously we were broken in various ways if a color attachment reference were VK_ATTACHMENT_UNUSED


tobine:

PR #486, opened 2016-05-05
Add some missing spaces in messages

baldurk:
There are a couple of error messages passed to debug report that have missing spaces that are kind of hidden because the string literals are broken up into multiple literals on different lines.


ianelliottus:
Looks good to me.  @tobin, are you okay with this as well?

ianelliottus:
Per LunarG process, I cherry-picked and pushed this change to the master branch (modifying your commit message a little, but keeping the old one as another line).

PR #489, opened 2016-05-05
Fix, that decrement resources counter do not get negative.

McNopper:
The decrement resource counter do get negative in specific cases, resulting in wrong reported errors and warnings.

The current code change does fix this by checking the counter for not being zero.

Also, a fence is kept in its state and not forced into a signaled satet, which also resulted in an error.

The behaviour can be reproduced with this example:
https://github.com/McNopper/Vulkan/tree/master/VKTS_Example09


lentinem:
-1. I would prefer fixing this by not calling decrement on the same cmdbuffer twice which is being looked at. In any other case there should be an equal number of increments and decrements.

McNopper:
Hmm, it is valid to call e.g. vkGetFenceStatus several times and if it is signaled, it would reduce the counter every time.
Do you fix it or should I create two issues?

lentinem:
That's a good question. I don't think it's wrong to do that but it is a bit odd to me. It should be an easy fix to remove the command buffers from the fence list after they are processed. I'll upload a fix for that one shortly.

McNopper:
If you look at my example 9 - just the update loop - you will see that e.g. the fence is decremented to many times because of vkWaitForFences,  vkQueueWaitIdle and the submits. And I am not using vkGetFenceStatus there.
Also, in a complex program you never know if a function (e.g. vkGetFenceStatus) is called many times - even it does maybe make no sense.

Anyway, thanks for fixing this.

lentinem:
Yea I agree, there is an open issue related to the multiple queuewait calls that is being investigated. I submitted the fix for getfencestatus as well.

McNopper:
Does not occur in master branch anymore.

PR #490, opened 2016-05-05
layers: Fix core_validation issues GH103 and GH240

MarkY-LunarG:
Perform pipeline sample and attachment count validation at draw
time instead of vkCmdNextSubpass() time.  Also validate that
an active renderpass exists.

Change-Id: I912947287eef29d532519220750c3a93a3554565


chrisforbes:
I see that @tobine pushed this to master but didn't close the PR. Doing that now.

PR #491, opened 2016-05-05
Clear cmdbufs and prior fences after processed in decrement.

lentinem:


tobine:

PR #492, opened 2016-05-05
tests: Add tests focussed on the Image validation layer

Tony-LunarG:
Adding 4 new layer validation tests focussed on the image validation layer.  Tested on 16.04 Mesa and Win 7 Nvidia.


tests/layer_validation_tests.cpp
```diff
@@ -8103,6 +8103,349 @@ TEST_F(VkLayerTest, CopyImageLayerCountMismatch) {
...
+            break;
+        }
+    }
+    if (unsupported != VK_FORMAT_UNDEFINED) {
+        m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                                             "vkCreateImage parameter, "
+                                             "VkFormat pCreateInfo->format, "
+                                             "contains unsupported format");
+        // Create image with unsupported format - Expect FORMAT_UNSUPPORTED
+        VkImageCreateInfo imgInfo = image.create_info();
```
chrisforbes: Creating another image just to use its CreateInfo is a bit obtuse.

tests/layer_validation_tests.cpp
```diff
@@ -8103,6 +8103,349 @@ TEST_F(VkLayerTest, CopyImageLayerCountMismatch) {
...
+        att.finalLayout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL;
+
+        VkRenderPassCreateInfo rp_info = {};
+        VkRenderPass rp;
+        rp_info.sType = VK_STRUCTURE_TYPE_RENDER_PASS_CREATE_INFO;
+        rp_info.attachmentCount = 1;
+        rp_info.pAttachments = &att;
+        rp_info.subpassCount = 0;
+        rp_info.pSubpasses = NULL;
+        vkCreateRenderPass(m_device->handle(), &rp_info, NULL, &rp);
```
chrisforbes: I don't think this part of the test is reasonable:

I don't see any requirement in the spec that attachment descriptions (even for attachments consumed by some subpass, which this isn't) have a format that's actually supported. The spec only requires that it be a valid `VkFormat` value.

Tony-LunarG:
So are you saying that the corresponding check in the image layer is invalid?  Should I remove both the test and the validation?

tests/layer_validation_tests.cpp
lentinem: I would expect that the later (consumed) is an oversight in the spec. If the format is not supported the attachment cannot be used by the render pass. I would instead of removing it file a bug with the spec.

chrisforbes: @lentinem, I don't think there's really much of a gap. The requirement is enforced through a different path:
- Image creation requires a supported combination of usage & format
- Framebuffer creation requires correct usage flags on images used

lentinem: Sure, but I still can't think of a reason why we would allow the creation of a pass with attachments using an unsupported format. If there is a use case here then we should remove the check and this test but it's difficult for me to think of one.

chrisforbes:
@Tony-LunarG after talking with @lentinem about this, I think it's fine as-is. We might have to adjust it later when the layer gets more precise about validating framebuffer and renderpass creation.

Tony-LunarG:
Ok thanks - I fixed the create info and left the renderpass test as is.

Issue #432, opened 2016-04-27
Add tests for image layout validation


tobine:
Currently no test(s) are identified for INVALID_IMAGE_LAYOUT case in [vk_validation_layer_details.md](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/vk_validation_layer_details.md).

This case should be examined according to the existing validation code and the Vulkan specification and test(s) should be added as appropriate.

tobine:
Fix was cherry-picked and pushed to master

PR #493, opened 2016-05-06
tests: Add InvalidImageLayout tests

tobine:
Addresses #432
These hit 16/18 cases for INVALID_IMAGE_LAYOUT enum in core_validation.
The other 2 cases were already hit by object_tracker and another test.


mark-lunarg:
+1

tests/layer_validation_tests.cpp
```diff
@@ -6142,6 +6142,177 @@ TEST_F(VkLayerTest, VtxBufferBadIndex) {
...
+    rpci.subpassCount = 1;
+    rpci.pSubpasses = &subpass;
+    rpci.attachmentCount = 1;
+    VkAttachmentDescription attach_desc = {};
+    attach_desc.format = VK_FORMAT_UNDEFINED;
+    rpci.pAttachments = &attach_desc;
+    VkRenderPass rp;
+    m_errorMonitor->SetDesiredFailureMsg(
+        VK_DEBUG_REPORT_PERFORMANCE_WARNING_BIT_EXT,
+        "Layout for input attachment is GENERAL but should be READ_ONLY_OPTIMAL.");
```
chrisforbes: I'm not sure about complaining about attachments in GENERAL layout. The current code isn't smart enough to /not/ emit the perf warning in cases where GENERAL is the only valid layout -- attachment used as both input attachment and color attachment, for example.

tobine: This set of tests is about hitting the cases that are in place.
You bring up a good point about perf warning in GENERAL case, but I believe it's separate from this test commit.
I'll file a separate ? issue where we can discuss what to do with these GENERAL layout perf warnings.

tobine:
Filed #508 to address @chrisforbes point regarding GENERAL layout perf warning.

tobine:

Issue #371, opened 2016-04-15
all validation layers need to use app provided memory allocators


karl-lunarg:
Moved from gitlab 112

Many object creation calls provide memory allocation callbacks. Validation layers should be using these callbacks when provided.

chrisforbes:
This is pretty nasty with all the STL use...

tobine:
This is not required. See [LoaderAndLayerInterface.md](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/loader/LoaderAndLayerInterface.md#layer-conventions-and-rules)

When a layer requires host memory, it is free to scope the allocations to itself, bypassing the provided allocators entirely.

tobine:
Keeping this open amid ongoing discussion about whether or not this should be a requirement. Current priority is low regardless so leaving un-assigned for time being.

jon-lunarg:
Resolution is that layers can choose to ignore the allocators or not. Recommendation is that layers intended to be run in production environments should use the allocators if they are provided.  Thus validation layers can ignore the allocators if they wish to.

PR #495, opened 2016-05-06
doc: allow layers to ignore pAllocator

olvaffe:


loader/LoaderAndLayerInterface.md
```diff
@@ -1115,6 +1115,10 @@ command to wrap the objects created by its lower layers, it must make sure its
 lower layers never see the wrapping objects, directly from itself or
 indirectly from its upper layers.
 
+When a layer requires host memory, it may ignore the provided allocators.  It
+is recommended to use the allocators when it is expected to be used as an
+implicit layer that is always enabled.
```
courtney-g: How about:
... Layers should use allocators if the layer is intended to run in a production environment, such as an implicit layer that is always enabled. That will allow applications to include the layer's memory usage.

olvaffe:
That is much better.  Updated.

ianelliottus:
+1.  LGTM

tobine:
LGTM :+1:
Did this get Khronos workgroup agreement? Can we close #371?

courtney-g:
We discussed during the WSI call. We should leave this open and make it an agenda topic for next Tuesday. We can then review the findings of the WSI group, get agreement from the rest of the WG and then be done.

jon-lunarg:
committed to master and closing #371

PR #496, opened 2016-05-06
Update VK_LAYER_LUNARG_swapchain

olvaffe:
Similar to pull request #443, but for VK_LAYER_LUNARG_swapchain.  Mainly, reorganize the code to move layer commands into a namesapce and improve vkGetInstanceProcAddr.


layers/swapchain.cpp
```diff
@@ -29,13 +29,13 @@
...
 #include "vk_layer_utils.h"
 
+namespace swapchain {
+
 static std::mutex global_lock;
 
 // The following is for logging error messages:
 static std::unordered_map<void *, layer_data *> layer_data_map;
 
-template layer_data *get_my_data_ptr<layer_data>(void *data_key, std::unordered_map<void *, layer_data *> &data_map);
```
courtney-g: where did this code go? Is it now unused?

layers/swapchain.cpp
```diff
@@ -2292,7 +2293,8 @@ vkEnumerateDeviceLayerProperties(VkPhysicalDevice physicalDevice, uint32_t *pCou
 VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkEnumerateDeviceExtensionProperties(VkPhysicalDevice physicalDevice,
                                                                                     const char *pLayerName, uint32_t *pCount,
                                                                                     VkExtensionProperties *pProperties) {
-    return swapchain::EnumerateDeviceExtensionProperties(physicalDevice, pLayerName, pCount, pProperties);
+    // the layer command handles VK_NULL_HANDLE just fine
+    return swapchain::EnumerateDeviceExtensionProperties(VK_NULL_HANDLE, pLayerName, pCount, pProperties);
```
courtney-g: Why aren't we using physicalDevice in the call to swapchain::EnumerateDeviceExtensionProperties?

layers/swapchain.cpp
```diff
@@ -2196,54 +2185,85 @@ VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL GetInstanceProcAddr(VkInstance instance
...
+        { "vkGetPhysicalDeviceSurfaceSupportKHR", reinterpret_cast<PFN_vkVoidFunction>(GetPhysicalDeviceSurfaceSupportKHR) },
+        { "vkGetPhysicalDeviceSurfaceCapabilitiesKHR", reinterpret_cast<PFN_vkVoidFunction>(GetPhysicalDeviceSurfaceCapabilitiesKHR) },
+        { "vkGetPhysicalDeviceSurfaceFormatsKHR", reinterpret_cast<PFN_vkVoidFunction>(GetPhysicalDeviceSurfaceFormatsKHR) },
+        { "vkGetPhysicalDeviceSurfacePresentModesKHR", reinterpret_cast<PFN_vkVoidFunction>(GetPhysicalDeviceSurfacePresentModesKHR) },
+    };
 
-    if (pTable->GetInstanceProcAddr == NULL)
-        return NULL;
-    return pTable->GetInstanceProcAddr(instance, funcName);
+    // do not check if VK_KHR_*_surface is enabled (why?)
```
courtney-g: That check is done at the CreateSurface call. If we checked here and didn't hook in swapchain it would be possible to call WSI functions and no one would know that WSI hadn't been enabled.

olvaffe: I would expect the extension to be checked here.  A warning should be logged if the extension is not enabled and the app will get NULL.

layers/swapchain.cpp
olvaffe: This is called by the loader at layer discovery phase.  physicalDevice is always NULL.  Even if it is non-NULL (a loader bug), we do not want to call to other layers accidentally.

layers/swapchain.cpp
olvaffe: Yes, it was never(?) needed.

layers/swapchain.cpp
courtney-g: If we returned NULL when the extension wasn't set we ran into a problem that validation couldn't detect. That may no longer be an issue with recent changes in the loader, but wouldn't hurt to keep the code in the layer as well.
The WSI extensions are an exception to the rule that only core API entrypoints are available to be linked to the application. So, it's possible to link to vkCreateSurfaceKHR and call it without enabling the extension. On older loaders this call would succeed and not get a validation error. We fixed the validation issue by having swapchain always hook in the WSI calls and then check in the call that the extension had been enabled. I think we still want that behavior in the swapchain validation layer.

olvaffe: That is good to know.  I do not introduce any behavior change here.  I also have no intention to introduce any behavior change in these pull requests except those explicitly pointed out in the commit messages.

But I think the fix is at the wrong place.  A loader is not allowed to call

  vkGetInstanceProcAddr(isntance, "vkCreateXxxSurfaceKHR")

in the first place when the extension is not enabled.  The loader's statically exported vkCreateXxxSurfaceKHR must do something when it is called without enabling the extension.

courtney-g: Hmm. I don't really like "A loader is not allowed to call ...".
The loader, or any app, can call GetInstanceProcAddr with any string it wants. However, only some will return a non-null function pointer.

We want the app to receive a notification callback if they call a WSI function without enabling WSI. This would be easy if the only way to get those functions was via vkGet*ProcAddr, but because of the special linking exception we have to deal with an app that can call a WSI function without enabling the extension. What would you recommend?

olvaffe: Sorry.  A loader is _allowed_ to make the call, but it should expect that may generate a debug_report error and return an invalid function pointer.  Something should be done in the loader's statically exported vkCreateXxxSurfaceKHR.

olvaffe:
Any objection?  I will merge this tomorrow unless I hear otherwise.

olvaffe:
Rebased and pushed to master.

PR #497, opened 2016-05-06
Update VK_LAYER_LUNARG_core_validation

olvaffe:
Similar to pull request #443, but for VK_LAYER_LUNARG_core_validation. Mainly, reorganize the code to move layer commands into a namesapce and improve vkGetInstanceProcAddr.


layers/core_validation.cpp
```diff
@@ -10043,18 +10044,57 @@ VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vkGetInstanceProcAddr(V
     return pTable->GetInstanceProcAddr(instance, funcName);
 }
 
+} // namespace core_validation
+
+// vk_layer_logging.h expects these to be defined
+
+VKAPI_ATTR VkResult VKAPI_CALL
+vkCreateDebugReportCallbackEXT(VkInstance instance, const VkDebugReportCallbackCreateInfoEXT *pCreateInfo,
```
courtney-g: What required us to leave the DebugReport entrypoints visible in the layer?
I think all the queries are done via GetInstanceProcAddr and thus wouldn't need the vk entrypoints in the layer. That would simplify things a little.

courtney-g:
+1

layers/core_validation.cpp
```diff
@@ -9804,6 +9789,31 @@ VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL GetInstanceProcAddr(VkInstance instance
...
+        { "vkEnumerateInstanceExtensionProperties", reinterpret_cast<PFN_vkVoidFunction>(vkEnumerateInstanceExtensionProperties) },
+        { "vkEnumerateDeviceLayerProperties", reinterpret_cast<PFN_vkVoidFunction>(vkEnumerateDeviceLayerProperties) },
+        { "vkEnumerateDeviceExtensionProperties", reinterpret_cast<PFN_vkVoidFunction>(EnumerateDeviceExtensionProperties) },
+    };
+
+    for (size_t i = 0; i < ARRAY_SIZE(core_instance_commands); i++) {
+        if (!strcmp(core_instance_commands[i].name, name))
+            return core_instance_commands[i].proc;
+    }
+
```
tobine: We could migrate these various intercept_* functions to use an unordered_map. Makes the look-up a little cleaner. Not sure of my preference, though, and current code is fine. Just putting the thought out there for future consideration.

tobine:
LGTM :+1:

layers/core_validation.cpp
olvaffe: They are required by debug_report_get_instance_proc_addr defined in vk_layer_logging.h.  They need to be defined, but not necessarily exported.  After all layers are updated, I plan to either add -fvisibility=hidden or enclose all but interface functions in an anonymous namespace.

courtney-g: I see. Hiding the entries sounds good.

tobine:

PR #501, opened 2016-05-06
layers: GH494 Add constructors for IMAGE and BUFFER nodes

### LINKED ISSUE #494: [Dota 2] vkBindImageMemory/vkBindBufferMemory validation errors due to uninitialized variables
I've been seeing intermittent errors with vkBindImageMemory/vkBindBufferMemory such as:

ERROR MEM 576:15: In vkBindImageMemory, attempting to bind memory (0x2e8a) to object (0x356b) which has already been bound to mem object 0xa

LunarG has not been able to repro it, so I looked into it and found the root cause of the problem.  There are two bugs in core_validation.cpp, one in vkCreateImage and one in vkCreateBuffer.  The problem is with this:

        dev_data->imageMap[*pImage].createInfo = *pCreateInfo;

Here the rest of dev_data->imageMap[*pImage] is uninitialized, and in particular the mem field, so I added this:

         dev_data->imageMap[*pImage].mem = 0;

There is a similar bug in vkCreateBuffer where this is done:

        dev_data->bufferMap[*pBuffer].createInfo = *pCreateInfo;
        dev_data->bufferMap[*pBuffer].in_use.store(0);

There needs to also be this:

       dev_data->bufferMap[*pBuffer].mem = 0;

Really, there should be robust initialization of all members.  The reason the bug was happening was depending on the random value in mem it would be reflected in *pMemBinding in set_mem_binding():

DEVICE_MEM_INFO *pPrevBinding = get_mem_obj_info(dev_data, *pMemBinding);

tobine:
Fixes #494 
Add proper constructors for IMAGE_NODE and BUFFER_NODE classes and
correctly initialize all of their data members.


mark-lunarg:
+1.  Pushed to Master.  Thanks Tobin.

Issue #498, opened 2016-05-06
"No active render pass found at draw-time in Pipeline" reported wrong for secondary buffers


McNopper:
Using secondary command buffers, the vaildation report is wrong. Can be reproduced with my example 7:

https://github.com/McNopper/Vulkan/tree/master/VKTS_Example07

Seems that VkCommandBufferInheritanceInfo is not taken into account.

tobine:
Fix was cherry-picked and pushed to master

PR #502, opened 2016-05-06
layers: Track secondary cmd buffer inherited render pass.

MarkY-LunarG:
We failed to track an inherited render pass for a secondary
command buffer, so we were failing one of my new checks.  This
could have potentially caused invalid failures in other cases
as well.

Change-Id: I01579d889dde8e0d331ed5a65eb48ce7185d3d82


MarkY-LunarG:
My previous changes introduced a validation error on Hologram.  The error was really because we weren't tracking the secondary command buffer inherited state.  This resolves that issue.

layers/core_validation.cpp
```diff
@@ -6137,12 +6137,22 @@ vkBeginCommandBuffer(VkCommandBuffer commandBuffer, const VkCommandBufferBeginIn
...
         }
+
         // Set updated state here in case implicit reset occurs above
         pCB->state = CB_RECORDING;
         pCB->beginInfo = *pBeginInfo;
+
         if (pCB->beginInfo.pInheritanceInfo) {
             pCB->inheritanceInfo = *(pCB->beginInfo.pInheritanceInfo);
             pCB->beginInfo.pInheritanceInfo = &pCB->inheritanceInfo;
+
```
tobine: You're killing me with the 3 added lines of whitespace... :)

MarkY-LunarG: That's almost as bad as people I know who complain about comments. ;-)

layers/core_validation.cpp
```diff
@@ -6137,12 +6137,22 @@ vkBeginCommandBuffer(VkCommandBuffer commandBuffer, const VkCommandBufferBeginIn
...
+            // If we are a secondary command-buffer and inheriting a RenderPass, make sure we record it
+            // as our active RenderPass so we can perform proper validation at draw-time.
+            if ((pCB->createInfo.level != VK_COMMAND_BUFFER_LEVEL_PRIMARY) &&
+                (pCB->beginInfo.flags & VK_COMMAND_BUFFER_USAGE_RENDER_PASS_CONTINUE_BIT) &&
+                (NULL != pCB->beginInfo.pInheritanceInfo->renderPass)) {
```
tobine: No need for NULL check here. According to spec if VK_COMMAND_BUFFER_USAGE_RENDER_PASS_CONTINUE_BIT is set then we should be copying over renderPass.

We should also copy over activeSubpass and insert framebuffer to framebuffers vector: `pCB->framebuffers.insert()`

tobine:
This is a good start and will fix #498 but can be better.
Please address the comments on skipping NULL check and copying over activeSubpass and framebuffer and then 👍

MarkY-LunarG:
Here you go.  Plus, I removed your whitespace concerns. :-)

chrisforbes:
:+1:

tobine:

PR #503, opened 2016-05-06
layers: Save swapchain image format for later validation compares

Tony-LunarG:
This fixes the cube resize with validation enabled bug, but should we be saving more/all of the createInfo for future checks?


lentinem:
I would vote for saving all of createInfo unless we change the data storage to not store the whole thing. It's going to be easy for others to just assume that they information is filled when writing a check.

Tony-LunarG:
There isn't really a 1 to 1 mapping from SwapchainCreateInfo to ImageCreateInfo, but I added a couple more fields that seemed to make sense.

Tony-LunarG:
Cherry picked and pushed to master

PR #504, opened 2016-05-06
Add RequiredParameter Test for parameter_validation Layer

dustin-lunarg:
Adds VkLayerTest.RequiredParameter test to check the parameter_validation layer functionality for identifying required handle, pointer, array, and array count parameters that are specified as VK_NULL_HANDLE, NULL, and 0.

Also adds the parameter_validation namespace to the parameter_validation layer, with an error code enumeration.


mark-lunarg:
+1.  Issue #436.  Pushed to Master.

PR #507, opened 2016-05-09
Jla debug marker header

jon-lunarg:


vk_helper.py
```diff
@@ -1589,7 +1589,7 @@ def _generateSafeStructDecls(self):
                 if is_type(m_type, 'struct') and self._hasSafeStruct(m_type):
                     m_type = self._getSafeStructName(m_type)
                 if self.struct_dict[s][m]['array_size'] != 0 and not self.struct_dict[s][m]['dyn_array']:
-                    ss_decls.append("    %s %s[%s];" % (m_type, self.struct_dict[s][m]['name'], self.struct_dict[s][m]['array_size']))
+                    ss_decls.append("    %s *%s; //array of size %s" % (m_type, self.struct_dict[s][m]['name'], self.struct_dict[s][m]['array_size']))
```
tobine: This happens to work in the VkDebugMarkerMarkerInfoEXT case b/c the static array is the last element of the struct. If the static array was not the last element, the safe struct would be broken b/c safe struct would not appear exactly the same as original struct.
As an alternative I recommend adding 'VkDebugMarkerMarkerInfoEXT' to the exceptions list on line 1551. There are no Vulkan objects in the struct so I believe it will work fine to just not generate the safe struct. I ran a quick test with that change and it seems to work. Unless you know that we needs safe structs for VkDebugMarkerMarkerInfoEXT I recommend the exception route.

jon-lunarg: okay I will do the exception (line 1551) for debug_marker. I don't think safe_struct is needed for debug_marker.  Although if an array gets added to the spec then the same compilation problem will occur so safe_structs needs to get fixed in this area.

stroyan:
This is updating vulkan.h without updating vk.xml.
Those two files should be kept in sync here.

jon-lunarg:
Dustin is working on updating vk.xml as their was some serious breakage with the AMD extension.
The changes to safe_structs have been merged.

Issue #431, opened 2016-04-27
Add test for push constant validation


tobine:
Currently no test(s) are identified for PUSH_CONSTANTS_ERROR case in [vk_validation_layer_details.md](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/vk_validation_layer_details.md).

This case should be examined according to the existing validation code and the Vulkan specification and test(s) should be added as appropriate.

karl-lunarg:
Closed https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/510

PR #510, opened 2016-05-09
tests: GH431 Add layer checks and tests for push constant

karl-lunarg:
Change-Id: I46eb5f9adbe702a278cded0f73dc18d1617489f0

Fix windows compilation issues

Change-Id: Ic9b9795e34d1e1e68bae9838b8538bb2e37d18b4

For Issue: Add test for push constant validation #431


layers/core_validation.cpp
```diff
@@ -5781,19 +5834,39 @@ VKAPI_ATTR VkResult VKAPI_CALL CreatePipelineLayout(VkDevice device, const VkPip
...
+        uint32_t i, j;
+        for (i = 0; i < pCreateInfo->pushConstantRangeCount; ++i) {
+            for (j = i + 1; j < pCreateInfo->pushConstantRangeCount; ++j) {
+                const uint32_t minA = pCreateInfo->pPushConstantRanges[i].offset;
+                const uint32_t maxA = minA + pCreateInfo->pPushConstantRanges[i].size;
+                const uint32_t minB = pCreateInfo->pPushConstantRanges[j].offset;
+                const uint32_t maxB = minB + pCreateInfo->pPushConstantRanges[j].size;
+                if ((minA <= minB && maxA > minB) || (minB <= minA && maxB > minA)) {
+                    skipCall |=
+                        log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
```
tobine: Is there spec language that dictates this to be an error?
I can see issuing a warning here, but I can also see instances where apps intentionally overlap ranges so that a superset range applies to multiple stages and a subset of that range applies to fewer stages.

layers/core_validation.cpp
```diff
@@ -7970,10 +8043,36 @@ VKAPI_ATTR void VKAPI_CALL CmdPushConstants(VkCommandBuffer commandBuffer, VkPip
...
+        for (uint32_t i = 0; i < ranges.size(); ++i) {
+            if ((offset >= ranges[i].offset) &&
+                ((uint64_t)offset + (uint64_t)size <= (uint64_t)ranges[i].offset + (uint64_t)ranges[i].size)) {
+                contained_in_a_range = true;
+                break;
+            }
+        }
+        if (!contained_in_a_range) {
+            skipCall |=
+                log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
```
tobine: What spec requirement is driving this check?

The primary requirement that I see for correlation between the pipeline layout push constant ranges and what is sent here is in valid usage requirement `stageFlags must match exactly the shader stages used in layout for the range specified by offset and size`

That requirement is not being checked here, and the source of the check that is performed is unclear to me. Push constants in general seem a bit unclear, though, so if I'm missing something in the spec that's driving this check please point it out.

layers/core_validation.cpp
karl-lunarg: Making this a warning is probably correct.  This code replaces a comment:
// TODO : Add warning if ranges overlap.
There's nothing in the spec that disallows overlap, but I had inferred that the shader inputs won't overlap, so these shouldn't either.  I can change it to a warning, for the reasons you mention.

layers/core_validation.cpp
karl-lunarg: Section 13.2.2 says: `Similarly, the push constant block declared in each shader (if present) must only place variables at offsets that are each included in a push constant range with stageFlags including the bit corresponding to the shader stage that uses it. `

But I think that instead of checking all the ranges, I should just find the range(s) selected by the incoming stageFlags argument and use those to see if the incoming offset+size are in those range(s).  Does that seem better?

tobine: The section 13.2.2 block that you indicate is at least partially covered by "validate_push_constant_block_against_pipeline()" function to validate between spirv push constant use and the corresponding pipeline.

Regarding checking ranges based on stageFlags, yes, since stageFlags must match exactly, checking only ranges with matching stageFlags is a good idea. There are still tricky corner cases to consider, though.

For example, a pipeline layout could include 3 separate ranges with the same stageFlags that are contiguous in memory (say the ranges cover 0-9,10-19,20-29). A single CmdPushConstants call could then update all three of those ranges with a single update that covers 1-29.

I think this is only really an issue if contiguous ranges have identical stageFlags, so maybe that's the only tricky corner case. If non-contiguous blocks have identical stageFlags, they can be checked independently as a single update couldn't write both without also writing intervening block that doesn't have identical stageFlags.

layers/core_validation.cpp
```diff
@@ -7970,10 +8043,36 @@ VKAPI_ATTR void VKAPI_CALL CmdPushConstants(VkCommandBuffer commandBuffer, VkPip
...
+    }
+
+    // Check if push constant update is within any of the ranges specified in pipeline layout.
+    auto pipeline_layout_it = dev_data->pipelineLayoutMap.find(layout);
+    if (pipeline_layout_it == dev_data->pipelineLayoutMap.end()) {
+        skipCall |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
+                            DRAWSTATE_PUSH_CONSTANTS_ERROR, "DS", "vkCmdPushConstants() Pipeline Layout %" PRIu64 " not found.",
+                            (uint64_t)layout);
+    } else {
+        auto ranges = pipeline_layout_it->second.pushConstantRanges;
```
chrisforbes: `const &` here -- you probably didn't want to copy this vector.

karl-lunarg: got it - thanks!

karl-lunarg:
OK to commit?

mark-lunarg:
Looks good to me.

layers/core_validation.cpp
tobine: This still doesn't appear to be a warning

tobine:
Ok, after looking through the commits more closely I think this is ok. It does look like you've addressed the feedback from Chris and me. Have you run a mustpass on it to check for regressions? Would be a good sanitycheck as there's a fair amount of new code here.
If it's testing clean I'm fine with these's changes going in now.

karl-lunarg:
cts mustpass shows no regressions and other apps we test with look good too.  Thanks!

karl-lunarg:
pushed to master

Issue #449, opened 2016-04-28
False DrawState positive - CB 0x24bda40 was submitted with secondary buffer 0x24ca270 but that buffer has subsequently been bound to primary cmd buffer 0x24c4400


mark-lunarg:
Running CTS dEQP-VK.api.command_buffers.secondary_execute_twice. The test creates and records a secondary command buffer, then executes it in recording primary command buffer 1, and also executes it in recording command buffer 2. Then, when it QueueSubmits command buffer 1, it gets the validation error in the title

tobine:
Were all the cmd buffers recorded w/ VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT? Note that I have an [open issue](https://github.com/KhronosGroup/Vulkan-Docs/issues/182) in Vulkan-Docs on this behavior. If all of the CBs in question are recorded w/ VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT then I agree that it's probably a false positive but spec can be clarified on this point. I'll file a PR for my proposed spec change and if there are no issues with it I'll update this behavior.

tobine:
Just submitted [MR1305](https://gitlab.khronos.org/vulkan/vulkan/merge_requests/1305) for Vulkan spec update.

Tony-LunarG:
The secondary command buffer was recorded with USAGE_SIMULTANEOUS_USE, but neither of the primary command buffers specified it.

tobine:
Spec change was approved for inclusion today so just submitted #511 to address this issue.

tobine:
Fix was cherry-picked and pushed to master

PR #511, opened 2016-05-09
layers: Update behavior for secondary cmdbuffers w/ *_SIMULTANEOUS_US…

tobine:
…E_BIT set

Spec clarification has been approved and secondary command buffers with
VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT may be in flight in multiple primary
command buffers at the same time.

This changes updates the error concerning secondary command buffers to be in-flight
on multiple primary command buffers to only appli if
VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT is not set.

Addresses #449


lentinem:
+1

tobine:

PR #513, opened 2016-05-10
layers: Start cutting needless deps between SC and layer_data

chrisforbes:
This is the first step toward moving SC out into its own module.


lentinem:
+1

PR #516, opened 2016-05-10
layers: Validate no active render pass when ending command buffer

cnorthrop:
We hit this case while debugging an application.  Per the spec:
"vkEndCommandBuffer must not be called inside a render pass instance"

Tested on Android, built from OSX, could use a test pass on desktop.


lentinem:
+1

mark-lunarg:
Can you add a TEST_DESCRIPTION() to the LVT?  LGTM, +1.

cnorthrop:
TEST_DESCRIPTION() added and re-pushed.

layers/core_validation.cpp
```diff
@@ -6153,6 +6153,12 @@ VKAPI_ATTR VkResult VKAPI_CALL EndCommandBuffer(VkCommandBuffer commandBuffer) {
     std::unique_lock<std::mutex> lock(global_lock);
     GLOBAL_CB_NODE *pCB = getCBNode(dev_data, commandBuffer);
     if (pCB) {
+        if (pCB->activeRenderPass != VK_NULL_HANDLE) {
+            skipCall |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
+                                DRAWSTATE_END_BUFFER_ACTIVE_RENDERPASS, "DS",
+                                "vkEndCommandBuffer called with an active render pass: renderPass (%#" PRIxLEAST64 ")",
+                                (uint64_t)pCB->activeRenderPass);
+        }
```
tobine: Existing "insideRenderPass()" function should be used here instead. Search code for "insideRenderPass(" to see examples.

layers/core_validation_error_enums.h
```diff
@@ -140,6 +140,7 @@ typedef enum _DRAW_STATE_ERROR {
                                                 // RenderPass is active
     DRAWSTATE_NO_ACTIVE_RENDERPASS,             // Rendering cmd submitted without an active
                                                 // RenderPass
+    DRAWSTATE_END_BUFFER_ACTIVE_RENDERPASS,     // Ending Command Buffer with active RenderPass
```
tobine: No need for new enum. You can piggy-back on the existing "DRAWSTATE_INVALID_RENDERPASS_CMD" enum.

layers/vk_validation_layer_details.md
```diff
@@ -70,6 +70,7 @@ The Draw State portion of the core validation layer tracks state leading into Dr
 | Valid RenderPass | Flag error if attempt made to Begin/End/Continue a NULL or otherwise invalid RenderPass object | INVALID_RENDERPASS | vkCmdBeginRenderPass vkCmdEndRenderPass vkBeginCommandBuffer | NullRenderPass | NA |
 | RenderPass Compatibility | Verify that active renderpass is compatible with renderpass specified in secondary command buffer, and that renderpass specified for a framebuffer is compatible with renderpass specified in secondary command buffer | RENDERPASS_INCOMPATIBLE | vkCmdExecuteCommands vkBeginCommandBuffer | TBD | None |
 | Framebuffer Compatibility | If a framebuffer is passed to secondary command buffer in vkBeginCommandBuffer, then it must match active renderpass (if any) at time of vkCmdExecuteCommands | FRAMEBUFFER_INCOMPATIBLE | vkCmdExecuteCommands | TBD | None |
+| No Active Render Pass when Ending Command Buffer | Verify that there is no active render pass when ending command buffer | END_BUFFER_ACTIVE_RENDERPASS | vkEndCommandBuffer | EndCommandBufferWithActiveRenderPass | None |
```
tobine: Can update existing table entry for "INVALID_RENDERPASS_CMD" (line 68) with vkEndCommandBuffer and new testname

cnorthrop:
Thanks for the pointers Tobin.  Update pushed.

chrisforbes:
:+1:

layers/core_validation.cpp
```diff
@@ -6153,6 +6153,7 @@ VKAPI_ATTR VkResult VKAPI_CALL EndCommandBuffer(VkCommandBuffer commandBuffer) {
     std::unique_lock<std::mutex> lock(global_lock);
     GLOBAL_CB_NODE *pCB = getCBNode(dev_data, commandBuffer);
     if (pCB) {
+        skipCall |= insideRenderPass(dev_data, pCB, "vkEndCommandBuffer");
```
tobine: This check has revealed an interesting issue, which I believe is a spec oversight. Smoketest hits this check a bunch because it uses secondary command buffers that inherit the activeRenderpass from the primary command buffer. Technically I believe these secondary command buffers are "inside a renderpass instance" which means, according to the spec, that it's illegal to call vkEndCommandBuffers() on such a secondary command buffer.

This seems ridiculous, however, so I believe the spec valid usage language for vkEndCommandBuffers() needs to be updated to note something to the effect of `vkEndCommandBuffer must not be called inside a render pass instance for a primary command buffer`.

Additionally, I believe that this check should be guarded by:

    if (VK_COMMAND_BUFFER_LEVEL_PRIMARY == pCB->createInfo.level)

tobine: Chris has pointed out that secondary command buffers can be executed inside or outside of a renderpass, and they may fully contain their own renderpass. That means that my proposed solution is insufficient. I now believe we only need to run the "insideRenderPass()" check for EndCommandBuffers() if it's a primary cmd buffer OR secondary w/o the RP_CONTINUE_BIT set, then we should perform the check:

    if ((VK_COMMAND_BUFFER_LEVEL_PRIMARY == pCB->createInfo.level) || !(pCB->beginInfo.flags & VK_COMMAND_BUFFER_USAGE_RENDER_PASS_CONTINUE_BIT))

This really needs some spec clarification, though.

cnorthrop:
Updated the code to reflect current thinking, thanks Tobin.  Tests passing on Android.

tobine:

PR #517, opened 2016-05-10
layers: Add output file option to vk-generate scripts.

null77:
With some build systems, it's not easy to pipe stdout into a file.
This solution allows the scripts to use the last argument as an
output file option. This is a problem in partiuclar for GN.

PTAL, this is a necessary feature for building with Chromium/ANGLE.


vk-layer-generate.py
```diff
@@ -1678,7 +1682,7 @@ def main():
     vk_helper.typedef_rev_dict = hfp.get_typedef_rev_dict()
     vk_helper.types_dict = hfp.get_types_dict()
```
stroyan: This could use an update to the usage message at line 1667.
There really are no "[options]" for this script now, so that could be replaced by "[outfile]".
The passing  of sys.argv[3:] and its storage as self.argv are also vestigial and could be removed.

null77: Changed usage message, and removed self.argv as requested.

PTAL, thank you!

stroyan:
Looks good to me.

tobine:

PR #518, opened 2016-05-10
Remove Xlib dependency when XCB is used

pavolzetor:


Tony-LunarG:
The intent was to have both Xcb and Xlib available to run either path by specifying --xlib at runtime. Your change would exclude having both at once.

pavolzetor:
Then it might help to have cmake show xlib and xcb as one option.

Tony-LunarG:
I tweaked your commit to allow xlib and xcb to co-exist and pushed to master.  Thanks!

PR #520, opened 2016-05-10
layers: GH458 Fix cleanup of queue lastFences

### LINKED ISSUE #458: Core Validation Layer issue when using VK_FENCE_CREATE_SIGNALED_BIT
vkCreateFence sets in_use to 0 and vkWaitForFences() decrements in_use. When using VK_FENCE_CREATE_SIGNALED_BIT it leads to this:
vkCreateFence(VK_FENCE_CREATE_SIGNALED_BIT) in_use = 0
vkWaitForFences() in_use = -1

I'm guessing in_use should be set to 1 when VK_FENCE_CREATE_SIGNALED_BIT is passed -- or possibly vkWaitForFences needs to take into account firstTimeFlag/needsSignaled, and not decrement.

tobine:
Fixes #458 
When waiting on fences, they were not being correctly removed from the
queue's lastFences vector. Updated the remove code and verified that
it's now working correctly.


layers/core_validation.cpp
```diff
@@ -4244,14 +4244,14 @@ static void decrementResources(layer_data *my_data, VkCommandBuffer cmdBuffer) {
 //  decrementResources for all priorFences and cmdBuffers associated with fence.
 static bool decrementResources(layer_data *my_data, uint32_t fenceCount, const VkFence *pFences) {
     bool skip_call = false;
-    std::vector<VkFence> fences;
+    std::vector<std::pair<VkFence, FENCE_NODE*>> fence_pairs;
```
lentinem: It looks like we only need VkQueue not the entire FENCE_NODE.

layers/core_validation.cpp
```diff
@@ -4264,13 +4264,12 @@ static bool decrementResources(layer_data *my_data, uint32_t fenceCount, const V
...
     }
-    for (auto fence : fences) {
-        for (auto queue_data : my_data->queueMap) {
-            auto last_fence_data = std::find(queue_data.second.lastFences.begin(), queue_data.second.lastFences.end(), fence);
-            if (last_fence_data != queue_data.second.lastFences.end()) {
-                queue_data.second.lastFences.erase(last_fence_data);
-                break;
-            }
+    for (auto fence_pair : fence_pairs) {
+        auto queue_pair = my_data->queueMap.find(fence_pair.second->queue);
```
lentinem: I'm a bit confused as to how this works. In the case of multiple queues with semaphores this fence would be on multiple queue lastFences lists. In that case this will only delete one.

layers/core_validation.cpp
tobine: agreed

layers/core_validation.cpp
tobine: It passes layer validation tests on my Linux/Nvidia which features multiple queues but it's possible that we don't have the case you're referencing covered.

If a fence can be on multiple queue lastFences list, should we change "queue" data member of FENCE_NODE into an unordered_set and then for this clean-up we can iterate over all queues in the set?

lentinem: I'm fine with that solution but the queue element needs to be updated to include all relevant queues. I believe that element is largely unused right now.

tobine:
Force pushed update that migrates from single VkQueue tracking per fence (queue) to unordered_set<VkQueue> (queues).

The one thing I'm not certain of if when to remove a queue from queues. Currently only removing when ResetFence() called. Can't remove queues as a fence is waited on with the current code or back-to-back WaitForFences on same fence issues error.

Note that I intentionally left FENCE_NODE* in the std::pair during decrementResources as I felt the syntax was cleaner than sticking a whole unordered_set in there.

Also added in a positive test to prevent regressions in this behavior.

lentinem:
+1. I would prefer a pointer to the set instead of the node but am fine with the way it is.

tobine:
This fix was pushed to master. I excluded the test b/c it's currently failing on android. Moved just the test CL to separate branch "tobin_gh458_fence_test" where I'll get it working on android before pushing.

PR #523, opened 2016-05-10
layers:LX#470 Address poor use of hex vs dec output & w/a convention

MarkMLunarG:
1) Introduce convention of explicitly placing "0x" before hex
    format requests for clarity (don't use "%#")
2) All lower case in hex output (except w/stringstream which refuses
    to do anything but uppercase, possibly only on windows).
3) Decorated pointers are printed for all Vulkan structure types.
3) Some intelligence in generators based on Vulkan variable name:
     if the Vulkan variable name contains ("flag", "bit", "offset",
     "handle", "buffer", "object", "mask") it will be output in
     hexadecimal format

Change-Id: Idbae73bfdaa3bc059543d43b209373cd0bcbc099

### DIFF
--- a/layers/core_validation.h
+++ b/layers/core_validation.h
@@ -59,9 +59,6 @@
 #include <unordered_set>
 #include <vector>
 
-using std::vector;
-using std::unordered_set;
-
 #if MTMERGE
 
 /*
@@ -147,7 +144,7 @@ class PIPELINE_NODE {
     uint32_t active_shaders;
     uint32_t duplicate_shaders;
     // Capture which slots (set#->bindings) are actually used by the shaders of this pipeline
-    unordered_map<uint32_t, unordered_set<uint32_t>> active_slots;
+    std::unordered_map<uint32_t, std::unordered_set<uint32_t>> active_slots;
     // Vtx input info (if any)
     std::vector<VkVertexInputBindingDescription> vertexBindingDescriptions;
     std::vector<VkVertexInputAttributeDescription> vertexAttributeDescriptions;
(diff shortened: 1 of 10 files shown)


mark-lunarg:
Tobin, we are moving the layers to use standard output formats for some items, can you take a look to see if it's agreeable?

layers/core_validation.h
```diff
@@ -147,7 +147,7 @@ class PIPELINE_NODE {
     uint32_t active_shaders;
     uint32_t duplicate_shaders;
     // Capture which slots (set#->bindings) are actually used by the shaders of this pipeline
-    unordered_map<uint32_t, unordered_set<uint32_t>> active_slots;
+    std::unordered_map<uint32_t, std::unordered_set<uint32_t>> active_slots;
```
tobine: can you kill the 2 "using" statements on lines 62&63 above as well?

tobine:
Yeah, I like this. I haven't run the code, but based on visual inspection LGTM

layers/object_tracker.h
```diff
@@ -392,7 +392,7 @@ static void destroy_surface_khr(VkInstance dispatchable_object, VkSurfaceKHR obj
         numObjs[objIndex]--;
         log_msg(mdd(dispatchable_object), VK_DEBUG_REPORT_INFORMATION_BIT_EXT, pNode->objType, object_handle, __LINE__,
                 OBJTRACK_NONE, "OBJTRACK",
-                "OBJ_STAT Destroy %s obj 0x%" PRIxLEAST64 " (%" PRIu64 " total objs remain & %" PRIu64 " %s objs).",
+                "OBJ_STAT Destroy %s obj 0x%" PRIxLEAST64 " (0x%" PRIX64 " total objs remain & 0x%" PRIX64 " %s objs).",
```
karl-lunarg: This looks like upper case.

PR #524, opened 2016-05-10
Sync XML registry files from Vulkan-Docs

dustin-lunarg:
Sync the vk.xml and related Python scripts with the latest revisions from
Vulkan-Docs:
 - Update local copy of vk.xml from Vulkan-Docs
 - Merge minor changes to genvk.py and generator.py from Vulkan-Docs to
   local copies
 - Update parameter_validation code generation to support use of types
   across API features; The core API feature needs to be aware of a
   structure defined by the AMD_rasterization_order feature for the
   purpose of pNext validation
 - Modify thread_check code generation to add TODO messages for debug
   marker support


generator.py
```diff
@@ -2760,6 +2763,9 @@ def genCmd(self, cmdinfo, name):
         if "KHR" in name:
             self.appendSection('command', '// TODO - not wrapping KHR function ' + name)
             return
+        if "DebugMarker" in name:
+            self.appendSection('command', '// TODO - not wrapping KHR function ' + name)
```
jon-lunarg: minor nit pick
debugMarker is a EXT rather than KHR function

include/vulkan/vk_layer.h
```diff
@@ -162,6 +162,11 @@ typedef struct VkLayerDispatchTable_ {
     PFN_vkGetSwapchainImagesKHR GetSwapchainImagesKHR;
     PFN_vkAcquireNextImageKHR AcquireNextImageKHR;
     PFN_vkQueuePresentKHR QueuePresentKHR;
+    PFN_vkDebugMarkerSetObjectTagEXT DebugMarkerSetObjectTagEXT;
+    PFN_vkDebugMarkerSetObjectNameEXT DebugMarkerSetObjectNameEXT;
+    PFN_vkCmdDebugMarkerBeginEXT CmdDebugMarkerBeginEXT;
+    PFN_vkCmdDebugMarkerEndEXT CmdDebugMarkerEndEXT;
+    PFN_vkCmdDebugMarkerInsertEXT CmdDebugMarkerInsertEXT;
 } VkLayerDispatchTable;
```
jon-lunarg: I think these shouldn't go in here.  The vk_layer.h partially defines the loader<->layer interface.  The loader doesn't have to change for device extensions.  If a layer wants to intercept a device extension such as debug_marker, it should create it's own dispatch table structure for the extension. Probably, for our layers that use hash maps for storing their dispatch table getting the vk_layer.h the loader is built with versus a layer wouldn't matter.  But for an object wrapping layer I think it would matter since the wrapped object dispatch table would be different sizes.

generator.py
dustin-lunarg: Done

include/vulkan/vk_layer.h
dustin-lunarg: Done

jon-lunarg:
thanks pushed to master!

PR #526, opened 2016-05-11
layers: Fix get_descriptor_binding

chrisforbes:
Not sure where the `return` went... this worked by accidents of calling
conventions in debug, but was rightfully eaten by the optimizer in release.


lentinem:
+1. Did this build before? Using Wall and Werror should have caused build failures on all platforms - it's a bit concerning if it didn't.

chrisforbes:
@lentinem desktop build I checked with clearly doesn't have this stuff cranked up high enough.

chrisforbes:
Cherry-picked to master.

karl-lunarg:
We do have -Wall and -Wextra set and it still doesn't give a warning when the return is left out, which is bad.
Adding -Werror would only have helped if it generated a warning.
We compile (on linux/gcc) without warnings today, so I'll probably add -Werror soon.

It may be worth looking at doing some static analysis, which should help avoid this sort of thing and find other latent problems.

lentinem:
I verified with ndk-build if I delete the return at least android fails. This indicates that with the right compiler it would. I'm using clang 3.6. I'm not sure there is anything extra we can do here other than try to make sure at least one of the compilers we use for regular build tests triggers a failure.

karl-lunarg:
When we do our gerrit submissions, an android build is performed, so that would have caught it on our end.

PR #530, opened 2016-05-11
layers: Fix Android build

dustin-lunarg:
Remove unused variable that was generating a warning, treated as
error, for the Android build.


tobine:
I already pushed this fix. Thanks, Dustin.

PR #531, opened 2016-05-11
layers: Fix VkBool32 return in vk-layer-generate.py.

null77:
This regressed in f4542e618, but is only visible when defining
VK_USE_PLATFORM_WIN32_KHR, so might have been missed. Fixes a
warning in MS 2015, as well as an incorrect return value.

PTAL, a build fix.


tobine:
LGTM :+1:

tobine:

PR #532, opened 2016-05-11
layers: Fix include order in decriptor_sets.h

null77:
The vk_loader_platform header must come first, since it fixes the
Windows Winsock.h include order problem. Without this the build
produces a spew of warnings.

PTAL, another build fix.


dustin-lunarg:
We should consider adding -DWIN32_LEAN_AND_MEAN to the compiler options or #define WIN32_LEAN_AND_MEAN before any windows.h includes, so that the include order for other headers doesn't matter.

null77:
Agreed, I wasn't aware this was a solution. Using that locally for now, someone may want to make a change on the SDK-side.

PR #534, opened 2016-05-12
tests: Add invalid pNext test

dustin-lunarg:
Add test for case where a non-NULL pNext value points to a structure
that is not valid for the current struct type.

VkPipelineRasterizationStateCreateInfo::pNext is currently only
allowed to point to VkPipelineRasterizationStateRasterizationOrderAMD.
Setting pNext to a structure of any other type should generate an
error.


mark-lunarg:
LGTM

tobine:
LGTM :+1:

PR #538, opened 2016-05-12
Additional Image Parameter Validation

dustin-lunarg:
Additional parameter validation for the VkCreateImage and VkCreateImageView functions.  

This branch also includes a commit to fix the error message reported for queue index validation (performed for vkCreateImage), to report the correct parameter names and use language that is consistent with the language used in the spec's valid usage section.


layers/parameter_validation.cpp
```diff
@@ -1289,6 +1267,57 @@ static bool validate_string(debug_report_data *report_data, const char *apiName,
...
 }
 
+static bool validate_queue_family_index(layer_data *device_data, const char *function_name, const char *parameter_name,
+                                        uint32_t index) {
+    assert(device_data != nullptr);
+    debug_report_data *report_data = device_data->report_data;
+    bool skip_call = false;
+
+    if (index == VK_QUEUE_FAMILY_IGNORED) {
+        skip_call |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__, 1,
```
tobine: @mark-lunarg This line brings up a couple of questions related to the Vulkan Validation Layer Authoring Guide, which I just commented on.
It's using a c-style cast and not providing an object handle, I'm fine with that in both of these cases as well as 3 similar cases below. Is that ok? If so can we amend Vulkan Validation Layer Authoring Guide to be a bit more lenient?

dustin-lunarg: I'll do a pass to clean up the casts.  I've been trying to replace the instances of '(VkDebugReportObjectTypeEXT)0' in the current code with 'VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT'.

tobine:
LGTM :+1:

dustin-lunarg:
Pushed to master.  I have a follow on change in progress that will address the C-style casts.

PR #542, opened 2016-05-13
layers: Move `fb` from RENDER_PASS_NODE to GLOBAL_CB_NODE

chrisforbes:
This is part of a renderpass instance, not of the renderpass itself.
If stored where MT put it, things go horribly wrong when recording
command buffers in parallel.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -3755,6 +3755,7 @@ static void resetCB(layer_data *dev_data, const VkCommandBuffer cb) {
             }
         }
         pCB->framebuffers.clear();
+        pCB->activeFramebuffer = nullptr;
 
     }
 }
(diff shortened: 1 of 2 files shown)


lentinem:
+1. I am curious why things go horribly wrong though. Only one thread should be calling begin or end for that render pass at a time.

tobine:
LGTM :+1:

chrisforbes:
@lentinem for a given command buffer. Multiple threads may record instances of the same renderpass into /different/ command buffers concurrently, though.

chrisforbes:
@lentinem we don't even need to bring threads into it -- a single thread, doing interleaved recording of two command buffers which contain instances of the same renderpass, is enough.

chrisforbes:
Cherry-picked to master.

PR #543, opened 2016-05-13
[For Review Only] layers: LX518, Fix ObjectTracker memory leaks

mark-lunarg:
Several places in object_tracker were not properly deleting data
structures causing significant memory leaks.

Note that we have a change in flight to move the OT maps into device/instance space, and that this layer is in need of a general re-write -- this only addresses the memory leaks.


tobine:
In regards to fixing the memory leaks, this LGTM 👍 

Agree lots more work to be done on this layer in general

mark-lunarg:
Pushed to Master, 

commit 86411429f89ddae704116364cb71e8556363e68b
Date:   Thu May 12 15:14:07 2016 -0600

    layers: LX518, Fix ObjectTracker memory leaks

    Several places in object_tracker were not properly deleting data
    structures, which caused signicant memory leaks.

PR #546, opened 2016-05-15
Fix allows to return actual storage_ and not default false value.

McNopper:
Without this virtual method, class ImageDescriptor does return the default value "false" from class Descriptor and not its actual value in "storage_".


chrisforbes:
:+1:

lentinem:
+1

tobine:

Issue #544, opened 2016-05-14
Incorrect CB_INVALID leading to confusing 'You must call vkBeginCommandBuffer() before this call to ...' messages


ShaneYCG:
I'm still trying to track down the precise cause, so this initial post is mostly to see if anyone else is running into this issue, and also to bring attention to the misleading warning message.

GLOBAL_CB_NODE::state is sometimes incorrectly being set to CB_INVALID, which causes subequent calls to nearly everything to emit the "You must call vkBeginCommandBuffer()" message.

As best I can tell so far, sometimes resetCB() is not always calling RemoveBoundCommandBuffer (uniqueBoundSets is sometimes incorrectly empty), causing vkUpdateDescriptorSets to call invalidateBoundCmdBuffers, setting the command buffer to the INVALID state.

ShaneYCG:
It seems related to my reuse of VkDescriptorSet. I have a pool of them and I reset to the head when it is safe, if I simply let my pool regrow infinitely the error disappears. Otherwise, the error seems to appear on the first frame that re-use occurs. Unfortunately I have not yet been able to reproduce this in a minimal sample, it may be timing sensitive.

ShaneYCG:
This appears to a problem with ping-ponging command buffers: https://gist.github.com/ShaneYCG/b2b7361f1ddcab0313da31eaf52d726c

tobine:
Debugging this now. Have the repro working. Thanks, Shane.

tobine:
Just submitted #549 to address this issue. That commit message explains the fix, but in general there was a hole where a CB that was placed into INVALID state due to having a bound descriptor set updated was not being implicitly reset when vkBeginCommandBuffer() was called on it.

ShaneYCG:
Will verify as soon as the PR is accepted (I am admittedly new to git and not sure how to grab it before! :)

tobine:
Fix was cherry-picked and pushed to master

ShaneYCG:
Did the trick, thanks!

cippyboy:
I'm using the 1.0.17 validation and I get this and have no idea what it might be related to as I am 1000% sure I'm calling Begin/EndCommandBuffer properly. Is this fix in the 1.0.17 release ?

tobine:
@cippyboy I'm pretty sure this fix should have been included in 1.0.17 SDK as the fix was in on May 18th and that SDK came out in late June.

You have a few options:

[...]

PR #549, opened 2016-05-16
layers: GH544 Allow INVALID cmd buffers to be reset

tobine:
Fix for #544 
BeginCommandBuffer performs an implicit reset of a command buffer.
This allows command buffers in the INVALID state to undergo this
implicit reset.

In order to make sure that an invalid command buffer in the INVALID
state has had EndCommandBuffer called, this commit adds a CMD_END
command that's inserted at the end of a command. If a command buffer
in the INVALID state has had EndCommandBuffer called and it is now
having BeginCommandBuffer called, it's ok to implicitly reset the
command buffer.
buffer's vector of commands.


mark-lunarg:
Looks good.

tobine:
ping. I just had to add a 2nd CL to prevent what I believe was an invalid error in smoketest due to incorrect comands in a subpass.

layers/core_validation.cpp
```diff
@@ -6251,9 +6254,7 @@ VKAPI_ATTR VkResult VKAPI_CALL EndCommandBuffer(VkCommandBuffer commandBuffer) {
             // https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/516#discussion_r63013756
             skipCall |= insideRenderPass(dev_data, pCB, "vkEndCommandBuffer");
         }
-        if (pCB->state != CB_RECORDING) {
```
lentinem: Why was this if removed? It still is true that it must have begun before being ended.

tobine: addCmd() call includes that check and more

lentinem:
+1

tobine:

PR #551, opened 2016-05-16
layers: GH485, Strengthen image format validation

mark-lunarg:
CreateImage was checking that format caps weren't null, but was not
validating against color or depth/stencil attachment types and their
corresponding format properties.  Also added valid-usage check for
FORMAT_UNDEFINED and a test to go with that.

Change-Id: Ic8996829033f552f6ef7477bdefdaba702ae2403


layers/image.cpp
```diff
@@ -243,13 +243,45 @@ CreateImage(VkDevice device, const VkImageCreateInfo *pCreateInfo, const VkAlloc
...
+            if (((pCreateInfo->tiling == VK_IMAGE_TILING_OPTIMAL) &&
+                 ((properties.optimalTilingFeatures & VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT) == 0)) ||
+                ((pCreateInfo->tiling == VK_IMAGE_TILING_LINEAR) &&
+                 ((properties.linearTilingFeatures & VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT) == 0))) {
+                std::stringstream ss;
+                ss << "vkCreateImage: VkFormat for image (" << string_VkFormat(pCreateInfo->format)
+                   << ") does not support requested Image usage type VK_IMAGE_USAGE_COLOR_ATTACHMENT";
+                skip_call |=
+                    log_msg(device_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0,
+                            __LINE__, IMAGE_INVALID_FORMAT, "IMAGE", "%s", ss.str().c_str());
```
tobine: Should these be broken into separate error cases and also report to the user that their specific requested tiling format is unsupported?

layers/image.cpp
```diff
@@ -243,13 +243,45 @@ CreateImage(VkDevice device, const VkImageCreateInfo *pCreateInfo, const VkAlloc
...
+                 ((properties.optimalTilingFeatures & VK_FORMAT_FEATURE_DEPTH_STENCIL_ATTACHMENT_BIT) == 0)) ||
+                ((pCreateInfo->tiling == VK_IMAGE_TILING_LINEAR) &&
+                 ((properties.linearTilingFeatures & VK_FORMAT_FEATURE_DEPTH_STENCIL_ATTACHMENT_BIT) == 0))) {
+                std::stringstream ss;
+                ss << "vkCreateImage: VkFormat for image (" << string_VkFormat(pCreateInfo->format)
+                   << ") does not support requested Image usage type VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT";
+                skip_call |=
+                    log_msg(device_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0,
+                            __LINE__, IMAGE_INVALID_FORMAT, "IMAGE", "%s", ss.str().c_str());
+            }
```
tobine: Same question here as above

layers/image.cpp
```diff
@@ -243,13 +243,45 @@ CreateImage(VkDevice device, const VkImageCreateInfo *pCreateInfo, const VkAlloc
...
         }
+    } else {
+        skip_call |= log_msg(device_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
```
tobine: Set explicit type VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT

layers/image.cpp
```diff
@@ -282,9 +314,9 @@ CreateImage(VkDevice device, const VkImageCreateInfo *pCreateInfo, const VkAlloc
         break;
     }
     if (failedMinSize) {
-        skipCall |=
+        skip_call |=
             log_msg(phy_dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT,
```
tobine: Should type here be VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT since we don't have a VkImage and are passing "0" handle?

layers/image.cpp
```diff
@@ -293,8 +325,8 @@ CreateImage(VkDevice device, const VkImageCreateInfo *pCreateInfo, const VkAlloc
     if ((pCreateInfo->extent.depth > ImageFormatProperties.maxExtent.depth) ||
         (pCreateInfo->extent.width > ImageFormatProperties.maxExtent.width) ||
         (pCreateInfo->extent.height > ImageFormatProperties.maxExtent.height)) {
-        skipCall |= log_msg(phy_dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT,
-                            (uint64_t)pImage, __LINE__, IMAGE_INVALID_FORMAT_LIMITS_VIOLATION, "Image",
+        skip_call |= log_msg(phy_dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT,
+                            0, __LINE__, IMAGE_INVALID_FORMAT_LIMITS_VIOLATION, "Image",
```
tobine: Same comment for all of these cases (5 more below): Should type here be VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT since we don't have a VkImage and are passing "0" handle?

layers/image.cpp
```diff
@@ -367,23 +399,6 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateRenderPass(VkDevice device, const VkRenderP
...
-            if ((properties.linearTilingFeatures) == 0 && (properties.optimalTilingFeatures == 0)) {
-                std::stringstream ss;
-                ss << "vkCreateRenderPass parameter, VkFormat in pCreateInfo->pAttachments[" << i
-                   << "], contains unsupported format";
-                // TODO: Verify against Valid Use section of spec. Generally if something yield an undefined result, it's invalid
-                skipCall |= log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
-                                    IMAGE_FORMAT_UNSUPPORTED, "IMAGE", "%s", ss.str().c_str());
-            }
-        }
-    }
```
tobine: Why is this code being removed? Is not clear to me based on commit message.

tobine:
One thing to think about going forward, and something I'm trying to be better about, is breaking commit into separate CLs as appropriate for simplicity. Here I could see at least 3 CLs: 1 for CreateImage updates, 1 for CreateRenderPass updates, and 1 for test updates.

layers/image.cpp
mark-lunarg: Had them that way, will split them again.

layers/image.cpp
mark-lunarg: Had already corrected these locally (guidelines!).

layers/image.cpp
mark-lunarg: We talked about this here and with Dan Ginsburg.  These checks sit in CreateImage, and there's an opportunity to do same/similar checks at CreateRenderPass time to check that all the attachments have valid format properties, but it seemed clear that between the CreateImage checks and the usage_flags checks that this should cover these cases as well.  Unless you think it's worth checking things again here (or that there's added value) I'll put a note in when I split this out into its own commit.

mark-lunarg:
Will split into three commits.  Thanks for the quick response Tobin.

layers/image.cpp
tobine: Yes, this makes sense and I'm ok removing these CreateRenderPass() checks.

mark-lunarg:
Updated and pushed to Master. 

commit 42192f7fa75ec2fbe2d1f15ae846de92778caaaf
Date:   Tue May 17 08:14:00 2016 -0600

    tests: GH485, Update valid format property tests

    Validation for image format properties was tightened up which resulted
    in changes to the existing tests and a valid usage addition.

    Change-Id: Id87c599a60229b38cbbe759d988f96f947b396ed

commit 7e10fc8b340627bfce818583809f53e6708b0478
Date:   Tue May 17 08:44:09 2016 -0600

    layers: GH485, Remove redundant format props check

    With new checks in place for CreateImage, old check in CreateRenderPass
    is redundant and can be removed.

    Change-Id: I53e5201f453041b3c8ba17edc193c1f3b142f4cb

commit bcf8bceadcddfe4725ce825bd8ef7c8d3bc5fc4f
Date:   Tue May 17 08:42:03 2016 -0600

    layers: GH485, Strengthen image format validation

    CreateImage was checking that format caps weren't null, but was not
    validating against color or depth/stencil attachment types.

    Change-Id: Ic8996829033f552f6ef7477bdefdaba702ae2403

PR #552, opened 2016-05-16
layers: Ignore 'noautovalidity' params w/codgen

dustin-lunarg:
Add special handling to the parameter_validation layer's code
generator for vk.xml parameters with the 'noautovalidty' attribute:
- Ignore vk.xml parameters with the 'noautovalidity' attribute
  when generating code for the parameter_validation layer.
- Implement custom validation routines for parameters with
  'noautovalidity' tags.


mark-lunarg:
Looks great.

PR #557, opened 2016-05-17
layers: Fix TexelDescriptor to have correct IsStorage()

tobine:
Similar to Norbert's previous fix for ImageDescriptor, TexelDescriptor
was also missing correct IsStorage() override function and so would always
return false instead of actual storage_ value.


mark-lunarg:
+1

mark-lunarg:
Pushed to Master:
commit 13138d0514445fc70b23ecb2ce639326f76c7883
Author: Tobin Ehlis <tobine@google.com>
Date:   Tue May 17 06:43:48 2016 -0600

    layers: Add correct IsStorage() to TexelDescriptor

    Similar to Norbert's previous fix for ImageDescriptor, TexelDescriptor
    was also missing correct IsStorage() override function and so would always
    return false instead of actual storage_ value.

    Change-Id: I9006cdb614c5b78db0a68b71dcc4300f4c5c0d7f

Issue #401, opened 2016-04-21
MEM: Stencil attachment memory not marked as valid in render pass


yavn:
I'm using a render pass with a single stencil attachment (no depth, color, etc.):

```
VkAttachmentDescription
{
    flags:          0
    format:         VK_FORMAT_S8_UINT (127)
    samples:        VK_SAMPLE_COUNT_1_BIT (1)
    loadOp:         VK_ATTACHMENT_LOAD_OP_DONT_CARE (2)
    storeOp:        VK_ATTACHMENT_STORE_OP_DONT_CARE (1)
    stencilLoadOp:  VK_ATTACHMENT_LOAD_OP_CLEAR (1)
    stencilStoreOp: VK_ATTACHMENT_STORE_OP_STORE (0)
    initialLayout:  VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL (3)
    finalLayout:    VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL (3)
}
```

I'm also passing in a `VkClearValue` when starting a render pass, so it's expected that contents of attachment memory are well-defined at the begining (CLEAR) as well as at the end (STORE) of the render pass.

However, layer reports this error:

> MEM(ERROR): object: 0x4 type: 8 location: 421 msgCode: 16: vkCmdCopyImageToBuffer(): Cannot read invalid memory 4, please fill the memory before using.

I think the root cause of this is in `vkCreateRenderPass()` in `core_validation.cpp`, where attachment description is copied. Only `loadOp` and `storeOp` fields are handled but in this case layer should use `stencilLoadOp` and `stencilStoreOp`.

In this case, attachment gets assigned DONT_CARE which later results in marking the memory as not valid.

mark-lunarg:
Hi Maciej, I'm working on a fix but it'd be great if you could share an app that reproduces the failure OR can either add a case to the layer_validation_tests.cpp or modify one of the vulkansamples to make it show the behavior.  Debugging will go a lot quicker.   Thanks!

mark-lunarg:
Bump.

mark-lunarg:
Pushed to Master:

commit 1a65650f856376768d7b03ea2d080aaff87cacfd
Author: Mark Lobodzinski <mark@lunarg.com>
Date:   Fri Jun 10 15:28:17 2016 -0600

[...]

PR #558, opened 2016-05-17
layers: GH401 -- Separate out stencil store/loadOps

mark-lunarg:
Tobin, can you look to see if you think this would be sufficient for GH401?  I don't have a repro case and was unsure if it needs some special-casing of the attachment types.  Can you take a quick peek?  The guideline-related stuff I have lined up -- right now I'm just wondering about the validity.  Thanks!

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -8719,6 +8719,8 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateRenderPass(VkDevice device, const VkRenderP
             MT_PASS_ATTACHMENT_INFO pass_info;
             pass_info.load_op = desc.loadOp;
             pass_info.store_op = desc.storeOp;
+            pass_info.stencil_load_op = desc.stencilLoadOp;
+            pass_info.stencil_store_op = desc.stencilStoreOp;
             pass_info.attachment = i;
             render_pass->attachments.push_back(pass_info);
         }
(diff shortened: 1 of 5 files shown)


tobine:
I believe this is better than what is currently in place and may functionally address #401. However, I think this change highlights another hole in the code that I'm not certain about. From spec section 7.1 Render Pass Creation note `If the format has depth and/or stencil components, loadOp and storeOp apply only to the depth data, while stencilLoadOp and stencilStoreOp define how the stencil data is handled.`

For the code in question, a depth-stencil format with loadOp of CLEAR and stencilLoadOp of DONT_CARE will queue a function call to set the memory valid due to the single CLEAR loadOp, but no record will be kept of the fact that stencil data is invalid.

So in general it seems that we may need to update the code in question, as well as memory validity tracking, for depth-stencil formats. That doesn't have to be done now, but if others are in agreement we should at least create an issue to track this hole until it can be properly addressed.

mark-lunarg:
OK, then we might as well use this issue to track the necessary changes.

karl-lunarg:
LGTM +1  Are you going to keep this issue to track the memory validity tracking as Tobin mentioned and IAW with the TODO in the new code?

mark-lunarg:
Cherry-picked to Master.  

commit 1a65650f856376768d7b03ea2d080aaff87cacfd
Date:   Fri Jun 10 15:28:17 2016 -0600

    layers: GH401, Break out stencil load/storeOp check

    Stencil attachment load/storeOp settings were being ignored. Added
    format appropriate checks for these flags.

    Change-Id: I6a917fc8a28cbb0d0441152e8d6630defb1b81e3

PR #562, opened 2016-05-18
FOR REVIEW : Tobin modularize update descriptor sets

tobine:
For Review. Primary CL of interest here is last one which adds a cleaner UpdateDescriptorSets() interface to the DescriptorSet() class and moves the core_validation.cpp UpdateDescriptorSets() function to use the new interface in-line with Vulkan Validation Layer Authoring Guide.


layers/descriptor_sets.cpp
```diff
@@ -803,134 +705,51 @@ cvdescriptorset::ImageSamplerDescriptor::ImageSamplerDescriptor(
...
-        *error = error_str.str();
-        return false;
-    }
-    // Update is clean so process it
+void cvdescriptorset::ImageDescriptor::WriteUpdate(const VkWriteDescriptorSet *update, const uint32_t index) {
     updated = true;
-    image_view_ = image_view;
-    image_layout_ = image_layout;
-    return true;
+    auto image_info = update->pImageInfo[index];
```
chrisforbes: `const &`

layers/descriptor_sets.cpp
```diff
@@ -942,70 +761,361 @@ cvdescriptorset::BufferDescriptor::BufferDescriptor(const VkDescriptorType type,
...
-        *error = error_str.str();
-        return false;
-    }
+void cvdescriptorset::BufferDescriptor::WriteUpdate(const VkWriteDescriptorSet *update, const uint32_t index) {
     updated = true;
-    buffer_ = buffer;
-    offset_ = update->pBufferInfo[index].offset;
-    range_ = update->pBufferInfo[index].range;
-    return true;
+    auto buffer_info = update->pBufferInfo[index];
```
chrisforbes: `const &`

layers/descriptor_sets.cpp
```diff
@@ -942,70 +761,361 @@ cvdescriptorset::BufferDescriptor::BufferDescriptor(const VkDescriptorType type,
...
-        return false;
-    }
+void cvdescriptorset::BufferDescriptor::CopyUpdate(const Descriptor *src) {
+    auto buff_desc = static_cast<const BufferDescriptor *>(src);
     updated = true;
-    buffer_ = buffer;
-    offset_ = static_cast<const BufferDescriptor *>(src)->offset_;
-    range_ = static_cast<const BufferDescriptor *>(src)->range_;
-    return true;
+    buffer_ = buff_desc->buffer_;
```
chrisforbes: Do we catch the source descriptor never having been updated somewhere?

Can we just do `*this = *static_cast<BufferDescriptor const *>(src);` here rather than open-coding it?

layers/descriptor_sets.h
```diff
@@ -257,11 +242,15 @@ class BufferDescriptor : public Descriptor {
...
-    const std::unordered_map<VkBuffer, BUFFER_NODE> *buffer_map_;
 };
-// Helper function for Updating descriptor sets since it crosses multiple sets
-void UpdateDescriptorSets(VkDevice device, uint32_t descriptorWriteCount, const VkWriteDescriptorSet *pDescriptorWrites,
-                          uint32_t descriptorCopyCount, const VkCopyDescriptorSet *pDescriptorCopies);
+// Helper functions for Updating descriptor sets since it crosses multiple sets
+// Validate will make sure an update is ok without actually performing it
+bool ValidateUpdateDescriptorSets(const debug_report_data *,
+                                  const std::unordered_map<VkDescriptorSet, cvdescriptorset::DescriptorSet *> &, const uint32_t,
+                                  const VkWriteDescriptorSet *, const uint32_t, const VkCopyDescriptorSet *);
```
chrisforbes: Not a lot of value in by-value parameters being passed as const.

layers/descriptor_sets.cpp
```diff
@@ -598,28 +543,34 @@ bool cvdescriptorset::DescriptorSet::CopyUpdate(debug_report_data *report_data,
...
+    auto src_start_idx = src_set->GetGlobalStartIndexFromBinding(update->srcBinding) + update->srcArrayElement;
+    auto dst_start_idx = p_layout_->GetGlobalStartIndexFromBinding(update->dstBinding) + update->dstArrayElement;
     // Update parameters all look good so perform update
     for (uint32_t di = 0; di < update->descriptorCount; ++di) {
-        if (!descriptors_[dst_start_idx]->CopyUpdate(src_set->descriptors_[src_start_idx].get(), error))
-            return false;
+        descriptors_[dst_start_idx + di]->CopyUpdate(src_set->descriptors_[src_start_idx + di].get());
         ++num_updates;
     }
     if (num_updates != 0) {
```
chrisforbes: This tracking is silly now -- how about just:

`if (update->descriptorCount) some_update_ = true;`

layers/descriptor_sets.cpp
```diff
@@ -598,28 +543,34 @@ bool cvdescriptorset::DescriptorSet::CopyUpdate(debug_report_data *report_data,
...
         some_update_ = true;
     }
     InvalidateBoundCmdBuffers();
-    return true;
 }
-cvdescriptorset::SamplerDescriptor::SamplerDescriptor(
-    const std::unordered_map<VkSampler, std::unique_ptr<SAMPLER_NODE>> *sampler_map)
-    : sampler_(VK_NULL_HANDLE), immutable_(false), sampler_map_(sampler_map) {
+cvdescriptorset::SamplerDescriptor::SamplerDescriptor() : sampler_(VK_NULL_HANDLE), immutable_(false) {
     updated = false;
```
chrisforbes: May as well pull this stuff into the initializer list too

layers/descriptor_sets.cpp
```diff
@@ -726,75 +677,26 @@ bool cvdescriptorset::ValidateImageUpdate(const VkImageView image_view, const Vk
...
-        if (!ValidateSampler(sampler_, sampler_map_)) {
-            std::stringstream error_str;
-            error_str << "Attempted write update to immutable sampler descriptor which has invalid immutable sampler: " << sampler_
-                      << ".";
-            *error = error_str.str();
-            return false;
-        }
-        // TODO : Do we want a way to warn here in case of updating immutable sampler (which can't be updated)?
-    }
+void cvdescriptorset::SamplerDescriptor::WriteUpdate(const VkWriteDescriptorSet *update, const uint32_t index) {
```
chrisforbes: Can we leave a blank line between functions please?

layers/descriptor_sets.cpp
```diff
@@ -803,134 +705,51 @@ cvdescriptorset::ImageSamplerDescriptor::ImageSamplerDescriptor(
...
-        error_str << "Attempted write update to combined image sampler descriptor failed due to: " << error->c_str();
-        *error = error_str.str();
-        return false;
-    }
+void cvdescriptorset::ImageSamplerDescriptor::WriteUpdate(const VkWriteDescriptorSet *update, const uint32_t index) {
     updated = true;
-    image_view_ = image_view;
-    image_layout_ = image_layout;
-    return true;
+    auto image_info = update->pImageInfo[index];
```
chrisforbes: `const &`

layers/descriptor_sets.cpp
```diff
@@ -942,70 +761,361 @@ cvdescriptorset::BufferDescriptor::BufferDescriptor(const VkDescriptorType type,
...
+    }
+    // Now validate copy updates
+    for (i = 0; i < copy_count; ++i) {
+        auto dst_set = p_cds[i].dstSet;
+        auto src_set = p_cds[i].srcSet;
+        auto src_pair = set_map.find(src_set);
+        auto dst_pair = set_map.find(dst_set);
+        if (src_pair == set_map.end()) {
+            skip_call |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DESCRIPTOR_SET_EXT,
+                                 reinterpret_cast<uint64_t &>(src_set), __LINE__, DRAWSTATE_DOUBLE_DESTROY, "DS",
```
chrisforbes: I don't think `DRAWSTATE_DOUBLE_DESTROY` is really what you wanted here?

layers/descriptor_sets.cpp
```diff
@@ -942,70 +761,361 @@ cvdescriptorset::BufferDescriptor::BufferDescriptor(const VkDescriptorType type,
...
+                                 "Cannot call vkUpdateDescriptorSets() to copy from descriptor set 0x%" PRIxLEAST64
+                                 " that has not been allocated.",
+                                 reinterpret_cast<uint64_t &>(src_set));
+        } else if (dst_pair == set_map.end()) {
+            skip_call |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DESCRIPTOR_SET_EXT,
+                                 reinterpret_cast<uint64_t &>(dst_set), __LINE__, DRAWSTATE_DOUBLE_DESTROY, "DS",
```
chrisforbes: Ditto

layers/descriptor_sets.cpp
```diff
@@ -942,70 +761,361 @@ cvdescriptorset::BufferDescriptor::BufferDescriptor(const VkDescriptorType type,
...
+//  sets, and then calls their respective Validate[Write|Copy]Update functions.
+// If the update hits an issue for which the callback returns "true", meaning that the call down the chain should
+//  be skipped, then true is returned.
+// If there is no issue with the update, then false is returned.
+bool cvdescriptorset::ValidateUpdateDescriptorSets(
+    const debug_report_data *report_data, const std::unordered_map<VkDescriptorSet, cvdescriptorset::DescriptorSet *> &set_map,
+    const uint32_t write_count, const VkWriteDescriptorSet *p_wds, const uint32_t copy_count, const VkCopyDescriptorSet *p_cds) {
+    bool skip_call = false;
+    // Validate Write updates
+    uint32_t i = 0;
```
chrisforbes: Just declare this inline in the loops. Widening the scope like this makes people go look for tricky uses.

chrisforbes:
Various minor inline comments; LGTM otherwise.

layers/descriptor_sets.cpp
tobine: yep

layers/descriptor_sets.cpp
tobine: yep

layers/descriptor_sets.cpp
tobine: VerifyCopyUpdateContents() makes sure source contents are good. I can see some obscure corner cases where a non-updated source could get through that check, though, so I'll add "updated?" check as well.

Need to add "=" operator to do straight descriptor copy so will save that for follow-on commit

layers/descriptor_sets.cpp
tobine: yep

layers/descriptor_sets.h
tobine: true dat

layers/descriptor_sets.cpp
tobine: These members live in abstract base class so can't initialize in derived class

layers/descriptor_sets.cpp
tobine: yeah, there were a few of these.

layers/descriptor_sets.cpp
tobine: INVALID_DESCRIPTOR_SET is better, but need to re-work error codes at some point and move DESCRIPTOR stuff to its own set of error codes apart from old "DRAWSTATE" prefix.

tobine:

tobine:

PR #563, opened 2016-05-19
Assorted core validation cleanup

chrisforbes:


layers/core_validation.cpp
```diff
@@ -1955,14 +1955,30 @@ static bool validate_status(layer_data *my_data, GLOBAL_CB_NODE *pNode, CBStatus
...
-static PIPELINE_NODE *getPipeline(layer_data const *my_data, const VkPipeline pipeline) {
+static PIPELINE_NODE *getPipeline(layer_data const *my_data, VkPipeline pipeline) {
     auto it = my_data->pipelineMap.find(pipeline);
     if (it == my_data->pipelineMap.end()) {
         return nullptr;
     }
     return it->second;
 }
 
+static RENDER_PASS_NODE *getRenderPass(layer_data const *my_data, VkRenderPass renderpass) {
```
lentinem: I would prefer getRenderPassNode

layers/core_validation.cpp
```diff
@@ -1955,14 +1955,30 @@ static bool validate_status(layer_data *my_data, GLOBAL_CB_NODE *pNode, CBStatus
...
+    auto it = my_data->renderPassMap.find(renderpass);
+    if (it == my_data->renderPassMap.end()) {
+        return nullptr;
+    }
+    return it->second;
+}
+
+static FRAMEBUFFER_NODE *getFramebuffer(layer_data *my_data, VkFramebuffer framebuffer) {
```
lentinem: getFramebufferNode

layers/core_validation.cpp
```diff
@@ -4100,7 +4118,7 @@ static bool ValidateCmdBufImageLayouts(VkCommandBuffer cmdBuffer) {
                 if (cb_image_data.first.hasSubresource) {
                     skip_call |= log_msg(
                         dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
-                        reinterpret_cast<uint64_t &>(cmdBuffer), __LINE__, DRAWSTATE_INVALID_IMAGE_LAYOUT, "DS",
+                        reinterpret_cast<uint64_t &>(pCB->commandBuffer), __LINE__, DRAWSTATE_INVALID_IMAGE_LAYOUT, "DS",
```
lentinem: I'm curious what the motivation for this is. Do you expect the two not to be the same?

layers/core_validation.cpp
```diff
@@ -4110,7 +4128,7 @@ static bool ValidateCmdBufImageLayouts(VkCommandBuffer cmdBuffer) {
                 } else {
                     skip_call |= log_msg(
                         dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
-                        reinterpret_cast<uint64_t &>(cmdBuffer), __LINE__, DRAWSTATE_INVALID_IMAGE_LAYOUT, "DS",
+                        reinterpret_cast<uint64_t &>(pCB->commandBuffer), __LINE__, DRAWSTATE_INVALID_IMAGE_LAYOUT, "DS",
```
lentinem: Ditto.

tobine:
I amended the getRenderPass helper change to rename [primary|secondary]RenderPass to lowercase. Changes look good and no test issues.

layers/core_validation.cpp
chrisforbes: As an alternative, could we move toward dropping `node` everywhere? All this naming needs to change to conform to the style guide, and appending `node` is no better than appending `thing` or `object` -- there's no value there.

lentinem: Having something to differentiate the "node" from the vulkan object is probably a positive so it's easier to tell what is what. I agree that node is not descriptive. Something like data or tracking might be better.

tobine: Or "state" such as image layer which uses IMAGE_STATE.

PR #564, opened 2016-05-19
docs: update v0 languages for device layer deprecation

olvaffe:
Clarify that a layer's

  vkEnumerateInstanceLayerProperties
  vkEnumerateInstanceExtensionProperties
   - should enumerate the layer itself

  vkEnumerateDeviceLayerProperties
   - is deprecated

  vkEnumerateDeviceExtensionProperties
   - must handle all layers by chaining

  vkCreateInstance
   - `pNext` handling is covered by the spec

  vkCreateDevice
   - allow validation layers to validate layer and extension names

Clarify that the layer library's

  vkEnumerateInstanceLayerProperties
  vkEnumerateInstanceExtensionProperties
   - are not used by the desktop loader
   - can be aliases to the layer's versions when the layer library
     contains only one layer

  vkEnumerateDeviceLayerProperties
  vkEnumerateDeviceExtensionProperties
   - are not used by the desktop loader

  vkGetInstanceProcAddr
   - mention that the special cases for vkCreateDevice and device commands
     are for compatibility


jon-lunarg:
Commited to master. Thanks!

I think we need to add vkCreateInstance to the list of functions that a layer's GIPA can ignore instance on. I am seeing this in some third-party layers GIPA(NULL, "vkCreateInstance"). At one point in time this was in our layers and also in the docs as an example in the CreateInstance chaining.

olvaffe:
`GIPA(NULL, "vkCreateInstance")` is valid.  In this specific case, `GIPA(valid_instance, "vkCreateInstance")` is actually invalid :smiley:

jon-lunarg:
hmm?  The documents just says this about GIPA:
 - when `pName` is `vkCreateDevice`, it ignores `instance`.

I think a similar statement should be added:
 - when `pName` is `vkCreateInstance`, it ignores `instance`.

olvaffe:
`vkCreateDevice` is called out because it requires a valid instance to be queried but we have to relax that requirement.  `vkCreateInstance`, on the other hand, requires an invalid instance (`VK_NULL_HANDLE`) to be queried.  I do not think we need to say anything here.

jon-lunarg:
Okay I forgot that GIPA(NULL, "vkCreateInstance") is the only valid way for an app to get the function pointer.

PR #565, opened 2016-05-19
Loader-layer interface v1

olvaffe:
The version 1 interface is a complete redesign, but it should be very easy for layers to support both v0 and v1 (demonstrated with device_limits).


olvaffe:
Closing.  No intent to merge this at the moment, just to show the vision on v1.

loader/LoaderAndLayerInterface.md
```diff
@@ -1293,6 +1293,66 @@ vkGetDeviceProcAddr. Layers do not need to implement commands that are not going
...
+
+   ```
+   uint32_t
+   vk_layerGetLayerNames(const char* const** pppLayerNames);
+   ```
+
+   The number of layers in the layer libraries is returned.  When
+   `pppLayerNames` is non-`NULL`, it is set to an array of `const char*`,
+   pointing to the layer names of the layers.
+
```
jon-lunarg: Wouldn't it be better to return an array of VkLayerProperties to get more than just the names?

loader/LoaderAndLayerInterface.md
```diff
@@ -1293,6 +1293,66 @@ vkGetDeviceProcAddr. Layers do not need to implement commands that are not going
...
+   choice to use the interface version returned from the layer library or to
+   neglect the layer library.
+
+   Otherwise, `VK_ERROR_LAYER_NOT_PRESENT` is returned.
+
+   `vk_layerNegotiateLoaderLayerInterfaceVersion` is always called and only
+   called before any other function, unless the layer library also supports
+   interface version 0.  When interface version 0 is supported, the loader may
+   assume version 0 without calling this function first.
+
```
jon-lunarg: My question on this NegotiateLoaderLayerVersion is this the entire loader<->layer interface or only the layer library interface  (either introspection or manifest file)?  If it is only  the later then it simplifies things since the introspection/ manifest files  are used to get info about the layer libraries and layers contained within the library.  However, then if the chaining interface needs to be changed this version negotiating interface doesn't apply. If it is the entire interface, that is more future looking for possible changes in the chaining interface.  However, it seems then we might need a different vk_layerNegotiateLoaderLayerInterfaceVersion to more easily handle  layers with different chaining interfaces being mixed onto the same chain.  For example, perhaps chaining interface 0 is not compatible with chaining interface 1  then the loader would either need to makes sure all layers  in a chain used 0  or all used 1.  Or perhaps not advertise layers of the wrong chaining interface.  Anyway there is a lot of options that would need to be discussed.

loader/LoaderAndLayerInterface.md
olvaffe: That would make things easier.  The downside is that there will be another source of VkLayerProperties:

  (pseudo code)
  name = vk_layerGetLayerNames();
  gipa = vk_layerGetLayerProcAddr(name, "vkGetInstanceProcAddr");
  enumerate_instance_layer_properties = gipa(NULL, "vkEnumerateInstanceLayerProperties);
  props = enumerate_instance_layer_properties();

I am thinking making this `vk_layerGetLayerCount` instead.  The layer name is not meaningful here.  Then `vk_layerGetLayerProcAddr` will take a `layerIndex` instead of `pLayerName`.

loader/LoaderAndLayerInterface.md
olvaffe: The entire loader<->layer interface, including layer<->layer interface.  You raised a valid concern.

v1 as is does not introduce any change to vk_layer.h.  That is intentional.  The loader can safely mix v0 and v1 layers in any order.  But, say, v5 introduced a non-backward compatible change to vk_layer.h, then it should explicitly call out that a loader may:

 - configure v5 layers as v0~v4 layers, or
 - ignore or even drop support for v0~v4 layers due to incompatibility

So the expectation is, newer interface versions are backward compatible with older ones.

jon-lunarg:
I am worried there will be a good reason at some point in time to break backwards compatibility.  For that reason, I think the versionNegotiation should be two calls.  First call would be give the layers, list of supported interface versions.  Then the loader could enumerate all the versions that all the layers it is going to use support.  Then, the second call could be made to tell the layers which interface version to use.  This way,  the loader would have a better chance of not eliminating a layer because it negotiated an interface version with that layer that is incompatible with other layers.

courtney-g:
I believe that should work with the existing interface. First pass is to discover what the highest level supported by all layers is (via returned pSupportedVersion) and then the second pass would be setting that as the target version for all layers. It could take more than one pass if some layers don't support all previous versions. I.e. one layer might require N-1 for the interface, but another layer may only support N and N-2. First pass would find N-1 to be the version to use, until it gets to the layer that doesn't support it and then N-2 because the target version.

jon-lunarg:
Yeah I think the existing interface would work but require potentially many more calls than with a dedicated call that just enumerates all versions that a given layer supports.

olvaffe:
The number of calls should not be an issue.  I prefer vk_layerNegotiateLoaderLayerInterfaceVersion because it mirrors vk_icdNegotiateLoaderIcdInterfaceVersion.  Compatibility breakage may also happen with ICDs.  The loader could use similar logic to negotiate ICD and layer interface versions.

PR #567, opened 2016-05-19
layers: Add default layer error message config

mark-lunarg:
Allows layers to output error messages even if no vk_layer_settings.txt
config file is present. Sets defaults to LOG_MSG, error, stdout.

Change-Id: I49f37189665816df58c258b9e9629f2bf76751c8


tobine:
Hi Mark,
Code looks reasonable as far as I can tell. Two questions:

1. Have you done a reasonable amount of testing and hit this with various fail cases from bad vk_layer_settings?
2. Is there someway we can add automated testing for these features?

Automated testing something to think about longer term, but wanted to bring it up. If you've tested this to where you're comfortable with it, then I'm fine with the changes.

mark-lunarg:
Jon suggested that we kill any default debug_callbacks if an app registers its own callback, so we're adding that now.

mark-lunarg:
Tobin, with Jon's suggested change, if there is not a vk_layer_settings.txt file default callbacks will get unloaded if an app registers a debug callback.

mark-lunarg:
I have a couple of small cleanup things to hit, just giving you a chance to check out things and make recommendations before we put them up.

layers/vk_layer_utils.cpp
```diff
@@ -604,11 +604,19 @@ VkStringErrorFlags vk_string_validate(const int max_length, const char *utf8) {
...
+// Debug callbacks get created in three ways:
+//   o  Application-defined debug callbacks
+//   o  Through settings in a vk_layer_settings.txt file
+//   o  By default, if neither a debug callback or vk_layer_settings.txt file is present
+//
+// At layer initialization time, default logging callbacks are created to output layer error messages.
+// Any settings found in a vk_layer_settings.txt file will override any default settings. If an application
+// later registers a debug callback function and no vk_layer_settings.txt file was located, the default
+// logging callbacks will be removed. Any settings in a vk_layer_settings.txt file will override
+// those in an application's debug callback function.
```
tobine: The behavior described in this comment is acceptable to me.
I'm not 100% sold on vk_layer_settings.txt overriding an applications debug callback, but I can see the benefit.
If/when a vk_layer_settings.txt causes an applications debug callback to be overridden, can we make a callback just before disabling the callback to indicate that the vk_layer_settings.txt file is causing the callback to be disabled?
That avoids the head-scratching case of a callback never being hit b/c a dev has forgotten or is unaware that the vk_layer_settings.txt file is present.

tobine:
Good update overall. Made one enhancement request for a single callback notification just before callbacks are disabled due to the presence of a vk_layer_settings.txt file.

layers/vk_layer_utils.cpp
mark-lunarg: As I understand it, the layer_settings.txt fill will not disable any callbacks but will cause additional callbacks to be created.  So regardless of what the settings file says the app-defined callbacks are intact.  Maybe the wording in the comment above can be improved to make this more clear.

mark-lunarg:
Reworded the comment in vk_layer_utils.cpp, and moved some header stuff around at JonA's request.

loader/vk_loader_layer.h
```diff
@@ -0,0 +1,32 @@
...
+
+// Linked list node for tree of debug callback functions
+typedef struct VkLayerDbgFunctionNode_ {
+    VkDebugReportCallbackEXT msgCallback;
+    PFN_vkDebugReportCallbackEXT pfnMsgCallback;
+    VkFlags msgFlags;
+    void *pUserData;
+    bool default_callback;
+    struct VkLayerDbgFunctionNode_ *pNext;
+} VkLayerDbgFunctionNode;
```
courtney-g: Note: Github complains about missing newline at end of vk_loader_layer.h

layers/vk_layer_logging.h
```diff
@@ -158,21 +137,63 @@ static inline void layer_destroy_msg_callback(debug_report_data *debug_data, VkD
                 debug_data->g_pDbgFunctionHead = pTrav->pNext;
             }
             debug_report_log_msg(debug_data, VK_DEBUG_REPORT_DEBUG_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEBUG_REPORT_EXT,
-                                 (uint64_t)pTrav->msgCallback, 0, VK_DEBUG_REPORT_ERROR_CALLBACK_REF_EXT, "DebugReport",
-                                 "Destroyed callback");
-        } else {
+                (uint64_t)pTrav->msgCallback, 0, VK_DEBUG_REPORT_ERROR_CALLBACK_REF_EXT, "DebugReport",
```
courtney-g: Do we need to restore the default callback once all the app callbacks have been removed?

courtney-g:
I'm wondering if this could be simpler.
What if in debug_report_log_msg if the list of registered callbacks is empty, call the "default" callback. Otherwise dispatch messages to the callbacks on the list and do not call the default callback.
I.e.
if (pTrav) {
  while (pTrav) ...
} else {
  default_callback()...
}

mark-lunarg:
Pushed to Master:

commit e249e6e3eb6a10ff46127e0712dc6eebb8f8fb99
Date:   Tue May 24 16:04:56 2016 -0600

    layers: Move debug action enums from vk_layer.h

    Moved debug-report related items from vk_layer.h. Enums are now in
    vk_layer_logging.h and DbgFunctionNode has been moved into a new
    loader header file vk_loader_layer.h.

    Change-Id: I6031146ba474ff01ca039da44ad5d42d054383a3

commit c34ab18ac7c064590a1020a8db4b6abf5a8d3a18
Date:   Thu May 19 15:27:18 2016 -0600

    layers: Add default layer error message config

    Allows layers to output error messages even if no vk_layer_settings.txt
    config file is present. Sets defaults to LOG_MSG, error, stdout.
    A layer settings file will override any default values. If no settings
    file is present and an app creates a debug callback, the default
    callbacks will be removed and unregistered.

    Change-Id: I49f37189665816df58c258b9e9629f2bf76751c8

PR #572, opened 2016-05-20
Fix event tracking to differentiate reading and writing

lentinem:


layers/core_validation.cpp
```diff
@@ -4133,6 +4137,10 @@ static bool validateAndIncrementResources(layer_data *my_data, GLOBAL_CB_NODE *p
             eventNode->second.in_use.fetch_add(1);
         }
     }
+    for (auto event : pCB->writeEventsBeforeWait) {
+        auto eventNode = my_data->eventMap.find(event);
+        eventNode->second.write_in_use++;
```
chrisforbes: I see increment and test of this -- should there be a reset somewhere?

lentinem: Good point. There shouldn't be a reset but there should be a decrement which was added in decrementResources.

tobine:

PR #573, opened 2016-05-20
Rename things still using imageArraySize instead of imageArrayLayers

corngood:
I'm not sure if this was left intentionally to avoid changing SWAPCHAIN_ERROR, but it seems like an oversight.


mark-lunarg:
Cherry-picked and pushed to Master.  Thanks David.

commit 339b6d556bc4a2d5d120fbd3a7a8ad651e7059a5
Author: David McFarland <corngood@gmail.com>
Date:   Fri May 20 18:26:28 2016 -0300

    layers: PR573, Handle imageArraySize rename

    Swapchain layer had stale references -- rename things still using
    imageArraySize instead of imageArrayLayers.

    Change-Id: Iaa4500ab8cf55e50b96636a50136605dd3e6d83e

PR #577, opened 2016-05-23
Mark fence in use after the check.

lentinem:


chrisforbes:
:+1: for the focused fix, but this function is still poorly-behaved in validation and driver failure cases.

tobine:

PR #579, opened 2016-05-23
loader: doc update for merge android and desktop layer interface

jon-lunarg:
Change-Id: Ibbffc0bf05bcbd1aa3b9272caa0b2c1c3a54037d

### DIFF
--- a/loader/LoaderAndLayerInterface.md
+++ b/loader/LoaderAndLayerInterface.md
@@ -1020,11 +1020,29 @@ is itself.  In other cases, it should normally chain to other layers.
 extension names.  It may assume the layer names and extension names have been
 validated.
 
-`vkGetInstanceProcAddr` can intercept a command by returning the function
-pointer of the layer's associated intercept function, otherwise it should return the value obtained through chaining or NULL.
+`vkGetInstanceProcAddr` intercepts a Vulkan command by returning a local entry point,
+otherwise it returns the value obtained by calling down the instance chain.
+   Detailed specification of this function,
+   - when `pName` is `vkEnumerateInstanceLayerProperties`,
+     `vkEnumerateInstanceExtensionProperties`, or
+     `vkEnumerateDeviceLayerProperties` (but _not_
+     `vkEnumerateDeviceExtensionProperties`), it returns a function pointer to
+     the corresponding introspection function defined by this interface.
+   - when `pName` is `vkGetInstanceProcAddr`, it returns a function pointer
+     to itself.
+   - when `pName` is `vkCreateInstance`, it returns a function pointer to the
+     layer's vkCreateInstance, which is required for instance chaining.
+   - when `pName` is `vkCreateDevice`, it returns a function pointer to the
+     layer's vkCreateDevice, which is only required for any device chaining.
 
-`vkGetDeviceProcAddr` can intercept a command by returning the function pointer
-of the layer's associated intercept function, otherwise it should return the value obtained through chaining or NULL.
+   For compatibility with older layer libraries,
+   - when `pName` is `vkCreateDevice`, it ignores `instance`.
+
+`vkGetDeviceProcAddr` intercepts a Vulkan command by returning a local entry point,
+otherwise it returns the value obtained by calling down the device chain.
+
+Vulkan commands that are defined by an extension that is disabled, must not be intercepted in either
+`vkGetInstanceProcAddr` or `vkGetDeviceProcAddr`.
 
 [\*]: The intention is for layers to have a well-defined baseline behavior.
 Some of the conventions or rules, for example, may be considered abuses of the
@@ -1064,42 +1082,12 @@ The introspection functions are not used by the desktop loader.
 
 It must also define and export these functions one for each layer in the library:
 
- - `<layerName>GetInstanceProcAddr(instance, pName)` either returns NULL, a local entry point or the value obtained
-    by calling down the instance chain (the next layer/ICD's GetInstanceProcAddr).
-
-   Detailed specification of this function,
-   - when `pName` is `vkEnumerateInstanceLayerProperties`,
-     `vkEnumerateInstanceExtensionProperties`, or
-     `vkEnumerateDeviceLayerProperties` (but _not_
-     `vkEnumerateDeviceExtensionProperties`), it returns a function pointer to
-     the corresponding introspection function defined by this interface.
-   - when `pName` is `vkGetInstanceProcAddr`, it returns a function pointer
-     to itself.
-   - when `pName` is `vkCreateInstance`, it returns a function pointer to the
-     layer's vkCreateInstance, which is required for instance chaining.
-   - when `pName` is `vkCreateDevice`, it returns a function pointer to the
-     layer's vkCreateDevice, which is only required for any device chaining.
-   - when `pName` is any Vulkan command including any enabled Vulkan extension command that the layer intercepts,
-     it returns a function pointer to the intercept routine.
-   - when `pName` is any disabled Vulkan extension command that the layer intercepts, it returns NULL.
-   - when `pName` is any Vulkan command that the layer does NOT intercept,
-     it returns the value from calling the next entity's GetInstanceProcAddr.
-
-   For compatibility with older layer libraries,
-   - when `pName` is `vkCreateDevice`, it ignores `instance`.
+ - `<layerName>GetInstanceProcAddr(instance, pName)` behaves identically to a <layerName>'s vkGetInstanceProcAddr except it is exported.
 
    When a layer library contains only one layer, this function may
    alternatively be named `vkGetInstanceProcAddr`.
 
- - `<layerName>GetDeviceProcAddr` either returns NULL, a local entry point or the value obtained
-    by calling down the device chain (the next layer/ICD's GetDeviceProcAddr).
-
-   Detailed specification of this function,
-   - when `pName` is a device level Vulkan command including any enabled Vulkan extension command that the layer intercepts,
-     it returns a function pointer to the intercept routine.
-   - when `pName` is any disabled Vulkan extension command that the layer intercepts, it returns NULL.
-   - when `pName` is an device level Vulkan command that the layer does NOT intercept,
-     it returns the value from calling the next entity's GetDeviceProcAddr.
+ - `<layerName>GetDeviceProcAddr`  behaves identically to a <layerName>'s vkGetDeviceProcAddr except it is exported.
 
    When a layer library contains only one layer, this function may
    alternatively be named `vkGetDeviceProcAddr`.
@@ -1143,7 +1131,7 @@ of the "file_format_version" and includes the semantics of the nodes in the JSON
 For interface version 0 the file format version must be "1.0.0"
 
 Note 1: Prior to deprecation, the "type" node was used to indicate which layer chain(s)
-to activate the layer upon: instnace, device, or both.
+to activate the layer upon: instance, device, or both.
 Distinct instance and device layers are deprecated; there are now just layers.
 Allowable values for type (both before and after deprecation) are "INSTANCE", "GLOBAL" and, "DEVICE."
 "DEVICE" layers are skipped over by the loader as if they were not found.
--- a/loader/LoaderAndLayerInterface.md
+++ b/loader/LoaderAndLayerInterface.md
@@ -1022,18 +1022,10 @@ validated.
 
 `vkGetInstanceProcAddr` intercepts a Vulkan command by returning a local entry point,
 otherwise it returns the value obtained by calling down the instance chain.
-   Detailed specification of this function,
-   - when `pName` is `vkEnumerateInstanceLayerProperties`,
-     `vkEnumerateInstanceExtensionProperties`, or
-     `vkEnumerateDeviceLayerProperties` (but _not_
-     `vkEnumerateDeviceExtensionProperties`), it returns a function pointer to
-     the corresponding introspection function defined by this interface.
-   - when `pName` is `vkGetInstanceProcAddr`, it returns a function pointer
-     to itself.
-   - when `pName` is `vkCreateInstance`, it returns a function pointer to the
-     layer's vkCreateInstance, which is required for instance chaining.
-   - when `pName` is `vkCreateDevice`, it returns a function pointer to the
-     layer's vkCreateDevice, which is only required for any device chaining.
+    These commands must be intercepted
+   - vkGetInstanceProcAddr
+   - vkCreateInstance
+   - vkCreateDevice (only required for any device-level chaining)
 
    For compatibility with older layer libraries,
    - when `pName` is `vkCreateDevice`, it ignores `instance`.
@@ -1041,8 +1033,9 @@ otherwise it returns the value obtained by calling down the instance chain.
 `vkGetDeviceProcAddr` intercepts a Vulkan command by returning a local entry point,
 otherwise it returns the value obtained by calling down the device chain.
 
-Vulkan commands that are defined by an extension that is disabled, must not be intercepted in either
-`vkGetInstanceProcAddr` or `vkGetDeviceProcAddr`.
+The specification requires `NULL` to be returned from `vkGetInstanceProcAddr` and
+`vkGetDeviceProcAddr` for disabled commands.  A layer may return `NULL` itself or
+rely on the following layers to do so.
 
 [\*]: The intention is for layers to have a well-defined baseline behavior.
 Some of the conventions or rules, for example, may be considered abuses of the
@@ -1082,12 +1075,12 @@ The introspection functions are not used by the desktop loader.
 
 It must also define and export these functions one for each layer in the library:
 
- - `<layerName>GetInstanceProcAddr(instance, pName)` behaves identically to a <layerName>'s vkGetInstanceProcAddr except it is exported.
+ - `<layerName>GetInstanceProcAddr(instance, pName)` behaves identically to a layer's vkGetInstanceProcAddr except it is exported.
 
    When a layer library contains only one layer, this function may
    alternatively be named `vkGetInstanceProcAddr`.
 
- - `<layerName>GetDeviceProcAddr`  behaves identically to a <layerName>'s vkGetDeviceProcAddr except it is exported.
+ - `<layerName>GetDeviceProcAddr`  behaves identically to a layer's vkGetDeviceProcAddr except it is exported.
 
    When a layer library contains only one layer, this function may
    alternatively be named `vkGetDeviceProcAddr`.
--- a/loader/LoaderAndLayerInterface.md
+++ b/loader/LoaderAndLayerInterface.md
@@ -693,6 +693,7 @@ Vulkan layer interface with the loader
 
 #### Windows
 
+<a name="ManifestFileExample"></a>
 ##### Properly-Installed Layers
 
 In order to find properly-installed layers, the Vulkan loader will use a
@@ -732,71 +733,8 @@ full pathname to the manifest file.
 The Vulkan loader will open each info file to obtain information about the
 layer, including the name or pathname of a shared library (".dll") file.
 
-This manifest file is in the JSON format and contains the following
-information:
-
-- (required) "file\_format\_version" - same as for ICDs, except that the format
-version can vary independently for ICDs and layers.
-
-- (required) "name" - layer name
-
-- (required and deprecated) "type" - which layer chains should the layer be activated on.
-Distinct instance and device layers are deprecated; there are now just layers.
-Allowable values for type (both before and after deprecation) are "INSTANCE", "GLOBAL" and, "DEVICE."
-"DEVICE" layers are skipped over by the loader as if they were not found.
-Thus, layers must have a type of "GLOBAL" or "INSTANCE" for the loader to include the layer in it's discovery.
-
-- (required) "library\_path" - filename / full path / relative path to the
-library file
-
-- (required) "api\_version" - same as for ICDs.
-
-- (required) "implementation\_version" - layer version, a single number
-increasing with backward compatible changes.
-
-- (required) "description" - informative description of the layer.
-
-- (optional) "device\_extensions" or "instance\_extensions" - array of
-extension information as follows
-
-    - (required) extension "name" - Vulkan registered name
-
-    - (required) extension "spec\_version" - extension specification version, a
-single number, increasing with backward compatible changes.
-
-    - (required for device\_extensions with entry points) extension
-"entrypoints" - array of device extension entry points; not used for instance
-extensions
-
-- (sometimes required) "functions" - mapping list of function entry points. If
-multiple layers exist within the same shared library (or if a layer is in the
-same shared library as an ICD), this must be specified to allow each layer to
-have its own vkGet\*ProcAddr entry points that can be found by the loader. At
-this time, only the following two functions are required:
-
-    - "vkGetInstanceProcAddr" name
-
-    - "vkGetDeviceProcAddr" name
-
-- (optional for implicit layers) "enable\_environment" requirement(s) -
-environment variable and value required to enable an implicit layer. This
-environment variable (which should vary with each "version" of the layer, as in
-"ENABLE\_LAYER\_FOO\_1") must be set to the given value or else the implicit
-layer is not loaded. This is for application environments (e.g. Steam) which
-want to enable a layer(s) only for applications that they launch, and allows
-for applications run outside of an application environment to not get that
-implicit layer(s).
-
-- (required for implicit layers) "disable\_environment" requirement(s) -
-environment variable and value required to disable an implicit layer. Note: in
-rare cases of an application not working with an implicit layer, the
-application can set this environment variable (before calling Vulkan commands)
-in order to "blacklist" the layer. This environment variable (which should vary
-with each "version" of the layer, as in "DISABLE\_LAYER\_FOO\_1") must be set
-(not particularly to any value). If both the "enable\_environment" and
-"disable\_environment" variables are set, the implicit layer is disabled.
-
-For example:
+This manifest file is in the JSON format as shown in the following example.
+See the section [Layer Library Manifest File](#LayerLibraryManifestFile) for more information about each of the nodes in the JSON file.
 
 ```
 {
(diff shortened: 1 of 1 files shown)


loader/LoaderAndLayerInterface.md
```diff
@@ -1180,72 +1064,112 @@ The introspection functions are not used by the desktop loader.
...
 
    When a layer library contains only one layer, this function may
    alternatively be named `vkGetDeviceProcAddr`.
 
-All contained layers must support [`vk_layer.h`][].  They do not need to
-implement commands that are not queryable.  They are recommended not to export
-any command.
+All layers contained within a library must support [`vk_layer.h`][].  They do not need to
+implement commands that they do not intercept.  They are recommended not to export
+any commands other than those required to be exported.
```
olvaffe: A layer library is required to export some functions.  A layer inside a layer library is never required to export any command, and we should keep recommending not to.

loader/LoaderAndLayerInterface.md
```diff
@@ -1180,72 +1064,112 @@ The introspection functions are not used by the desktop loader.
...
+ - `<layerName>GetDeviceProcAddr` either returns NULL, a local entry point or the value obtained
+    by calling down the device chain (the next layer/ICD's GetDeviceProcAddr).
+
+   Detailed specification of this function,
+   - when `pName` is `vkGetDeviceProcAddr`, it returns a function pointer
+     to itself.
+   - when `pName` is a device level Vulkan command that the layer intercepts,
+     it returns a function pointer to the intercept routine.
+   - when `pName` is an device level Vulkan command that the layer does NOT intercept,
+     it returns the value from calling the next entity's GetDeviceProcAddr.
```
olvaffe: If a layer does not intercept any device command, its GDPA should not need to return itself when queried.

Can a layer hide a device extension from the ICD?  If so, it needs to return NULL for device commands from hidden device extensions.

Generally speaking, I think we over-specify here.  We should instead say that a layer intercepts a command if its GIPA/GDPA returns a different pointer than the following layer would.  We can specify the implications when a layer intercepts certain commands, if any.

loader/LoaderAndLayerInterface.md
```diff
@@ -1180,72 +1064,112 @@ The introspection functions are not used by the desktop loader.
 
 It must also define and export these functions:
 
- - `<layerName>GetInstanceProcAddr` behaves as if `<layerName>`'s
-   `vkGetInstanceProcAddr` is called, except
+ - `<layerName>GetInstanceProcAddr(instance, pName)` either returns NULL, a local entry point or the value obtained
+    by calling down the instance chain (the next layer/ICD's GetInstanceProcAddr).
 
+   Detailed specification of this function,
+   - when `instance` is invalid the layer may return NULL.
```
olvaffe: Why?  This also conflicts with what would happen when `vkCreateInstance` is queried.

loader/LoaderAndLayerInterface.md
```diff
@@ -1180,72 +1064,112 @@ The introspection functions are not used by the desktop loader.
...
    - when `pName` is `vkEnumerateInstanceLayerProperties`,
      `vkEnumerateInstanceExtensionProperties`, or
      `vkEnumerateDeviceLayerProperties` (but _not_
      `vkEnumerateDeviceExtensionProperties`), it returns a function pointer to
      the corresponding introspection function defined by this interface.
    - when `pName` is `vkGetInstanceProcAddr`, it returns a function pointer
      to itself.
+   - when `pName` is `vkCreateInstance`, it returns a function pointer to the
+     layer's vkCreateInstance required for chaining.
+   - when `pName` is `vkCreateDevice`, it returns a function pointer to the
```
olvaffe: A layser should not be required to intercept `vkCreateDevice` (though it will not be terribly useful).

loader/LoaderAndLayerInterface.md
```diff
@@ -1180,72 +1064,112 @@ The introspection functions are not used by the desktop loader.
...
+     layer's vkCreateDevice required for chaining.
+   - when `pName` is an instance level Vulkan command that the layer intercepts,
+     it returns a function pointer to the intercept routine.
+   - when `pName` is an instance level Vulkan command that the layer does NOT intercept,
+     it returns the value from calling the next entity's GetInstanceProcAddr.
+   - when `pName` is a device level Vulkan command, the return value may be any of
+     NULL, a local entry point, or the value from calling the next entity's GetInstanceProcAddr.
```
olvaffe: We need GIPA to be sane for device commands.  That is, the only difference between calling the pointers returned by GIPA("vkCreateImage") and GDPA("vkCreateImage") should be performance.  Same for all core or non-core commands.

See also my comment on GDPA.

loader/LoaderAndLayerInterface.md
courtney-g: This seems like a classic "undefined" behavior. If an application uses an invalid VkInstance, the result is "undefined", that is unpredictable, it may work, it may crash, don't know.

loader/LoaderAndLayerInterface.md
jon-lunarg: okay

loader/LoaderAndLayerInterface.md
jon-lunarg: Yes the GDPA(("vkGetDeviceProcAddr") is not strictly needed AFAIK. But I added it for completeness. We could remove it but then the layer's GDPA  is less true to the  Vulkan specs description of GDPA. I am flexible though.

On hiding a ICD's device extension, I am not sure if we should comment on that one way or another without  Khronos working group feedback.  Although in practice anyway can write a Vulkan layer that  does whatever it is possible for layer to do and if apps like it the behavior and use the layer no one is preventing that no matter what we specify in this document.

I think it is good to have more descriptive and explanatory words rather than just minimalistic language.  I did have one (very smart SW person in the Vulkan Working Group) comment that they read this spec and they couldn't understand what they needed to do to write a layer. Then they looked at validation layers and thought that they needed to do a bunch of stuff that
is common in validation  (eg debug_report and a bunch of other utilities).  Finally when I pointed then to some of the simple example layers in VulkanTools repo they were able to understand the requirements better.

This is an example of where I think more language is better.  I think it is better to definitively say that the layer must return the function pointer to the intercept routine  in GDPA in order to intercept.  Rather than say a layer intercepts a command if it returns a function pointer different than the next layer would is accurate but rather vague as far as what exactly how to code a layer's GDPA so it will work.  Can I just return  my layer's GDPA function pointer to intercept all Vulkan device commands?

loader/LoaderAndLayerInterface.md
jon-lunarg: By "invalid" I meant  !(instance == NULL if pname==createinstance or enumnerateinstance*properties  && instance == a valid VkInstance in other cases).

I am fine with undefined behavior.

loader/LoaderAndLayerInterface.md
jon-lunarg: okay

loader/LoaderAndLayerInterface.md
jon-lunarg: okay so we can just mimic here whatever we agree to as the language for intercept versus non-intercepted functions?

jon-lunarg: Is this GIPA handling device commands that are intercepted by a layer needed for version 0 or would this be something to put in version 1?  For desktop it doesn't matter if a layer's GIPA intercepts device commands or not.  I imagine for Android existing layers that don't intercept any of their device commands in GIPA probably wouldn't work?  I noticed you added this to most (maybe all) of the validation layers so they now intercept their device commands in GIPA.  But there are other layers out their that don't do this but limit themselves strictly to instance level command in their GIPA.  Therefore it might be best to make this a version 1 requirement rather than a version 0 requirement and leave it optional in version 0. But I am wondering what you think?

loader/LoaderAndLayerInterface.md
olvaffe: How is omitting the requirement for GDPA("vkGetDeviceProcAddr") be less true to the spec?  The last two bullets are fine.

I am also fine with more language.  But it is better to be clear what are definitions/requirements and what are examples.  In this case, a command is considered intercepted if GDPA returns a value different from what would be returned from chaining.  This is the definition.  We may go on to say that, generally, when a command is intercepted, GDPA returns the intercepting command.  When not, GDPA calls down the next layer.

loader/LoaderAndLayerInterface.md
olvaffe: I prefer to omit this and leave it undefined.

loader/LoaderAndLayerInterface.md
olvaffe: We have workarounds so that we are fine with making exceptions for device commands defined in Vulkan 1.0 and in WSI extensions.  But we do need GIPA to be sane for other non-WSI device extensions.

Do you know if those layers advertise non-WSI device extensions that require new entrypoints?

loader/LoaderAndLayerInterface.md
olvaffe: Since a layer's `vkEnumereate*Properties` is now documented in the previous section, it is actually possible to directly export a layer's `vkGetInstanceProcAddr`:

```diff
diff --git a/loader/LoaderAndLayerInterface.md b/loader/LoaderAndLayerInterface.md
index 4a986cd..2f3cfde 100644
--- a/loader/LoaderAndLayerInterface.md
+++ b/loader/LoaderAndLayerInterface.md
@@ -1178,39 +1178,36 @@ signatures, and other similarities:
 
 The introspection functions are not used by the desktop loader.
 
-It must also define and export these functions:
-
- - `<layerName>GetInstanceProcAddr` behaves as if `<layerName>`'s
-   `vkGetInstanceProcAddr` is called, except
-
-   - when `pName` is `vkEnumerateInstanceLayerProperties`,
-     `vkEnumerateInstanceExtensionProperties`, or
-     `vkEnumerateDeviceLayerProperties` (but _not_
-     `vkEnumerateDeviceExtensionProperties`), it returns a function pointer to
-     the corresponding introspection function defined by this interface.
-   - when `pName` is `vkGetInstanceProcAddr`, it returns a function pointer
-     to itself.
-
-   For compatibility with older layer libraries,
-
-   - when `pName` is `vkCreateDevice`, it ignores `instance`.
-   - when `pName` is a device command defined by Vulkan 1.0 or
-     `VK_KHR_swapchain` (but _not_ other device commands), it may chain to
-     other layers without intercepting.  A loader should avoid querying such
-     device commands.
-
-   When a layer library contains only one layer, this function may
-   alternatively be named `vkGetInstanceProcAddr`.
-
- - `<layerName>GetDeviceProcAddr` behaves as if `<layerName>`'s
-   `vkGetDeviceProcAddr` is called.
-
-   When a layer library contains only one layer, this function may
-   alternatively be named `vkGetDeviceProcAddr`.
-
-All contained layers must support [`vk_layer.h`][].  They do not need to
-implement commands that are not queryable.  They are recommended not to export
-any command.
+Each layer's `vkGetInstanceProcAddr` must be exported as
+`<layerName>GetInstanceProcAddr`, unless there is only one layer.  When the
+layer library contains only one layer, it may alternatively export the layer's
+`vkGetInstanceProcAddr` simply as `vkGetInstanceProcAddr`.
+
+The layer library must also export each layer's `vkGetDeviceProcAddr` as
+`<layerName>GetDeviceProcAddr` when it is intercepted.  Similar to
+`<layerName>GetInstanceProcAddr`, when the layer library contains only one
+layer, it may alternatively export the layer's `vkGetInstanceProcAddr` simply
+as `vkGetInstanceProcAddr`.
+
+For compatibility with older layer libraries, a layer's
+`vkGetInstanceProcAddr`
+
+ - must ignore `instance` when `pName` is `vkCreateDevice`
+ - may be undefined when `pName` is one of the device commands defined by
+   Vulkan 1.0 or `VK_KHR_swapchain`.  A loader/layer should not use
+   `vkGetInstanceProcAddr` to query such device commands.
+
+   `vkGetInstanceProcAddr` must be well-defined otherwise.  Noticeably, it
+   must properly handle device commands defined by extensions other than
+   `VK_KHR_swapchain`.  In other words, the only difference between calling
+   the function pointers returned by `vkGetInstanceProcAddr` and
+   `vkGetDeviceProcAddr` should be performance.
+
+   A future interface version will require `vkGetInstanceProcAddr` to be
+   well-defined in all cases.
+
+All contained layers must support [`vk_layer.h`][].  They are recommended not
+to export any command except for those explicitly specified.
 
 ###### Layer Library Interface Version 0 (Windows and Linux)
 On Windows and Linux (desktop), the loader uses manifest files to discover
```

courtney-g: In the Olv's paragraph starting with "The layer library must also export each layer's `vkGetDeviceProcAddr`", I believe the last two vkGetInstanceProcAddr's should be vkGetDeviceProcAddr.

jon-lunarg: okay

loader/LoaderAndLayerInterface.md
jon-lunarg: What if a layer A returned  NULL for a GDPA  query but the next layer B returned an intercepted function for the same GDPA query. This meets your definition of intercept for layer A but it is clearly not what is intended behavior for layer A intercepting.

loader/LoaderAndLayerInterface.md
jon-lunarg: okay

loader/LoaderAndLayerInterface.md
jon-lunarg: There is at least one (non-sample) layer I know of that advertises non-WSI device extensions (EXT_debug_marker). I am pretty sure this layer limits it GIPA to instance level commands.

jon-lunarg:
I pushed an updated patch that addresses many of the comments. I went ahead and added the spec for a layer's GIPA that it returns function pointers for device level commands in version 0 interface. But we may not want this.

loader/LoaderAndLayerInterface.md
olvaffe: We will still say layer A intercept that function.  The function does behaves differently because of layer A.  Other components (loaders/layers/drivers) do not need to know what layer A intends to do.

On the other hand, the spec rules that layer A can do so only for un-enabled device extension commands.  Since the function belongs to an un-enabled device extension, layer B would have to return NULL as well.  The fact layer A returning NULL for the function makes no difference.

loader/LoaderAndLayerInterface.md
olvaffe: Is it open sourced?  I am willing to fix it.  If not, I hope I can at least talk with the developers.

loader/LoaderAndLayerInterface.md
```diff
@@ -1178,74 +1062,109 @@ signatures, and other similarities:
...
- - `<layerName>GetDeviceProcAddr` behaves as if `<layerName>`'s
-   `vkGetDeviceProcAddr` is called.
+ - `<layerName>GetDeviceProcAddr` either returns NULL, a local entry point or the value obtained
+    by calling down the device chain (the next layer/ICD's GetDeviceProcAddr).
+
+   Detailed specification of this function,
+   - when `pName` is a device level Vulkan command that the layer intercepts,
+     it returns a function pointer to the intercept routine.
+   - when `pName` is an device level Vulkan command that the layer does NOT intercept,
+     it returns the value from calling the next entity's GetDeviceProcAddr.
```
olvaffe: If a layer intercepts VK_KHR_swapchain but the extension is not enabled, this implies valid function pointers to be returned for swapchain functions.  That is a violation with the spec.

The real issue is we should not define `<layerName>GetDeviceProcAddr` out of nowhere.  There will always be gray area.  We should use the spec's `vkGetDeviceProcAddr` as the base and expand on it.

loader/LoaderAndLayerInterface.md
```diff
@@ -1178,74 +1062,109 @@ signatures, and other similarities:
...
    - when `pName` is `vkEnumerateInstanceLayerProperties`,
      `vkEnumerateInstanceExtensionProperties`, or
      `vkEnumerateDeviceLayerProperties` (but _not_
      `vkEnumerateDeviceExtensionProperties`), it returns a function pointer to
      the corresponding introspection function defined by this interface.
    - when `pName` is `vkGetInstanceProcAddr`, it returns a function pointer
      to itself.
+   - when `pName` is `vkCreateInstance`, it returns a function pointer to the
+     layer's vkCreateInstance required for instance chaining.
+   - when `pName` is `vkCreateDevice`, it returns a function pointer to the
```
olvaffe: A layer may not intercept `vkCreateDevice`.

loader/LoaderAndLayerInterface.md
```diff
@@ -1178,74 +1062,109 @@ signatures, and other similarities:
...
+   Detailed specification of this function,
    - when `pName` is `vkEnumerateInstanceLayerProperties`,
      `vkEnumerateInstanceExtensionProperties`, or
      `vkEnumerateDeviceLayerProperties` (but _not_
      `vkEnumerateDeviceExtensionProperties`), it returns a function pointer to
      the corresponding introspection function defined by this interface.
    - when `pName` is `vkGetInstanceProcAddr`, it returns a function pointer
      to itself.
+   - when `pName` is `vkCreateInstance`, it returns a function pointer to the
+     layer's vkCreateInstance required for instance chaining.
```
olvaffe: Do we require `vkCreateInstance` to be intercepted elsewhere?  If so, this bullet is not needed.

loader/LoaderAndLayerInterface.md
```diff
@@ -1178,74 +1062,109 @@ signatures, and other similarities:
...
+    by calling down the instance chain (the next layer/ICD's GetInstanceProcAddr).
 
+   Detailed specification of this function,
    - when `pName` is `vkEnumerateInstanceLayerProperties`,
      `vkEnumerateInstanceExtensionProperties`, or
      `vkEnumerateDeviceLayerProperties` (but _not_
      `vkEnumerateDeviceExtensionProperties`), it returns a function pointer to
      the corresponding introspection function defined by this interface.
    - when `pName` is `vkGetInstanceProcAddr`, it returns a function pointer
      to itself.
```
olvaffe: If you envision `<layerName>GetInstanceProcAddr` to be the `<layerName>`'s `vkGetInstanceProcAddr`, the first two bullets can be removed.  But then we should be explicit that `<layerName>GetInstanceProcAddr` just exports the layer's `vkGetInstanceProcAddr`.

loader/LoaderAndLayerInterface.md
jon-lunarg: A layer on it's own is already possibly non-conformant to the spec  (eg a layer's GIPA doesn't have to return a valid function pointer for every Vulkan command). Also the spec may change independently of this document, thus it is better to over specify in this document if it means repetition.
But you bring up a good point we need to specify what happens with disabled extensions.

loader/LoaderAndLayerInterface.md
jon-lunarg: okay making it optional.

loader/LoaderAndLayerInterface.md
jon-lunarg: I couldn't find this listed elsewhere

loader/LoaderAndLayerInterface.md
jon-lunarg: I believe you mean in the case where a layer library has just one layer?

jon-lunarg:
I updated the doc to change the GIPA and GDPA description's in the section:
##### Layer Library API Version 0

loader/LoaderAndLayerInterface.md
olvaffe: It should not matter.  The behavior of the exported `vkEnumerateInstanceLayerProperties` is defined by this section.  The behavior of the function pointer returned by

 `<layerName>GetInstanceProcAddr(VK_NULL_HANDLE, "vkEnumerateInstanceLayerProperties")`

would be defined by section `Layer Conventions and Rules`.

But it is true that they will differ when there are more than one layer in the library.  I think saying `<layerName>GetInstanceProcAddr` just export `<layerName>`'s vkGetInstanceProcAddr is better, in that

 - we do not need to define the behavior of `<layerName>GetInstanceProcAddr` ourselves
 - it is explicitly stated so that layer developers do not need to figure that out
 - it will be slightly less work for layer libraries that contain multiple layers

olvaffe:
I think in `Layer Library API Version 0`, we can say that each layer's GIPA and GDPA should be exported as `<layerName>GIPA` and `<layerName>GDPA` respectively (or simply as GIPA and GDPA when there is only one layer).

Then in `Layer Conventions and Rules`, we can say

     `vkGetInstanceProcAddr` can intercept a command by returning a function
    pointer different from what would be returned through chaining.  Detailed specification of this function,

    - when `pName` is...

loader/LoaderAndLayerInterface.md
jon-lunarg: This sounds good to me!

jon-lunarg:
Okay I think we are in agreement mostly  except for one last point, regarding this in the Layers Conventions and Rules section:
`vkGetInstanceProcAddr` can intercept a command by returning a function
pointer different from what would be returned through chaining.  Detailed specification of this function,

- when `pName` is...


I still don't like the wording "can intercept a command by returning a function
pointer different from what would be returned through chaining."  since it defines an allowed behavior based on the behavior of an external component (the next layer of entity in the chain).   To strictly abide by this requirement as a layer author, I would need to go investigate every possible layer and ICD that might be the next entity in the chain and make sure I don't return a value that would be the same as what was returned by the next entity.  How about this wording:
GIPA either returns NULL, a local entry point or the value obtained
    by calling down the instance chain (the next layer/ICD's GetInstanceProcAddr).

olvaffe:
That is fine, but we would fail to specify how to intercept a command.  How about "GIPA can intercept a command by returning a local entry point rather than calling down the instance chain"?

jon-lunarg:
I moved the GIPA and GDPA descriptions into the Layer Conventions section.  The Layer Library section refers to the Layer Conventions section.  Look over the language for GIPA and GDPA but  I think this should now be acceptable or very close?

Note:  for the final push to master we can squash the commits.  I added a new commit for this change as it is easier to see the differences and Github doesn't get confused about finding old comments when they are associated with a commit that has been removed from the history due to a force push.

loader/LoaderAndLayerInterface.md
```diff
@@ -1136,17 +1020,35 @@ is itself.  In other cases, it should normally chain to other layers.
...
+     `vkEnumerateInstanceExtensionProperties`, or
+     `vkEnumerateDeviceLayerProperties` (but _not_
+     `vkEnumerateDeviceExtensionProperties`), it returns a function pointer to
+     the corresponding introspection function defined by this interface.
+   - when `pName` is `vkGetInstanceProcAddr`, it returns a function pointer
+     to itself.
+   - when `pName` is `vkCreateInstance`, it returns a function pointer to the
+     layer's vkCreateInstance, which is required for instance chaining.
+   - when `pName` is `vkCreateDevice`, it returns a function pointer to the
+     layer's vkCreateDevice, which is only required for any device chaining.
```
olvaffe: The first bullet is not needed anymore.  I wonder if we should instead say that

    These commands must be intercepted
     - ...

loader/LoaderAndLayerInterface.md
```diff
@@ -1136,17 +1020,35 @@ is itself.  In other cases, it should normally chain to other layers.
...
+   For compatibility with older layer libraries,
+   - when `pName` is `vkCreateDevice`, it ignores `instance`.
+
+`vkGetDeviceProcAddr` intercepts a Vulkan command by returning a local entry point,
+otherwise it returns the value obtained by calling down the device chain.
 
-`vkGetDeviceProcAddr` can intercept a command by returning a function pointer
-different from what would be returned through chaining.
+Vulkan commands that are defined by an extension that is disabled, must not be intercepted in either
+`vkGetInstanceProcAddr` or `vkGetDeviceProcAddr`.
```
olvaffe: This is weaker than the spec.  The spec actually says NULL should be returned.  We should rephrase what the spec says or remove this.

loader/LoaderAndLayerInterface.md
jon-lunarg: Okay, something like this?

These commands must be intercepted
vkGetInstanceProcAddr
vkCreateInstance
vkCreateDevice (only required for any device-level chaining)

loader/LoaderAndLayerInterface.md
jon-lunarg: Except I am pretty sure ALL our layers in Khronos and LunarG Github do exactly what I wrote. See debug_report_get_instance_proc_addr() for an example.  In our current implementation, the entity at the bottom of the chain ultimately must return NULL  for a disabled extension.  This NULL then gets propagated up the chain thru all the layers.  I would be in favor modifying the layers to be closer to the spec:  If a layer implements an extension  and the extension is disabled, it shouldn't call down the chain for GIPA/GDPA query but should return NULL.  Now a layer that doesn't implement an extension but just intercepts the extension entry points  (eg swapchain layer) could do one of two behaviors:  1) call down chain  (this is the current behavior), 2)  return NULL.  For consistency and simplicity, I like option 2).

If we do go with changing the layer implementation then I would still be in favor of having some sort of statement in here since the layer has three options for handling disabled extensions it intercepts ( 1) rely on some other entity to return NULL always, 2) return NULL if it implements the extension, 3) return NULL always if the extension is disabled)

olvaffe: I think we can consider it layer bug if a layer returns the local entrypoint for a disabled command instead of `NULL`.  The bug is minor enough that we do not need to bend the doc for it.  As for wordings, how about

    The specification requires `NULL` to be returned from `vkGetInstanceProcAddr` and
    `vkGetDeviceProcAddr` for disabled commands.  A layer may return `NULL` itself or
    rely on the following layers to do so.

jon-lunarg: Okay I used this exact sentence.

jon-lunarg:
Olv,
I think this was the last issue; fixed it and pushed a commit to master. Please let me know if something looks wrong.

Thanks!

PR #580, opened 2016-05-23
layers: Handle static arrays in codegen

### LINKED ISSUE #415: debug report tries to read from possibly stack allocated memory
in cube.c, VkPipelineColorBlendAttachmentState att_state[1] is allocated on the stack.
The debug report will try to read from it later, from vkCmdDraw, reporting bogus values.

tobine:
Fixes #415 
Update safe_struct codegen to handle the case of a static array
in a struct. This removes exceptions for VkDebugMarkerMarkerInfoEXT
and VkPipelineColorBlendStateCreateInfo so they can be treated
as all other structs.


jon-lunarg:
fix pushed to master

PR #581, opened 2016-05-24
SC: match_sorted

chrisforbes:
Most of this has already been reviewed; this is just a rebase and cleanup of a branch from a few weeks back.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -1423,6 +1423,48 @@ static void collect_interface_by_descriptor_slot(debug_report_data *report_data,
     }
 }
 
+/* Walks two sorted sequences (typically maps) together. Pairs of entries are
+ * compared via the `compare` function, which should return negative if a < b,
+ * positive if a > b, zero otherwise.
+ * The function `unmatched_a` will be called for each entry in the first sequence
+ * which does not have a corresponding entry in the second.
+ * The function `unmatched_b` will be called for each entry in the second
+ * sequence which does not have a corresponding entry in the first.
+ * The function `matched` will be called for each pair of matched entries.
+ */
+template<typename AIterator, typename BIterator>
+void match_sorted(AIterator a_begin, AIterator a_end,
+                  BIterator b_begin, BIterator b_end,
+                  std::function<int (AIterator, BIterator)> compare,   // b - a
+                  std::function<void (AIterator)> unmatched_a,
+                  std::function<void (BIterator)> unmatched_b,
+                  std::function<void (AIterator, BIterator)> matched)
+{
+    auto it_a = a_begin;
+    auto it_b = b_begin;
+
+    while (it_a != a_end && it_b != b_end) {
+        auto cmp = compare(it_a, it_b);
+        if (cmp > 0) {
+            unmatched_a(it_a++);
+        }
+        else if (cmp < 0) {
+            unmatched_b(it_b++);
+        }
+        else {
+            matched(it_a++, it_b++);
+        }
+    }
+
+    while (it_a != a_end) {
+        unmatched_a(it_a++);
+    }
+
+    while (it_b != b_end) {
+        unmatched_b(it_b++);
+    }
+}
+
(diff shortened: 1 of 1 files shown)


layers/core_validation.cpp
```diff
@@ -1627,56 +1672,59 @@ static bool validate_vi_against_vs_inputs(debug_report_data *report_data, VkPipe
...
+                 compare, unmatched_attrib, unmatched_input, matched);
 
     return pass;
 }
 
 static bool validate_fs_outputs_against_render_pass(debug_report_data *report_data, shader_module const *fs,
                                                     spirv_inst_iter entrypoint, RENDER_PASS_NODE const *rp, uint32_t subpass) {
-    std::map<location_t, interface_var> outputs;
-    std::map<uint32_t, VkFormat> color_attachments;
+    const std::vector<VkFormat> &color_formats = rp->subpassColorFormats[subpass];
```
tobine: Doesn't build on android due to unused variable

layers/core_validation.cpp
```diff
@@ -1423,75 +1423,117 @@ static void collect_interface_by_descriptor_slot(debug_report_data *report_data,
...
+ * The function `unmatched_a` will be called for each entry in the first sequence
+ * which does not have a corresponding entry in the second.
+ * The function `unmatched_b` will be called for each entry in the second
+ * sequence which does not have a corresponding entry in the first.
+ * The function `matched` will be called for each pair of matched entries.
+ */
+template<typename AIterator, typename BIterator>
+void match_sorted(AIterator a_begin, AIterator a_end,
+                  BIterator b_begin, BIterator b_end,
+                  std::function<int (AIterator, BIterator)> compare,   // b - a
```
tobine: This "// b - a" comment confuses me. Function comment says "a < b" is negative so should this comment be "// a - b"?
What does this comment mean? Is the function comment above sufficient so that this comment can be killed?

tobine:
Still chewing on this. I know I said I liked your initial prototype. It's nice to unify these checks in match_sorted function. On down side it does grow the code and is a bit more complex than original code to achieve same result. Do you anticipate further uses for match_sorted?

@chrisforbes assigning back to you to fix android build error to keep you in the habit of verifying that.

@lentinem your thoughts on this?

layers/core_validation.cpp
```diff
@@ -1423,75 +1423,117 @@ static void collect_interface_by_descriptor_slot(debug_report_data *report_data,
...
-
-    /* maps sorted by key (location); walk them together to find mismatches */
-    while ((outputs.size() > 0 && a_it != outputs.end()) || (inputs.size() && b_it != inputs.end())) {
-        bool a_at_end = outputs.size() == 0 || a_it == outputs.end();
-        bool b_at_end = inputs.size() == 0 || b_it == inputs.end();
-        auto a_first = a_at_end ? std::make_pair(0u, 0u) : a_it->first;
-        auto b_first = b_at_end ? std::make_pair(0u, 0u) : b_it->first;
-
-        if (b_at_end || ((!a_at_end) && (a_first < b_first))) {
+    std::function<int(var_map::iterator, var_map::iterator)> compare =
```
lentinem: So if a_loc < b_loc we increment b?

lentinem:
Overall I think this style is more readable than the original as it compartmentalizes the logic into functions instead of giant if blocks.

layers/core_validation.cpp
lentinem: I would expect b - a to be correct. Basically if b is smaller increment it, if b is bigger increment a. There is one statement that seems to be doing a - b but I don't think the comparator is right - I could be wrong though and just don't completely understand what is going on.

layers/core_validation.cpp
chrisforbes: That doesn't sound right. Let me double-check i didn't botch this while updating it.

chrisforbes:
This might actually not be quite the right direction. It's worked so far, but #604 breaks things.

PR #582, opened 2016-05-24
Added loader support for KHR_swapchain and KHR_display_swapchain.

pbantolas:
The loader now discovers if the swapchain or the display_swapchain
extensions are enabled on vkCreateInstance.

Created missing terminator functions for the swapchain extension.

Created missing loader support for the display_swapchain extension by
handling the vkCreateSharedSwapchainsKHR entry point.


pbantolas:
@jon-lunarg Is this what you had in mind?

jon-lunarg:
Petros:

Thanks for the PR.  Actually the KHR_swapchain and KHR_display_swapchain are device extensions rather than instance extensions so they are treated differently. There is some subtlties that would be probably as hard to describe as just doing the code.  If you want to forge ahead that is fine, I can try to describe some of the issues. But maybe it is better if someone that is more familiar with the loader code just adds support for KHR_display_swapchain extension.

pbantolas:
Jon, agreed, it will be better if someone with more experience did the device extensions. We are mainly interested in support for KHR_display functionality.

jon-lunarg:
okay thanks, closing this

PR #584, opened 2016-05-24
layers: Fix ImageSampler immut sampler check

tobine:
Was missing IsImmutableSampler() override function for the
ImageSamplerDescriptor class.


Tony-LunarG:
This fixes the issue I was seeing with the sample.

chrisforbes:
@tobine had a bad run with this function... :+1:

PR #585, opened 2016-05-25
layers: simplify v0 function implementations

olvaffe:
Since each of the layer library contains only a layer, we can simplify v0
functions by making them wrappers to the contained layer's corresponding
commands.

While we have to make changes to each layer individually, all the changes
should be the same in spirit.  The exception is vk-layer-generate.py where
the missing vkEnumerateDeviceExtensionProperties is also added.


olvaffe:
I am assigning this to myself and will commit in a day or two, unless @tobine wants to take a look.

olvaffe:
I just pushed this.  Sorry that I did not get notifications that you assigned this to you.  I can do follow-ups or revert if there is an issue with the change.

jon-lunarg:
That is alright to push and revisit as needed. I haven't been able to be very timely in the last several weeks do to various other responsibilities, meetings, offsites, etc.

PR #586, opened 2016-05-25
Modify warning msgs.

lentinem:


tobine:
LGTM :+1:
Assigning to @mark-lunarg in case he @jon-lunarg or @karl-lunarg  have any input prior to commit.
Guys, note that this adds some postfix text to warnings in an effort to reduce user support requests of the nature "I'm getting this warning but my code is ok." We are aware that such cases exist, but want to keep the warnings to make certain that users know what they're doing.

mark-lunarg:
Hey Michael, we'd prefer a one-time header or legend that defines what error, info, warn and perf-warn mean and think this would be preferable to outputting a disclaimer with each message.  We will put something together this week and you can let us know if it will cover the warning-related complaints that have been coming in.

mark-lunarg:
See #633.

PR #587, opened 2016-05-25
DescriptorSet image & buffer usage bit validation

tobine:
Verify that correct usage bits are set for both image and buffer descriptor updates.
Add test that triggers all the possible usage bit errors.


tobine:

PR #588, opened 2016-05-25
tests: Add test for fence submit/wait over multiple frames

tobine:
There was previously a bug in how old fence dependencies were cleaned
up so this test is being added to prevent any regressions in that
behavior.

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -2971,6 +2971,93 @@ TEST_F(VkLayerTest, ResetEventThenSet) {
 }
 
 // This is a positive test.  No errors should be generated.
+TEST_F(VkLayerTest, TwoFencesThreeFrames) {
+    TEST_DESCRIPTION("Two command buffers with two separate fences are each "
+                     "run through a Submit & WaitForFences cycle 3 times. This "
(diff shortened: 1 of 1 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -2971,6 +2971,93 @@ TEST_F(VkLayerTest, ResetEventThenSet) {
...
+        err = vkCreateFence(m_device->device(), &fence_ci, nullptr, &fences[i]);
+        ASSERT_VK_SUCCESS(err);
+    }
+
+    for (uint32_t frame = 0; frame < NUM_FRAMES; ++frame) {
+        // Create empty cmd buffer
+        VkCommandBufferBeginInfo cmdBufBeginDesc = {};
+        cmdBufBeginDesc.sType = VK_STRUCTURE_TYPE_COMMAND_BUFFER_BEGIN_INFO;
+
+        err = vkBeginCommandBuffer(cmd_buffers[0], &cmdBufBeginDesc);
```
chrisforbes: This could be a loop over NUM_OBJECTS and half as long?

chrisforbes:
:+1: one inline comment.

tobine:
Added loop to reduce code and pushed to master

PR #589, opened 2016-05-26
Tobin android val layer test fixes

tobine:
These changes get 4 layer validation tests that were failing on Nexus Player to pass.


tests/layer_validation_tests.cpp
```diff
@@ -9896,42 +9896,24 @@ TEST_F(VkLayerTest, InvalidImageView) {
...
-    image_create_info.extent.depth = 1;
-    image_create_info.mipLevels = 1;
-    image_create_info.samples = VK_SAMPLE_COUNT_1_BIT;
-    image_create_info.tiling = VK_IMAGE_TILING_LINEAR;
-    image_create_info.usage = VK_IMAGE_USAGE_SAMPLED_BIT;
-    image_create_info.flags = 0;
-
-    err = vkCreateImage(m_device->device(), &image_create_info, NULL, &image);
-    ASSERT_VK_SUCCESS(err);
+    VkImageObj image(m_device);
```
chrisforbes: :+1: as a practical workaround, but I don't see a requirement for images to have memory bound before creating views -- so this is really a driver bug?

tobine:

tests/layer_validation_tests.cpp
tobine: Hung onto this in case it was driver bug but is actually spec requirement. Spec section 11.6:

`Non-sparse resources must be bound completely and contiguously to a single VkDeviceMemory object before the resource is passed as a parameter to any of the following operations:

creating image or buffer views
updating descriptor sets
recording commands in a command buffer`

PR #590, opened 2016-05-26
Tobin alloc descriptor interface refactor

tobine:
Migrates DescriptorSet class (and cvdescriptorset interface) to have interface that closely matches AllocateDescriptorSets() so all validation and state updates can be performed through a single call.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5865,10 +5865,9 @@ ResetDescriptorPool(VkDevice device, VkDescriptorPool descriptorPool, VkDescript
     }
     return result;
 }
-
 // Ensure the pool contains enough descriptors and descriptor sets to satisfy
-// an allocation request. Fills requiredDescriptorsByType with the total number
-// of descriptors of each type required, for later update.
+// an allocation request. Fills common_data with the total number of descriptors of each type required,
+// as well as DescriptorSetLayout ptrs used for later update.
 static bool PreCallValidateAllocateDescriptorSets(layer_data *dev_data, const VkDescriptorSetAllocateInfo *pAllocateInfo,
                                                   cvdescriptorset::AllocateDescriptorSetsData *common_data) {
     // All state checks for AllocateDescriptorSets is done in single function
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5869,110 +5869,36 @@ ResetDescriptorPool(VkDevice device, VkDescriptorPool descriptorPool, VkDescript
 // Ensure the pool contains enough descriptors and descriptor sets to satisfy
 // an allocation request. Fills requiredDescriptorsByType with the total number
 // of descriptors of each type required, for later update.
-static bool PreCallValidateAllocateDescriptorSets(const layer_data *dev_data, const DESCRIPTOR_POOL_NODE *pPoolNode, uint32_t count,
-                                                  std::vector<cvdescriptorset::DescriptorSetLayout const *> const &layout_nodes,
-                                                  uint32_t requiredDescriptorsByType[]) {
-    bool skipCall = false;
-
-    // Track number of descriptorSets allowable in this pool
-    if (pPoolNode->availableSets < count) {
-        skipCall |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DESCRIPTOR_POOL_EXT,
-                            reinterpret_cast<const uint64_t &>(pPoolNode->pool), __LINE__, DRAWSTATE_DESCRIPTOR_POOL_EMPTY, "DS",
-                            "Unable to allocate %u descriptorSets from pool 0x%" PRIxLEAST64
-                            ". This pool only has %d descriptorSets remaining.",
-                            count, reinterpret_cast<const uint64_t &>(pPoolNode->pool), pPoolNode->availableSets);
-    }
-
-    // Count total descriptors required per type
-    for (auto layout_node : layout_nodes) {
-        if (layout_node) {
-            for (uint32_t j = 0; j < layout_node->GetBindingCount(); ++j) {
-                const auto &binding_layout = layout_node->GetDescriptorSetLayoutBindingPtrFromIndex(j);
-                uint32_t typeIndex = static_cast<uint32_t>(binding_layout->descriptorType);
-                requiredDescriptorsByType[typeIndex] += binding_layout->descriptorCount;
-            }
-        }
-    }
-
-    // Determine whether descriptor counts are satisfiable
-    for (uint32_t i = 0; i < VK_DESCRIPTOR_TYPE_RANGE_SIZE; i++) {
-        if (requiredDescriptorsByType[i] > pPoolNode->availableDescriptorTypeCount[i]) {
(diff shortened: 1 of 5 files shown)


layers/descriptor_sets.cpp
```diff
@@ -1131,4 +1131,97 @@ bool cvdescriptorset::DescriptorSet::VerifyCopyUpdateContents(const VkCopyDescri
...
+                            reinterpret_cast<uint64_t &>(pool_it->second->pool), pool_it->second->availableDescriptorTypeCount[i]);
+            }
+        }
+    }
+    return skip_call;
+}
+// Decrement allocated sets from the pool and insert new sets into set_map
+void cvdescriptorset::PerformAllocateDescriptorSets(
+    const VkDescriptorSetAllocateInfo *p_alloc_info, const VkDescriptorSet *descriptor_sets,
+    std::unordered_map<VkDescriptorPool, DESCRIPTOR_POOL_NODE *> *pool_map,
```
chrisforbes: Why these two by ptr rather than ref-to-nonconst? It leads to nasty (*pool_map)[...]

layers/descriptor_sets.cpp
```diff
@@ -1131,4 +1131,97 @@ bool cvdescriptorset::DescriptorSet::VerifyCopyUpdateContents(const VkCopyDescri
...
+    const std::unordered_map<VkDescriptorSetLayout, cvdescriptorset::DescriptorSetLayout *> &layout_map,
+    const std::unordered_map<VkBuffer, BUFFER_NODE> &buffer_map,
+    const std::unordered_map<VkDeviceMemory, DEVICE_MEM_INFO> &mem_obj_map,
+    const std::unordered_map<VkBufferView, VkBufferViewCreateInfo> &buffer_view_map,
+    const std::unordered_map<VkSampler, std::unique_ptr<SAMPLER_NODE>> &sampler_map,
+    const std::unordered_map<VkImageView, VkImageViewCreateInfo> &image_view_map,
+    const std::unordered_map<VkImage, IMAGE_NODE> &image_map,
+    const std::unordered_map<VkImage, VkSwapchainKHR> &image_to_swapchain_map,
+    const std::unordered_map<VkSwapchainKHR, SWAPCHAIN_NODE *> &swapchain_map) {
+    auto pool_state = (*pool_map)[p_alloc_info->descriptorPool];
```
chrisforbes: This was previously tolerant to the app choosing not to skip the call. Now you'll insert junk into this map.

layers/descriptor_sets.cpp
```diff
@@ -1131,4 +1131,97 @@ bool cvdescriptorset::DescriptorSet::VerifyCopyUpdateContents(const VkCopyDescri
...
+    /* Account for sets allocated from pool */
+    pool_state->availableSets -= p_alloc_info->descriptorSetCount;
+    /* Create tracking object for each descriptor set; insert into
+     * global map and the pool's set.
+     */
+    for (uint32_t i = 0; i < p_alloc_info->descriptorSetCount; i++) {
+        auto layout_state = layout_map.find(p_alloc_info->pSetLayouts[i])->second;
```
chrisforbes: We're doing this twice again :(

layers/descriptor_sets.cpp
```diff
@@ -1131,4 +1131,97 @@ bool cvdescriptorset::DescriptorSet::VerifyCopyUpdateContents(const VkCopyDescri
...
+        // Account for individual descriptors allocated from pool
+        for (uint32_t j = 0; j < layout_state->GetBindingCount(); ++j) {
+            const auto &binding_layout = layout_state->GetDescriptorSetLayoutBindingPtrFromIndex(j);
```
chrisforbes: Just `auto` here, this returns a ptr. -- this ends up being `VkDescriptorSetBinding const * const &`.

layers/descriptor_sets.cpp
```diff
@@ -1131,4 +1131,97 @@ bool cvdescriptorset::DescriptorSet::VerifyCopyUpdateContents(const VkCopyDescri
...
+// Verify that the state at allocate time is correct, but don't actually allocate the sets yet
+bool cvdescriptorset::ValidateAllocateDescriptorSets(
+    const debug_report_data *report_data, const VkDescriptorSetAllocateInfo *p_alloc_info,
+    const std::unordered_map<VkDescriptorSetLayout, cvdescriptorset::DescriptorSetLayout *> &set_layout_map,
+    const std::unordered_map<VkDescriptorPool, DESCRIPTOR_POOL_NODE *> &pool_map) {
+    bool skip_call = false;
+    uint32_t requiredDescriptorsByType[VK_DESCRIPTOR_TYPE_RANGE_SIZE]{};
+
+    for (uint32_t i = 0; i < p_alloc_info->descriptorSetCount; i++) {
+        auto layout_it = set_layout_map.find(p_alloc_info->pSetLayouts[i]);
```
chrisforbes: Perhaps we should make `getFoo(layer_data *)` (with `layer_data` being left an incomplete type) visible to this code. We've ended up with map iterator noise everywhere again.

layers/descriptor_sets.cpp
chrisforbes: I think the model is deficient. We should have a way of passing data from validate to update, so we're not forced to do things twice.

layers/core_validation.cpp
```diff
@@ -5883,110 +5883,36 @@ ResetDescriptorPool(VkDevice device, VkDescriptorPool descriptorPool, VkDescript
 // Ensure the pool contains enough descriptors and descriptor sets to satisfy
 // an allocation request. Fills requiredDescriptorsByType with the total number
```
chrisforbes: This comment is no longer true.

layers/descriptor_sets.cpp
tobine: Sole reason is Google C++ style guide on [reference arguments](https://google.github.io/styleguide/cppguide.html#Reference_Arguments)

layers/descriptor_sets.cpp
tobine: This is a tricky line to walk. With migration to "PreValidation" & "PostRecord" functions, the assumption with calling the "Perform" function is that the "Validate" function has passed. Otherwise all bets are off and you may, indeed, get junk in the map.

layers/descriptor_sets.cpp
tobine: I don't like looping twice and agree this is a deficiency. With the current proposed architecture (PreValidate, PostRecord), I don't see a way around this, but I'm open to altering the proposed architecture. We should come up with an alternate proposal that cleanly allows passing "Pre" work into the "Post" update.

layers/descriptor_sets.cpp
tobine: Yes, I agree that hoisting getFoo() helpers to use here is the right thing to do. I'll go that route.

layers/descriptor_sets.cpp
chrisforbes: OK, fair enough.

tobine:
Tweaked this PR to include some shared data between PreCallValidateAllocateDescriptorSets() and PostCallRecordAllocateDescriptorSets(). This eliminates repeating the same work (map look-ups and per-type descriptor count calculations) in each function.

tobine:

PR #595, opened 2016-05-30
layers: LX506

MarkMLunarG:
Correct invalid test of size variable to correctly test size passed by application

Change-Id: I6d855482985512b5096696f5c37346dde8a391ac


mark-lunarg:
Hey Mark, don't forget to update the commit message.

lentinem:
+1

tobine:
LGTM :+1:
Assigning back to @MarkMLunarG to update commit message

MarkMLunarG:
I believe this now properly checks the conditions that lentinem is referring to. It's not clear to me from the spec what offset means when the size is WHOLE_SIZE in calls to vkMapMemory so this assumes that offset is 0 in MapMemory cases where size was WHOLE_SIZE. Limited testing does produce an MapMemory error if offset is not 0.

lentinem:
+1. Would it be possible to add your limited tests to the test suite?

MarkMLunarG:
All that I did was change the cube.c code to do vkMapMemory with WHOLE_SIZE and an offset of 0x10 and the vkMapMemory call failed. It doesn't sound like a useful test and it isn't related to vkFlushMappedMemoryRanges. Given that it was related to ambiguity in the spec for an odd case it seemed like a slippery slope to go any further with

lentinem:
These are still cases we want to test the validation layer against. I would recommend checking this in and opening up a new test bug for this.

MarkMLunarG:
Resolved with:
SHA-1: 4b70ef75f98d2a6196132d8c48d43d1d5e04992e

* layers: LX506 fix vkFlushMappedMemoryRanges validation error

Fix LX506: vkFlushMappedMemoryRanges validation error
 with VkMappedMemoryRange::size = VK_WHOLE_SIZE.
Resolution: Change test of size variable to correctly test the
 size passed by application against the relevant mapped memory

Change-Id: I6d855482985512b5096696f5c37346dde8a391ac

PR #596, opened 2016-05-30
Expand noautovalidity processing for structs

dustin-lunarg:
Apply 'noautovalidity' code generation logic to struct members.
Previously, struct pointers tagged as noautovalidity were ignored by
the code generator, but the members of non-NULL struct pointers
were still processed.  Now both the struct pointer and the struct
members are ignored when the struct pointer has the 'noautovalidity'
attribute.


mark-lunarg:
+1

PR #597, opened 2016-05-31
layers: take better advantage of emplace_back in descriptor sets

chrisforbes:
Addresses a "you could reduce the noise a bit if..." comment from the original series that added the descriptor set classes.


lentinem:
+1

tobine:

PR #599, opened 2016-05-31
Validate compute descriptor use

chrisforbes:
This series finishes wiring up the tracking of descriptor use by compute pipelines, and throws away the various hacks that were in use previously.


tobine:

Issue #601, opened 2016-05-31
Missing validation check: vkCmdBeginRenderPass clearValueCount check


danginsburg:
It does not appear to me that validation is checking the following valid usage of clearValueCount for VkRenderPassBeginInfo:

"clearValueCount must be greater than or equal to the number of attachments in renderPass that specify a loadOp of VK_ATTACHMENT_LOAD_OP_CLEAR" (Vulkan spec 1.0.15 section 7.4)

One of the IHVs has reported that we are violating this requirement.  I'm still working to track it down, but wanted to report this as a missing validation check because this is something validation should be able to catch.

tobine:
Submitted #603 to fill this hole.

danginsburg:
I reviewed the change and it looks good to me.  Thanks.

tobine:
Fix was cherry-picked and pushed to master

mre4ce:
I have three attachments (VkFramebufferCreateInfo::attachmentCount = 3):

1. MSAA color buffer with VK_ATTACHMENT_LOAD_OP_CLEAR
2. resolve color buffer with VK_ATTACHMENT_LOAD_OP_DONT_CARE
3. MSAA depth buffer with VK_ATTACHMENT_LOAD_OP_CLEAR

If I set the VkRenderPassBeginInfo::clearValueCount to 2 I do not get a validation error.

tobine:
From spec:

`clearValueCount must be greater than or equal to the number of attachments in renderPass that specify a loadOp of VK_ATTACHMENT_LOAD_OP_CLEAR`

What do you expect the error to be?

mre4ce:
With a VkRenderPassBeginInfo::clearValueCount of 2 things don't render correctly on all drivers/hardware I've tried.

jeffbolznv:
The spec says:

  * pname:clearValueCount is the number of elements in pname:pClearValues.
  * pname:pClearValues is an array of slink:VkClearValue structures that
    contains clear values for each attachment, if the attachment uses a
    pname:loadOp value of ename:VK_ATTACHMENT_LOAD_OP_CLEAR. **The array is
    indexed by attachment number**. Only elements corresponding to cleared
    attachments are used. Other elements of pname:pClearValues are ignored.

The validity rule looks wrong, it should care about the attachment indices. I think it should be more like:

[...]

PR #603, opened 2016-05-31
Tobin gh601 clear attach count

tobine:
Addresses missing check as requested in #601


layers/core_validation.cpp
```diff
@@ -8926,13 +8927,19 @@ CmdBeginRenderPass(VkCommandBuffer commandBuffer, const VkRenderPassBeginInfo *p
...
+                                                              "must therefore be greater than or equal to %u.",
+                    pRenderPassBegin->clearValueCount, reinterpret_cast<uint64_t &>(renderPass), clear_op_count, clear_op_count);
+            }
             skipCall |= VerifyRenderAreaBounds(dev_data, pRenderPassBegin);
             skipCall |= VerifyFramebufferAndRenderPassLayouts(dev_data, pCB, pRenderPassBegin);
             skipCall |= insideRenderPass(dev_data, pCB, "vkCmdBeginRenderPass");
-            if (renderPass) {
-                skipCall |= ValidateDependencies(dev_data, framebuffer, renderPass);
-            }
+            skipCall |= ValidateDependencies(dev_data, framebuffer, renderPass);
```
lentinem: If there is no render pass why are we validating dependencies.

tobine: See line 8899 above, we already checked to make sure we had renderPass so this check was redundant.

lentinem:
+1

tobine:

PR #606, opened 2016-05-31
cmake: Fix MSVCRT override

dustin-lunarg:
Fix the CMake code that overrides the /MD flag with the /MT flag for
the Windows demo and loader projects.  The CMake debug and release
variables were set to /MTd and /MT, overriding all existing debug
and release settings.

Now replacing /MD with /MT in the debug and release variables,
without losing other debug/release settings.  Also performing
the MD/MT substitution for all release and debug configurations.

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -28,8 +28,13 @@ if(WIN32)
         set (BUILDTGT_DIR build32)
     endif()
 
-    set(CMAKE_C_FLAGS_DEBUG "-MTd")
-    set(CMAKE_C_FLAGS_RELEASE "-MT")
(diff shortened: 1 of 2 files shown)


jon-lunarg:
Does this change or David's change cause this warning or was it already there before the change(s)?
Warning	1	warning LNK4098: defaultlib 'LIBCMTD' conflicts with use of other libs; use /NODEFAULTLIB:library	C:\Users\jon\vk\vk\dbuild\loader\LINK	vulkan-1

Other than this potentially new warning, the loader CMake change looks fine to me as far as the specific concern I had with the original patch.

dustin-lunarg:
That warning predates both changes.

PR #612, opened 2016-06-02
GitHub 611: Fix dynamic offset indexing

### LINKED ISSUE #611: Dynamic offsets incorrectly bound with dynamic descriptors
In `descriptor_sets.cpp`, `ValidateDrawState()` function has `dyn_offset_index` set to 0 for each binding, while it should not.

iostrowsINTEL:
Fix #611


tobine:
Good catch. Fix pushed to master, thanks!

PR #613, opened 2016-06-02
Add helper functions and use them in DescriptorSet class

tobine:
Kill almost all of the maps in DescriptorSet class and use utility functions to look-up data blocks from generic layer_data ptr.
Also migrated a few of the maps to use unique_ptrs which kills some delete calls and unifies the helper function interfaces.


chrisforbes:
:+1:

tobine:

PR #620, opened 2016-06-03
layers: Remove erroneous check re: setting an event in a cmdbuffer

Tony-LunarG:
Remove the layer test for it as well

This was causing the WaitEventThenSet layer validation test to TDR on my Windows Nvidia system, since it waited for the event in the command buffer, but the call to vkSetEvent got a validation error and never made it to the driver to signal the event, so the driver timed out.

As an aside, this test is using queue 1, and only queue 1, making it necessary that a device have 2 queues to run it.  If that wasn't intentional, I would propose using queue 0 and removing the need for 2 queues to let it run on far more hardware.


tobine:
If the WaitEventThenSet test is hitting this case, the error monitor system won't return "true" from the callback unless this specific error was being waited on so this error should just be logged, skip_call should remain "false" and operation proceed as normal. Is there some aliasing between this check and the check that WaitEventThenSet is looking for that causes the callback for this check to fire and skip_call to be set to true? If so, then the output error messages and/or the message being check for by the error monitor.

:-1: for removing this check and test.

Tony-LunarG:
WaitEventThenSet claims to be a positive test that gets no errors, but it does due to the check in question

tobine:
It sounds like WaitEventThenSet should be evaluated and potentially changed rather than removing the check.

Tony-LunarG:
Nevermind - you're correct - I misunderstood the check and this isn't the answer to the problem - different fix coming

PR #622, opened 2016-06-05
Mac OS X support to compile layers

mre4ce:
This obviously does not make Vulkan work on Mac OS X but it allows compiling layers for Mac OS X using the common headers.


mre4ce:
Removed the dependency on vk_loader_platform.h instead.

PR #625, opened 2016-06-06
Added --suppress_popup flag

lenny-lunarg:
This flag is needed so that if a test fails on windows, the programs can fail instead of hanging.

### DIFF
--- a/demos/cube.c
+++ b/demos/cube.c
@@ -61,7 +61,8 @@
 #ifdef _WIN32
 #define ERR_EXIT(err_msg, err_class)                                           \
     do {                                                                       \
-        MessageBox(NULL, err_msg, err_class, MB_OK);                           \
+        if (!demo->suppress_popups)                                            \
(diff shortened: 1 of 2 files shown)


davidlunarg:
Looks good to me, but I would ask that the commit description be changed to conform to the norm of beginning with the subsystem name, and add a little more detail to the description. I'll suggest:

      demos: Added --suppress_popup flag to cube

lenny-lunarg:
Pushed to master

commit 20a497b97273bbbc0f853f12c0cf6db57a2aef9d
Author: lenny-lunarg <lenny@lunarg.com>
Date:   Mon Jun 6 11:07:53 2016 -0600

    demos: Added --suppress_popups flag

    Added a command line option to cube and tri to suppress
    windows message boxes to facilitate Windows testing.

    Change-Id: I069f9b199c9740e1e21c42100ceec42d5fb8f0e2

PR #629, opened 2016-06-08
Remove warning for implicit sub pass dependencies.

### LINKED ISSUE #628: Remove warning for implicit subpass dependencies
The validation layer checks whether there's a dependency between sub passes using the same attachments. This is good, but it also issues a warning if there are "only" implicit dependencies between two sub passes (rather than a direct dependency). I don't think this should be a warning since the notion of a subpass dependency chain is well defined in the spec:

> A subpass dependency chain is a sequence of subpass dependencies in a render pass, where the source subpass of each subpass dependency (after the first) equals the destination subpass of the previous dependency."

and:
> Render passes must include subpass dependencies (either directly or via a subpass dependency chain) between any two subpasses that operate on the same attachment or aliasing attachments

janharaldfredriksen-arm:
Fix #628 

Change-Id: I1039955d6d445e1ee234116c6da2e4dbaf9c7dd9

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -8199,15 +8199,10 @@ static bool CheckDependencyExists(const layer_data *my_data, const int subpass, 
         auto prev_elem = std::find(node.prev.begin(), node.prev.end(), dependent_subpasses[k]);
         auto next_elem = std::find(node.next.begin(), node.next.end(), dependent_subpasses[k]);
         if (prev_elem == node.prev.end() && next_elem == node.next.end()) {
-            // If no dependency exits an implicit dependency still might. If so, warn and if not throw an error.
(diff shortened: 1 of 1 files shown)


mark-lunarg:
Cherry-picked and pushed to Master.

commit 7655cb8b5eb52badee0b011729a05afa36316d69
Author: Jan-Harald Fredriksen <jan-harald.fredriksen@arm.com>
Date:   Wed Jun 8 14:20:50 2016 +0200

    layers: PR629, Remove subpass warnings

    The validation layer checks for dependency between subpasses using the
    same attachments, but it also issues warnings if there are "only"
    implicit dependencies between two subpasses. This is explicitly
    allowed by the spec and should not cause validation warnings.

PR #631, opened 2016-06-08
Remove error of incorrect use of vkEnumeratePhysicalDevices

### LINKED ISSUE #630: Remove error of incorrect use of vkEnumeratePhysicalDevices
The validation layer throws an error if vkEnumeratePhysicalDevices is called w/non-NULL pPhysicalDevices if pPhysicalDeviceCount is not been previously retrieved by a previous call to the same function.

I don't think there's any justification for this in the spec. It may be good practice, but the spec allows you use vkEnumeratePhysicalDevices to retrieve a subset of the devices (and that VK_INCOMPLETE is returned if you didn't retrieve them all) so I don't think the validation layer should complain about this.
(See https://www.khronos.org/registry/vulkan/specs/1.0/xhtml/vkspec.html#devsandqueues-physical-device-enumeration)

janharaldfredriksen-arm:
Fix #630 

Change-Id: I7a3ff1fcee702263727154b1be67067d22687739

### DIFF
--- a/layers/device_limits.cpp
+++ b/layers/device_limits.cpp
@@ -165,22 +165,6 @@ EnumeratePhysicalDevices(VkInstance instance, uint32_t *pPhysicalDeviceCount, Vk
         if (NULL == pPhysicalDevices) {
             my_data->instanceState->vkEnumeratePhysicalDevicesState = QUERY_COUNT;
         } else {
-            if (UNCALLED == my_data->instanceState->vkEnumeratePhysicalDevicesState) {
-                // Flag error here, shouldn't be calling this without having queried count
-                skipCall |=
-                    log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, 0,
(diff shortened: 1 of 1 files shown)


layers/device_limits.cpp
```diff
@@ -165,22 +165,6 @@ EnumeratePhysicalDevices(VkInstance instance, uint32_t *pPhysicalDeviceCount, Vk
         if (NULL == pPhysicalDevices) {
             my_data->instanceState->vkEnumeratePhysicalDevicesState = QUERY_COUNT;
         } else {
-            if (UNCALLED == my_data->instanceState->vkEnumeratePhysicalDevicesState) {
-                // Flag error here, shouldn't be calling this without having queried count
-                skipCall |=
-                    log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, 0,
```
tobine: See #630 for why I think it would be useful to keep this as a WARNING.

janharaldfredriksen-arm:
Closing this request. Will replace with a new one that keeps the error as a warning.

PR #632, opened 2016-06-08
Replace call sequence errors with warnings for device query APIs

### LINKED ISSUE #630: Remove error of incorrect use of vkEnumeratePhysicalDevices

janharaldfredriksen-arm:
Fixes #630 

The issue was created for vkEnumeratePhysicalDevices, but also applies to GetPhysicalDeviceQueueFamilyProperties. This MR addresses both.

It adds a new enum to distinguish these warnings from one remaining case (vkCreateDevice) where you really must have called an enumeration API first.

Change-Id: I1505f4a2b339ac7ef435da8b3ec2f85612026a42

### DIFF
--- a/layers/device_limits.cpp
+++ b/layers/device_limits.cpp
@@ -166,15 +166,16 @@ EnumeratePhysicalDevices(VkInstance instance, uint32_t *pPhysicalDeviceCount, Vk
             my_data->instanceState->vkEnumeratePhysicalDevicesState = QUERY_COUNT;
         } else {
             if (UNCALLED == my_data->instanceState->vkEnumeratePhysicalDevicesState) {
-                // Flag error here, shouldn't be calling this without having queried count
+                // Flag warning here. You can call this without having queried the count, but it may not be
+                // robust on platforms with multiple physical devices.
                 skipCall |=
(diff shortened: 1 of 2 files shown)


mark-lunarg:
Cherry-picked and pushed to Master.

commit 2af1e0d106432f6ac46c5f9596687eed8a7a13b4
Author: Jan-Harald Fredriksen <jan-harald.fredriksen@arm.com>
Date:   Wed Jun 8 18:51:52 2016 +0200

    layers: PR632, Change devlimits errors to warnings

    Replace call sequence errors with warnings for
    vkEnumeratePhysicalDevices and GetPhysicalDeviceQueueFamilyProperties.

    Change-Id: I1505f4a2b339ac7ef435da8b3ec2f85612026a42

PR #633, opened 2016-06-09
Modify warning msgs: add debug output message header

mark-lunarg:
This PR should print a message header to stdout with the first validation message that gets printed.  Looking for wording changes to the error/warn/perf warn definitions in particular.

### DIFF
--- a/layers/vk_layer_config.cpp
+++ b/layers/vk_layer_config.cpp
@@ -32,6 +32,17 @@
 
 #define MAX_CHARS_PER_LINE 4096
 
+// Flag for ONE-TIME debug output initialization
+static bool log_message_initialized = false;
+
+// Accessor functions to ensure the correct instance of the shared flag gets used
+const bool LogMessageInitialized() {
+    return log_message_initialized;
+}
+void SetLogMessageInitialized() {
+    log_message_initialized = true;
+}
+
 class ConfigFile {
   public:
     ConfigFile();
--- a/layers/vk_layer_logging.h
+++ b/layers/vk_layer_logging.h
@@ -104,17 +104,52 @@ static inline void RemoveAllMessageCallbacks(debug_report_data *debug_data, VkLa
     *list_head = NULL;
 }
 
+static inline bool OutputLogMsgHeader(const debug_report_data *debug_data, VkFlags msgFlags) {
+    bool output = false;
+    VkLayerDbgFunctionNode *pTrav =
+        (debug_data->debug_callback_list != NULL) ? debug_data->debug_callback_list : debug_data->default_debug_callback_list;
+
+    while (pTrav) {
+        const char *message_header = "Legend:\n\n"
+                               "*******************************************************************************************\n"
+                               "* Vulkan Validation Layer Debug Output                                                    *\n"
+                               "*                                                                                         *\n"
+                               "* Debug Output Type Definitions:                                                          *\n"
+                               "* ------------------------------                                                          *\n"
+                               "* ERROR: Errors are output when a validation layer detects that some application behavior *\n"
+                               "*        has violated the Vulkan Specification.                                           *\n"
+                               "* WARN:  Warnings are output in cases where mistakes are commonly made and do NOT         *\n"
+                               "*        necessarily indicate that an app has violated the Vulkan Specification.          *\n"
+                               "*        Warnings basically translate to 'Did you really mean to do this?'                *\n"
+                               "* PERF:  Performance Warnings are output in cases where a possible inefficiency has been  *\n"
+                               "*        detected.  These also do NOT imply that the specification was violated.          *\n"
+                               "* INFO:  These log messages are for informational purposes only. For instance, the        *\n"
+                               "*        core_validation layer can print out lists of memory objects and their bindings   *\n"
+                               "*        which may help with debugging or improving application efficiency.               *\n"
+                               "*******************************************************************************************\n";
+
+        if (pTrav->msgFlags & msgFlags) {
+            pTrav->pfnMsgCallback(VK_DEBUG_REPORT_DEBUG_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEBUG_REPORT_EXT, 0, 0,
+                VK_DEBUG_REPORT_ERROR_CALLBACK_REF_EXT, "", message_header, pTrav->pUserData);
+            output |= true;
+        }
+        pTrav = pTrav->pNext;
+    }
+    return output;
+}
+
 // Utility function to handle reporting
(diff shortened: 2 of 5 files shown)


layers/vk_layer_logging.h
```diff
@@ -104,17 +104,52 @@ static inline void RemoveAllMessageCallbacks(debug_report_data *debug_data, VkLa
...
+
+    while (pTrav) {
+        const char *message_header = "Legend:\n\n"
+                               "*******************************************************************************************\n"
+                               "* Vulkan Validation Layer Debug Output                                                    *\n"
+                               "*                                                                                         *\n"
+                               "* Debug Output Type Definitions:                                                          *\n"
+                               "* ------------------------------                                                          *\n"
+                               "* ERROR: Errors are output when a validation layer detects that some application behavior *\n"
+                               "*        has violated the Vulkan Specification.                                           *\n"
```
stroyan: Returning "true" will only prevent that call from proceeding to the lower
layers and driver.
It won't in itself change the behavior of the application and future calls
to vulkan.
To be really safe the application would need to change its implicit
expectation that the bad call was made.
And earlier layers may not know that the call did not complete not all such
calls return results.
The only simple path to safety is to stop making vulkan calls.
That would be likely to be done by just shutting down the application.

On Mon, Jun 13, 2016 at 10:40 AM, Tobin Ehlis <notifications@github.com>
wrote:

> In layers/vk_layer_logging.h
> <https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/633#discussion_r66825347>
> :
>
> > @@ -104,17 +104,52 @@ static inline void RemoveAllMessageCallbacks(debug_report_data *debug_data, VkLa
> >      *list_head = NULL;
> >  }
> >
> > +static inline bool OutputLogMsgHeader(const debug_report_data *debug_data, VkFlags msgFlags) {
> > +    bool output = false;
> > +    VkLayerDbgFunctionNode *pTrav =
> > +        (debug_data->debug_callback_list != NULL) ? debug_data->debug_callback_list : debug_data->default_debug_callback_list;
> > +
> > +    while (pTrav) {
> > +        const char *message_header = "Legend:\n\n"
> > +                               "*******************************************************************************************\n"
> > +                               "* Vulkan Validation Layer Debug Output                                                    *\n"
> > +                               "*                                                                                         *\n"
> > +                               "* Debug Output Type Definitions:                                                          *\n"
> > +                               "* ------------------------------                                                          *\n"
> > +                               "* ERROR: Errors are output when a validation layer detects that some application behavior *\n"
> > +                               "*        has violated the Vulkan Specification.                                           *\n"
>
> For ERROR I would like to add something similar to the following:
>
> When an error is encountered it is strongly recommended that the user call
> back function return "true" in order to halt execution, otherwise behavior
> becomes undefined and a crash becomes probable.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/633/files/5715283cbca9d249707608b2632fa8ce6d6a5898#r66825347>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AEV7uKgz0NHBZ2yl8LpvzdK3ZTqENs6Vks5qLYf0gaJpZM4IyKFi>
> .
>



-- 
Mike Stroyan, stroyan@google.com

layers/vk_layer_config.cpp
```diff
@@ -32,6 +32,17 @@
 
 #define MAX_CHARS_PER_LINE 4096
 
+// Flag for ONE-TIME debug output initialization
+static bool log_message_initialized = false;
+
+// Accessor functions to ensure the correct instance of the shared flag gets used
+const bool LogMessageInitialized() {
```
tobine: When I build this on Linus I get:

```vk_layer_config.h:66:34: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
 const bool LogMessageInitialized();```

File: layers/vk_layer_logging.h (hunk 1 above)
Comment (mark-lunarg) [inline layers/vk_layer_logging.h]: Well, Tobin's answer was shorter.  Mike is right, even returning TRUE might not keep you out of trouble, but validation-wise you'll get more bang for your buck.  How about something like "When an error is encountered it is recommended that the user call back function return "true" to get the best validation results, though doing so can still result in undefined behavior."

File: layers/vk_layer_config.cpp (hunk 2 above)
Comment (mark-lunarg) [inline layers/vk_layer_config.cpp]: Ah, thanks.  Apparently we don't have warnings_as_errors set yet, maybe soon.

File: layers/vk_layer_logging.h (hunk 1 above)
Comment (tobine) [inline layers/vk_layer_logging.h]: Yeah, good point, @stroyan. I'm not sure exactly what I want to say yet. @mark-lunarg recommendation is better. Key idea is that when errors are encountered we want users to fix them ASAP and understand that while there are errors, bad things can (and often are likely) happen. Still thinking of ideal way to say that bug Mark's suggestion is a good start if you want to proceed with that.

**Comment by mark-lunarg** [comment]
Fixed up ERROR definition per @tobine and pushed to Master.

File: layers/vk_layer_logging.h [hunk 3]
```diff
@@ -104,17 +104,52 @@ static inline void RemoveAllMessageCallbacks(debug_report_data *debug_data, VkLa
...
+                               "* WARN:  Warnings are output in cases where mistakes are commonly made and do NOT         *\n"
+                               "*        necessarily indicate that an app has violated the Vulkan Specification.          *\n"
+                               "*        Warnings basically translate to 'Did you really mean to do this?'                *\n"
+                               "* PERF:  Performance Warnings are output in cases where a possible inefficiency has been  *\n"
+                               "*        detected.  These also do NOT imply that the specification was violated.          *\n"
+                               "* INFO:  These log messages are for informational purposes only. For instance, the        *\n"
+                               "*        core_validation layer can print out lists of memory objects and their bindings   *\n"
+                               "*        which may help with debugging or improving application efficiency.               *\n"
+                               "*******************************************************************************************\n";
```
chrisforbes: Do we really need the box of stars? This wraps on an 80 column terminal (it's 91 wide), and looks awful when it lands somewhere other than a terminal.

layers/vk_layer_logging.h
```diff
@@ -104,17 +104,52 @@ static inline void RemoveAllMessageCallbacks(debug_report_data *debug_data, VkLa
...
+
+        if (pTrav->msgFlags & msgFlags) {
```
chrisforbes: This ends up delivering a message with `VK_DEBUG_REPORT_DEBUG_BIT_EXT` to callbacks which have filtered out that bit. I don't think that's OK.

PR #634, opened 2016-06-10
Fix semaphore tracking for CB simultaneously inflight

chrisforbes:
This series tidies up around CB submission a bit, and moves the semaphore sets into a new CB_SUBMISSION structure. This allows correct tracking of the different semaphore sets associated with each simultaneous submission of the same CB.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -318,6 +318,22 @@ VkBufferViewCreateInfo *getBufferViewInfo(const layer_data *my_data, VkBufferVie
     return bv_it->second.get();
 }
 
+FENCE_NODE *getFenceNode(layer_data *dev_data, VkFence fence) {
+    auto it = dev_data->fenceMap.find(fence);
+    if (it == dev_data->fenceMap.end()) {
+        return nullptr;
+    }
+    return &it->second;
+}
+
+QUEUE_NODE *getQueueNode(layer_data *dev_data, VkQueue queue) {
+    auto it = dev_data->queueMap.find(queue);
+    if (it == dev_data->queueMap.end()) {
+        return nullptr;
+    }
+    return &it->second;
+}
+
 static VkDeviceMemory *get_object_mem_binding(layer_data *my_data, uint64_t handle, VkDebugReportObjectTypeEXT type) {
     switch (type) {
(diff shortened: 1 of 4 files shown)


mark-lunarg:
Chris, this looks great and I'll try to give it a run through our build server and do some more strenuous testing.  We can fix up the commit messages before they get pushed.

mark-lunarg:
So far, have only seen that on Windows, running the layer validation tests results in:
`[ RUN ] VkLayerTest.CommandBufferTwoSubmits
Assertion failed: !pFence->submissions.size(), file C:\dev\xgl\Vulkan-LoaderAndValidationLayers\layers\core_validation.cpp, line 4352`

layers/core_validation.cpp
```diff
@@ -318,6 +318,22 @@ VkBufferViewCreateInfo *getBufferViewInfo(const layer_data *my_data, VkBufferVie
     return bv_it->second.get();
 }
```
mark-lunarg: These changes might also need to be run through clang-format.

tests/layer_validation_tests.cpp
```diff
@@ -2237,6 +2237,61 @@ TEST_F(VkLayerTest, ResetUnsignaledFence) {
...
 }
 
+TEST_F(VkLayerTest, CommandBufferSimultaneousUseSync)
+{
+    m_errorMonitor->ExpectSuccess();
+
+    ASSERT_NO_FATAL_FAILURE(InitState());
+    VkResult err;
+
+    // record (empty!) command buffer that can be submitted multiple times
```
mark-lunarg: Can these comments get capitalized?  That'd make them more consistent with the rest of the file.

chrisforbes:
@mark-lunarg This has uncovered some more missing initialization -- will fix on Monday.

chrisforbes:
@mark-lunarg fixed the test. In addition to the behavior it was supposed to exercise, `CommandBufferTwoSubmits` was resubmitting the /fence/  and blundering past the error for that. I've fixed it to be more precise.

mark-lunarg:
Passes the tests now, thanks Chris.

tobine:

PR #636, opened 2016-06-12
vkjson: Add whole-instance support

critsec:
Previously vkjson only represented, imported, and exported a single VkPhysicalDevice at a time. Now it can also handle an entire instance: instance extensions, instance layers (and their extensions), as well as a set of devices.

This change remains source compatible, and the vkjson_info command takes the same arguments and produces the same output as before; whole-instance is a new mode. It is not binary compatible, but a simple recompile of the client will work; binary compatibility could be retained easily, but since it's intended to be used as a static rather than shared library this didn't seem important.


libs/vkjson/vkjson_instance.cc
```diff
@@ -0,0 +1,160 @@
...
+      nullptr};
+  VkInstance vkinstance;
+  result = vkCreateInstance(&instance_info, nullptr, &vkinstance);
+  if (result != VK_SUCCESS)
+    return VkJsonInstance();
+
+  count = 0;
+  result = vkEnumeratePhysicalDevices(vkinstance, &count, nullptr);
+  if (result != VK_SUCCESS)
+    return VkJsonInstance();
```
pimanttr: Destroy instance in this case - otherwise it leaks.

libs/vkjson/vkjson_instance.cc
```diff
@@ -0,0 +1,160 @@
...
+  std::vector<VkPhysicalDevice> devices(count, VK_NULL_HANDLE);
+  result = vkEnumeratePhysicalDevices(vkinstance, &count, devices.data());
+  if (result != VK_SUCCESS)
+    return VkJsonInstance();
```
pimanttr: ditto.

pimanttr:
LGTM after couple of issues fixed

critsec:
Missing vkDestroyInstance fixed; thanks Antoine.

Jon, I don't have a Windows machine to check whether this builds. Anyone there who can check?

jon-lunarg:
Runs on Windows just fine even with multiple ICDs. Cherry picked commits to master branch. Thanks!

PR #638, opened 2016-06-14
layers: fix CMakeLists.txt to support non MSVC win builds

MarkMLunarG:
Fix a recent change broke windows builds that don't use the VC IDE or decorate build trees like VC does

Change-Id: I93c8a559c4255408a6fa890b2a3e4d81822d1cea


mark-lunarg:
+1.  Fixed commit message and pushed to Master.

PR #642, opened 2016-06-15
Tobin various fixes

tobine:
Primary change here is adding a new check to verify that PSO renderpass at draw time is compatible with the command buffer's active renderpass. Also added a test for this case.
Also have a bug fix to prevent NULL ptr issues and some minor cleanup.


mark-lunarg:
LGTM.

layers/core_validation.cpp
```diff
@@ -2133,6 +2133,15 @@ static bool attachment_references_compatible(const uint32_t index, const VkAttac
                                              const uint32_t primaryCount, const VkAttachmentDescription *pPrimaryAttachments,
                                              const VkAttachmentReference *pSecondary, const uint32_t secondaryCount,
                                              const VkAttachmentDescription *pSecondaryAttachments) {
+    // Check potential NULL cases first to avoid nullptr issues later
+    if (pPrimary == NULL) {
+        if (pSecondary == NULL) {
```
chrisforbes: Please use `nullptr`

tobine:

Issue #644, opened 2016-06-15
Validate DS image view aspectMask restriction


tobine:
According to spec `When using an imageView of a depth/stencil image to populate a descriptor set (e.g. for sampling in the shader, or for use as an input attachment), the aspectMask must only include one bit and selects whether the imageView is used for depth reads (i.e. using a floating-point sampler or input attachment in the shader) or stencil reads (i.e. using an unsigned integer sampler or input attachment in the shader). When an imageView of a depth/stencil image is used as a depth/stencil framebuffer attachment, the aspectMask is ignored and both depth and stencil image subresources are used.`

Need to validate this. Should be simple to add for textures, which I'll do first, and then will add input attachment support second.

DominikWitczakAMD:
Please note https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/issues/650. On the surface, that ticket seems to be slightl related to what you're working on ATM.

tobine:
Fix was cherry-picked and pushed to master

PR #646, opened 2016-06-15
Tobin gh644 ds tex aspect mask restriction

tobine:
Fix for #644. Add a check to make sure any time a DS image is used in a descriptorset that it DOES NOT have both DEPTH and STENCIL aspect bits set.
Also added test to verify the check and updated documentation.


mark-lunarg:
+1

tobine:

PR #355, opened 2016-04-15
Adding DisplayPlane Surface support for object tracking

pbantolas:
Having no WSI was not handled in the layer generation scripts. Added a
new DisplayServer named 'Null'.

Added a missing layer function for vkCreateDisplayPlaneSurfaceKHR. This
fixes errors when using a NULL window system reporting an invalid
VkSurface object.

PR #656, opened 2016-06-17
Add validation support for WSI  khr display extension

jon-lunarg:
Mark, Tobin, and Ian:
Can you review these commits for adding VK_KHR_display extension support to object_tracker, unique_objects  and swapchain?  The swapchain validation is pretty basic, there are probably more checks that could be added later.

Petros:
I squashed your series of commits from PR #355 into one commit that is in this PR.  Then I finished out the support for the extension (primarily in object_tracker and unique_objects).  Please review.  I think as far as object_tracker and unique_object layers they should completely support this extension now.

### DIFF
--- a/layers/swapchain.cpp
+++ b/layers/swapchain.cpp
@@ -951,8 +951,6 @@ GetPhysicalDeviceDisplayPropertiesKHR(VkPhysicalDevice physicalDevice, uint32_t 
     return VK_ERROR_VALIDATION_FAILED_EXT;
 }
 
-static uint32_t gDisplayPlanePropertyCount = 0;
-static bool gDisplayPlanePropertyCountInit = false;
 VKAPI_ATTR VkResult VKAPI_CALL 
 GetPhysicalDeviceDisplayPlanePropertiesKHR(VkPhysicalDevice physicalDevice, uint32_t *pPropertyCount, VkDisplayPlanePropertiesKHR *pProperties) {
     VkResult result = VK_SUCCESS;
@@ -1016,7 +1014,7 @@ GetDisplayPlaneSupportedDisplaysKHR(VkPhysicalDevice physicalDevice, uint32_t pl
         skipCall |= LOG_ERROR_NULL_POINTER(VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, pPhysicalDevice->pInstance, "pDisplayCount");
     }
 
-    if (!gDisplayPlanePropertyCountInit)
+    if (!pPhysicalDevice->gotDisplayPlanePropertyCount)
     {
         LOG_WARNING(VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, pPhysicalDevice->pInstance, "planeIndex",
                 SWAPCHAIN_GET_SUPPORTED_DISPLAYS_WITHOUT_QUERY,
@@ -1024,14 +1022,14 @@ GetDisplayPlaneSupportedDisplaysKHR(VkPhysicalDevice physicalDevice, uint32_t pl
                 __FUNCTION__);
     }
 
-    if (gDisplayPlanePropertyCountInit && planeIndex >= gDisplayPlanePropertyCount)
+    if (pPhysicalDevice->gotDisplayPlanePropertyCount && planeIndex >= pPhysicalDevice->gotDisplayPlanePropertyCount)
     {
         skipCall |= LOG_ERROR(VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, pPhysicalDevice->pInstance, "planeIndex",
                 SWAPCHAIN_PLANE_INDEX_TOO_LARGE,
                 "%s(): %s must be in the range [0, %d] that was returned by vkGetPhysicalDeviceDisplayPlanePropertiesKHR. Do you have the plane index hardcoded?",
                 __FUNCTION__,
                 "planeIndex",
-                gDisplayPlanePropertyCount - 1);
+                pPhysicalDevice->displayPlanePropertyCount - 1);
     }
     lock.unlock();
 
@@ -1125,7 +1123,7 @@ GetDisplayPlaneCapabilitiesKHR(VkPhysicalDevice physicalDevice, VkDisplayModeKHR
                               VK_KHR_DISPLAY_EXTENSION_NAME);
     }
 
-    if (!gDisplayPlanePropertyCountInit)
+    if (!pPhysicalDevice->gotDisplayPlanePropertyCount)
     {
         LOG_WARNING(VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, pPhysicalDevice->pInstance, "planeIndex",
                 SWAPCHAIN_GET_SUPPORTED_DISPLAYS_WITHOUT_QUERY,
@@ -1133,14 +1131,14 @@ GetDisplayPlaneCapabilitiesKHR(VkPhysicalDevice physicalDevice, VkDisplayModeKHR
                 __FUNCTION__);
     }
 
-    if (gDisplayPlanePropertyCountInit && planeIndex >= gDisplayPlanePropertyCount)
+    if (pPhysicalDevice->gotDisplayPlanePropertyCount && planeIndex >= pPhysicalDevice->displayPlanePropertyCount)
     {
         skipCall |= LOG_ERROR(VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, pPhysicalDevice->pInstance, "planeIndex",
                 SWAPCHAIN_PLANE_INDEX_TOO_LARGE,
                 "%s(): %s must be in the range [0, %d] that was returned by vkGetPhysicalDeviceDisplayPlanePropertiesKHR. Do you have the plane index hardcoded?",
                 __FUNCTION__,
                 "planeIndex",
-                gDisplayPlanePropertyCount - 1);
+                pPhysicalDevice->displayPlanePropertyCount - 1);
     }
 
     if (!pCapabilities) {
--- a/layers/swapchain.cpp
+++ b/layers/swapchain.cpp
@@ -129,11 +129,26 @@ static void createInstanceRegisterExtensions(const VkInstanceCreateInfo *pCreate
         (PFN_vkGetPhysicalDeviceSurfaceFormatsKHR)gpa(instance, "vkGetPhysicalDeviceSurfaceFormatsKHR");
     pDisp->GetPhysicalDeviceSurfacePresentModesKHR =
         (PFN_vkGetPhysicalDeviceSurfacePresentModesKHR)gpa(instance, "vkGetPhysicalDeviceSurfacePresentModesKHR");
+    pDisp->CreateDisplayPlaneSurfaceKHR =
+        (PFN_vkCreateDisplayPlaneSurfaceKHR)gpa(instance, "vkCreateDisplayPlaneSurfaceKHR");
+    pDisp->GetPhysicalDeviceDisplayPropertiesKHR =
+        (PFN_vkGetPhysicalDeviceDisplayPropertiesKHR)gpa(instance, "vkGetPhysicalDeviceDisplayPropertiesKHR");
+    pDisp->GetPhysicalDeviceDisplayPlanePropertiesKHR =
+        (PFN_vkGetPhysicalDeviceDisplayPlanePropertiesKHR)gpa(instance, "vkGetPhysicalDeviceDisplayPlanePropertiesKHR");
+    pDisp->GetDisplayPlaneSupportedDisplaysKHR =
+        (PFN_vkGetDisplayPlaneSupportedDisplaysKHR)gpa(instance, "vkGetDisplayPlaneSupportedDisplaysKHR");
+    pDisp->GetDisplayModePropertiesKHR =
+        (PFN_vkGetDisplayModePropertiesKHR)gpa(instance, "vkGetDisplayModePropertiesKHR");
+    pDisp->CreateDisplayModeKHR =
+        (PFN_vkCreateDisplayModeKHR)gpa(instance, "vkCreateDisplayModeKHR");
+    pDisp->GetDisplayPlaneCapabilitiesKHR =
+        (PFN_vkGetDisplayPlaneCapabilitiesKHR)gpa(instance, "vkGetDisplayPlaneCapabilitiesKHR");
 
     // Remember this instance, and whether the VK_KHR_surface extension
     // was enabled for it:
     my_data->instanceMap[instance].instance = instance;
     my_data->instanceMap[instance].surfaceExtensionEnabled = false;
+    my_data->instanceMap[instance].displayExtensionEnabled = false;
 #ifdef VK_USE_PLATFORM_ANDROID_KHR
     my_data->instanceMap[instance].androidSurfaceExtensionEnabled = false;
 #endif // VK_USE_PLATFORM_ANDROID_KHR
@@ -165,6 +180,10 @@ static void createInstanceRegisterExtensions(const VkInstanceCreateInfo *pCreate
         if (strcmp(pCreateInfo->ppEnabledExtensionNames[i], VK_KHR_SURFACE_EXTENSION_NAME) == 0) {
 
             my_data->instanceMap[instance].surfaceExtensionEnabled = true;
+        }
+        if (strcmp(pCreateInfo->ppEnabledExtensionNames[i], VK_KHR_DISPLAY_EXTENSION_NAME) == 0) {
+
+            my_data->instanceMap[instance].displayExtensionEnabled = true;
         }
 #ifdef VK_USE_PLATFORM_ANDROID_KHR
         if (strcmp(pCreateInfo->ppEnabledExtensionNames[i], VK_KHR_ANDROID_SURFACE_EXTENSION_NAME) == 0) {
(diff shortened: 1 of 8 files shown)


layers/swapchain.cpp
```diff
@@ -129,11 +129,26 @@ static void createInstanceRegisterExtensions(const VkInstanceCreateInfo *pCreate
...
+    pDisp->GetPhysicalDeviceDisplayPlanePropertiesKHR =
+        (PFN_vkGetPhysicalDeviceDisplayPlanePropertiesKHR)gpa(instance, "vkGetPhysicalDeviceDisplayPlanePropertiesKHR");
+    pDisp->GetDisplayPlaneSupportedDisplaysKHR =
+        (PFN_vkGetDisplayPlaneSupportedDisplaysKHR)gpa(instance, "vkGetDisplayPlaneSupportedDisplaysKHR");
+    pDisp->GetDisplayModePropertiesKHR =
+        (PFN_vkGetDisplayModePropertiesKHR)gpa(instance, "vkGetDisplayModePropertiesKHR");
+    pDisp->CreateDisplayModeKHR =
+        (PFN_vkCreateDisplayModeKHR)gpa(instance, "vkCreateDisplayModeKHR");
+    pDisp->GetDisplayPlaneCapabilitiesKHR =
+        (PFN_vkGetDisplayPlaneCapabilitiesKHR)gpa(instance, "vkGetDisplayPlaneCapabilitiesKHR");
```
ianelliottus: Question: are the commands/functions from VK_KHR_display supposed to be statically exported from the loader, like the "core WSI extensions" (e.g. VK_KHR_surface)?  I honestly don't remember.  We should check, in case they need to be protected by an enabled-check.

layers/swapchain.cpp
```diff
@@ -129,11 +129,26 @@ static void createInstanceRegisterExtensions(const VkInstanceCreateInfo *pCreate
...
 
     // Remember this instance, and whether the VK_KHR_surface extension
     // was enabled for it:
     my_data->instanceMap[instance].instance = instance;
     my_data->instanceMap[instance].surfaceExtensionEnabled = false;
+    my_data->instanceMap[instance].displayExtensionEnabled = false;
```
ianelliottus: This should be protected by #ifdef.  That's because not all platforms will support the VK_KHR_display[_swapchain] extensions.

This comment can effectively be "ditto'd" throughout this pull request.

layers/swapchain.cpp
ianelliottus: This should be protected by #ifdef.  That's because not all platforms will support the VK_KHR_display[_swapchain] extensions.

This comment can effectively be "ditto'd" throughout this pull request.

ianelliottus: I believe that I'll be filing a Khronos-private bug about whether the VK_KHR_display[_swapchain] extension will be statically exported.  The language in the spec was adopted before these extensions were added.  Overtime, I can see some other extensions being created that may not be supported across Android, Linux, and/or Windows.  Thus, we might need some caveats about which WSI extensions will have their commands/functions statically exported.

layers/swapchain.cpp
```diff
@@ -901,6 +920,292 @@ VKAPI_ATTR VkBool32 VKAPI_CALL GetPhysicalDeviceXlibPresentationSupportKHR(VkPhy
...
+        skipCall |= LOG_ERROR(VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, pPhysicalDevice->pInstance, "VkInstance",
+                              SWAPCHAIN_EXT_NOT_ENABLED_BUT_USED,
+                              "%s() called even though the %s extension was not enabled for this VkInstance.", __FUNCTION__,
+                              VK_KHR_DISPLAY_EXTENSION_NAME);
+    }
+
+    if (!pPropertyCount) {
+        skipCall |= LOG_ERROR_NULL_POINTER(VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, pPhysicalDevice->pInstance, "pPropertyCount");
+    }
+    lock.unlock();
```
ianelliottus: Please look at the implementation of GetPhysicalDeviceSurfaceFormatsKHR() in this file, for the "else" case (i.e. if pPropertyCount is non-NULL).

layers/swapchain.cpp
ianelliottus: I did a little research and filed a private, spec issue (380) to address this.  At the time we discussed this, I believe we only had what I'll call the "core WSI extensions" (VK_KHR_{surface|swapchain}), which are fully supported on Android, Linux, and Windows.  Since then, we've added the 6 "platform extensions", which probably will only be supported on one OS each, and the 2 "display extensions", which will never be supported on Android.  The issue is to change the spec language a bit to clarify that only "relevant extensions" will have their functions statically exported by a given loader.

layers/swapchain.cpp
```diff
@@ -901,6 +920,292 @@ VKAPI_ATTR VkBool32 VKAPI_CALL GetPhysicalDeviceXlibPresentationSupportKHR(VkPhy
...
+
+    if (!skipCall) {
+        result = my_data->instance_dispatch_table->GetPhysicalDeviceDisplayPropertiesKHR(physicalDevice, pPropertyCount, pProperties);
```
ianelliottus: if ((result == VK_SUCCESS) && ...) {
    // Record the result ...
}

layers/swapchain.cpp
```diff
@@ -901,6 +920,292 @@ VKAPI_ATTR VkBool32 VKAPI_CALL GetPhysicalDeviceXlibPresentationSupportKHR(VkPhy
...
+    if (pPhysicalDevice && pPhysicalDevice->pInstance && !pPhysicalDevice->pInstance->displayExtensionEnabled) {
+        skipCall |= LOG_ERROR(VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, pPhysicalDevice->pInstance, "VkInstance",
+                              SWAPCHAIN_EXT_NOT_ENABLED_BUT_USED,
+                              "%s() called even though the %s extension was not enabled for this VkInstance.", __FUNCTION__,
+                              VK_KHR_DISPLAY_EXTENSION_NAME);
+    }
+
+    if (!pPropertyCount) {
+        skipCall |= LOG_ERROR_NULL_POINTER(VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, pPhysicalDevice->pInstance, "pPropertyCount");
+    }
```
ianelliottus: Please look at the implementation of GetPhysicalDeviceSurfaceFormatsKHR() in this file, for the "else" case (i.e. if pPropertyCount is non-NULL).

layers/swapchain.cpp
```diff
@@ -901,6 +920,292 @@ VKAPI_ATTR VkBool32 VKAPI_CALL GetPhysicalDeviceXlibPresentationSupportKHR(VkPhy
...
+    lock.unlock();
+
+    if (!skipCall) {
+        result = my_data->instance_dispatch_table->GetPhysicalDeviceDisplayPlanePropertiesKHR(physicalDevice, pPropertyCount, pProperties);
+
+        lock.lock();
+        if (!pPhysicalDevice->gotDisplayPlanePropertyCount)
+        {
+            pPhysicalDevice->displayPlanePropertyCount = *pPropertyCount;
+            pPhysicalDevice->gotDisplayPlanePropertyCount = true;
```
ianelliottus: Should should be recording the VkDisplayKHR handles, for future validation against.

layers/swapchain.cpp
```diff
@@ -901,6 +920,292 @@ VKAPI_ATTR VkBool32 VKAPI_CALL GetPhysicalDeviceXlibPresentationSupportKHR(VkPhy
...
+                "%s(): %s must be in the range [0, %d] that was returned by vkGetPhysicalDeviceDisplayPlanePropertiesKHR. Do you have the plane index hardcoded?",
+                __FUNCTION__,
+                "planeIndex",
+                pPhysicalDevice->displayPlanePropertyCount - 1);
+    }
+    lock.unlock();
+
+    if (!skipCall) {
+        result = my_data->instance_dispatch_table->GetDisplayPlaneSupportedDisplaysKHR(physicalDevice, planeIndex, pDisplayCount, pDisplays);
+
```
ianelliottus: Assuming VK_SUCCESS, you should should be recording the VkDisplayKHR handles, for future validation against.

layers/swapchain.cpp
```diff
@@ -901,6 +920,292 @@ VKAPI_ATTR VkBool32 VKAPI_CALL GetPhysicalDeviceXlibPresentationSupportKHR(VkPhy
...
+        }
+        lock.unlock();
+
+        return result;
+    }
+    return VK_ERROR_VALIDATION_FAILED_EXT;
+}
+
+VKAPI_ATTR VkResult VKAPI_CALL 
+GetDisplayPlaneSupportedDisplaysKHR(VkPhysicalDevice physicalDevice, uint32_t planeIndex, uint32_t* pDisplayCount, VkDisplayKHR* pDisplays) {
```
ianelliottus: Ditto some of the other comments from other queries

ianelliottus: Also, please look at the "Valid Usage" sections of the spec for things to validate against, which come from other queries.

layers/swapchain.cpp
```diff
@@ -901,6 +920,292 @@ VKAPI_ATTR VkBool32 VKAPI_CALL GetPhysicalDeviceXlibPresentationSupportKHR(VkPhy
...
+    if (!skipCall) {
+        result = my_data->instance_dispatch_table->GetDisplayPlaneSupportedDisplaysKHR(physicalDevice, planeIndex, pDisplayCount, pDisplays);
+
+        return result;
+    }
+    return VK_ERROR_VALIDATION_FAILED_EXT;
+}
+
+VKAPI_ATTR VkResult VKAPI_CALL 
+GetDisplayModePropertiesKHR(VkPhysicalDevice physicalDevice, VkDisplayKHR display, uint32_t* pPropertyCount, VkDisplayModePropertiesKHR* pProperties) {
```
ianelliottus: Ditto previous comments (which I won't keep repeating:-)

layers/swapchain.cpp
```diff
@@ -901,6 +920,292 @@ VKAPI_ATTR VkBool32 VKAPI_CALL GetPhysicalDeviceXlibPresentationSupportKHR(VkPhy
...
+        auto it = my_data->physicalDeviceMap.find(physicalDevice);
+        pPhysicalDevice = (it == my_data->physicalDeviceMap.end()) ? NULL : &it->second;
+    }
+
+    if (pPhysicalDevice && pPhysicalDevice->pInstance && !pPhysicalDevice->pInstance->displayExtensionEnabled) {
+        skipCall |= LOG_ERROR(VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, pPhysicalDevice->pInstance, "VkInstance",
+                              SWAPCHAIN_EXT_NOT_ENABLED_BUT_USED,
+                              "%s() called even though the %s extension was not enabled for this VkInstance.", __FUNCTION__,
+                              VK_KHR_DISPLAY_EXTENSION_NAME);
+    }
```
ianelliottus: Please validate against previous queries (e.g. per Valid Usage).

layers/swapchain.h
```diff
@@ -161,6 +164,9 @@ struct SwpInstance {
     // Set to true if VK_KHR_SURFACE_EXTENSION_NAME was enabled for this VkInstance:
     bool surfaceExtensionEnabled;
 
+    // Set to true if VK_KHR_DISPLAY_EXTENSION_NAME was enabled for this VkInstance:
```
ianelliottus: Please protect with #ifdef.

layers/unique_objects.h
```diff
@@ -81,6 +83,16 @@ static void createInstanceRegisterExtensions(const VkInstanceCreateInfo *pCreate
         (PFN_vkGetPhysicalDeviceSurfaceFormatsKHR)gpa(instance, "vkGetPhysicalDeviceSurfaceFormatsKHR");
     pDisp->GetPhysicalDeviceSurfacePresentModesKHR =
         (PFN_vkGetPhysicalDeviceSurfacePresentModesKHR)gpa(instance, "vkGetPhysicalDeviceSurfacePresentModesKHR");
+
+    // KHR_display
```
ianelliottus: Ditto on #ifdef.

vk-layer-generate.py
```diff
@@ -54,7 +54,14 @@ def proto_is_global(proto):
         "GetPhysicalDeviceMirPresentationSupportKHR",
         "CreateAndroidSurfaceKHR",
         "CreateWin32SurfaceKHR",
-        "GetPhysicalDeviceWin32PresentationSupportKHR"
+        "GetPhysicalDeviceWin32PresentationSupportKHR",
```
ianelliottus: Need some way of #ifdef'ing the display extension functions

ianelliottus:
-1.  Hi @jon-lunarg and @pbantolas.  I did a review (stopped after a bit).  This looks like a really good start, with a few issues I outlined in comments.  Best wishes!

layers/swapchain.cpp
jon-lunarg: Hi Ian:
Thanks for the code review!  I am not sure how the issue of weather the loader exports the KHR_display extension  entry points or not effects this code review.  The loader exports or doesn't export the entry points independently of any layer's decision to intercept  the extension entry points.  The above code is just initializing the layer's dispatch table for these functions.  So if the layer intercepts these functions  it can call down the chain by having it's dispatch table pointers initialized.

layers/swapchain.cpp
jon-lunarg: There is no ifdef in vulkan.h that pertains to this extension VK_KHR_display. Perhaps that is an oversight but I don't think so.  I think, the reason the VK_KHR_xlib_surface, VK_KHR_win32_surface, etc. extensions are protected by an ifdef in vulkan.h is because  these extensions # include platform specific headers (such as #include <X11/Xlib.h>) that will not compile on foreign platforms.  The VK_KHR_display extension doesn't # include any headers so it can compile on ANY platform even though it may not be supported on various platforms at run time by the driver.  Thus, if we ifdef this code we would have to invent a #define and the only advantage would be the layer library size would be smaller for platforms that don't support this extension. I think it is better to always compile this extension and let the app determine at run time if this extension is supported or not on the platform.

layers/swapchain.cpp
jon-lunarg: got it!

layers/swapchain.cpp
jon-lunarg: okay

layers/swapchain.cpp
jon-lunarg: okay

layers/swapchain.cpp
jon-lunarg: okay

layers/swapchain.cpp
jon-lunarg: okay

jon-lunarg:
Petros:

I pushed two new commits to this branch. One that fixed the issue you pointed out in unique_objects. The other commit fixes object_tracker. 

As for your comment:
"In a similar fashion, the object_tracker layer requires that vkGetDisplayPlaneSupportedDisplaysKHR is called for any subsequent VkDisplayKHR validation to pass. Do you think create_display_khr should be introduced to other display-related functions, or should vkGetDisplayPlaneSupportedDisplaysKHR always be called first?"

My closer reading of the spec indicates that GetPhysicalDeviceDisplayPropertiesKHR actually creates the DisplayKHR objects so it must be called first. GetDisplayPlaneSupportedDisplaysKHR returns DisplayKHR objects that have already been created. So I skipped validating these output
parameters  nor putting them in the map.  Do you think this is correct?

ashleysmithgpu:
Hi, I believe that vkGetPhysicalDeviceDisplayPropertiesKHR need an explicit_ version too, like vkGetDisplayModePropertiesKHR:

```
    if (result == VK_SUCCESS && pProperties)
    {   
        for (uint32_t idx0=0; idx0<*pPropertyCount; ++idx0) {
            
            uint64_t unique_id = global_unique_id++;
            my_map_data->unique_id_mapping[unique_id] = reinterpret_cast<uint64_t &>(local_pProperties[idx0].display);
            pProperties[idx0].display = reinterpret_cast<VkDisplayKHR&>(unique_id);
            pProperties[idx0].displayName = local_pProperties[idx0].displayName;
            pProperties[idx0].physicalDimensions = local_pProperties[idx0].physicalDimensions;
            pProperties[idx0].physicalResolution = local_pProperties[idx0].physicalResolution;
            pProperties[idx0].supportedTransforms = local_pProperties[idx0].supportedTransforms;
            pProperties[idx0].planeReorderPossible = local_pProperties[idx0].planeReorderPossible;
            pProperties[idx0].persistentContent = local_pProperties[idx0].persistentContent;
        }       
    }
```
As the pProperties is not copied over in the function at the moment. I also needed to modify terminator_CreateDisplayPlaneSurfaceKHR because it doesn’t map the pSurface using the unique_id_mapping system.

jon-lunarg:
AshleySmith:
Yes it looks like unique_objects needs an explicit_GetPhysicalDeviceDisplayPropertiesKHR like you said and the existing  explicit_GetDisplayPlaneSupportedDisplaysKHR() isn't quite right since it should reuse the existing mapping to unique_id rather than create a new one.

As for the explicit_CreateDisplayPlaneSurfaceKHR within unique_objects.cpp, what did you have to change? The surface mapping looks okay to me:

    if (VK_SUCCESS == result) {
// CODEGEN : file /home/jon/vk/vk/vk-layer-generate.py line #1594
        uint64_t unique_id = global_unique_id++;
        my_map_data->unique_id_mapping[unique_id] = reinterpret_cast<uint64_t &>(*pSurface);
        *pSurface = reinterpret_cast<VkSurfaceKHR&>(unique_id);
    }

jon-lunarg:
Added patch to fix the unique_objects  GetPhysicalDeviceDisplayProperties and GetDisplayPlaneSupportedDisplays.

layers/unique_objects.h
```diff
@@ -410,18 +410,56 @@ VkResult explicit_GetSwapchainImagesKHR(VkDevice device, VkSwapchainKHR swapchai
...
     if (VK_SUCCESS == result) {
         if ((*pDisplayCount > 0) && pDisplays) {
             std::lock_guard<std::mutex> lock(global_lock);
-            uint64_t unique_id;
             for (uint32_t i = 0; i < *pDisplayCount; i++) {
-                unique_id = global_unique_id++;
-                my_map_data->unique_id_mapping[unique_id] = reinterpret_cast<uint64_t &>(*pDisplays);
-                pDisplays[i] = reinterpret_cast<VkDisplayKHR&>(unique_id);
+                auto it = my_map_data->unique_id_mapping.find(reinterpret_cast<uint64_t> (pDisplays[i]));
+		assert (it !=  my_map_data->unique_id_mapping.end());
```
mark-lunarg: missingaspacehere.

pbantolas:
Jon:

That looks correct.

Although I think we should file a spec bug, it seems strange that you have to call GetPhysicalDeviceDisplayPropertiesKHR to initialise the displays. The spec is quite dubious on that aspect.

jon-lunarg:
Petros:

Go ahead and file a spec bug if you like for clarification.  I found the example usage in the spec does show calling GetPhysicalDeviceDisplayPropertiesKHR to get the display objects and then
GetDisplayPlaneSupportedDisplaysKHR to find which displays support  a given plane.

Ian:
I currently don't have time for adding all the validation checks to swapchain for this extension.  I did add TODOs in the code for the missing checks to be added later.

I fixed some compilation issues on Android and Windows  and pushing this series of patches to master branch.

PR #657, opened 2016-06-17
layers: Fix CreateRenderpass validation

mark-lunarg:
In the core_validation layer, much of the validation was occurring
AFTER the API call had been made. Separated validation and tracking.

Change-Id: I6f8502d52dd2861a18254de48141a3736ced257f

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -8640,25 +8640,86 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateShaderModule(VkDevice device, const VkShade
     return res;
 }
 
-VKAPI_ATTR VkResult VKAPI_CALL CreateRenderPass(VkDevice device, const VkRenderPassCreateInfo *pCreateInfo,
-                                                const VkAllocationCallbacks *pAllocator,
-                                                VkRenderPass *pRenderPass) {
+static bool ValidateAttachmentIndex(layer_data *dev_data, uint32_t attachment, uint32_t attachment_count, const char *type) {
     bool skip_call = false;
-    layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
-    // Create DAG
-    std::vector<bool> has_self_dependency(pCreateInfo->subpassCount);
-    std::vector<DAGNode> subpass_to_node(pCreateInfo->subpassCount);
-    {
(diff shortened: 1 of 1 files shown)


chrisforbes:
Will look at this on Monday.

layers/core_validation.cpp
```diff
@@ -8640,25 +8640,86 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateShaderModule(VkDevice device, const VkShade
...
-        // Validate
-        skip_call |= ValidateLayouts(dev_data, device, pCreateInfo);
-        if (skip_call) {
-            return VK_ERROR_VALIDATION_FAILED_EXT;
-        }
+
+    std::unique_lock<std::mutex> lock(global_lock);
+
+    skip_call |= ValidateLayouts(dev_data, device, pCreateInfo);
+#if MTMERGESOURCE
```
tobine: Why is this guarded by "MTMERGESOURCE"?

tobine:
As it stands this code looks ok and is an improvement on existing code by moving some validation checks that were after a call down the chain to before a call down the chain. All of this is good and I'm ok with current code going in.

That said, "ValidateLayouts" function already has a loop that parses all of the subpasses and various attachment types. I think it makes sense in a soon future commit to merge "ValidateLayouts" and the new "ValidateRenderpassAttachmentUsage" function. For this commit, can you add a "TODO" noting that such work should be done?

LGTM :+1:

cnorthrop:
+1
This fixes a crash on NexusPlayer

layers/core_validation.cpp
mark-lunarg: My inclination was to not put it in, but that chunk of validation was previously guarded similarly and as you point out it still needs to be reconciled to remove duplication and integrate things more uniformly. I'll remove the guards and add the TODO as you suggested.

mark-lunarg:
Fixes added, pushed to Master.

commit 4f76794409854faf16dd51f37d5fa21dff1c5d44
Date:   Fri Jun 17 15:24:01 2016 -0600

    layers: GH656, Fix CreateRenderpass validation

    In the core_validation layer, much of the validation was occurring
    AFTER the API call had been made. Separated validation and tracking.

    Change-Id: I6f8502d52dd2861a18254de48141a3736ced257f

PR #652, opened 2016-06-16
More fence tracking fixes

chrisforbes:
This series unhacks various weird things in fence handling.

PR #665, opened 2016-06-20
Tobin disable invalid mem type test

tobine:
I'd like to turn this test off for now as it crashed on at least 1 android driver and looking at it closer I don't like the way that it induces the error when allocating memory. Based on the current spec it's not clear that the way the test is allocating is illegal (just picking an arbitrary memTypeIndex), but I believe it should be which may require a spec update.
I noted a way that the test can be updated, which is left as a later exercise.


mark-lunarg:
I'll push it Tobin.  We're code freezing today, is there anything else you want/need to get into 1.17?

tobine:
#652 would be good. I already pushed #666 and I'll also shortly push #642 & #646.

mark-lunarg:
Also, if it's not already there (and it sounds like it might not be) we should probably have a validation check to make sure that any memoryTypeIndex is less than the number of memoryTypes returned by the driver.

mark-lunarg:
Pushed to Master.

commit 8c3801f19bd7924484128095ead518323ddb64e4
Author: Tobin Ehlis <tobine@google.com>
Date:   Mon Jun 20 11:32:13 2016 -0600

    tests: Disable BindImageInvalidMemoryType test

    This test causes a crash on at least one android gfx driver and it's not clear
    that the way it allocates memory is within spec. Commented the test to note an
    allternative method that can be used to get the desired result but also make
    sure that memory is allocated in a device-supported manner.

PR #668, opened 2016-06-21
layers: Fix attachment_references_compatible() crash in Windows

mark-lunarg:
In attachment_references_compatible(), having array elements set
to VK_ATTACHMENT_UNUSED was not handled resulting in in invalid
memory references.

Tobin, this fixes the crashes but I'm getting validation errors in Dota2.  This seems to be fine spec-wise, can you take a look?

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -2134,7 +2134,7 @@ static bool validate_draw_state_flags(layer_data *dev_data, GLOBAL_CB_NODE *pCB,
 
 // Verify attachment reference compatibility according to spec
 //  If one array is larger, treat missing elements of shorter array as VK_ATTACHMENT_UNUSED & other array much match this
-//  If both AttachmentReference arrays have requested index, check their corresponding AttachementDescriptions
+//  If both AttachmentReference arrays have requested index, check their corresponding AttachmentDescriptions
 //   to make sure that format and samples counts match.
 //  If not, they are not compatible.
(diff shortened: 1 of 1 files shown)


layers/core_validation.cpp
```diff
@@ -2156,7 +2156,14 @@ static bool attachment_references_compatible(const uint32_t index, const VkAttac
     } else if (index >= secondaryCount) { // Check primary as if secondary is VK_ATTACHMENT_UNUSED
         if (VK_ATTACHMENT_UNUSED == pPrimary[index].attachment)
             return true;
-    } else { // format and sample count must match
+    } else { // Format and sample count must match
+        if (pPrimary[index].attachment == VK_ATTACHMENT_UNUSED) {
+            if (pSecondary[index].attachment == VK_ATTACHMENT_UNUSED) {
+                return true;
+            }
```
tobine: I believe there should be an early "return false" here for case where pPrimary is UNUSED but pSecondary is not. In that case we don't want to proceed with the format checks.

mark-lunarg:
Cherry-picked and pushed to Master (and SDK) branches.

commit 5ca3c071afc98f9144f489b74aa3affaa93633b0
Date:   Mon Jun 20 18:49:25 2016 -0600

    layers: Fix Windows crash in Dota2

    In core_validation attachment_references_compatible(), having array
    elements set to VK_ATTACHMENT_UNUSED was not handled resulting in
    invalid memory references.

    Change-Id: If78aea1467d191db905272bece16cc9f56643136

PR #672, opened 2016-06-21
Do not clobber CXXFLAGS in various CMakeLists.txt files

jcowgill:
This variable usually allows users to add custom compile options which can be very useful, but currently does not work for the validation layers because the flags are erased in the `CMakeLists.txt` file.


karl-lunarg:
Some additional changes are required beyond what is in the pull request.  By inheriting the CXXFLAGS from the top-level CMakeLists.txt file, a large number of warnings and errors appear in the build.  This is because the top-level sets -Wall and turns off a selected few warnings.  When these three CMakeLists.txt files start inheriting these warning settings, the additional warnings and errors appear.

While the Right way to fix this would be to fix the code, much of this code is auto-generated and some of the compilation problems will be difficult to fix.  The alternative would be to go ahead and inherit CXXFLAGS, but add additional options to turn off the problematic warnings.

We'll probably end up doing some combination of the above two options, but it may take some time.  I just wanted to mention that this isn't a trivial change.  But this is a good thing to address.

jcowgill:
Thanks, that sounds reasonable. I think I must have overlooked the extra warnings generated by this :/

karl-lunarg:
Fixed with https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/commit/e8a100f135c260ad3021b630d7715dc14e09d6a6

PR #673, opened 2016-06-21
tests: Use defined but still incompatible format in RenderPassIncompa…

cnorthrop:
…tible

Some implementation don't like undefined.  This gets all devices green on master.


tobine:

PR #674, opened 2016-06-21
Layers: Improve locking in swapchain layer.

ianelliottus:
This fixes cases where the swapchain layer wasn't always unlocking before
returning from a function.


ianelliottus:
I cherry-picked and pushed this commit.

Issue #616, opened 2016-06-03
Crash if binding point is used with no resource has been bound


janharaldfredriksen-arm:
Specifically, descriptor_sets.cpp::GetGlobalStartIndexFromBinding returns 0xFFFFFFFF if the binding doesn't exist:
```
// For the given binding, return start index
uint32_t cvdescriptorset::DescriptorSetLayout::GetGlobalStartIndexFromBinding(const uint32_t binding) const {
    assert(binding_to_global_start_index_map_.count(binding));
    const auto &btgsi_itr = binding_to_global_start_index_map_.find(binding);
    if (btgsi_itr != binding_to_global_start_index_map_.end()) {
        return btgsi_itr->second;
    }
    // In error case max uint32_t so index is out of bounds to break ASAP
    return 0xFFFFFFFF;
}
```

This index is then out of bounds value is then used to index an array - and for non-immutable samplers, to create very long loop:
```
bool cvdescriptorset::DescriptorSet::ValidateDrawState(const std::unordered_set<uint32_t> &bindings,
                                                       const std::vector<uint32_t> &dynamic_offsets, std::string *error) const {
    auto dyn_offset_index = 0;
    for (auto binding : bindings) {
        auto start_idx = p_layout_->GetGlobalStartIndexFromBinding(binding);
        if (descriptors_[start_idx]->IsImmutableSampler()) {
            // Nothing to do for strictly immutable sampler
        } else {
            auto end_idx = p_layout_->GetGlobalEndIndexFromBinding(binding);
            for (uint32_t i = start_idx; i <= end_idx; ++i) {
                if (!descriptors_[i]->updated) {
```

Similar issues are present in other places where GetGlobalStartIndexFromBinding is called.

tobine:
Submitted #675 to address these issues

tobine:
Fix was cherry-picked and pushed to master

PR #675, opened 2016-06-21
layers: Improve descriptor set invalid binding handling

tobine:
Fix for #616 

Prior to calls to get an global index from a binding, make sure that
HasBinding() is called to verify that the binding exists.
Added HasBinding() calls to a couple of validation functions for
early exist.
Updated code comment to indicate that HasBinding() call should
precede GetGlobal*IndexFromBinding() calls.
Note that Perform*Update() functions already indicate that they should
be preceded by their matching Validate*Update() functions, which include
the appropriate checks for HasBinding().
case the Validate*Update() funciont


tobine:

PR #678, opened 2016-06-22
Wip tobin gh643 fb valid usage

tobine:
WIP of CreateFramebuffer validation checks. Still need to enable 1 more check and add a couple of tests, but putting this out for early review.

My current main concern is whether or not I'm doing the dimensions check correctly.


layers/core_validation.cpp
```diff
@@ -142,7 +142,7 @@ struct layer_data {
     unordered_map<VkQueryPool, QUERY_POOL_NODE> queryPoolMap;
     unordered_map<VkSemaphore, SEMAPHORE_NODE> semaphoreMap;
     unordered_map<VkCommandBuffer, GLOBAL_CB_NODE *> commandBufferMap;
-    unordered_map<VkFramebuffer, FRAMEBUFFER_NODE> frameBufferMap;
+    unordered_map<VkFramebuffer, unique_ptr<FRAMEBUFFER_NODE>> frameBufferMap;
```
chrisforbes: What are we gaining here?

layers/core_validation.cpp
```diff
@@ -8267,15 +8266,10 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateFramebuffer(VkDevice device, const VkFrameb
         // Shadow create info and store in map
         std::lock_guard<std::mutex> lock(global_lock);
 
+        dev_data->frameBufferMap.insert(
+            std::make_pair(*pFramebuffer, unique_ptr<FRAMEBUFFER_NODE>(new FRAMEBUFFER_NODE(
+                                              pCreateInfo, dev_data->renderPassMap[pCreateInfo->renderPass]->pCreateInfo))));
         auto & fbNode = dev_data->frameBufferMap[*pFramebuffer];
```
chrisforbes: This is weird. You've just done an explicit insert and then you look it up again?

chrisforbes: Can accomplish the same thing but lose the additional unique_ptr layer and allocation by using emplace here. [I /assume/ that's what drove you to add the unique_ptr-- wanting to use FRAMEBUFFER_NODE's ctor?]

layers/core_validation.cpp
```diff
@@ -8267,15 +8266,10 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateFramebuffer(VkDevice device, const VkFrameb
...
-        fbNode.createInfo = *pCreateInfo;
-        if (pCreateInfo->pAttachments) {
-            auto attachments = new VkImageView[pCreateInfo->attachmentCount];
```
chrisforbes: Nice that this goes away.

layers/core_validation.cpp
```diff
@@ -8228,11 +8228,21 @@ static bool MatchUsage(layer_data *dev_data, uint32_t count, const VkAttachmentR
...
 }
 
-static bool ValidateAttachmentImageUsage(layer_data *dev_data, const VkFramebufferCreateInfo *pCreateInfo) {
+// Validate VkFramebufferCreateInfo which includes:
+// 1. attachmentCount equals renderPass attachmentCount
+static bool ValidateFramebufferCreateInfo(layer_data *dev_data, const VkFramebufferCreateInfo *pCreateInfo) {
     bool skip_call = false;
 
     const VkRenderPassCreateInfo *rpci = getRenderPass(dev_data, pCreateInfo->renderPass)->pCreateInfo;
     if (rpci != nullptr) {
```
chrisforbes: Iffy: this is guaranteed never `nullptr`. `getRenderPass` may return `nullptr` if the renderpass doesn't exist, but you already took a segfault on the previous line if that happened.

layers/core_validation.cpp
```diff
@@ -8246,6 +8256,12 @@ static bool ValidateAttachmentImageUsage(layer_data *dev_data, const VkFramebuff
                                         VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT);
             }
         }
+    } else {
+        skip_call |=
```
chrisforbes: Unreachable, see above.

layers/core_validation.cpp
```diff
@@ -8228,11 +8228,21 @@ static bool MatchUsage(layer_data *dev_data, uint32_t count, const VkAttachmentR
...
 
-static bool ValidateAttachmentImageUsage(layer_data *dev_data, const VkFramebufferCreateInfo *pCreateInfo) {
+// Validate VkFramebufferCreateInfo which includes:
+// 1. attachmentCount equals renderPass attachmentCount
+static bool ValidateFramebufferCreateInfo(layer_data *dev_data, const VkFramebufferCreateInfo *pCreateInfo) {
     bool skip_call = false;
 
     const VkRenderPassCreateInfo *rpci = getRenderPass(dev_data, pCreateInfo->renderPass)->pCreateInfo;
     if (rpci != nullptr) {
+        if (rpci->attachmentCount != pCreateInfo->attachmentCount) {
```
chrisforbes: :+1:

tests/layer_validation_tests.cpp
```diff
@@ -3267,11 +3269,33 @@ TEST_F(VkLayerTest, AttachmentUsageMismatch) {
...
+    subpass.pDepthStencilAttachment = &attach;
+    subpass.pColorAttachments = NULL;
+    VkRenderPass rp_ds;
+    err = vkCreateRenderPass(m_device->device(), &rpci, NULL, &rp_ds);
+    ASSERT_VK_SUCCESS(err);
+    // Set correct attachment count, but attachment as COLOR usage bit set
+    fb_info.attachmentCount = 1;
+    fb_info.renderPass = rp_ds;
+
+    m_errorMonitor->SetDesiredFailureMsg(
```
chrisforbes: General grumbling about run-on tests.

layers/core_validation.cpp
```diff
@@ -8263,11 +8263,17 @@ static bool ValidateFramebufferCreateInfo(layer_data *dev_data, const VkFramebuf
                         i, string_VkFormat(ivci->format), string_VkFormat(rpci->pAttachments[i].format),
                         reinterpret_cast<const uint64_t &>(pCreateInfo->renderPass));
                 }
-#if 0 // Enabling 1 new check/test at a time
                 const VkImageCreateInfo *ici = &getImageNode(dev_data, ivci->image)->createInfo;
```
chrisforbes: ref to const here, and lose the &?

layers/core_validation.cpp
```diff
@@ -9971,6 +9977,7 @@ GetSwapchainImagesKHR(VkDevice device, VkSwapchainKHR swapchain, uint32_t *pCoun
             image_ci.arrayLayers = swapchain_node->createInfo.imageArrayLayers;
             image_ci.usage = swapchain_node->createInfo.imageUsage;
             image_ci.format = swapchain_node->createInfo.imageFormat;
+            image_ci.samples = VK_SAMPLE_COUNT_1_BIT;
```
chrisforbes: :+1:

tests/layer_validation_tests.cpp
```diff
@@ -3317,6 +3319,26 @@ TEST_F(VkLayerTest, FramebufferCreateErrors) {
     if (err == VK_SUCCESS) {
         vkDestroyFramebuffer(m_device->device(), fb, NULL);
     }
+    vkDestroyRenderPass(m_device->device(), rp, NULL);
+
+    // Create new renderpass with alternate sample count from fb
+    attach_desc.format = VK_FORMAT_B8G8R8A8_UNORM;
+    attach_desc.samples = VK_SAMPLE_COUNT_2_BIT;
```
chrisforbes: Probably a better idea to use 4 samples (is required to be supported) unless you check for 2 and skip if not.

tests/layer_validation_tests.cpp
```diff
@@ -3341,6 +3342,45 @@ TEST_F(VkLayerTest, FramebufferCreateErrors) {
...
+    ivci.subresourceRange.levelCount = 2;
+    ivci.subresourceRange.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
+    err = vkCreateImageView(m_device->device(), &ivci, NULL, &view);
+    ASSERT_VK_SUCCESS(err);
+    // Re-create renderpass to have matching sample count
+    attach_desc.samples = VK_SAMPLE_COUNT_1_BIT;
+    err = vkCreateRenderPass(m_device->device(), &rpci, NULL, &rp);
+    ASSERT_VK_SUCCESS(err);
+
+    // Cause error due to mis-matched sample count between rp & fb
```
chrisforbes: This comment is bogus

layers/core_validation_error_enums.h
```diff
@@ -131,6 +131,7 @@ enum DRAW_STATE_ERROR {
     DRAWSTATE_FRAMEBUFFER_INCOMPATIBLE,         // Incompatible framebuffer between
                                                 // secondary cmdBuffer and active
                                                 // renderPass
+    DRAWSTATE_INVALID_FRAMEBUFFER_CREATE_INFO,  // Invalid VkFramebufferCreateInfo state
```
chrisforbes: Doc for this is in the next commit rather than here?

layers/core_validation.cpp
```diff
@@ -8282,17 +8282,31 @@ static bool ValidateFramebufferCreateInfo(layer_data *dev_data, const VkFramebuf
                               "but only a single mip level (levelCount ==  1) is allowed when creating a Framebuffer.",
                         i, ivci->subresourceRange.levelCount);
                 }
-#if 0 // Enabling 1 new check/test at a time
                 const uint32_t mip_level = ivci->subresourceRange.baseMipLevel;
-                if ((ivci->subresourceRange.layerCount < pCreateInfo->layers) ||
-                    ((ici->extent.width >> mip_level) < pCreateInfo->width) ||
-                    ((ici->extent.height >> mip_level) < pCreateInfo->height)) {
-                    // TODO
+                uint32_t mip_width = (ici->extent.width >> mip_level);
```
chrisforbes: This should be `max(1, ici->extent.width >> mip_level);`; similarly for height.

layers/core_validation.cpp
chrisforbes: Scratch that, it doesn't quite work. Want C++17's try_emplace (but we don't have that) or proper move support on the safe_* types (which is a pile of work). Might be best to just keep this as you've done for now.

layers/core_validation.cpp
```diff
@@ -8267,15 +8266,10 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateFramebuffer(VkDevice device, const VkFrameb
         // Shadow create info and store in map
         std::lock_guard<std::mutex> lock(global_lock);
 
+        dev_data->frameBufferMap.insert(
+            std::make_pair(*pFramebuffer, unique_ptr<FRAMEBUFFER_NODE>(new FRAMEBUFFER_NODE(
+                                              pCreateInfo, dev_data->renderPassMap[pCreateInfo->renderPass]->pCreateInfo))));
```
chrisforbes: `dev_data->renderPassMap[pCreateInfo->renderPass]->pCreateInfo` should be `getRenderPass(dev_data, pCreateInfo->renderPass)->pCreateInfo`

tests/layer_validation_tests.cpp
tobine: yep

tests/layer_validation_tests.cpp
tobine: yep

layers/core_validation_error_enums.h
tobine: updated commit

layers/core_validation.cpp
tobine: yep

layers/core_validation.cpp
tobine: unique_ptr is the future, so by using them here we're ushering in the future. Don't fight the future...

Actually, I'll shift this code to emplace and kill unique_ptr as you recommend below. That's cleaner. Maybe emplace is the future and I was wrong about unique_ptr all along?

layers/core_validation.cpp
tobine: Yeah I'll tweak this

layers/core_validation.cpp
tobine: Or maybe I was right about unique_ptr all along?

Yes, now I recall. Without proper move on safe_* types I resorted to the unique_ptr to use ctor followed by item look-up. Not as clean as try_emplace but I'll keep for now and add comment for future improvement.

tests/layer_validation_tests.cpp
tobine: To me the run-ons in this case are a reasonable trade-off. The test cases are all related so this makes good re-use of existing code. We can have a discussion about the best way to balance this trade-off in general. My preference is not to have a separate test for every individual error check.

tobine:
Created #683 which addresses review feedback and finishes up adding valid usage checks and tests.

layers/core_validation.cpp
lentinem: Just to comment on this. If possible, using NODE as the type in the maps is proffered but it is not always possible due to too much copying around (basically objects that need to live in multiple locations). In this case we should use unique_ptr<NODE>. Either of these departs from NODE* which had been used often in the past and lead to memory leaks due to improper freeing of memory.

PR #685, opened 2016-06-23
tests: vkUpdateDescriptorSet parameter validation

dustin-lunarg:
Add a positive test to ensure that the parameter_validation layer ignores
vkUpdateDescriptorSet parameters that are not related to the type of
descriptor set being updated.

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -1016,6 +1016,404 @@ TEST_F(VkLayerTest, FillBufferAlignment) {
 
     EndCommandBuffer();
 }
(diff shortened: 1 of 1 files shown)


mark-lunarg:
:+1:

tobine:
This is a good case and useful beyond just parameter_checker.
Could be enhanced to include *TEXEL_BUFFER and *BUFFER updates that validate that verify that other struct members are properly ignored.

dustin-lunarg:
@tobine I have updated this test to include TEXEL_BUFFER and BUFFER updates.

Issue #376, opened 2016-04-16
Validate that image/buffer memory is bound before creating views


baldurk:
Currently on master there doesn't seem to be any validation that memory is bound to a buffer or image before trying to create views - but the spec requires that this in section 11.6.

I haven't checked if the other requirements are validated (memory bound before use in command buffers is maybe already validated, but not sure if memory bind before use in descriptor sets is checked).

DominikWitczakAMD:
+1. I can recall at least one application in the wild where this problem took us several hours to hunt down.

PR #686, opened 2016-06-24
layers: Verify memory present on vkCreateXXXView

MarkY-LunarG:
Verify that a vkImage has memory bound to it when
vkCreateImageView is called.  Likewise, verify vkBuffer
memory at vkCreateBufferView time.

Change-Id: If0cf1c972f721e5596f7b8a1c7652f6165401b2c


MarkY-LunarG:
Fix for GH Issue #376

MarkY-LunarG:
Fixed commit message so it includes GH #.  Don't know how I didn't catch that before.

layers/core_validation.cpp
```diff
@@ -5552,7 +5552,29 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateBuffer(VkDevice device, const VkBufferCreat
...
+            get_mem_binding_from_object(dev_data, (uint64_t)pCreateInfo->buffer, VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT, &mem);
+        DEVICE_MEM_INFO *pMemObj = getMemObjInfo(dev_data, mem);
+        if (!pMemObj) {
+            skip_call |=
+                log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_MEMORY_EXT,
+                        (uint64_t)(mem), __LINE__, MEMTRACK_INVALID_USAGE_FLAG, "MEM",
+                        "vkCreateBufferView called with invalid memory 0x%" PRIx64 " bound to buffer 0x%" PRIx64 ".  Memory "
+                        "must be bound prior to creating a view to the buffer.",
+                        (uint64_t)(mem), (uint64_t)pCreateInfo->buffer);
+        }
```
tobine: Avoid c-style casts. Use `reinterpret_cast<unit64_t &>()` instead.

layers/core_validation.cpp
```diff
@@ -5624,19 +5646,33 @@ static void ResolveRemainingLevelsLayers(layer_data *dev_data, uint32_t *levels,
...
+        if (mem != MEMTRACKER_SWAP_CHAIN_IMAGE_KEY) {
+            DEVICE_MEM_INFO *pMemObj = getMemObjInfo(dev_data, mem);
+            if (!pMemObj) {
+                skip_call |=
+                    log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_MEMORY_EXT,
+                            (uint64_t)(mem), __LINE__, MEMTRACK_INVALID_USAGE_FLAG, "MEM",
+                            "vkCreateImageView called with invalid memory 0x%" PRIx64 " bound to image 0x%" PRIx64 ".  Memory "
+                            "must be bound prior to creating a view to the image.",
+                            (uint64_t)(mem), (uint64_t)pCreateInfo->image);
+            }
```
tobine: Avoid c-style casts. Use `reinterpret_cast<unit64_t &>()` instead.

layers/core_validation.cpp
```diff
@@ -5552,7 +5552,29 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateBuffer(VkDevice device, const VkBufferCreat
 VKAPI_ATTR VkResult VKAPI_CALL CreateBufferView(VkDevice device, const VkBufferViewCreateInfo *pCreateInfo,
                                                 const VkAllocationCallbacks *pAllocator, VkBufferView *pView) {
     layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
-    VkResult result = dev_data->device_dispatch_table->CreateBufferView(device, pCreateInfo, pAllocator, pView);
+    bool skip_call = false;
+    VkResult result = VK_ERROR_VALIDATION_FAILED_EXT;
```
tobine: You can kill this line if you change early exit below to be:

```
if (skip_call)
    return VK_ERROR_VALIDATION_FAILED_EXT;
```

layers/core_validation.cpp
```diff
@@ -5552,7 +5552,29 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateBuffer(VkDevice device, const VkBufferCreat
...
+                        "vkCreateBufferView called with invalid memory 0x%" PRIx64 " bound to buffer 0x%" PRIx64 ".  Memory "
+                        "must be bound prior to creating a view to the buffer.",
+                        (uint64_t)(mem), (uint64_t)pCreateInfo->buffer);
+        }
+    }
+
+    if (!skip_call) {
+        result = dev_data->device_dispatch_table->CreateBufferView(device, pCreateInfo, pAllocator, pView);
+    }
+
```
tobine: See note above about potential modification here.

layers/core_validation.cpp
```diff
@@ -5624,19 +5646,33 @@ static void ResolveRemainingLevelsLayers(layer_data *dev_data, uint32_t *levels,
 
 VKAPI_ATTR VkResult VKAPI_CALL CreateImageView(VkDevice device, const VkImageViewCreateInfo *pCreateInfo,
                                                const VkAllocationCallbacks *pAllocator, VkImageView *pView) {
-    bool skipCall = false;
+    bool skip_call = false;
     VkResult result = VK_ERROR_VALIDATION_FAILED_EXT;
```
tobine: You can kill this line if you change early exit below to be:

```
if (skip_call)
    return VK_ERROR_VALIDATION_FAILED_EXT;
```

layers/core_validation.cpp
```diff
@@ -5624,19 +5646,33 @@ static void ResolveRemainingLevelsLayers(layer_data *dev_data, uint32_t *levels,
...
+                            "must be bound prior to creating a view to the image.",
+                            (uint64_t)(mem), (uint64_t)pCreateInfo->image);
+            }
+        }
     }
 
-    if (!skipCall) {
+    if (!skip_call) {
         result = dev_data->device_dispatch_table->CreateImageView(device, pCreateInfo, pAllocator, pView);
     }
```
tobine: See note above about potential modification here

tobine:
The logic for the checks look good.
As a general architectural practice, if we're adding new checks like this I'd like to see them spun off into a PreCallValidate*() function as outlined in the layer authoring guide spearheaded by @mark-lunarg .

Also, there are no tests for these new checks. With tests and a bit of clean-up this should be ok, but for now 👎

layers/core_validation.cpp
mark-lunarg: I know it's a style thing, but some folks prefer a single point of return.

tobine: Early return avoids risk of any unwanted side effects later in the function. Currently for these two cases the code is guarded so that no such side effects presently exist, but I could see code being added in the future that is not guarded by "VK_SUCCESS" return result but that would be undesirable in the event of a validation error that set `skip_call = true`.

MarkY-LunarG: You could make the same argument about unwanted side effects on early returns as well.  In fact, I'm still in the middle of cleaning some up in the loader where we didn't properly free items.  I think either way, it's an exercise for the developer (and reviewers) to catch such cases.  Personally, I'm more in favor of the single point of return as Mark is due to the ease in debugging issues and cleanliness of code.

layers/core_validation.cpp
```diff
@@ -5549,18 +5549,43 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateBuffer(VkDevice device, const VkBufferCreat
...
+                             reinterpret_cast<uint64_t &>(mem), __LINE__, MEMTRACK_INVALID_USAGE_FLAG, "MEM",
+                             "vkCreateBufferView called with invalid memory 0x%" PRIx64 " bound to buffer 0x%" PRIx64 ".  Memory "
+                             "must be bound prior to creating a view to the buffer.",
+                             reinterpret_cast<uint64_t &>(mem), reinterpret_cast<const uint64_t &>(pCreateInfo->buffer));
+    }
+    return skip_call;
+}
+
+static void PostCallValidateCreateBufferView(layer_data *dev_data, const VkBufferViewCreateInfo *pCreateInfo, VkBufferView *pView) {
+    dev_data->bufferViewMap[*pView] = unique_ptr<VkBufferViewCreateInfo>(new VkBufferViewCreateInfo(*pCreateInfo));
```
tobine: This line can be inline in CreateBufferView() following successful bufferview creation.

layers/core_validation.cpp
```diff
@@ -5549,18 +5549,43 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateBuffer(VkDevice device, const VkBufferCreat
...
+    // In order to create a valid buffer view, the buffer must have been created with at least one of the
+    // following flags:  UNIFORM_TEXEL_BUFFER_BIT or STORAGE_TEXEL_BUFFER_BIT
+    validate_buffer_usage_flags(dev_data, pCreateInfo->buffer,
+                                VK_BUFFER_USAGE_UNIFORM_TEXEL_BUFFER_BIT | VK_BUFFER_USAGE_STORAGE_TEXEL_BUFFER_BIT, false,
+                                "vkCreateBufferView()", "VK_BUFFER_USAGE_[STORAGE|UNIFORM]_TEXEL_BUFFER_BIT");
+}
```
tobine: This code should precede the call down the chain, and can be included in the PreCallValidate* function and this PostCallValidate* function can be removed.

layers/core_validation.cpp
```diff
@@ -5622,28 +5647,46 @@ static void ResolveRemainingLevelsLayers(layer_data *dev_data, uint32_t *levels,
...
+                        reinterpret_cast<uint64_t &>(mem), reinterpret_cast<const uint64_t &>(pCreateInfo->image));
+        }
     }
+    return skip_call;
+}
 
-    if (!skipCall) {
-        result = dev_data->device_dispatch_table->CreateImageView(device, pCreateInfo, pAllocator, pView);
-    }
+static void PostCallValidateCreateImageView(layer_data *dev_data, const VkImageViewCreateInfo *pCreateInfo, VkImageView *pView) {
```
tobine: I've been using "PostCallRecord*" for function naming of post-creation state updates. "PostCallValidate*" is only appropriate in rare cases where validation is required following a call down the chain.

tobine:
This looks nice. Just a couple minor naming and organizational points and validation code will be good. Still need test cases for these checks, though, along with any associated documentation updates for tests.

chrisforbes:
This is only correct for non-sparse resources.

layers/core_validation.cpp
MarkY-LunarG: This functionality was previously after the call-down.  The note indicates it's validating the returned flags.  Should this really be part of the PreCall?

layers/core_validation.cpp
MarkY-LunarG: Just to be clear (since GitHub isn't indicating a precise line) you mean line 5568?

MarkY-LunarG:
Good catch Chris.

layers/core_validation.cpp
tobine: Yes, line 5568. With this code moved in-line and the validation code moved to the pre-call, this "PostCallValidate*" function can be destroyed.

layers/core_validation.cpp
tobine: Yes, this should be part of PreCall. The check concerns how the buffer was created, but doesn't depend on the bufferView being created, and should precede bufferView creation.

layers/core_validation.cpp
```diff
@@ -5530,18 +5530,52 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateBuffer(VkDevice device, const VkBufferCreat
...
 }
 
+static bool PreCallValidateCreateBufferView(layer_data *dev_data, const VkBufferViewCreateInfo *pCreateInfo) {
+    bool skip_call = false;
+    BUFFER_NODE *buf_node = getBufferNode(dev_data, pCreateInfo->buffer);
+    // If this isn't a sparse buffer, it needs to have memory backing it at CreateBufferView time
+    if (0 == (static_cast<uint32_t>(buf_node->createInfo.flags) & VK_BUFFER_CREATE_SPARSE_BINDING_BIT)) {
+        VkDeviceMemory *pMemBinding = get_object_mem_binding(dev_data, reinterpret_cast<const uint64_t &>(pCreateInfo->buffer),
+                                                             VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT);
+        if (pMemBinding) {
```
tobine: Kill this call, I'm about to remove this function in separate PR.

tobine: Use "if (buf_node->mem)" instead

layers/core_validation.cpp
```diff
@@ -5530,18 +5530,52 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateBuffer(VkDevice device, const VkBufferCreat
...
+            DEVICE_MEM_INFO *pMemObjInfo = getMemObjInfo(dev_data, *pMemBinding);
```
tobine: use buf_node->mem instead of *pMemBinding

layers/core_validation.cpp
```diff
@@ -5603,28 +5637,58 @@ static void ResolveRemainingLevelsLayers(layer_data *dev_data, uint32_t *levels,
...
+                                            VK_IMAGE_USAGE_SAMPLED_BIT | VK_IMAGE_USAGE_STORAGE_BIT |
+                                                VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT | VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT,
+                                            false, "vkCreateImageView()", "VK_IMAGE_USAGE_[SAMPLED|STORAGE|COLOR_ATTACHMENT]_BIT");
+    IMAGE_NODE *image_node = getImageNode(dev_data, pCreateInfo->image);
+    // If this isn't a sparse image, it needs to have memory backing it at CreateImageView time
+    if (0 == (static_cast<uint32_t>(image_node->createInfo.flags) & VK_IMAGE_CREATE_SPARSE_BINDING_BIT)) {
+        VkDeviceMemory *pMemBinding = get_object_mem_binding(dev_data, reinterpret_cast<const uint64_t &>(pCreateInfo->image),
+                                                             VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT);
+        if (MEMTRACKER_SWAP_CHAIN_IMAGE_KEY != *pMemBinding) {
+            if (pMemBinding) {
```
tobine: Same thing here as above. Remove call to get_object_mem_binding() and replace *pMemBinding w/ image_node->mem.

tests/layer_validation_tests.cpp
```diff
@@ -5461,6 +5461,43 @@ TEST_F(VkLayerTest, InvalidBufferViewObject) {
     vkDestroyDescriptorPool(m_device->device(), ds_pool, NULL);
 }
 
+TEST_F(VkLayerTest, CreateBufferViewNoMemoryBoundToBuffer) {
+    TEST_DESCRIPTION("Attempt to create a buffer view with a buffer that has"
```
tobine: missing space at end of line to "hasno" will run together.

tests/layer_validation_tests.cpp
```diff
@@ -5461,6 +5461,43 @@ TEST_F(VkLayerTest, InvalidBufferViewObject) {
...
+                     "no memory bound to it.");
+
+    VkResult err;
+    m_errorMonitor->SetDesiredFailureMsg(
+        VK_DEBUG_REPORT_ERROR_BIT_EXT,
+        "vkCreateBufferView called with invalid memory ");
+
+    ASSERT_NO_FATAL_FAILURE(InitState());
+
+    // Create a buffer & bufferView to be used for invalid updates
```
tobine: invalid comments

tests/layer_validation_tests.cpp
```diff
@@ -5461,6 +5461,43 @@ TEST_F(VkLayerTest, InvalidBufferViewObject) {
...
+    VkBufferCreateInfo buff_ci = {};
+    buff_ci.sType = VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO;
+    // This usage is not valid for any descriptor type
```
tobine: invalid comment and usage bit is serving no purpose for this test

tests/layer_validation_tests.cpp
```diff
@@ -5461,6 +5461,43 @@ TEST_F(VkLayerTest, InvalidBufferViewObject) {
...
+
+    VkBufferViewCreateInfo buff_view_ci = {};
+    buff_view_ci.sType = VK_STRUCTURE_TYPE_BUFFER_VIEW_CREATE_INFO;
+    buff_view_ci.buffer = buffer;
+    buff_view_ci.format = VK_FORMAT_R8_UNORM;
+    buff_view_ci.range = VK_WHOLE_SIZE;
+    VkBufferView buff_view;
+    err =
+        vkCreateBufferView(m_device->device(), &buff_view_ci, NULL, &buff_view);
+    ASSERT_VK_SUCCESS(err);
```
tobine: Do we expect this call to succeed?

tests/layer_validation_tests.cpp
```diff
@@ -5461,6 +5461,43 @@ TEST_F(VkLayerTest, InvalidBufferViewObject) {
...
+
+    m_errorMonitor->VerifyFound();
+    vkDestroyBuffer(m_device->device(), buffer, NULL);
+    vkDestroyBufferView(m_device->device(), buff_view, NULL);
```
tobine: If view creation doesn't succeed no need to destroy it

tobine:
This will have some conflicts with my #710, which drove my remove "get_object_mem_binding()" comments. A couple other conflicts too in usage bit validation, but if my changes land first (I'm still cleaning them up) I'm happy to rebase and push your changes.

It looks like one more iteration and you'll be good to go.

MarkY-LunarG:
I'll remove the get_object_mem_binding calls as I have other things to clean up.  Then I'll push it up again after I've run it through the testing rig.

tests/layer_validation_tests.cpp
MarkY-LunarG: Is this a stale comment?  " no" has the space in front of it now.

tobine:

PR #687, opened 2016-06-24
layers: GH618, Correct WSI exension flag usage

mark-lunarg:
Object_tracker had no notion of individual WSI extension flags,
and unique_objects was not checking them.

Change-Id: I93c1aa0c324aa602717f36e2975120dba8bc364e

### DIFF
--- a/layers/object_tracker.h
+++ b/layers/object_tracker.h
@@ -89,11 +89,17 @@ struct layer_data {
           tmp_dbg_create_infos(nullptr), tmp_callbacks(nullptr){};
 };
 
-struct instExts {
+struct instance_extension_enables {
     bool wsi_enabled;
+    bool xlib_enabled;
+    bool xcb_enabled;
+    bool wayland_enabled;
+    bool mir_enabled;
+    bool android_enabled;
+    bool win32_enabled;
 };
 
-static std::unordered_map<void *, struct instExts> instanceExtMap;
+static std::unordered_map<void *, struct instance_extension_enables> instanceExtMap;
 static std::unordered_map<void *, layer_data *> layer_data_map;
 static device_table_map object_tracker_device_table_map;
 static instance_table_map object_tracker_instance_table_map;
@@ -135,7 +141,6 @@ static void createDeviceRegisterExtensions(const VkDeviceCreateInfo *pCreateInfo
 }
 
 static void createInstanceRegisterExtensions(const VkInstanceCreateInfo *pCreateInfo, VkInstance instance) {
-    uint32_t i;
     VkLayerInstanceDispatchTable *pDisp = get_dispatch_table(object_tracker_instance_table_map, instance);
     PFN_vkGetInstanceProcAddr gpa = pDisp->GetInstanceProcAddr;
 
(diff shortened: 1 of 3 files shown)


layers/object_tracker.h
```diff
@@ -178,11 +183,49 @@ static void createInstanceRegisterExtensions(const VkInstanceCreateInfo *pCreate
...
+    instanceExtMap[pDisp] = {};
+
+    for (uint32_t i = 0; i < pCreateInfo->enabledExtensionCount; i++) {
+        if (strcmp(pCreateInfo->ppEnabledExtensionNames[i], VK_KHR_SURFACE_EXTENSION_NAME) == 0) {
             instanceExtMap[pDisp].wsi_enabled = true;
+        }
+#ifdef VK_USE_PLATFORM_XLIB_KHR
+        if (strcmp(pCreateInfo->ppEnabledExtensionNames[i], VK_KHR_XLIB_SURFACE_EXTENSION_NAME) == 0) {
+            instanceExtMap[pDisp].xlib_enabled = true;
+        }
```
ianelliottus: Hi Mark, perhaps I'm not looking carefully enough, and I see that it follows the pattern done in "swapchain.cpp".  This is coming after some lines above that set pDisp->Create*SurfaceKHR pointers.  Should those only work if the extension is enabled?  I haven't fully context-switched my brain back to this topic, and so I may not be thinking of this correctly.

jon-lunarg:
The initialization of the layer's extension dispatch table pointers could be done conditionally as Ian points out.  However, this pattern that Mark is adding here has been done repeatedly in the layers and it should function just fine.  The various wsi  enabled  flags are checked int he layer's Get*ProcAddr function instead of at CreateDevice  and CreateInstance.  This checking of the flags in GIPA/GDPA is the critical piece for the WSI extension intercept to work properly (ie only intercept if the extension is enabled).

layers/object_tracker.h
```diff
@@ -178,11 +183,49 @@ static void createInstanceRegisterExtensions(const VkInstanceCreateInfo *pCreate
...
+#ifdef VK_USE_PLATFORM_XCB_KHR
+        if (strcmp(pCreateInfo->ppEnabledExtensionNames[i], VK_KHR_XCB_SURFACE_EXTENSION_NAME) == 0) {
+            instanceExtMap[pDisp].xcb_enabled = true;
+        }
+#endif
+#ifdef VK_USE_PLATFORM_XCB_KHR
+        if (strcmp(pCreateInfo->ppEnabledExtensionNames[i], VK_KHR_XCB_SURFACE_EXTENSION_NAME) == 0) {
+            instanceExtMap[pDisp].xcb_enabled = true;
+        }
+#endif
```
jon-lunarg: lines 202-206 duplicate lines 197-201.  One set could be removed I think.

jon-lunarg:
Other than the one comment about code duplication, LGTM

mark-lunarg:
Cherry-picked to Master.  


commit 61d505ea40b189b7ae63f311e2f13210eef219d2
Date:   Fri Jun 24 09:57:32 2016 -0600'

    layers: GH618, Correct WSI exension flag usage

    Object_tracker had no notion of individual WSI extension flags,
    and unique_objects was not checking them.

Issue #91, opened 2016-03-08
Unify validation INVALID cmd buffer cases


tobine:
Currently cmd buffers can be made invalid by updating or freeing descriptor sets that they depend on or if a framebuffer they reference is deleted.
Really, anytime that an object that a cmd buffer depends on is destroyed, the cmd buffer should be made invalid or, if it's already in-flight then an error should be flagged at destroy time.
Then when/if an INVALID cmd buffer is reported, an error should be flagged with a nice message about why the cmd buffer is invalid.

tobine:
The complete list of objects that will invalidate a cmd buffer that references them when deleted, from the spec "Object Lifetime" section:

VkEvent
VkQueryPool
VkBuffer
VkBufferView
VkImage
VkImageView
VkPipeline
VkSampler
VkDescriptorPool
VkFramebuffer
VkRenderPass
VkCommandPool
VkDeviceMemory
VkDescriptorSet

tobine:
We handle CB Invalidation for many of these cases. Using this issue to track exact status. There are also indirect objects that need to be tracked such as images bound to framebuffer. I'll call these "children" for now. Currently we handle all CB invalid cases in original list except:
VkBufferView (and children)
VkImageView (and children)
Children of VkPipeline
VkDescriptorPool
Children of VkFramebuffer
VkRenderpass (and children)
VkCommandPool (and children) - definitely some handling of this but need to verify code
VkDeviceMemory
Children of VkDescriptorSet - Samplers are handled but need to make sure all related image/buffer/views/memory get bound and cause invalidation

Also I'm going to use this issue to track in-use errors which are related but slightly different as in-use is flagged for objects destroyed while bound to an in-flight CB and CB invalid is when object is destroyed for non in-flight CB.

Need to verify in-use errors for:
All of the cases above
VkBuffer
VkImage

tobine:
Once #1088 lands I believe that the only cases remaining from this issue are making sure that children of VkPipeline (renderpass) are bound and make sure that command pool is correctly bound.

tobine:
Fix was cherry-picked and pushed to master

PR #688, opened 2016-06-24
Tobin gh91 invalid cb cases

tobine:
This is mostly refactoring and clean-up in preparation for greatly expanding command buffer invalidation handling. No new checks are added in this series. Next task will be iterating through new object types (such as images and buffers), making sure the dependencies are tracked and that cmd buffer invalidation  due to those dependencies is correctly handled.


layers/core_validation_types.h
```diff
@@ -102,11 +108,18 @@ struct DESCRIPTOR_POOL_NODE {
 class BUFFER_NODE : public BASE_NODE {
   public:
     using BASE_NODE::in_use;
+    using BASE_NODE::cb_bindings;
```
chrisforbes: Why?

layers/core_validation_types.h
```diff
@@ -102,11 +108,18 @@ struct DESCRIPTOR_POOL_NODE {
...
-    BUFFER_NODE() : mem(VK_NULL_HANDLE), createInfo{} { in_use.store(0); };
-    BUFFER_NODE(const VkBufferCreateInfo *pCreateInfo) : mem(VK_NULL_HANDLE), createInfo(*pCreateInfo) { in_use.store(0); };
-    BUFFER_NODE(const BUFFER_NODE &rh_obj) : mem(rh_obj.mem), createInfo(rh_obj.createInfo) { in_use.store(rh_obj.in_use.load()); };
+    BUFFER_NODE() : buffer(VK_NULL_HANDLE), mem(VK_NULL_HANDLE), createInfo{} { in_use.store(0); };
+    BUFFER_NODE(VkBuffer buff, const VkBufferCreateInfo *pCreateInfo)
+        : buffer(buff), mem(VK_NULL_HANDLE), createInfo(*pCreateInfo) {
+        in_use.store(0);
+    };
+    BUFFER_NODE(const BUFFER_NODE &rh_obj) : buffer(rh_obj.buffer), mem(rh_obj.mem), createInfo(rh_obj.createInfo) {
+        in_use.store(rh_obj.in_use.load());
```
chrisforbes: Copying (effectively) reference counts doesn't make a lot of sense. Do you mean to make this moveable instead?

layers/core_validation.cpp
```diff
@@ -8079,14 +8079,14 @@ CmdCopyQueryPoolResults(VkCommandBuffer commandBuffer, VkQueryPool queryPool, ui
     GLOBAL_CB_NODE *pCB = getCBNode(dev_data, commandBuffer);
 #if MTMERGESOURCE
     VkDeviceMemory mem;
-    auto cb_data = dev_data->commandBufferMap.find(commandBuffer);
+    auto cb_node = getCBNode(dev_data, commandBuffer);
```
chrisforbes: Just did this a few lines prior

layers/core_validation.cpp
```diff
@@ -5367,10 +5367,10 @@ FreeCommandBuffers(VkDevice device, VkCommandPool commandPool, uint32_t commandB
     std::unique_lock<std::mutex> lock(global_lock);
 
     for (uint32_t i = 0; i < commandBufferCount; i++) {
-        auto cb_pair = dev_data->commandBufferMap.find(pCommandBuffers[i]);
+        auto cb_node = getCBNode(dev_data, pCommandBuffers[i]);
```
chrisforbes: Only reason I hadn't done this already is that getCBNode is a bit weird -- it produces validation errors.

layers/core_validation.cpp
```diff
@@ -5522,17 +5484,22 @@ VKAPI_ATTR VkResult VKAPI_CALL ResetFences(VkDevice device, uint32_t fenceCount,
...
     std::unique_lock<std::mutex> lock(global_lock);
     auto fb_node = getFramebuffer(dev_data, framebuffer);
     if (fb_node) {
-        for (auto cb_node : fb_node->cb_bindings) {
-            // Set CB as invalid and record destroyed framebuffer
-            cb_node->state = CB_INVALID;
-            cb_node->destroyedFramebuffers.insert(framebuffer);
-        }
+        invalidateCommandBuffers(fb_node->cb_bindings,
+                                 {static_cast<uint64_t>(fb_node->framebuffer), VK_DEBUG_REPORT_OBJECT_TYPE_FRAMEBUFFER_EXT});
```
chrisforbes: Is this cast going to do the right thing on all targets? If so, why are we using `reinterpret_cast<uint64_t [const] &>(ndhandle)` everywhere else?

layers/core_validation_types.h
tobine: At least 10 of the objects listed in #91 will need tracking for command buffer invalidation so adding to BASE_NODE and will make state structs for each of those objects inherit from BASE_NODE to standardize handling of CB invalidation.

layers/core_validation.cpp
tobine: No, needs to be reinterpret_cast. Not sure why I even put in static_cast, but I happened to only be testing on arm7 when I did and it happens to work there.

layers/core_validation_types.h
tobine: I agree. I was thinking moveable when I copied the reference counts.

tobine:
I pushed some fixes and ran this on our test farm where all configs are passing.

layers/core_validation_types.h
chrisforbes: My point is there's no templates in play here to complicate name resolution -- you can refer to members from the base just fine without `using`.

tobine:
Killed the unneeded "using" line and pushed to master.

PR #696, opened 2016-06-27
tests: assert vkWaitForFences result

stroyan:
Assert that vkWaitForFences returns VK_SUCCESS in
TwoQueueSubmitsOneQueueNullQueueSubmitWithFence .
If that call fails then later calls make otherwise unexplained
validation callbacks.  An assert clears up the mystery.


tobine:
LGTM :+1:

stroyan:
merged and pushed

Issue #690, opened 2016-06-25
False positive when layer is checking pipeline compatibility against a deleted render pass


hrydgard:
Let's say I do the following:

* Create a renderpass object
* Create a pipeline, passing in that renderpass object
* Delete the renderpass object
* Create another render pass object, compatible with the first (identical)
* Draw using the pipeline

This currently triggers

    ERROR: [DS] Pipeline Code 43 : At Draw time the active render pass (0x52) is incompatible
    w/ gfx pipeline (0x23f733aebb0) that was created w/ render pass (0x10) due to:
    invalid VkRenderPass (0000000000000010)

Presumably the validation layer compat check wants to compare the current render pass against a no-longer-existing VkRenderPass object, while the driver will certainly not be doing such validation, there's no reason it should go back and try to read the renderpass the pipeline was created with, so it really should be safe.

As in my previous reports, this can be easily reproduced by building PPSSPP in debug mode and switching to the Vulkan backend, then debug from within VS.

Appears to be new in 17.

chrisforbes:
Hey @tobine, another variation on what you just fixed for framebuffers.

tobine:
Submitted #697 that should fix this issue

tobine:
Update PR #717 fixes this and various related issues

tobine:
Fix was cherry-picked and pushed to master

PR #697, opened 2016-06-27
Tobin gh690 handle deleted renderpass

tobine:
Fix for #690 
Store creation info for renderpass that Pipeline is created with.


chrisforbes:
Change looks good; do you have tests?

tobine:
This is fixing a bug with no new checks so I didn't add new tests. RenderPassIncompatible test hits the case that was re-factored. I'm making other changes to add new checks on the deletion of object used by command buffers which will include new tests. I'll make sure to have a deleted renderPass test in that set of changes.

chrisforbes:
I think this may still be broken in the other family of renderpass compatibility functions.

tobine:
Yes, you're correct.  There's a call to validateRenderPassCompatibility() that's still using renderPass from framebuffer createInfo instead of the renderPassCreateInfo that I added to FRAMEBUFFER_NODE. I'll fix that as well.

tobine:
Pushed a follow-on commit. Not pretty, but fixes lifetime issues for now. I'll look at further clean-up in follow-on PRs.

layers/core_validation.cpp
```diff
@@ -2246,24 +2236,53 @@ static bool verify_renderpass_compatibility(const layer_data *my_data, const VkR
...
+    if (!secondary_render_pass) {
+        stringstream errorStr;
+        errorStr << "invalid VkRenderPass (" << secondaryRP << ")";
+        errorMsg = errorStr.str();
+        return false;
+    }
+    return verify_renderpass_compatibility(dev_data, primaryRPCI, secondary_render_pass->pCreateInfo, errorMsg);
+}
+// For given primary RenderPass object and secondary RPCI, verify that they're compatible
+static bool verify_renderpass_compatibility(const layer_data *dev_data, const VkRenderPass primaryRP,
```
chrisforbes: Not really happy with growing a pile of overloads of this. Can't we just pull the differences up into the callers? There's ~1 per variant.

layers/core_validation.cpp
```diff
@@ -9387,11 +9402,72 @@ static bool validateSubpassCompatibility(layer_data *dev_data, VkCommandBuffer p
...
+        for (uint32_t i = 0; i < primaryPassCI->subpassCount; ++i) {
+            skip_call |= validateSubpassCompatibility(dev_data, primaryBuffer, primaryPassCI, secondaryBuffer, secondaryPassCI, i,
+                                                      primaryPassCI->subpassCount > 1);
+        }
+    }
+    return skip_call;
+}
+
+// Overloaded version that takes primary renderPass CreateInfo and secondary renderPass handle
+static bool validateRenderPassCompatibility(layer_data *dev_data, VkCommandBuffer primaryBuffer,
```
chrisforbes: Ditto, I think this is crazy.

tobine:
Slightly saner version in #717

PR #701, opened 2016-06-27
tests: assert CreatePipelineRelaxedTypeMatch success

stroyan:
Check that vkCreateGraphicsPipelines succeeds in
CreatePipelineRelaxedTypeMatch.
Some devices incorrectly fail to return success on this call.
Both the driver and the layers should return VK_SUCCESS.


tobine:
LGTM :+1:

stroyan:
merged and pushed.

PR #702, opened 2016-06-27
Rework cross-queue work tracking, remove globalInFlightCmdBuffers set.

chrisforbes:
This series swaps out the work-stealing approach to cross-queue work tracking in favor of having semaphore waits track the signaling point.

It's not perfect -- QBS is assumed to be ordered when it is not really, but in non-sparse cases this should be a good step forward.

Fixes (and reenables) two tests.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -4219,182 +4219,92 @@ static inline void removeInFlightCmdBuffer(layer_data *dev_data, VkCommandBuffer
     }
 }
 
-static void decrementResources(layer_data *my_data, CB_SUBMISSION *submission) {
-    for (auto cb : submission->cbs) {
-        auto pCB = getCBNode(my_data, cb);
-        for (auto drawDataElement : pCB->drawData) {
-            for (auto buffer : drawDataElement.buffers) {
-                auto buffer_node = getBufferNode(my_data, buffer);
-                if (buffer_node) {
-                    buffer_node->in_use.fetch_sub(1);
-                }
-            }
-        }
-        for (uint32_t i = 0; i < VK_PIPELINE_BIND_POINT_RANGE_SIZE; ++i) {
-            for (auto set : pCB->lastBound[i].uniqueBoundSets) {
-                set->in_use.fetch_sub(1);
-            }
-        }
-        for (auto event : pCB->events) {
-            auto eventNode = my_data->eventMap.find(event);
-            if (eventNode != my_data->eventMap.end()) {
-                eventNode->second.in_use.fetch_sub(1);
-            }
-        }
-        for (auto event : pCB->writeEventsBeforeWait) {
-            auto eventNode = my_data->eventMap.find(event);
-            if (eventNode != my_data->eventMap.end()) {
-                eventNode->second.write_in_use--;
-            }
-        }
-        for (auto queryStatePair : pCB->queryToStateMap) {
-            my_data->queryToStateMap[queryStatePair.first] = queryStatePair.second;
-        }
-        for (auto eventStagePair : pCB->eventToStageMap) {
-            my_data->eventMap[eventStagePair.first].stageMask = eventStagePair.second;
-        }
-    }
 
-    for (auto semaphore : submission->semaphores) {
-        auto pSemaphore = getSemaphoreNode(my_data, semaphore);
-        if (pSemaphore) {
+static bool RetireWorkOnQueue(layer_data *dev_data, QUEUE_NODE *pQueue, uint64_t seq)
+{
+    bool skip_call = false; // TODO: extract everything that might fail to precheck
+    std::unordered_map<VkQueue, uint64_t> otherQueueSeqs;
+
+    // Roll this queue forward, one submission at a time.
+    while (pQueue->seq < seq) {
+        auto & submission = pQueue->submissions.front();
+
+        for (auto & wait : submission.waitSemaphores) {
+            auto pSemaphore = getSemaphoreNode(dev_data, wait.semaphore);
+            pSemaphore->in_use.fetch_sub(1);
+            auto & lastSeq = otherQueueSeqs[wait.queue];
+            lastSeq = std::max(lastSeq, wait.seq);
+        }
+
+        for (auto & semaphore : submission.signalSemaphores) {
+            auto pSemaphore = getSemaphoreNode(dev_data, semaphore);
             pSemaphore->in_use.fetch_sub(1);
         }
-    }
-}
-// For fenceCount fences in pFences, mark fence signaled, decrement in_use, and call
-//  decrementResources for all priorFences and cmdBuffers associated with fence.
-static bool decrementResources(layer_data *my_data, uint32_t fenceCount, const VkFence *pFences) {
-    bool skip_call = false;
-    std::vector<std::pair<VkFence, FENCE_NODE *>> fence_pairs;
-    for (uint32_t i = 0; i < fenceCount; ++i) {
-        auto pFence = getFenceNode(my_data, pFences[i]);
-        if (!pFence || pFence->state != FENCE_INFLIGHT)
-            continue;
 
-        fence_pairs.emplace_back(pFences[i], pFence);
-        pFence->state = FENCE_RETIRED;
+        for (auto cb : submission.cbs) {
+            auto pCB = getCBNode(dev_data, cb);
+            for (auto drawDataElement : pCB->drawData) {
+                for (auto buffer : drawDataElement.buffers) {
+                    auto buffer_node = getBufferNode(dev_data, buffer);
+                    if (buffer_node) {
+                        buffer_node->in_use.fetch_sub(1);
+                    }
+                }
+            }
+            for (uint32_t i = 0; i < VK_PIPELINE_BIND_POINT_RANGE_SIZE; ++i) {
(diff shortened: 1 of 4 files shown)


layers/core_validation.cpp
```diff
@@ -4206,198 +4205,98 @@ static bool cleanInFlightCmdBuffer(layer_data *my_data, VkCommandBuffer cmdBuffe
...
-        }
-        for (auto eventStagePair : pCB->eventToStageMap) {
-            my_data->eventMap[eventStagePair.first].stageMask = eventStagePair.second;
-        }
-    }
 
-    for (auto semaphore : submission->semaphores) {
-        auto pSemaphore = getSemaphoreNode(my_data, semaphore);
-        if (pSemaphore) {
+static bool RetireWorkOnQueue(layer_data *dev_data, QUEUE_NODE *pQueue, uint64_t seq)
```
tobine: This is a great improvement.
Add function header comment please.

tobine: Thinking of other potential names than "RetireWorkOnQueue" that more fully captures what function does. Not obvious to me that queue seq# is advanced by this function.

Potential alternate function names:
AdvanceQueueToSeqenceNum
AdvanceQueueWorkToSeqenceNum
RetireWorkOnQueueToSeqenceNum

layers/core_validation.cpp
```diff
@@ -4838,7 +4749,12 @@ VKAPI_ATTR VkResult VKAPI_CALL GetFenceStatus(VkDevice device, VkFence fence) {
...
     lock.lock();
     if (result == VK_SUCCESS) {
-        skipCall |= decrementResources(dev_data, 1, &fence);
+        auto pFence = getFenceNode(dev_data, fence);
+        if (pFence->signaler.first != VK_NULL_HANDLE) {
+            skipCall |= RetireWorkOnQueue(dev_data,
+                                          getQueueNode(dev_data, pFence->signaler.first),
+                                          pFence->signaler.second);
+        }
     }
```
tobine: Potential for helper function for code here and same code in WaitForFences()

layers/core_validation.cpp
```diff
@@ -9915,10 +9825,30 @@ QueueBindSparse(VkQueue queue, uint32_t bindInfoCount, const VkBindSparseInfo *p
...
+                                         bindIdx == bindInfoCount - 1 ? fence : VK_NULL_HANDLE);
+    }
+
+    if (pFence && !bindInfoCount) {
+        // No work to do, just dropping a fence in the queue by itself.
+        pQueue->submissions.emplace_back(std::vector<VkCommandBuffer>(),
+                                         std::vector<SEMAPHORE_WAIT>(),
+                                         std::vector<VkSemaphore>(),
+                                         fence);
     }
```
tobine: Potential for helper function(s) for large block of code here and same code in QueueSubmit()

layers/core_validation.h
```diff
@@ -241,11 +240,11 @@ class EVENT_NODE : public BASE_NODE {
...
     VkQueue queue;
-    VkDevice device;
-    std::vector<VkFence> lastFences;
-    std::vector<CB_SUBMISSION> untrackedSubmissions;
     std::unordered_map<VkEvent, VkPipelineStageFlags> eventToStageMap;
     std::unordered_map<QueryObject, bool> queryToStateMap; // 0 is unavailable, 1 is available
+
+    uint64_t seq;
+    std::deque<CB_SUBMISSION> submissions;
 };
```
tobine: Comment here would be nice to summarize seq/submission use.

layers/core_validation_types.h
```diff
@@ -483,12 +483,20 @@ struct GLOBAL_CB_NODE : public BASE_NODE {
...
 };
 
+struct SEMAPHORE_WAIT {
+    VkSemaphore semaphore;
+    VkQueue queue;
+    uint64_t seq;
+};
+
 struct CB_SUBMISSION {
-    CB_SUBMISSION(std::vector<VkCommandBuffer> const &cbs, std::vector<VkSemaphore> const &semaphores)
```
tobine: Comment header to summarize use of these structs would be nice.

tobine:
Great changes. Kills lots of complicated and buggy code and makes it simpler and cleaner.
Primary suggestion is to add a few comments to aid future civilizations who analyze our advanced technology.

Do you have a PixelC to test on? I'm seeing failures on that device in LongFenceChain test.

lentinem:
I haven't gone through all of this yet but I would far prefer the use of another class similar to the descriptor sets if the logic is changing to this extent.

chrisforbes:
Can discuss how best to split this into a module, but much of the code doesn't naturally belong to a particular object.

chrisforbes:
@tobine LongFenceChain is a little bit mean. It used to cause stack overflows in the layers, and while that's fixed, it's still stressful on the layers (and the driver, particularly if it internally tracks more than it ought to)

layers/core_validation_types.h
```diff
@@ -483,12 +483,20 @@ struct GLOBAL_CB_NODE : public BASE_NODE {
     ~GLOBAL_CB_NODE();
 };
 
+struct SEMAPHORE_WAIT {
+    VkSemaphore semaphore;
+    VkQueue queue;
+    uint64_t seq;
```
lentinem: Can we use a more descriptive name than seq?

layers/core_validation.cpp
```diff
@@ -3745,9 +3744,9 @@ static void resetCB(layer_data *dev_data, const VkCommandBuffer cb) {
         pCB->drawData.clear();
         pCB->currentDrawData.buffers.clear();
         pCB->primaryCommandBuffer = VK_NULL_HANDLE;
-        // Make sure any secondaryCommandBuffers are removed from globalInFlight
         for (auto secondary_cb : pCB->secondaryCommandBuffers) {
-            dev_data->globalInFlightCmdBuffers.erase(secondary_cb);
+            auto pSubCB = getCBNode(dev_data, secondary_cb);
+            if (pSubCB) pSubCB->in_use.fetch_sub(1);
```
lentinem: Tobin may be able to commend on this but I forget why this is necessary. Shouldn't the counts/global be removed when the command buffer has completed not on reset?

layers/core_validation.cpp
```diff
@@ -4811,7 +4715,14 @@ WaitForFences(VkDevice device, uint32_t fenceCount, const VkFence *pFences, VkBo
         lock.lock();
         // When we know that all fences are complete we can clean/remove their CBs
         if (waitAll || fenceCount == 1) {
-            skip_call |= decrementResources(dev_data, fenceCount, pFences);
+            for (uint32_t i = 0; i < fenceCount; i++) {
```
lentinem: I prefer if this is in a function as it's duplicated with the code below.

layers/core_validation.cpp
chrisforbes: Secondary command buffers have an extra ref per primary command buffer they are currently referenced from via CmdExecuteCommands.

I'm not really thrilled with it.

layers/core_validation.cpp
chrisforbes: OK, seems fair

lentinem:
-1. It's going to take a bit to get all the way through this and I'd rather do it before it was checked in. IMO it might be nicer to submit different PRs for the different CLs.

chrisforbes:
The last commit can be moved out to another PR, but the others are just cleaning the work area and then doing the work.

lentinem:
I would break out the first two, the next two and the last one into three separate ones. I'm fine with keeping it for this one but it's harder to go through so I'm ending up having to go through each CL separately which is taking longer. All changes other than the third one look mostly fine (minor comments) but am still going through that one.

layers/core_validation.cpp
```diff
@@ -4566,59 +4476,52 @@ QueueSubmit(VkQueue queue, uint32_t submitCount, const VkSubmitInfo *pSubmits, V
 
     // Mark the fence in-use.
     if (pFence) {
-        SubmitFence(pQueue, pFence);
+        SubmitFence(pQueue, pFence, std::max(1u, submitCount));
```
lentinem: Why do we use 1 when submitCount is 0? submitCount 0 should probably just be a no-op.

lentinem:
Ok; added some comments but after they are addressed it seems ok. As far as the class separation goes this would likely belong in a queue class but it's probably ok to not to do that; would be nice adding a TODO comment for that though.

layers/core_validation.cpp
chrisforbes: But it's /not/ quite a noop -- it delimits previously submitted work with the new fence. It's possible to not allocate a sequence number in some cases (where the last submitted work was unfenced), but that would add more complexity.

lentinem: Fair enough. I still find it odd to do this but you're right in that it's probably the easiest way to handle this in theory valid case.

chrisforbes:
Pushed the first two trivial commits to master. The third commit now has horrible conflicts.

chrisforbes:
Going to abandon the last commit (removing globalInFlightCmdBuffers), I don't really like it any more. The 'considered in flight from CmdExecuteCommands' bit doesn't sit well.

chrisforbes:
Closing; the good bits of this landed some time ago.

PR #693, opened 2016-06-26
demos: GH692 Added a supporting of wayland wsi

elongbug:
Added a supporting of wayland wsi to cube, tri and smoke.

Signed-off-by: Mun Gwan-gyeong <elongbug@gmail.com>

PR #703, opened 2016-06-28
tests: Fix compilation failure with Wayland/Mir platforms enabled

aejsmith:
Fix the following compilation error in
vkGetPhysicalDeviceWaylandPresentationSupportKHR:

  .../wrap_objects.cpp:1101:23: error: ‘inst’ was not declared in this scope

A similar error exists in vkGetPhysicalDeviceMirPresentationSupportKHR,
so fix that as well.


dustin-lunarg:
I pushed a fix for this yesterday when reviewing #693.  Sorry, didn't notice this pull request before now.

PR #708, opened 2016-06-28
layers: Don't produce spurious validation errors for sample count

### LINKED ISSUE #705: Core Validation: spurious error on sample count when rendering without attachments
The error
```
Num samples mismatch! At draw-time in Pipeline ... with 1 samples while current RenderPass ... w/ 0 samples!
```

Is spuriously emitted, when: 

- current pipeline has `VkPipelineMultisampleStateCreateInfo.rasterizationSamples == 1`
AND
- renderpass has no color attachments
AND
- render pass has no depth stencil attachments BUT subpass description has `pDepthStencilAttachment != NULL` (`attachment == VK_ATTACHMENT_UNUSED`). 

According to my spec reading, this situation should be valid.

chrisforbes:
Previously we would emit a spurious validation error if the only
color or depth attachment references were VK_ATTACHMENT_UNUSED.

Fixes #705


tobine:
LGTM :+1:

chrisforbes:
Picked to master.

PR #709, opened 2016-06-28
Gh705 followup - CreateRenderPass validation of subpass resolves

chrisforbes:


chrisforbes:
Added some more.

layers/core_validation.cpp
```diff
@@ -2852,32 +2851,21 @@ static bool validatePipelineDrawtimeState(layer_data const *my_data,
                                 reinterpret_cast<const uint64_t &>(pPipeline->pipeline));
             }
 
-            for (i = 0; i < subpass_desc->colorAttachmentCount; i++) {
-                VkSampleCountFlagBits samples;
-
-                if (subpass_desc->pColorAttachments[i].attachment == VK_ATTACHMENT_UNUSED)
-                    continue;
+            unsigned subpass_num_samples = 0;
```
tobine: Why unsigned here? Causes casting and pso_num_samples that it's compared against later is type "VkSampleCountFlagBits".

layers/core_validation.cpp
```diff
@@ -2852,32 +2851,21 @@ static bool validatePipelineDrawtimeState(layer_data const *my_data,
...
+                subpass_desc->pDepthStencilAttachment->attachment != VK_ATTACHMENT_UNUSED) {
+                auto attachment = subpass_desc->pDepthStencilAttachment->attachment;
+                subpass_num_samples |= (unsigned)render_pass_info->pAttachments[attachment].samples;
             }
 
-            if (((subpass_desc->colorAttachmentCount > 0) || (subpass_desc->pDepthStencilAttachment != NULL)) &&
-                (pso_num_samples != subpass_num_samples)) {
+            if (subpass_num_samples && pso_num_samples != subpass_num_samples) {
                 skip_call |=
                         log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_PIPELINE_EXT,
```
tobine: Not part of your commit but I see that this error message can be updated to use string_VkSampleCountFlagBits() helper for printing sample count enum values as strings.

layers/core_validation.cpp
```diff
@@ -8806,23 +8798,66 @@ static bool ValidateRenderpassAttachmentUsage(layer_data *dev_data, const VkRend
                 skip_call |= ValidateAttachmentIndex(dev_data, attachment, pCreateInfo->attachmentCount, "Preserve");
             }
         }
+
+        auto subpass_performs_resolve = subpass.pResolveAttachments && std::any_of(
+            subpass.pResolveAttachments, subpass.pResolveAttachments + subpass.colorAttachmentCount,
+            [](VkAttachmentReference ref) { return ref.attachment != VK_ATTACHMENT_UNUSED; });
```
tobine: nifty

layers/core_validation.cpp
tobine: Ah, I see. There's no 0 value in the enum. Unfortunate. I guess unsigned w/ casting better than alternative like init to "VK_SAMPLE_COUNT_FLAG_BITS_MAX_ENUM".

layers/core_validation.cpp
```diff
@@ -8806,23 +8798,66 @@ static bool ValidateRenderpassAttachmentUsage(layer_data *dev_data, const VkRend
...
+                }
             }
             attachment = subpass.pColorAttachments[j].attachment;
             skip_call |= ValidateAttachmentIndex(dev_data, attachment, pCreateInfo->attachmentCount, "Color");
+
+            if (!skip_call && attachment != VK_ATTACHMENT_UNUSED) {
+                sample_count |= (unsigned)pCreateInfo->pAttachments[attachment].samples;
+            }
+
+            if (!skip_call &&
```
tobine: Could embed this block as a sub-block in "if" check above to avoid re-checking skip_call & ATT_UNUSED

tests/layer_validation_tests.cpp
```diff
@@ -3230,6 +3230,180 @@ TEST_F(VkLayerTest, UnusedPreserveAttachment) {
     }
 }
 
+TEST_F(VkLayerTest, CreateRenderPassResolveRequiresColorMsaa) {
+    ASSERT_NO_FATAL_FAILURE(InitState());
```
tobine: Please add TEST_DESCRIPTION("Test for blah...") macro at start of test

tests/layer_validation_tests.cpp
```diff
@@ -3230,6 +3230,180 @@ TEST_F(VkLayerTest, UnusedPreserveAttachment) {
...
+    VkResult err = vkCreateRenderPass(m_device->device(), &rpci, nullptr, &rp);
+
+    m_errorMonitor->VerifyFound();
+
+    if (err == VK_SUCCESS)
+        vkDestroyRenderPass(m_device->device(), rp, nullptr);
+}
+
+TEST_F(VkLayerTest, CreateRenderPassResolveRequiresSingleSampleDest) {
+    ASSERT_NO_FATAL_FAILURE(InitState());
```
tobine: TEST_DESCRIPTION("") here please

tests/layer_validation_tests.cpp
```diff
@@ -3230,6 +3230,180 @@ TEST_F(VkLayerTest, UnusedPreserveAttachment) {
...
+    VkResult err = vkCreateRenderPass(m_device->device(), &rpci, nullptr, &rp);
+
+    m_errorMonitor->VerifyFound();
+
+    if (err == VK_SUCCESS)
+        vkDestroyRenderPass(m_device->device(), rp, nullptr);
+}
+
+TEST_F(VkLayerTest, CreateRenderPassSubpassSampleCountConsistency) {
+    ASSERT_NO_FATAL_FAILURE(InitState());
```
tobine: TEST_DESCRIPTION("") here please

layers/core_validation.cpp
```diff
@@ -8806,23 +8798,66 @@ static bool ValidateRenderpassAttachmentUsage(layer_data *dev_data, const VkRend
...
+                subpass_performs_resolve &&
+                attachment != VK_ATTACHMENT_UNUSED &&
+                pCreateInfo->pAttachments[attachment].samples == VK_SAMPLE_COUNT_1_BIT) {
+                skip_call |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VkDebugReportObjectTypeEXT(0), 0,
+                                     __LINE__, DRAWSTATE_INVALID_RENDERPASS, "DS",
+                                     "CreateRenderPass:  Subpass %u requests multisample resolve, but renders to "
```
tobine: This message is slightly confusing to me as I was thinking "renders to" would be where the resolve would occur which you would want to be 1_BIT. Does the following seem clearer?

`CreateRenderPass: Subpass %u requests multisample resolve from attachment %u which has VK_SAMPLE_COUNT_1_BIT, but must instead have a multi-bit sample count.`

layers/core_validation.cpp
```diff
@@ -8786,6 +8774,10 @@ static bool ValidateAttachmentIndex(layer_data *dev_data, uint32_t attachment, u
     return skip_call;
 }
 
+static bool IsPowerOfTwo(unsigned x) {
+    return x && !(x & (x-1));
+}
+
 static bool ValidateRenderpassAttachmentUsage(layer_data *dev_data, const VkRenderPassCreateInfo *pCreateInfo) {
```
tobine: I'm not sure what I think of the various early "!skip_call" checks in this function. Part of me feels that if we're going to have the early checks, may as well just return early and skip the extra work. I know that we rarely do that in validation, though, and other have advocated for single point of return when possible.

I think I lean towards early returns, but I'm fine with the existing code also. Just wanted to record my thoughts here in case anyone has further input.

tobine: Make it so

chrisforbes: I'll tackle this in a follow-up PR.

layers/core_validation.cpp
chrisforbes: Done.

layers/core_validation.cpp
chrisforbes: When the point is to accumulate bits, ^ wouldn't work very well :)

layers/core_validation.cpp
chrisforbes: Will address this in a follow-up too.

chrisforbes:
Fixes done, picked to master.

Issue #641, opened 2016-06-15
Destroying an Image buffer used in a command buffer?


Mervill:
This might just be me misunderstanding how Vulkan works, but consider the following:

```C
// Imagine that previously, `imageBuffer` has been filled 
// with an image we want to copy to an Image handle.

// Begin a command buffer
vkBeginCommandBuffer(cmdBuffer, ...)

// Record that we want to copy `imageBuffer` to the target image
vkCmdCopyBufferToImage(cmdBuffer, imageBuffer, targetImage, ...)

// End the command buffer
vkEndCommandBuffer(cmdBuffer);

// We can destroy `imageBuffer` now? How does that make sense?
vkDestroyBuffer(imageBuffer);

// Some time later, submit for execution
vkSubmit(cmdBuffer, ...)

// Image will render correctly
```

As I understand it, command buffers simply 'record' commands to be executed later (when passed to `vkSubmit`). Yet the `imageBuffer` these commands will copy can seemingly be destroyed long before these commands are ever actually executed. Shouldn't I have to wait until this particular command buffer has been executed or freed before I can destroy `imageBuffer`? I've tested this under `VK_LAYER_LUNARG_standard_validation` and it doesn't seem to complain.

courtney-g:
vkDestroyBuffer's valid usage says:

...
All submitted commands that refer to buffer, either directly or via a VkBufferView, must have completed execution
...

Yes, you need to wait until the command buffer is complete before destroying the buffer. The validation layers should throw an error if you were running with them. Were you?

Since the buffer was still in use at vkDestroyBuffer time you violated the Valid Use of vkDestroyBuffer and you enter the realm of "Undefined". Undefined behavior may be nothing happens, things happen to work, segfault the driver, corrupt data, etc. Here you got lucky and though you destroyed the imageBuffer that didn't change the data that was there and so the copy could "work".

tobine:
This is a hole in validation. Currently we'll flag errors if CB dependencies (such as buffers) are destroyed while any CB using such dependencies is in-flight, but only a couple cases are handled of CB dependencies being destroyed prior to CB submission (descriptor sets and framebuffers are the two I know we handle).

There are a few ways to handle this, but my general notion is that anytime such a dependency is destroyed, the CB state for all dependent CBs should be set to CB_INVALID and the error will then be triggered at submission time.

I've got old issue #91 related to this. It's been lower priority until now but I think it now deserves some attention.

[...]

PR #710, opened 2016-06-29
Tobin gh91 new cb invalid cases

tobine:
Reworking of image and buffer binding to cmd buffer.
Now correctly handle case where cmd buffer image/buffer dependencies that are destroyed will cause cmd buffer to become INVALID w/ and error flagged at submit time. Added tests for image/buffer cases.

Still more cases like this to add as well as their tests, but this is a good start and will address #641


layers/core_validation.cpp
```diff
@@ -416,6 +416,16 @@ static bool validate_buffer_usage_flags(layer_data *dev_data, VkBuffer buffer, V
     return skip_call;
 }
 
+// Helper function to validate usage flags for buffers
+// For given buffer_node send actual vs. desired usage off to helper above where
+//  an error will be flagged if usage is not correct
+static bool validateBufferUsageFlags(layer_data *dev_data, BUFFER_NODE const *buffer_node, VkFlags desired, VkBool32 strict,
```
chrisforbes: Can this completely replace `validate_buffer_usage_flags` just above?

tobine: Yes, pushed. I'll rework this with #686 when they collide.

tests/layer_validation_tests.cpp
```diff
@@ -5267,6 +5267,83 @@ TEST_F(VkLayerTest, WriteDescriptorSetIntegrityCheck) {
     vkDestroyDescriptorPool(m_device->device(), ds_pool, NULL);
 }
 
+TEST_F(VkLayerTest, InvalidCmdBufferImageDestroyed) {
+    TEST_DESCRIPTION("Attempt to draw with a command buffer that is invalid "
+                     "due to an image dependency being destroyed.");
+    ASSERT_NO_FATAL_FAILURE(InitState());
+
+    VkImage image;
```
chrisforbes: This test seems far more complex than it needs to be.

tobine: Yeah, I thought I'd be clever and use "vkCmdCopyBufferToImage()" so I could share code between buffer/image tests, but I agree it's too complex for what it's doing.

Pushed a simpler version that just uses vkCmdClearColorImage()

tests/layer_validation_tests.cpp
```diff
@@ -5819,40 +5819,20 @@ TEST_F(VkLayerTest, InvalidCmdBufferImageDestroyed) {
...
-                           image, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, 1,
-                           &region);
+    VkClearColorValue ccv;
+    ccv.float32[0] = 1.0f;
+    ccv.float32[1] = 1.0f;
+    ccv.float32[2] = 1.0f;
+    ccv.float32[3] = 1.0f;
+    VkImageSubresourceRange isr = {};
+    isr.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
+    isr.baseArrayLayer = 1;
```
chrisforbes: baseArrayLayer and baseMiplevel should be 0.

chrisforbes:
With the test fixed, :+1:

tobine:

PR #718, opened 2016-07-01
Validate buffer barrier memory ranges against backing memory not buffer size

### LINKED ISSUE #648: Core_validation: "Buffer Barrier X has offset Y and size Z whose sum is greater than total size W" error refers to the wrong size
Say X is a buffer of size 4 which is going to be used as an uniform buffer. On our implementation, vkGetBufferMemoryRequirements() will tell the application to bind a memory block of 16b for that buffer. Let's assume a memory block of size 16 is bound to that buffer.

Later on, if I schedule execution of a command buffer, which issues a buffer barrier requesting a flush of all 16 bytes, I'm getting the following call-back from the core_validation layer:

image

This is incorrect, since I'm absolutely fine flushing all 16 bytes there, as per the following spec language, describing arguments of the vkBindBufferMemory() function:

> memoryOffset is the start offset of the region of memory which is to be bound to the buffer. ***The number of bytes returned in the VkMemoryRequirements::size member in memory, starting from memoryOffset bytes, will be bound to the specified buffer.***

Edit: This is being reported for the latest version from master, checked out approximately half an hour ago.

chrisforbes:
Fixes #648

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -7887,8 +7887,7 @@ static bool ValidateBarriers(const char *funcName, VkCommandBuffer cmdBuffer, ui
 
         auto buffer_node = getBufferNode(dev_data, mem_barrier->buffer);
         if (buffer_node) {
-            VkDeviceSize buffer_size =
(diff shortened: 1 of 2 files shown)


tobine:
LGTM :+1:

tobine:

PR #726, opened 2016-07-05
tests: Fix subpass self-dependency and barrier in test

chrisforbes:
As @ratchetfreak points out, this self-dependency and barrier were
invalid per 6.5.1: "Self-dependencies must only include pipeline stage
bits that are graphics stages".


tobine:

PR #733, opened 2016-07-06
Allow location aliasing on FS->CB interface

### LINKED ISSUE #731: Validation: shader checker doesn't understand component qualifiers on FS output
Same idea as #604, but for FS outputs.

There are multiple things broken here:
- glslang gets confused about test code (see KhronosGroup/glslang#372)
- Nvidia gets confused and refuses to compile a correct shader which does this
- Validation didn't consider this case at all, producing spurious errors like #604.

chrisforbes:
Fixes #731.

DO NOT MERGE until KhronosGroup/glslang#372 is merged, and ideally drivers are fixed (at least Nvidia is broken at this time)

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -11211,6 +11211,65 @@ TEST_F(VkLayerTest, CreatePipelineAttribArrayType)
     m_errorMonitor->VerifyNotFound();
 }
 
+TEST_F(VkLayerTest, CreatePipelineAttribComponents)
+{
+    m_errorMonitor->ExpectSuccess();
+
+    ASSERT_NO_FATAL_FAILURE(InitState());
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+
+    VkVertexInputBindingDescription input_binding;
+    memset(&input_binding, 0, sizeof(input_binding));
+
+    VkVertexInputAttributeDescription input_attribs[3];
+    memset(input_attribs, 0, sizeof(input_attribs));
+
+    for (int i = 0; i < 3; i++) {
+        input_attribs[i].format = VK_FORMAT_R32G32B32A32_SFLOAT;
+        input_attribs[i].location = i;
+    }
+
+    char const *vsSource =
+        "#version 450\n"
+        "\n"
+        "layout(location=0) in vec4 x;\n"
+        "layout(location=1) in vec3 y1;\n"
+        "layout(location=1, component=3) in float y2;\n"
+        "layout(location=2) in vec4 z;\n"
+        "out gl_PerVertex {\n"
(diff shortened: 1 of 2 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -11251,6 +11310,48 @@ TEST_F(VkLayerTest, CreatePipelineSimplePositive)
     m_errorMonitor->VerifyNotFound();
 }
 
+TEST_F(VkLayerTest, CreatePipelineFsOutputLocationAliasingPositive)
```
tobine: This test fails on ubuntu nvidia w/ message:

> vk_layer_validation_tests: /home/buildbot/LVL-buildbot/LVL-ubuntu-slave/LVL_ubuntu/build/tests/vkrenderframework.cpp:1165: VkShaderObj::VkShaderObj(VkDeviceObj*, const char*, VkShaderStageFlagBits, VkRenderFramework*, const char*): Assertion `VK_SUCCESS == err' failed.
> Warning, version 450 is not yet complete; most version-specific features are present, but some are missing.
> ERROR: 0:4: 'location' : overlapping use of location 0
> ERROR: 1 compilation errors.  No code generated.

tobine:
LGTM :+1:

I assume the test fail is due to the issue you note so we should sit on this as you recommend.

tests/layer_validation_tests.cpp
chrisforbes: Yes, this will fail everywhere, which is why it needs to not be merged (yet). That glslang change is required for this shader to compile. Once /that's/ fixed, nvidia's driver still doesn't like it-- but the spec allows this.

chrisforbes:
glslang issues are now long sorted. I'll poke at this again and see where we are with drivers.

chrisforbes:
NVIDIA still rejects this:

```
Internal error: assembly compile error for fragment shader at offset 350:
-- error message --
line 12, column 8:  error: redeclared identifier
-- internal assembly text --
!!NVfp5.0
OPTION NV_bindless_texture;
# cgc version 3.4.0001, build date Aug 17 2016
# command line args: 
#vendor NVIDIA Corporation
#version 3.4.0.1 COP Build Date Aug 17 2016
#profile gp5fp
#program main
#var float3 rgb : $vout.COL0 : COL0[0] : -1 : 1
#var float alpha : $vout.COL0.w : COL0[0] : -1 : 1
OUTPUT result_color0 = result.color;
OUTPUT result_color0 = result.color;
MOV.F result_color0.xyz, {1, 0, 0, 0}.x;
MOV.F result_color0.w, {0, 0, 0, 0}.x;
END
# 2 instructions, 0 R-regs
```

mark-lunarg:
Closing due to inactivity, and labeled as such.  Please reopen if necessary.

Issue #699, opened 2016-06-27
Layer Validation: Add Valid Usage Cases and Tests for INVALID_MEM_OBJ and vkFreeMemory


MarkMLunarG:
The layer validation code does not validate against application challenges of usage cases mentioned in the spec for vkFreeMemory, nor do related tests exist for layer verification

mark-lunarg:
The referenced valid usage checks are complete.

PR #734, opened 2016-07-06
tests: GH422 Addition of Buffer Validation Tests

MarkMLunarG:
Add classes to support vertex buffer creation and binding
to the pipe and command buffer. Create VertexBufferInvalid
Test to verify:
1) INVALID_BUFFER
2) DOUBLE_DELETE
3) INVALID_BUFFER_MEMORY_OFFSET
4) INVALID_TEXEL_BUFFER_OFFSET
5) INVALID_STORAGE_BUFFER_OFFSET
6) INVALID_UNIFORM_BUFFER_OFFSET
7) INVALID_MEM_OBJ

Remove dead code by getting rid of deleteMemObjInfo function
and incorrect validation test

Change-Id: I117b0bbf71deeb7d9b7e03d93e7d968152425f28


tests/layer_validation_tests.cpp
```diff
@@ -10294,6 +10556,180 @@ TEST_F(VkLayerTest, VtxBufferBadIndex) {
     vkDestroyDescriptorSetLayout(m_device->device(), ds_layout, NULL);
     vkDestroyDescriptorPool(m_device->device(), ds_pool, NULL);
 }
+
+TEST_F(VkLayerTest, VertexBufferInvalid) {
```
tobine: This test crashes on multiple android targets

layers/core_validation.cpp
```diff
@@ -5245,7 +5234,7 @@ BindBufferMemory(VkDevice device, VkBuffer buffer, VkDeviceMemory mem, VkDeviceS
                 skipCall |=
                     log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_PHYSICAL_DEVICE_EXT,
                             0, __LINE__, DRAWSTATE_INVALID_TEXEL_BUFFER_OFFSET, "DS",
-                            "vkBindBufferMemory(): memoryOffset is 0x%" PRIxLEAST64 " but must be a multiple of "
+                            "vkBindBufferMemory(): texel memoryOffset is 0x%" PRIxLEAST64 " but must be a multiple of "
                             "device limit minTexelBufferOffsetAlignment 0x%" PRIxLEAST64,
                             memoryOffset, dev_data->phys_dev_properties.properties.limits.minTexelBufferOffsetAlignment);
             }
```
tobine: I'd like to see the "if (usage &..." block surrounding this error case and the two error cases below spun off into a utility function. It looks like the code is common with just a difference in the bit(s) being checked against and the min alignment variable being used.

tobine:
:-1: Need to get tests passing on android.

For future work, it's preferable to split changes like this into multiple CLs for simpler review and partitioning of changes. Core validation changes in one CL, test infrastructure in another, each new test in a separate CL... Then you can include them all together in the same PR to make sure they're merged to master as a group.

tests/layer_validation_tests.cpp
MarkMLunarG: Are you able to provide any more crash information? A stack trace or line number perhaps? As far as I am aware, we do not have a means to recreate the android failure here.

tobine:
All of the test cases pass on android until the very last one: eFreeInvalidHandle

MarkMLunarG:
This will hopefully address all 3 comments. I polished up a few things in hopes of addressing the android failure but can not yet confirm that that problem was resolved

layers/vk_validation_layer_details.md
```diff
@@ -150,7 +150,7 @@ The Mem Tracker portion of the VK_LAYER_LUNARG_core_validation layer tracks memo
...
-| Valid Memory Object | Verifies that the memory object was properly created and is currently valid | INVALID_MEM_OBJ | vkCmdDrawIndirect vkCmdDrawIndexedIndirect vkCmdDispatchIndirect vkCmdCopyBuffer vkCmdCopyImage vkCmdBlitImage vkCmdCopyBufferToImage vkCmdCopyImageToBuffer vkCmdUpdateBuffer vkCmdFillBuffer vkCmdClearColorImage vkCmdClearDepthStencilImage vkCmdResolveImage vkFreeMemory vkBindBufferMemory vkBindImageMemory vkQueueBindSparse | TODO | NA |
+| Valid Memory Object | Verifies that the memory object was properly created and is currently valid | INVALID_MEM_OBJ | vkCmdDrawIndirect vkCmdDrawIndexedIndirect vkCmdDispatchIndirect vkCmdCopyBuffer vkCmdCopyImage vkCmdBlitImage vkCmdCopyBufferToImage vkCmdCopyImageToBuffer vkCmdUpdateBuffer vkCmdFillBuffer vkCmdClearColorImage vkCmdClearDepthStencilImage vkCmdResolveImage vkFreeMemory vkBindBufferMemory vkBindImageMemory vkQueueBindSparse | INVALID_MEM_OBJ | NA |
```
tobine: This is causing the doc validator to fail. 2nd to last column should be test name such as "VertexBufferInvalid".

MarkMLunarG: Rats. A copy paste mistake... fixed now

MarkMLunarG:
Hey Chris. I changed the FreeMemory function which calls freeMemObjInfo to test the return value and if it fails, the lower layer is not dispatched. Is this the correct behavior for the command buffer clear and memory reference changes that you made recently? Thanks for checking this.

chrisforbes:
> Hey Chris. I changed the FreeMemory function which calls freeMemObjInfo to test the return value and if it fails, the lower layer is not dispatched. Is this the correct behavior for the command buffer clear and memory reference changes that you made recently? Thanks for checking this.

It looks better, but still wobbly -- there's side effects within the layer in the case where the call is skipped.

MarkMLunarG:
> It looks better, but still wobbly -- there's side effects within the layer in the case where the call is skipped.

Earlier I incorrectly implied that clear_cmd_buf_and_mem_references changed the value of skip_call. Only the reportMemReferencesAndCleanUp changes skip_call, outside of my change, where, by my understanding of the convention, skip_call must be set to true to prevent lower layers from acting on an invalid handle. I believe the same convention implies that reportMemReferencesAndCleanUp should only set skip_call to true for conditions where lower layers should not be dispatched.

layers/core_validation.cpp
```diff
@@ -587,53 +587,41 @@ static bool reportMemReferencesAndCleanUp(layer_data *dev_data, DEVICE_MEM_INFO
...
+                clear_cmd_buf_and_mem_references(dev_data, cb);
+            }
+        }
+
+        // Now verify that no references to this mem obj remain and remove bindings
+        if (pInfo->commandBufferBindings.size() || pInfo->objBindings.size()) {
+            skip_call |= reportMemReferencesAndCleanUp(dev_data, pInfo);
+        }
+        // Delete mem obj info
+        dev_data->memObjMap.erase(dev_data->memObjMap.find(mem));
```
chrisforbes: @MarkMLunarG here's one of the problems -- if you've decided to skip the call, you still ruin the layer's state here.

layers/core_validation.cpp
```diff
@@ -4760,11 +4748,13 @@ FreeMemory(VkDevice device, VkDeviceMemory mem, const VkAllocationCallbacks *pAl
     // undefined behavior.
 
     std::unique_lock<std::mutex> lock(global_lock);
-    freeMemObjInfo(my_data, device, mem, false);
+    bool skip_call = freeMemObjInfo(my_data, device, mem, false);
     print_mem_list(my_data);
     printCBList(my_data);
     lock.unlock();
-    my_data->device_dispatch_table->FreeMemory(device, mem, pAllocator);
+    if (!skip_call) {
```
chrisforbes: This bit here is clearly a win over doing it unconditionally :)

layers/core_validation.cpp
```diff
@@ -587,53 +587,41 @@ static bool reportMemReferencesAndCleanUp(layer_data *dev_data, DEVICE_MEM_INFO
...
-    }
-    return skip_call;
-}
-
 static bool freeMemObjInfo(layer_data *dev_data, void *object, VkDeviceMemory mem, bool internal) {
     bool skip_call = false;
     // Parse global list to find info w/ mem
     DEVICE_MEM_INFO *pInfo = getMemObjInfo(dev_data, mem);
     if (pInfo) {
-        if (pInfo->allocInfo.allocationSize == 0 && !internal) {
```
chrisforbes: Is this logic actually dead? It's nice if you /can/ just nuke it.

layers/core_validation.cpp
```diff
@@ -587,53 +587,41 @@ static bool reportMemReferencesAndCleanUp(layer_data *dev_data, DEVICE_MEM_INFO
...
-        }
+        // TODO: Verify against Valid Use section
+        // Clear any CB bindings for completed CBs
+        //   TODO : Is there a better place to do this?
+
+        assert(pInfo->object != VK_NULL_HANDLE);
+        // clear_cmd_buf_and_mem_references removes elements from
+        // pInfo->commandBufferBindings -- this copy not needed in c++14,
+        // and probably not needed in practice in c++11
+        auto bindings = pInfo->commandBufferBindings;
```
chrisforbes: This whole block is really weird, but you probably don't want to tackle it in this change..

layers/core_validation.cpp
```diff
@@ -5282,38 +5272,37 @@ BindBufferMemory(VkDevice device, VkBuffer buffer, VkDeviceMemory mem, VkDeviceS
...
         // Validate device limits alignments
+        static const VkBufferUsageFlagBits usage_list[3] = {
+            static_cast<VkBufferUsageFlagBits>(VK_BUFFER_USAGE_UNIFORM_TEXEL_BUFFER_BIT | VK_BUFFER_USAGE_STORAGE_TEXEL_BUFFER_BIT),
+            VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT,
+            VK_BUFFER_USAGE_STORAGE_BUFFER_BIT};
+        static const char *memory_type[3] = {"texel",
+                                             "uniform",
+                                             "storage"};
+        static const VkDeviceSize offset_requirement[3] = {
+            dev_data->phys_dev_properties.properties.limits.minTexelBufferOffsetAlignment,
```
chrisforbes: This is subtly broken when there are multiple GPUs

layers/core_validation.cpp
MarkMLunarG: The logic is that I could not find language in the spec that prevents freeing an allocation of size 0. Plus internal is initialize to false and never changed anywhere in the current code. I don't recall finding a clear explanation of what internal is used for but I stopped short of removing it altogether.

layers/core_validation.cpp
MarkMLunarG: Yes. GH issue #699 was added for that critical rugby play. I think that incorrectly assumed that you were also familiar with reportMemReferencesAndCleanUp from git blame, but you only did the CB bindings, sorry.

layers/core_validation.cpp
chrisforbes: Seems fair.

layers/core_validation.cpp
chrisforbes: This is pulling it from the right place -- you just can't stash it forever like this.

MarkMLunarG: Oh yes, of course. Too subtle for me to catch. Nice, thanks!

layers/core_validation.cpp
MarkMLunarG: Given that skip_call is set, a spec violation has occurred so some level of undefined behavior seems reasonable. If there's a cleaner resolution in this case, that's good, but I'm not sure what that would be

chrisforbes: Well, the ideal thing is to move all the side effects after the call, and only perform them if skip_call wasn't set. If the app requests to skip the call, the app, layer, and driver should all agree that it didn't happen.

chrisforbes: (I know this is a pain to get perfect in practice, for the calls which don't return failure status for issues found further down the chain -- but you can get pretty close)

chrisforbes: I think that might be a problem for a follow-up change though, this PR is a solid step forward, so has my :+1:

MarkMLunarG: I can see your point especially given that the app makes the decision. In my limited experience, I haven't run across anything within core_validation that supports this though.  Probably the first step would be to assess what's involved with supporting that concept

chrisforbes: A lot of the newer code does -- but you're right, there's a /lot/ of bad behavior still.

tobine:
LGTM :+1:

PR #735, opened 2016-07-07
Tobin gh689 fix

### LINKED ISSUE #689: "Layout for input image should be TRANSFER_SRC_OPTIMAL instead of GENERAL" is emitted even for LINEAR textures.
LINEAR tiled textures should only use GENERAL, so this performance warning is misleading.

tobine:
Fixes #689 

Don't issue perf warning if GENERAL layout used in linear tiling case, as GENERAL layout is only valid option in that case.


chrisforbes:
:+1:

PR #736, opened 2016-07-07
Tobin gh695 pipeline layout lifetime

### LINKED ISSUE #695: Core Validation mishandling of pipeline layout objects
Vulkan spec 2.3.1 says:

>The following object types are consumed when they are passed into a Vulkan command and not further accessed by the objects they are used to create. They can be destroyed at any time they are not in use by an API command:
>
>VkShaderModule
>VkPipelineCache
>VkPipelineLayout

However, core validation retains the pipeline layout handle and uses it in descriptor set compatibility checks at bind & draw time etc.

tobine:
Fixes #695 

Save PIPELINE_LAYOUT_NODE struct for last bound VkPipelineLayout and use that to perform compatibility checks. This makes sure that validation doesn't rely on VkPipelineLayout node existing beyond its spec-defined lifetime.

As it turns out, we weren't removing entries from pipelineLayoutMap() ad VkPipelineLayout destroy time, so that code was added as well.

Also updated an existing test to destroy VkPipelineLayout node earlier and verify that it's not required to be alive in order to perform compatibility validation.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -1981,7 +1981,7 @@ static void mark_accessible_ids(shader_module const *src, spirv_inst_iter entryp
 }
 
 static bool validate_push_constant_block_against_pipeline(debug_report_data *report_data,
-                                                          std::vector<VkPushConstantRange> const *pushConstantRanges,
+                                                          std::vector<VkPushConstantRange> const *push_constant_ranges,
                                                           shader_module const *src, spirv_inst_iter type,
                                                           VkShaderStageFlagBits stage) {
     bool pass = true;
@@ -2001,7 +2001,7 @@ static bool validate_push_constant_block_against_pipeline(debug_report_data *rep
                 auto size = 4; /* bytes; TODO: calculate this based on the type */
 
                 bool found_range = false;
-                for (auto const &range : *pushConstantRanges) {
+                for (auto const &range : *push_constant_ranges) {
                     if (range.offset <= offset && range.offset + range.size >= offset + size) {
                         found_range = true;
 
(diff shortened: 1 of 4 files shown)


lentinem:
+1

layers/core_validation_types.h
```diff
@@ -426,10 +426,25 @@ template <> struct hash<ImageSubresourcePair> {
...
 
+// Store layouts and pushconstants for PipelineLayout
+struct PIPELINE_LAYOUT_NODE {
+    VkPipelineLayout layout;
+    std::vector<cvdescriptorset::DescriptorSetLayout const *> set_layouts;
+    std::vector<VkPushConstantRange> push_constant_ranges;
+
+    PIPELINE_LAYOUT_NODE() : layout(VK_NULL_HANDLE), set_layouts{}, push_constant_ranges{} {}
+
+    void reset() {
```
chrisforbes: Needing this suggests something isn't quite right in how we do LAST_BOUND_STATE...

layers/core_validation.cpp
```diff
@@ -5342,9 +5342,10 @@ DestroyPipeline(VkDevice device, VkPipeline pipeline, const VkAllocationCallback
 
 VKAPI_ATTR void VKAPI_CALL
 DestroyPipelineLayout(VkDevice device, VkPipelineLayout pipelineLayout, const VkAllocationCallbacks *pAllocator) {
-    get_my_data_ptr(get_dispatch_key(device), layer_data_map)
-        ->device_dispatch_table->DestroyPipelineLayout(device, pipelineLayout, pAllocator);
-    // TODO : Clean up any internal data structures using this obj.
+    layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
+    dev_data->device_dispatch_table->DestroyPipelineLayout(device, pipelineLayout, pAllocator);
+
+    dev_data->pipelineLayoutMap.erase(pipelineLayout);
```
chrisforbes: Better hold the lock while you do this.

chrisforbes:
With the locking issue fixed, :+1:

tobine:

PR #748, opened 2016-07-12
layers: Don't use instance pointer after it has been destroyed

cnorthrop:
In unique_objects' explicit_DestroyInstance, the recently introduced
second get_dispatch_table was using a destroyed instance as a lookup
key, failing to find an entry (since the driver controls that memory
and probably freed it), and was crashing consistently on PixelC.

Instead, use the instance pointer to cleanup instanceExtMap before
calling down to the ICD's DestroyInstance.

This was causing random crashes on PixelC so we had to take it offline until we could debug it.  This now passes on all platforms in our test farm.

### DIFF
--- a/layers/unique_objects.h
+++ b/layers/unique_objects.h
@@ -192,10 +192,10 @@ VkResult explicit_CreateInstance(const VkInstanceCreateInfo *pCreateInfo, const 
 
 void explicit_DestroyInstance(VkInstance instance, const VkAllocationCallbacks *pAllocator) {
     dispatch_key key = get_dispatch_key(instance);
-    get_dispatch_table(unique_objects_instance_table_map, instance)->DestroyInstance(instance, pAllocator);
-    layer_data_map.erase(key);
     VkLayerInstanceDispatchTable *pDisp = get_dispatch_table(unique_objects_instance_table_map, instance);
     instanceExtMap.erase(pDisp);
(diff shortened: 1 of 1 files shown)


mark-lunarg:
:+1:

layers/unique_objects.h
```diff
@@ -192,10 +192,10 @@ VkResult explicit_CreateInstance(const VkInstanceCreateInfo *pCreateInfo, const
 
 void explicit_DestroyInstance(VkInstance instance, const VkAllocationCallbacks *pAllocator) {
     dispatch_key key = get_dispatch_key(instance);
-    get_dispatch_table(unique_objects_instance_table_map, instance)->DestroyInstance(instance, pAllocator);
-    layer_data_map.erase(key);
     VkLayerInstanceDispatchTable *pDisp = get_dispatch_table(unique_objects_instance_table_map, instance);
     instanceExtMap.erase(pDisp);
+    get_dispatch_table(unique_objects_instance_table_map, instance)->DestroyInstance(instance, pAllocator);
```
stroyan: This could use "pDisp" instead of calling "get_dispatch_table(unique_objects_instance_table_map, instance)" again.

cnorthrop: Yes, I was trying to make it clear that this change is just a re-order.  I can optimize it too, if the original change has buyoff.

cnorthrop: Added as a second commit for testing, but can be squished on master.

cnorthrop:
cherry-picked and squished to master.

PR #749, opened 2016-07-12
layers: Handle special layerCount case

dustin-lunarg:
Make parameter validation logic for VkImageSubresource ignore
layerCount when it is set to VK_REMAINING_ARRAY_LAYERS.


tobine:
LGTM :+1:

PR #751, opened 2016-07-13
layers: Rework Destroy/Free data structure cleanup

### LINKED ISSUE #750: core_validation may fail due to reuse of dispatchable handles in threads
If a dispatchable handle is reused after free or destroy, core_validation may lose track of it.
FreeCommandBuffers calls the down the dispatch table to free before updating its data structures.
Another thread may reuse those command buffer handles before the call returns and the data is updated.
All calls releasing dispatchable handles should update data structures before calling down the the next layer or icd.  This problem may effect other functions beyond FreeCommandBuffers.

tobine:
Fixes #750 

When destroying dispatchable objects we must cleanup internal data
structures prior to calling down the chain in the event that another
thread were to prempt execution and re-allocate a freed handle prior
to the layer cleanup code executing.

For non-dispatchable handles this shouldn't be a requirement due to
the fact that layers only see unique handles. However, moved some
non-dispatchable cleanup before the call to ICD as well for consistency
and to be extra safe.


stroyan:
Looks good to me.

tobine:

chrisforbes:
:+1:

PR #753, opened 2016-07-13
Stroyan thread tuning

stroyan:
These changes reduce the overhead of the threading layer.
The number of mutexes is increased to reduce contention between threads.
The parameter reference counters are only updated once multiple threads are seen overlapping calls to vulkan.


lentinem:
+1

tobine:
LGTM :+1:

chrisforbes:
Needs perf numbers. I'm seeing something odd on nvidia/z840 running smoketest (built as RelWithDebInfo) with just this threading layer -- total frame time jumps from 41.4ms to 41.8ms with this change.

stroyan:
Running "smoketest -b" on a very similar system with linux shows-
no validation:  1.5 milliseconds per frame
old threading layer: 69.5 milliseconds per frame
new threading layer:  33.0 milliseconds per frame

stroyan:
Pushed to master.

PR #754, opened 2016-07-13
Tobin validation test coverage

tobine:
These are initial documentation updates to bring documentation of validation tests in-line with actual behavior. Just trickling this one out as an initial FYI of what I'm doing. Will have more such updates but will probably buffer them into larger PRs as long as it's just documentation and non-functional code changes.


tobine:
This PR is replaced by #776

Issue #663, opened 2016-06-20
VkBuffer objects not bound to VkDeviceMemory objects are not detected on use


iamgreaser:
This was tested on the 1.0.13 release of the LunarG SDK, and a fresh Git checkout of the validation layer.

Vulkan spec 1.0.17, section 11.6 states:

> Non-sparse resources must be bound completely and contiguously to a single VkDeviceMemory object before the resource is passed as a parameter to any of the following operations:
> 
> *creating image or buffer views
> *updating descriptor sets
> *recording commands in a command buffer

It is currently possible for me to call `vkCmdUpdateBuffer` or `vkCmdFillBuffer` with no warning, and then an in-driver crash (I'm using Mesa-git anvil).

It is also currently possible for me to call `vkCmdBindVertexBuffers` with no warning, and then the call is ignored and the vertex data is read as all-zeroes.

tobine:
Note that same restrictions apply to VkImage object. @mark-lunarg we may want to break this out into separate issues.
I believe @MarkY-LunarG fix from last week handles the first case, `creating image or buffer views`
Descriptor should be an easy fix-up for me to own if there are any bugs there.
There are a bunch of holes for the last case `recording commands in a command buffer`. However, there are some uniform calls that already take place on images and buffers so we can hopefully find a clean, unified solution for making sure that memory is correctly bound to those objects when they're used while recording cmd buffer cmds.

tobine:
#758 will address `recording commands in command buffer` case.
This just leaves `updating descriptor sets` so assigning this to myself to make sure those cases are correctly handled.

mark-lunarg:
#758 has been pushed and closed, that covers the `recording commands in command buffer` case.

tobine:
Just looking at this issue and I see that I added checks for Buffer descriptor updates but not images. Just making a note here that only remaining work is to handle image descriptor updates that don't have memory bound to image as specified.

PR #758, opened 2016-07-14
Check that memory is bound to Images/Buffers before use

mark-lunarg:
Couldn't find in the spec where this was explicitly called out, but it will crash layers and drivers.


layers/core_validation.cpp
```diff
@@ -692,6 +692,32 @@ static bool clear_object_binding(layer_data *dev_data, uint64_t handle, VkDebugR
     return skip_call;
 }
 
+// Check to see if memory was ever bound to this image
+static bool ValidateMemoryIsBoundToImage(layer_data *dev_data, VkImage image, const char *api_name) {
+    bool result = false;
+    auto const image_node = getImageNode(dev_data, image);
```
tobine: In most of the places where you're using these functions the image or buffer node lookup has already occurred. Can you please change the function to just take the node ptr and avoid the extra lookup?

layers/core_validation.cpp
```diff
@@ -692,6 +692,32 @@ static bool clear_object_binding(layer_data *dev_data, uint64_t handle, VkDebugR
     return skip_call;
 }
 
+// Check to see if memory was ever bound to this image
+static bool ValidateMemoryIsBoundToImage(layer_data *dev_data, VkImage image, const char *api_name) {
```
tobine: See PreCallValidateCreateImageView() and PreCallValidateCreateBufferView(). They handle these checks but only for image/buffer view creation. There's some extra logic there, though that I believe is also required for these functions:

- Checking for sparse binding
- Handling swapchain images

Please update these new ValidateMemoryIsBoundTo* functions to have the extra logic, and update the PreCallValidateCreate*View functions to use these new ValidateMemoryIsBoundTo* functions.

I believe @MarkY-LunarG wrote those functions so you may want to coordinate with him.

tobine:
Good start but I flagged a couple spots where I'd like to see cleanup and consolidation.

layers/vk_validation_layer_details.md
```diff
@@ -162,6 +162,7 @@ The Mem Tracker portion of the VK_LAYER_LUNARG_core_validation layer tracks memo
...
 | Memory Map Range Checks | Validates that Memory Mapping Requests are valid for the Memory Object (in-range, not currently mapped on Map, currently mapped on UnMap, size is non-zero) | INVALID_MAP | vkMapMemory | InvalidMemoryMapping | NA |
 | Memory Type Index Checks | Validates that specified memory type indices are valid | INVALID_MEM_TYPE | vkBindImageMemory vkBindBufferMemory | TODO | Need to fix up and re-enable BindImageInvalidMemoryType test as noted in comment in test |
+| Memory Bound To Object Checks | Validates that memory was bound to image or buffer before use | OBJECT_NOT_BOUND | vkCmdCopyImage vkCmdBlitImage vkCmdCopyBufferToImage vkCmdCopyImageToBuffer vkCmdClearColorImage vkCmdClearDepthStencilImage vkCmdResolveImage vkCmdUpdateBuffer vkCmdFillBuffer vkCmdBindIndexBuffer vkCmdBindVertexBuffers vkCmdDrawIndirect vkCmdDrawIndexedIndirect vkCmdDispatchIndirect vkCmdCopyBuffer vkCmdCopyQueryPoolResults| ImageMemoryNotBound BufferMemoryNotBound | None |
```
MarkY-LunarG: Can you change the error I throw in my PreCallValidateCreateImageView and PreCallValidateCreateBufferView calls from MEMTRACK_INVALID_USAGE_FLAG to MEMTRACK_OBJECT_NOT_BOUND?  Once you do that, this new row should then include vkCreateBufferView and vkCreateImageView.

chrisforbes:
@mark-lunarg wrote:
> Couldn't find in the spec where this was explicitly called out, but it will crash layers and drivers.

If you want a spec reference, this is covered by *11.6 Resource Memory Association*, which says:
>
>- creating image or buffer views
>- updating descriptor sets
>- recording commands in a command buffer

layers/core_validation.cpp
mark-lunarg: Swapchain images should be fine since their image_node->mem field is set to MEMTRACKER_SWAP_CHAIN_IMAGE_KEY.  The rest is done, I'll push the changes after a bit more testing.  These were all really good suggestions -- thanks for the context.

mark-lunarg:
@tobine , I think this is good-to-go, but thought you might want to run it through your test farm.  If not, lemme know and I'll push it.

tobine:
LGTM :+1:

Clean on our farm

mark-lunarg:
Pushed to Master -- LX555 and #663. 

commit b3c2024e0ed61d7f68630d7eb7c8c03100689bb1
Date:   Thu Jul 14 09:54:11 2016 -0600

    layers: LX555/GH663, Add BindXxxMemory validation

    If the user neglected to call BindImageMemory or BindBufferMemory,
    validation gave not indication of the failure which then results in
    a driver or layer crash. This commit extends and generalizes the
    checks done for CreateImageView and CreateBufferView to cover all
    the bound memory checks for the 'recording commands in command
    buffer' case in spec section 11.6

    Change-Id: I3b4a3d130e9b6c242413fc0ad14c1624295b33b3

PR #761, opened 2016-07-14
Tobin gh756 unique obj free destroy cleanup

### LINKED ISSUE #756: Verify unique_objects Free/Destroy behavior
Unique_objects needs to handle all state clean-up for Destroy/Free of non-dispatchable objects prior to calling down chain.
This issue is similar to #750 but for non-dispatchable objects only unique_objects layer is affected.

tobine:
Fixes #756 

This is similar to #750 but it's an issue for non-dispatchable handles in unique_objects.
Update the codegen for Destroy functions to remove the handle from the map prior to calling down the chain.


dustin-lunarg:
This passes all of our builds and the diff between the old and new generated unique_objects.cpp files looks good.  The number of locks performed for the vkDestroy.* functions was cut in half.  Nice!  :+1:

chrisforbes:
:+1:

PR #765, opened 2016-07-15
Stroyan threading cmd reuse

stroyan:
Update threading layer data structures before calling down to FreeCommandBuffers.
That avoids a race condition with reuse of freed command buffer handles.

Use calls to vkEnumeratePhysicalDevices to reliably enable threading checks
in validation test before attempting to trigger a validation report.


tobine:
LGTM :+1:

PR #766, opened 2016-07-15
layers: Fix minor completeness/performance issue

MarkMLunarG:
The first time the condition is discovered it will
be the same as every subsequent time, so there is
no need to continue. Also, the whole structure must
be compared, so memcmp assures that will happen
with good efficiency

Change-Id: I71f608c71f83ebcd02212bb391a30bdf15279ff6


tobine:
LGTM :+1:

MarkMLunarG:
SHA-1: 06811df0256552cd7da9d7297672af377463fc4a

* layers: Fix minor completeness/performance issue

PR #768, opened 2016-07-15
windows: Return exit code from build script if any step fails

cnorthrop:
Dropping the /B from each exit allows the full script to exit(1) on failure, instead of just the current code block.  I've confirmed this change allows our system to detect when Win32 build failures.


dustin-lunarg:
Looks good to me! :+1:

cnorthrop:

PR #769, opened 2016-07-15
layers: Update clearValueCount check

### LINKED ISSUE #601: Missing validation check: vkCmdBeginRenderPass clearValueCount check

tobine:
Fixes #601 

The pClearValues array is indexed by the attachment index which
means that it must be at least as large as the last LOAD_OP_CLEAR
attachment in the renderpass. Updating this check to reflect this
requirement.


lentinem:
+1

chrisforbes:
:+1:

PR #771, opened 2016-07-16
Add NV_dedicated_allocation support to unique_objects layer

dustin-lunarg:
Add ID substitution support for the VkDedicatedAllocationMemoryAllocateInfoNV extension structure to the unique_objects layer.  The current implementation is specific to the case where the vkAllocateMemory function's pAllocateInfo parameter references a single VkDedicatedAllocationMemoryAllocateInfoNV struct.


tobine:
LGTM :+1:

PR #774, opened 2016-07-19
Tobin gh773 attach first use layout check

### LINKED ISSUE #773: Only first use layout should be validated for LOAD_OP_CLEAR
Only the first used layout should be validated for LOAD_OP_CLEAR:

> The **_first use_** of an attachment **must not** specify a layout equal to `VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL` or `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL` if the attachment specifies that the `loadOp` is `VK_ATTACHMENT_LOAD_OP_CLEAR`.

This commit probably introduced the layer false positive reports: https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/commit/be45dd763c94dfcb0b52709d8d613848c4c914e1
(previously only the first layout was checked -- now all the subpass layouts are checked)

E.g. consider the (valid) case that the image is `LAYOUT_COLOR_ATTACHMENT` with `LOAD_OP_CLEAR` in first subpass and `LAYOUT_READ_ONLY` in second. That situation produces a false positive.

tobine:
Fixes #773 

Only make sure that a subpass attachment that has LOAD_OP_CLEAR is not used in *READ_ONLY* layout for its first use. On subsequent uses this is fine.
Also add a positive test that hits this case.

Note that the diff is a bit messy because I moved inputAttachment loop to the bottom. The read code change, though, is checking attach_first_use[index] before calling ValidateLayoutVsAttachmentDescription() and then setting attach_first_use[index] to false.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -8956,8 +8956,56 @@ static bool ValidateLayoutVsAttachmentDescription(debug_report_data *report_data
 static bool ValidateLayouts(const layer_data *my_data, VkDevice device, const VkRenderPassCreateInfo *pCreateInfo) {
     bool skip = false;
 
+    // Track when we're observing the first use of an attachment
+    std::vector<bool> attach_first_use(pCreateInfo->attachmentCount, true);
     for (uint32_t i = 0; i < pCreateInfo->subpassCount; ++i) {
         const VkSubpassDescription &subpass = pCreateInfo->pSubpasses[i];
+        for (uint32_t j = 0; j < subpass.colorAttachmentCount; ++j) {
+            auto attach_index = subpass.pColorAttachments[j].attachment;
+            if (attach_index == VK_ATTACHMENT_UNUSED)
+                continue;
+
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -9017,8 +9000,36 @@ static bool ValidateLayouts(const layer_data *my_data, VkDevice device, const Vk
...
+                                                              attach_index, pCreateInfo->pAttachments[attach_index]);
+            }
+            attach_first_use[attach_index] = false;
+        }
+        for (uint32_t j = 0; j < subpass.inputAttachmentCount; ++j) {
+            auto attach_index = subpass.pInputAttachments[j].attachment;
+            if (attach_index == VK_ATTACHMENT_UNUSED)
+                continue;
+
+            if (subpass.pInputAttachments[j].layout != VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL &&
```
chrisforbes: Consider switch on layout?

chrisforbes:
:+1:

PR #779, opened 2016-07-19
layers: Change pNext validation error to warning

dustin-lunarg:
Change the severity level of the parameter_validation layer's pNext parameter validation messages from error to warning.  Also add the following explanation that the warning is derived from the Valid Usage
language from version X of the Vulkan spec/header: 
>This warning is based on the Valid Usage documentation for version %d of the Vulkan header.  It is possible that you are using a struct from a private extension or an extension that was added to a later
version of the Vulkan header, in which case your use of pNext is perfectly valid but is not guaranteed to work correctly with validation enabled

Issues-Addressed: LunarXchange 560


lentinem:
+1

PR #781, opened 2016-07-20
layers: Fix bug with renderpass state in cmd buffer

tobine:
When a CmdEndRenderPass is called, reset the command buffer's activeSubpassContents
to the default value of VK_SUBPASS_CONTENTS_INLINE.
This fixes a but where, if a previously renderpass in a cmd buffer had been in
VK_SUBPASS_CONTENTS_SECONDARY_COMMAND_BUFFERS mode, then a subsequent renderpass
in that same mode will incorrectly hit an error about an invalid call to
vkCmdBeginRenderPass() within a secondary cmd buffer subpass.


chrisforbes:
Is there a corresponding bug?

We've had a test for a while that requires this case to work correctly. Any code that's looking at the subpass contents state when there isn't an active renderpass instance is overreaching.

tobine:
Bug received via direct mail from a vendor. I see the RenderPassSecondaryCommandBuffersMultipleTimes test is currently passing.
I suspect the vendor isn't running with current code that includes following at the start of validateCmdsInCmdBuffer():
```

if (!pCB->activeRenderPass)
        return false;

```

I agree that should render my current PR unnecessary.

PR #782, opened 2016-07-20
Add struct for setting levels.

lentinem:
-1. This can't go in as is but gives an overview of what we have.


lentinem:
Assign to tobine to query about next steps on this.

layers/core_validation.cpp
```diff
@@ -4002,6 +4002,15 @@ CreateInstance(const VkInstanceCreateInfo *pCreateInfo, const VkAllocationCallba
     layer_data *instance_data = get_my_data_ptr(get_dispatch_key(*pInstance), layer_data_map);
     instance_data->instance = *pInstance;
     instance_data->instance_dispatch_table = new VkLayerInstanceDispatchTable;
+    instance_data->enabledLevels = 0x7FFFFFFF;
```
chrisforbes: Why this particular number?

lentinem: Its the enum max we use in vulkan.h

PR #783, opened 2016-07-20
tests: Remove duplicate pNext test

dustin-lunarg:
After adjustment for new extension structures, the
VkLayerTest.InvalidStructPNext vkAllocateMemory and
vkCreateGraphicsPipelines test cases both test the same
validation error.  The larger vkCreateGraphicsPipeline test
case has been removed.


mark-lunarg:
:+1:

PR #784, opened 2016-07-20
Tobin gh663 mem bound desc updates

### LINKED ISSUE #663: VkBuffer objects not bound to VkDeviceMemory objects are not detected on use

tobine:
Fixes #663 

Add validation to make sure that memory is bound to buffer before it's used to update a descriptor. 
Also create associated test to verify that error is correctly flagged.
Image descriptors require an imageView (vs. just image) and we already validate that memory is bound to an image at imageView creation time so no need to add another check for image memory binding at descriptor update time.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -693,28 +693,28 @@ static bool clear_object_binding(layer_data *dev_data, uint64_t handle, VkDebugR
 }
 
 // Check to see if memory was ever bound to this image
-static bool ValidateMemoryIsBoundToImage(layer_data *dev_data, IMAGE_NODE *image_node, const char *api_name) {
+bool ValidateMemoryIsBoundToImage(const layer_data *dev_data, const IMAGE_NODE *image_node, const char *api_name) {
     bool result = false;
     if (0 == (static_cast<uint32_t>(image_node->createInfo.flags) & VK_IMAGE_CREATE_SPARSE_BINDING_BIT)) {
         if (0 == image_node->mem) {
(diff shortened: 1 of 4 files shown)


lentinem:
+1

layers/core_validation.cpp
```diff
@@ -693,28 +693,28 @@ static bool clear_object_binding(layer_data *dev_data, uint64_t handle, VkDebugR
...
         }
     }
     return result;
 }
 
 // Check to see if memory was bound to this buffer
-static bool ValidateMemoryIsBoundToBuffer(layer_data *dev_data, BUFFER_NODE *buffer_node, const char *api_name) {
+bool ValidateMemoryIsBoundToBuffer(const layer_data *dev_data, const BUFFER_NODE *buffer_node, const char *api_name) {
     bool result = false;
     if (0 == (static_cast<uint32_t>(buffer_node->createInfo.flags) & VK_IMAGE_CREATE_SPARSE_BINDING_BIT)) {
```
chrisforbes: Not new, but this should really be `VK_BUFFER_CREATE_SPARSE_BINDING_BIT`

chrisforbes:
:+1:

tobine:

PR #786, opened 2016-07-21
Tobin image aspect mask fixes

tobine:
Ran across this in debugging a test.
Test framework didn't correctly set aspect bits and validation didn't correctly check them.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -8012,14 +8012,39 @@ static bool ValidateBarriers(const char *funcName, VkCommandBuffer cmdBuffer, ui
                 }
             }
             if (imageFound) {
-                if (vk_format_is_depth_and_stencil(format) &&
-                    (!(mem_barrier->subresourceRange.aspectMask & VK_IMAGE_ASPECT_DEPTH_BIT) ||
-                     !(mem_barrier->subresourceRange.aspectMask & VK_IMAGE_ASPECT_STENCIL_BIT))) {
-                    log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
-                            DRAWSTATE_INVALID_BARRIER, "DS", "%s: Image is a depth and stencil format and thus must "
-                                                             "have both VK_IMAGE_ASPECT_DEPTH_BIT and "
-                                                             "VK_IMAGE_ASPECT_STENCIL_BIT set.",
-                            funcName);
+                auto aspect_mask = mem_barrier->subresourceRange.aspectMask;
+                if (vk_format_is_depth_or_stencil(format)) {
+                    if (vk_format_is_depth_and_stencil(format)) {
+                        if (!(aspect_mask & VK_IMAGE_ASPECT_DEPTH_BIT) && !(aspect_mask & VK_IMAGE_ASPECT_STENCIL_BIT)) {
+                            log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0,
+                                    __LINE__, DRAWSTATE_INVALID_BARRIER, "DS",
+                                    "%s: Image is a depth and stencil format and thus must "
+                                    "have either one or both of VK_IMAGE_ASPECT_DEPTH_BIT and "
+                                    "VK_IMAGE_ASPECT_STENCIL_BIT set.",
+                                    funcName);
+                        }
+                    } else if (vk_format_is_depth_only(format)) {
+                        if (!(aspect_mask & VK_IMAGE_ASPECT_DEPTH_BIT)) {
+                            log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0,
+                                    __LINE__, DRAWSTATE_INVALID_BARRIER, "DS", "%s: Image is a depth-only format and thus must "
+                                                                               "have VK_IMAGE_ASPECT_DEPTH_BIT set.",
+                                    funcName);
+                        }
+                    } else { // stencil-only case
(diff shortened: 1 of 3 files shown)


tests/vkrenderframework.cpp
```diff
@@ -46,6 +47,35 @@ bool vk_format_is_depth_and_stencil(VkFormat format) {
...
+        break;
+    }
+
+    return is_depth;
+}
+
+// Return true if format is a depth or stencil format
+bool vk_format_is_depth_or_stencil(VkFormat format) {
+    return (vk_format_is_depth_and_stencil(format) ||
+            vk_format_is_depth_only(format) ||
```
chrisforbes: This seems like an unnecessary layer unless you use it somewhere else too. Just using the depth_and_stencil, depth_only and stencil_only functions directly in the code below would let you unnest it a bit.

tests/vkrenderframework.cpp
```diff
@@ -46,6 +47,35 @@ bool vk_format_is_depth_and_stencil(VkFormat format) {
     return is_ds;
 }
 
+// Return true if format is a stencil-only format
+bool vk_format_is_stencil_only(VkFormat format) {
+    return (format == VK_FORMAT_S8_UINT);
+}
+
+// Return true if format is a depth-only format
+bool vk_format_is_depth_only(VkFormat format) {
```
chrisforbes: There's a copy of these helpers in vk_layer_utils.h, and some bits already duplicated in renderframework. Can we just keep one?

layers/core_validation.cpp
```diff
@@ -8012,14 +8012,39 @@ static bool ValidateBarriers(const char *funcName, VkCommandBuffer cmdBuffer, ui
...
+                        }
+                    } else if (vk_format_is_depth_only(format)) {
+                        if (!(aspect_mask & VK_IMAGE_ASPECT_DEPTH_BIT)) {
+                            log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0,
+                                    __LINE__, DRAWSTATE_INVALID_BARRIER, "DS", "%s: Image is a depth-only format and thus must "
+                                                                               "have VK_IMAGE_ASPECT_DEPTH_BIT set.",
+                                    funcName);
+                        }
+                    } else { // stencil-only case
+                        if (!(aspect_mask & VK_IMAGE_ASPECT_DEPTH_BIT)) {
```
chrisforbes: You meant `VK_IMAGE_ASPECT_STENCIL_BIT`

tests/layer_validation_tests.cpp
```diff
@@ -9821,6 +9821,73 @@ TEST_F(VkLayerTest, InvalidBarriers) {
...
+    if (format_props.optimalTilingFeatures &
+        VK_FORMAT_FEATURE_DEPTH_STENCIL_ATTACHMENT_BIT) {
+        m_errorMonitor->SetDesiredFailureMsg(
+            VK_DEBUG_REPORT_ERROR_BIT_EXT,
+            "Image is a depth-only format and thus must "
+            "have VK_IMAGE_ASPECT_DEPTH_BIT set.");
+        VkDepthStencilObj d_image(m_device);
+        d_image.Init(m_device, 128, 128, VK_FORMAT_D16_UNORM);
+        ASSERT_TRUE(d_image.initialized());
+        img_barrier.oldLayout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL;
```
chrisforbes: A transition between these two layouts would never work, but the expected error is about the aspects?

tests/layer_validation_tests.cpp
```diff
@@ -9821,6 +9821,73 @@ TEST_F(VkLayerTest, InvalidBarriers) {
...
+        img_barrier.newLayout =
+            VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL;
+        img_barrier.image = d_image.handle();
+        // Leave aspectMask at COLOR on purpose
```
chrisforbes: Maybe it would be best to just reassign the aspect mask you want here, rather than explaining the coupling to a previous test?

tests/layer_validation_tests.cpp
```diff
@@ -9821,6 +9821,73 @@ TEST_F(VkLayerTest, InvalidBarriers) {
...
+    VkImageObj c_image(m_device);
+    c_image.init(128, 128, VK_FORMAT_B8G8R8A8_UNORM,
+                 VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, VK_IMAGE_TILING_OPTIMAL,
+                 0);
+    ASSERT_TRUE(c_image.initialized());
+    img_barrier.oldLayout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL;
+    img_barrier.newLayout = VK_IMAGE_LAYOUT_GENERAL;
+    img_barrier.image = c_image.handle();
+    // Set aspect to depth (non-color)
+    img_barrier.subresourceRange.aspectMask = VK_IMAGE_ASPECT_DEPTH_BIT;
```
chrisforbes: Should add a stencil-only case -- since that was broken in your code.

tobine: Stencil only case is just above but none of our test systems hit that path due to not supporting the only spec-defined stencil-only format.

tobine:
Pushed updates to address all of @chrisforbes input except for code duplication of helpers in vkrenderframework.cpp. I'll create a separate issue to clean that up by moving the format helpers into a general vk_format_helper.cpp file (or something like that) that's used by both layers and test framework.

chrisforbes:
:+1:

tests/vkrenderframework.cpp
Tony-LunarG: I tried to use vk_layer_utils.h in the framework, but it comes with a great deal of loader references as well.  I considered creating a vk_format_utils include file that could be included both places, but I don't remember why I didn't.

tobine:

PR #791, opened 2016-07-22
Tobin gh91 descriptor set cb invalid

### LINKED ISSUE #91: Unify validation INVALID cmd buffer cases

tobine:
Fixes #91 

I believe this completes the remaining cases where CB_INVALID was not correctly handled.
This PR handles descriptor sets and all of their bound sampler/image/buffer resources.
Added tests to verify command buffer set to invalid for destruction of bound descriptor set and descriptor set buffer/image/sampler resources.

A key for descriptor sets and their bound resources is to only consider the active set/bindings combinations at draw time. This means that a cmd buffer only goes invalid when resources that it's using are destroyed.


tobine:

chrisforbes:
:+1:

PR #793, opened 2016-07-22
generate one vk_dispatch_table_helper.h for all platforms

mre4ce:


MarkY-LunarG:
If your intent is to make this the default, you'll have to modify the CMakeLists.txt file in the layer sub-folder to generate the vk_dispatch_table_helper.h file with all defines.

mre4ce:
Making a single header file work for all platforms fixes CI issues when compiling for multiple different platforms in one go. Afaik there's no CI setup for this repository so it's not a problem, but it is a problem for Vulkan-Samples which includes several layers that use this header file. With these changes the header file will continue to be platform specific if you specify a particular platform. I didn't want to make any more changes than necessary, but I can certainly make this the default if that's desired for this repository.

MarkY-LunarG:
Changes committed with caeb0e1fd41fe854c9ceb2a65c75920fcae258fb

PR #794, opened 2016-07-22
layers: Fix bug in cmd buffer framebuffer binding

tobine:
If cmd buffers were freed via DestroyCommandPool, then the framebuffer links to the
individual cb_nodes were not getting cleared. Fix that issue.

This is initial, simple solution to avoid crash, need smarter solution long-term.


lentinem:
Please add a test that exposes the issue.

tobine:
Test added that exposes the issue.

lentinem:
+1

tobine:

PR #795, opened 2016-07-26
tests: Create APK version of layer validation tests

cnorthrop:
I haven't updated the documentation yet, but if you want to test the build:

./android-generate.sh
android update project -p . -t "android-23"
ndk-build
ant debug


lentinem:
+1. There is definitely simplification that can be done between the android_main and other platform equivalents. However, this can be done later (cube/tri were done similarly to this).

cnorthrop:
Agreed, this just gets us off the ground.  Thanks!

cherry-picked to master

PR #796, opened 2016-07-26
android: Lift global offset table size limit (-mxgot)

cnorthrop:
Currently the mips64 build for Android is broken due to size
increases of core_validation.  If you disabled optimizations (-O0),
you'll get the following error:

    relocation truncated to fit: R_MIPS_CALL16

The linker flag to lift the limit is -mxgot, documented here:

    https://gcc.gnu.org/onlinedocs/gcc-4.0.0/gcc/MIPS-Options.html

It is a slight de-optimization for global symbol lookups (3 inst
instead of 1).  Hopefully we aren't using global lookups in
performance critical code.


stroyan:
Looks good to me.

cnorthrop:
cherry-picked to master, thanks.

PR #799, opened 2016-07-27
Impose requirements on descriptors from spirv image types

### LINKED ISSUE #762: Validation suggestion: textureSamples() should not be called for single-sampled image bindings
We've just discovered a problem with one of the recently introduced CTS tests, where they first bind a single-sample MS image and then try to query the number of samples exposed by the image with a textureSamples() call. This would work in GLSL, but is illegal after being converted to SPIR-V due to the following restriction in the SPIR-V spec:

```
OpImageQuerySamples
Query the number of samples available per texel fetch in a multisample image.
Result Type must be a scalar integer type. The result is the number of samples.
Image must be an object whose type is OpTypeImage. Its Dim operand must be one
of 2D and MS of 1(multisampled)
```

SPIRV-Validation does not have shadow enough state information to determine this condition, so validation layers seem like a more suitable candidate.

chrisforbes:
Fixes #762, and lays a lot of groundwork for other image usage validation that @tobine has wanted for a while.

### DIFF
--- a/layers/descriptor_sets.cpp
+++ b/layers/descriptor_sets.cpp
@@ -335,10 +335,11 @@ bool cvdescriptorset::DescriptorSet::IsCompatible(const DescriptorSetLayout *lay
 //  This includes validating that all descriptors in the given bindings are updated,
 //  that any update buffers are valid, and that any dynamic offsets are within the bounds of their buffers.
 // Return true if state is acceptable, or false and write an error message into error string
-bool cvdescriptorset::DescriptorSet::ValidateDrawState(const std::unordered_set<uint32_t> &bindings,
+bool cvdescriptorset::DescriptorSet::ValidateDrawState(const std::unordered_map<uint32_t, descriptor_req> &bindings,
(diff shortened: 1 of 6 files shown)


layers/descriptor_sets.cpp
```diff
@@ -327,18 +327,31 @@ cvdescriptorset::DescriptorSet::~DescriptorSet() {
...
 }
+
+
+static char const * string_descriptor_req_view_type(descriptor_req req) {
+    for (unsigned i = 0; i <= VK_IMAGE_VIEW_TYPE_END_RANGE; i++) {
+        if (req & (1 << i)) {
+            return string_VkImageViewType(VkImageViewType(i));
+        }
+    }
+}
```
tobine: error: control may reach end of non-void function

lentinem: Can we auto upload build failures?

lentinem:
+1

chrisforbes:
Updated and pushed to master.

PR #805, opened 2016-07-28
tests & doc: GH427 - Addition of RP Validation tests

MarkMLunarG:
test: GH427 - Addition of Renderpass Validation tests

Addition of tests to challenge renderpass parts of the
validation layer.
1) INVALID_RENDER_AREA - InvalidRenderArea
2) RENDERPASS_INCOMPATIBLE - RenderPassClearOpMismatch
     RenderPassIncompatible FramebufferCreateErrors
3) FRAMEBUFFER_INCOMPATIBLE - FramebufferInvalid
4) INDEPENDENT_BLEND - FailDisabledIndependentBlend

An invalid object test was added as it was a side affect
of other testing in FramebufferInvalid

Also, basic shareable test classes were added using code
pulled from existing tests to support creating
a custom vulkan device for testing, a second framebuffer,
and a pipeline that works with the custom vulkan device
without unexpected errors, all of which are complex to
complete or not supported under the existing frameworks


tobine:
In general I think I like this direction. One main gripe is that I'd like to see the test changes split into multiple CLs.
I believe something needs to be done to make the test code simpler/shorter. Potential solutions are:

1. Split common code into functions
2. Split common code into classes with a bit more hand-holding than functions (what you've done here)
3. Update the test framework to handle much more variation and hide many/most test details from the test bodies

Since you've already taken the steps down the path of #2 I'm fine continuing down that path unless there are strenuous objects. I'm adding a few more reviewers who may have opinions so let's hold onto this PR through at least Monday to see what input others have, if any.

MarkMLunarG:
Thanks Tobin. My selection of solution 2 is driven primarily by:
1) very strict time constraints that I am under and still not meeting
2) the need for a methodology that has 0 immediate impact on the existing battery of tests (which
     makes solution 3 impractical)
3) the existing framework is complex, invarient, and it tightly intertwines lots of different aspects of the
     API which could be managed more independently
4) the original authors of the framework are no longer maintaining it and it significantly violates
    current coding standards

I am well aware that the classes that I have added with this PR are somewhat special case and limited - this is only meant to be a starting point (including the naming). The concept behind these classes is that there is a transient state management article designed to facilitate automatic, generic, and interpreted state initialization that opens the door for later addition of member functions to customize the state as needed.

The state management article is then directly coupled to a more persistent Vulkan object creation article or equivalent, which will create, hold, and manage the related Vulkan object handles for the duration of it's existence.

tests/layer_validation_tests.cpp
```diff
@@ -716,9 +718,667 @@ class VkBufferTest {
...
+
+        QueueCreateInfo.reserve(aVulkanDevice->queue_props.size());
+
+        for (uint32_t i = 0; i < (uint32_t)aVulkanDevice->queue_props.size();
+             i++) {
+            VkDeviceQueueCreateInfo qi = {};
+            qi.sType = VK_STRUCTURE_TYPE_DEVICE_QUEUE_CREATE_INFO;
+            qi.queueFamilyIndex = i;
+            qi.queueCount = aVulkanDevice->queue_props[i].queueCount;
+            QueuePrioriyList.emplace_back(qi.queueCount, 0.0f);
```
chrisforbes: `QueuePrioriyList` is misspelled.

Would prefer `push_back` here -- this isn't interesting in-place construction, this is copying a float.

chrisforbes: Nevermind, no it's not. I see what you're doing now.

tests/layer_validation_tests.cpp
```diff
@@ -716,9 +718,667 @@ class VkBufferTest {
...
+            LayerNames.push_back("VK_LAYER_LUNARG_image");
+            LayerNames.push_back("VK_LAYER_GOOGLE_unique_objects");
+
+            DeviceCreateInfo.enabledLayerCount = LayerNames.size();
+            DeviceCreateInfo.ppEnabledLayerNames = LayerNames.data();
+        }
+
+        if (aExtensionNames) {
+            DeviceCreateInfo.enabledExtensionCount = aExtensionNames->size();
+            DeviceCreateInfo.ppEnabledExtensionNames = aExtensionNames->data();
```
chrisforbes: Retaining pointers into caller-owned containers makes me really uncomfortable. Consider copying or moving aExtensionNames?

tests/layer_validation_tests.cpp
```diff
@@ -716,9 +718,667 @@ class VkBufferTest {
...
+    VkFramebufferCreateState(VkDeviceObj *aDeviceObj, VkRenderPass aRenderPass)
+        : FramebufferCreateInfo{} {
+        FramebufferCreateInfo.height = 32;
+        FramebufferCreateInfo.layers = 1;
+        FramebufferCreateInfo.renderPass = aRenderPass;
+        FramebufferCreateInfo.sType = VK_STRUCTURE_TYPE_FRAMEBUFFER_CREATE_INFO;
+        FramebufferCreateInfo.width = 32;
+    }
+
+    void SetImageView(unsigned aViewCount, VkImageView &aImageViews) {
```
chrisforbes: How many views are you passing in? This is a ref to one.

tests/layer_validation_tests.cpp
```diff
@@ -716,9 +718,667 @@ class VkBufferTest {
...
+        FramebufferCreateInfo.attachmentCount = aViewCount;
+        FramebufferCreateInfo.pAttachments = &aImageViews;
```
chrisforbes: Storing this ptr is trouble...

tests/layer_validation_tests.cpp
```diff
@@ -716,9 +718,667 @@ class VkBufferTest {
...
+
+
+        api_result = vkCreateRenderPass(VulkanDevice->GetDevice(),
+                                        &aGraphicsPipeCreateState.
+                                        RenderPassCreateInfo,
+                                        nullptr, &RenderPass);
+        if (VK_SUCCESS != api_result) {
+            vkDestroyPipelineLayout(VulkanDevice->GetDevice(),
+                                    PipelineLayout, nullptr);
+            LayoutCurrent = false;
```
chrisforbes: All this flag fiddling and explicit cleanup is murky. Consider an RAII container for the pipeline layout?

tests/layer_validation_tests.cpp
```diff
@@ -716,9 +718,667 @@ class VkBufferTest {
...
+            return;
+        }
+        PipelineCurrent = true;
+    }
+
+    ~VkGraphicsPipelineTest() {
+        for (auto interator =
```
chrisforbes: Consider range-based for? The iterator is not interesting in this case, and it's misspelled.

tests/layer_validation_tests.cpp
```diff
@@ -716,9 +718,667 @@ class VkBufferTest {
...
+        if (RenderPassCurrent) {
+            vkDestroyRenderPass(VulkanDevice->GetDevice(), RenderPass, nullptr);
+        }
+        if (LayoutCurrent) {
+            vkDestroyPipelineLayout(VulkanDevice->GetDevice(), PipelineLayout,
+                                    nullptr);
+        }
+    }
+
+    int AddShaderStage(const char *aShaderCode,
```
chrisforbes: Why not use VkShaderObj here?

tests/layer_validation_tests.cpp
```diff
@@ -11678,6 +12341,133 @@ TEST_F(VkLayerTest, VtxBufferBadIndex) {
     vkDestroyDescriptorPool(m_device->device(), ds_pool, NULL);
 }
 
+TEST_F(VkLayerTest, FramebufferInvalid) {
+    RecordProperty("description", "Test case where an invalid Framebuffer Index "
```
chrisforbes: Why not `TEST_DESCRIPTION()` in this case?

tests/layer_validation_tests.cpp
```diff
@@ -11678,6 +12341,133 @@ TEST_F(VkLayerTest, VtxBufferBadIndex) {
...
+    m_errorMonitor->VerifyFound();
+
+    vkCmdEndRenderPass(m_commandBuffer->handle());
+
+    vkDestroyPipelineLayout(m_device->device(), pipeline_layout, NULL);
+}
+
+TEST_F(VkLayerTest, FailDisabledIndependentBlend) {
+    TEST_DESCRIPTION("Test case where Independent Blend is disabled "
+                     "and color blend attachements don't match");
```
chrisforbes: sp: attachments.

tests/layer_validation_tests.cpp
```diff
@@ -11678,6 +12341,133 @@ TEST_F(VkLayerTest, VtxBufferBadIndex) {
...
+    rp_begin.framebuffer = m_framebuffer;
+    rp_begin.renderArea.offset.x = 0;
+    rp_begin.renderArea.offset.y = 0;
+    rp_begin.renderArea.extent.width = static_cast<uint32_t>(m_width);
+    rp_begin.renderArea.extent.height = static_cast<uint32_t>(m_height);
+    rp_begin.clearValueCount = m_renderPassClearValues.size();
+    rp_begin.pClearValues = m_renderPassClearValues.data();
+
+    // specifying VK_SUBPASS_CONTENTS_SECONDARY_COMMAND_BUFFERS means this
+    // render pass may
```
chrisforbes: pedantic, but: this subpass may...

tests/layer_validation_tests.cpp
```diff
@@ -716,9 +718,667 @@ class VkBufferTest {
...
+        aRenderFramework->GLSLtoSPV(aStageFlags, aShaderCode, spirv_code);
+        shader_module_create_info.pCode = spirv_code.data();
+        shader_module_create_info.codeSize = spirv_code.size() * sizeof(unsigned int);
+
+        VkShaderModule shader_module;
+        if (VK_SUCCESS == vkCreateShaderModule(VulkanDevice->GetDevice(),
+                                               &shader_module_create_info, nullptr,
+                                               &shader_module)) {
+            const unsigned shader_id = PipelineShaderStageInfo.size();
+            PipelineShaderStageInfo.resize(shader_id + 1);
```
chrisforbes: Why not `push_back`? resize-on-every-insert is generally a bad idea.

tests/layer_validation_tests.cpp
```diff
@@ -716,9 +718,667 @@ class VkBufferTest {
...
+                                  *aBlendAttachmentState = nullptr) {
+        RemoveColorBlendAttachmentState();
+        VkPipelineColorBlendAttachmentState *blend_attachments =
+                new VkPipelineColorBlendAttachmentState[aBlendAttachmentCount];
+        if (nullptr != blend_attachments) {
+            if (aBlendAttachmentState) {
+                memcpy(blend_attachments, aBlendAttachmentState,
+                       sizeof(VkPipelineColorBlendAttachmentState) *
+                       aBlendAttachmentCount);
+                ColorBlendCurrent = (true);
```
chrisforbes: Weird parens

tests/layer_validation_tests.cpp
```diff
@@ -716,9 +718,667 @@ class VkBufferTest {
...
+
+    bool IsStateCurrent() {
+        return ColorBlendCurrent && LayoutCurrent &&
+                RenderPassCurrent && ShadersCurrent;
+    }
+
+    void RemoveColorBlendAttachmentState() {
+        if (PipelineColorBlendCreateInfo.pAttachments) {
+            delete [] PipelineColorBlendCreateInfo.pAttachments;
+            PipelineColorBlendCreateInfo.pAttachments = 0;
```
chrisforbes: `nullptr`

tests/layer_validation_tests.cpp
```diff
@@ -716,9 +718,667 @@ class VkBufferTest {
...
+            }
+            return true;
+        }
+        return false;
+    }
+
+    void CompleteShaderStages(std::vector<VkPipelineShaderStageCreateInfo>
+                              &aPipelineShaderStageInfo) {
+        GraphicsPipeCreateInfo.stageCount = aPipelineShaderStageInfo.size();
+        GraphicsPipeCreateInfo.pStages = aPipelineShaderStageInfo.data();
```
chrisforbes: Stashing of pointers into caller-owned containers. I see you get away with it at the moment, but it's trouble if the test then accidentally invalidates them.

tests/layer_validation_tests.cpp
MarkMLunarG: This just has to do with the code guidelines being anti macro. I suppose the advantage to the macro is the occasional misspelling of description. How about an inline function to replace the macro?

chrisforbes: I really think just be consistent with the existing tests (and the other tests you're adding, which use TEST_DESCRIPTION).

mark-lunarg: All tests will have a TEST_DESCRIPTION.

tests/layer_validation_tests.cpp
MarkMLunarG: VkShaderObj requires a VkDeviceObj. The point of this exercise is to be able to create a pipeline that supports  a custom device (the first example being a device that was created with Independent Blend disabled), which is the VkCreateDevice test class

tests/layer_validation_tests.cpp
MarkMLunarG: yes. That is this macro: #define TEST_DESCRIPTION(desc) RecordProperty("description", desc). All I did was unroll the macro. This is really a precursor to a much larger problem which I should not be tackling here so I've changed this back

tests/layer_validation_tests.cpp
MarkMLunarG: Yes, I'd like to see this cleaned up too, but I'd prefer to leave that until later when things have matured a bit more. For now this seems to do OK as a quick and easy method

tests/layer_validation_tests.cpp
MarkMLunarG: I struggled with this because Vulkan depends heavily on transient create_info and other types of state structures, which the framework completely glosses over.  The new CreateState classes are intentionally designed to be volatile and valid only up to the point of creation of the object they are paired with. Plus, I have some concerns that having a copy of the data list could be more confusing then understanding that the CreateState classes are short lived. Ultimately, I decided that a protected opaque copy of the handles is probably OK.

MarkMLunarG:
Yesterday I pushed a new version that attempts to addresses Chris's review

PR #810, opened 2016-08-01
layers: GH650, Fix D/S image layout validation

mark-lunarg:
When an imageView of a depth/stencil image is used as a depth/stencil
framebuffer attachment, the aspectMask is ignored and both depth and
stencil image subresources are used. Validation was throwing errors
as only the specified aspectMask transition was tracked.

Change-Id: I4f45eec74040448ba3ed52b07268fab3b3cb9318

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -3592,6 +3592,12 @@ void SetLayout(const layer_data *dev_data, GLOBAL_CB_NODE *pCB, VkImageView imag
         for (uint32_t k = 0; k < subRange.layerCount; k++) {
             uint32_t layer = subRange.baseArrayLayer + k;
             VkImageSubresource sub = {subRange.aspectMask, level, layer};
+            // TODO: If ImageView was created with depth or stencil, transition both layouts as
+            // the aspectMask is ignored and both are used. Verify that the extra implicit layout
+            // is OK for descriptor set layout validation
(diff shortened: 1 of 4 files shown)


mark-lunarg:
We are working to pull together an LVT test case, but in the meantime I was hoping y'all could run this through the build farm.

layers/core_validation.cpp
```diff
@@ -3592,6 +3592,11 @@ void SetLayout(const layer_data *dev_data, GLOBAL_CB_NODE *pCB, VkImageView imag
         for (uint32_t k = 0; k < subRange.layerCount; k++) {
             uint32_t layer = subRange.baseArrayLayer + k;
             VkImageSubresource sub = {subRange.aspectMask, level, layer};
+            // If ImageView was created with depth or stencil, transition both layouts as
```
tobine: This should have a "TODO" caveat as the exception only applies to DS image bound as a framebuffer attachment (vs. something like a depth or stencil imageView for descriptor set resource).
In practice I don't *think* doing the transition of both aspects all the time as a workaround will cause any problems, as it's really just extra image state that should be ignored in most (hopefully all...) cases. However, for a "proper" fix we should really note a "TODO" comment that this could be cleaned up to be more precise.

mark-lunarg:
@tobine -- Made the suggested edit and added Tony's LVT test and supporting framework change.  Thanks.

tobine:

PR #813, opened 2016-08-02
layers: Fix object_tracker build

tobine:
This moves a static array of maps to be a vector of maps.
Building android from Windows crashes clang without this change.
clang in the NDK doesn't like arrays of unordered_maps.


lentinem:
+1

PR #815, opened 2016-08-03
tests: Modify ErrorMonitor for warn behavior

mark-lunarg:
For Positive tests, warnings and perf-warnings should not cause the test to fail.  This change makes the default ERROR only, and allows an individual LVT to specify exactly which flags are to be checked.

Using ExpectSuccess() is now configurable for custom combinations
of debug flags.

Change-Id: Ic0917cfa46fc15207d265d392020aadf2f66b675

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -106,6 +106,7 @@ myDbgFunc(VkFlags msgFlags, VkDebugReportObjectTypeEXT objType,
           uint64_t srcObject, size_t location, int32_t msgCode,
           const char *pLayerPrefix, const char *pMsg, void *pUserData);
 
+
 // ********************************************************
 // ErrorMonitor Usage:
 //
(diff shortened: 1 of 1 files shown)


tobine:
LGTM :+1:

chrisforbes:
I can see that you'd want the ability to customize the level, but I'm not really thrilled about the default being relaxed.

The existing positive tests were careful to be clean at all levels. The existing /negative/ tests have piles of problems where they sweep both warnings and unexpected errors under the rug.

What motivates this change?

mark-lunarg:
It is allowable to have warnings or performance warnings and having them
does not necessarily mean the tests are not running cleanly.  The positive
tests run without them currently but this is not a requirement. If folks
feel strongly about it, I can go through and change the current
ExpectSuccess calls to explicitly take warning/perf-warning as additional
args.

tobine:
I don't feel too strongly about it. I recommended ERROR only as the default because we advertise to users that WARN and PERF_WARN are allowed, but something to be aware of. If we make the default stricter that will catch any unexpected warnings that could then be handled as appropriate or the test could just be updated to explicitly set ERROR only if author knew that warnings were acceptable for that test.

PR #817, opened 2016-08-04
layers: Fix function interception list for param validation

chrisforbes:
A bunch of functions were implemented but not wired up to G*PA.


tobine:
LGTM :+1:

PR #823, opened 2016-08-04
Tobin gh821 multiple img mem mapping

### LINKED ISSUE #821: Core validation doesn't track multiple images bound to a single device memory block
Vulkan allows a single piece of device memory to be divided up and used for multiple things - in this case images.

The core validation layer only tracks one image (the last one bound) in its DEVICE_MEM_INFO structure. If you use multiple images per device memory block this can cause any number of false validation errors since the layer only knows about the last bound image.

tobine:
Fixes #821 
The first and last CLs are the important ones. Just some clean-up in-between.
Kill single "image" member from DEVICE_MEM_INFO struct as all bound images are now correctly tracked in object_bindings set.
Update the validation check at MapMemory time to verify that ANY images that overlap the mapped memory are in the correct image layout.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5347,7 +5347,6 @@ VKAPI_ATTR void VKAPI_CALL DestroyImage(VkDevice device, VkImage image, const Vk
         if (mem_info) {
             remove_memory_ranges(reinterpret_cast<uint64_t &>(image), img_node->mem, mem_info->imageRanges);
             clear_object_binding(dev_data, reinterpret_cast<uint64_t &>(image), VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT);
-            mem_info->image = VK_NULL_HANDLE;
         }
         // Remove image from imageMap
         dev_data->imageMap.erase(img_node->image);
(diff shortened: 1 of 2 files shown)


mark-lunarg:
:+1:

layers/core_validation.cpp
```diff
@@ -10030,32 +10030,42 @@ CmdExecuteCommands(VkCommandBuffer commandBuffer, uint32_t commandBuffersCount,
     if (!skip_call)
         dev_data->device_dispatch_table->CmdExecuteCommands(commandBuffer, commandBuffersCount, pCommandBuffers);
 }
+// Return true if given range intersects with offset and size, else false
+// Prereq : size > 0 and range->end - range->start > 0. Both of these cases should have already resulted
+//  in an error (During MapMemory and AllocateMemory respectively) so not checking them here
+static bool rangesIntersect(MEMORY_RANGE const *range, VkDeviceSize offset, VkDeviceSize size) {
+    auto range2_end = offset + size;
+    if ((range->start >= offset && range->start <= (range2_end)) || (range->end >= offset && range->end <= range2_end)) {
```
chrisforbes: We're collecting lots of weird implementations of this...

lentinem: Agreed. Why can't we use the already existing ones?

tobine: @lentinem @chrisforbes line numbers?

tobine:
There's some existing code to check intersection between linear and non-linear objects bound to the same allocation. I'll be integrating that code with this code in a future commit. Pushed this update for now.

PR #825, opened 2016-08-05
GH 818: Improve swapchain error for vkCreateSwapchainKHR()

ianelliottus:
When the the application calls vkCreateSwapchainKHR(), the
pCreateInfo->surface used must have been previously checked as supported,
by calling vkGetPhysicalDeviceSurfaceSupportKHR().  The error message needed
to be more clear about this.


ianelliottus:
Cherry-picked and pushed.

PR #826, opened 2016-08-05
Windows shared memory support for layers

mark-lunarg:
Adding Windows shared memory support to the layers.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -108,6 +108,7 @@ struct layer_data {
     VkInstance instance;
     unique_ptr<INSTANCE_STATE> instance_state;
 
+    layer_shared_memory_info shared_memory_info;
 
     debug_report_data *report_data;
     std::vector<VkDebugReportCallbackEXT> logging_callback;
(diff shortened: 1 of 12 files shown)


mark-lunarg:
@cnorthrop , can you please give this a run through the test farm?

stroyan:
What is the intended use of this shared memory?  Validation layers haven't ventured into inter-process communication before.  You will need an IPC mutex such as windows' CreateMutex or a pthread mutex set to PTHREAD_PROCESS_SHARED to protect references to shared memory such as `shared_data->reference_count++;`

cnorthrop:
We're getting some Android failures, sync'd with Mark offline.

mark-lunarg:
@stroyan -- Since this is not intended for cross-process sharing, do you still suggest that more robust protection is needed for this change?

stroyan:
While this is not intended for cross-process sharing, I don't see anything that will prevent different processes from sharing this memory.  You may need to use a unique name parameter to CreateFileMapping to avoid collisions.  You could concatenate the value of _getpid().

mark-lunarg:
Shelving this for now.

PR #838, opened 2016-08-10
layers: Validate Against Enabled Features in Parameter Validation Layer

mark-lunarg:
ParameterValidation was using the hardware supported feature set for
validation -- these are now replaced with the application-enabled
features at CreateDevice time.

Change-Id: If52d3ae3e55c69611e02437d33a651d88ba7043d

### DIFF
--- a/layers/parameter_validation.cpp
+++ b/layers/parameter_validation.cpp
@@ -1432,7 +1432,7 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDevices(VkInstance instance, uin
         if ((result == VK_SUCCESS) && (NULL != pPhysicalDevices)) {
             for (uint32_t i = 0; i < *pPhysicalDeviceCount; i++) {
                 layer_data *phy_dev_data = get_my_data_ptr(get_dispatch_key(pPhysicalDevices[i]), layer_data_map);
-                // Save the supported features for each physical device
+                // Save the default supported features for each physical device
                 VkLayerInstanceDispatchTable *disp_table = get_dispatch_table(pc_instance_table_map, pPhysicalDevices[i]);
                 disp_table->GetPhysicalDeviceFeatures(pPhysicalDevices[i], &(phy_dev_data->physical_device_features));
             }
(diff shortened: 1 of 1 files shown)


layers/parameter_validation.cpp
```diff
@@ -1432,7 +1432,7 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDevices(VkInstance instance, uin
         if ((result == VK_SUCCESS) && (NULL != pPhysicalDevices)) {
             for (uint32_t i = 0; i < *pPhysicalDeviceCount; i++) {
                 layer_data *phy_dev_data = get_my_data_ptr(get_dispatch_key(pPhysicalDevices[i]), layer_data_map);
-                // Save the supported features for each physical device
+                // Save the default supported features for each physical device
```
chrisforbes: `supported features` is still correct here, vs `enabled features`, but OK; no strong opinion.

layers/parameter_validation.cpp
```diff
@@ -1666,6 +1665,14 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDevice(VkPhysicalDevice physicalDevice,
             instance_dispatch_table->GetPhysicalDeviceProperties(physicalDevice, &device_properties);
             memcpy(&my_device_data->device_limits, &device_properties.limits, sizeof(VkPhysicalDeviceLimits));
             my_device_data->physical_device = physicalDevice;
+
+            // Replace hardware feature set with app-enabled features
+            layer_data *phy_dev_data = get_my_data_ptr(get_dispatch_key(physicalDevice), layer_data_map);
+            if (pCreateInfo->pEnabledFeatures) {
+                memcpy(&phy_dev_data->physical_device_features, &pCreateInfo->pEnabledFeatures, sizeof(VkPhysicalDeviceFeatures));
```
chrisforbes: Each side of this branch writes to a different structure. Is that what you intended?

chrisforbes: Extra `&` breaks this: `&pCreateInfo->pEnabledFeatures` is type `VkPhysicalDeviceFeatures**`, so you're actually copying the pointer value and whatever junk follows it in memory, rather than the struct you wanted.

Why not just copy the struct? This would work fine and be more obvious:

`phy_dev_data->physical_device_features = *pCreateInfo->pEnabledFeatures;`

mark-lunarg: Ugh, yes on both counts.

mark-lunarg:
Incorporated feedback and merged to Master.

commit 26e15dcde463371a8c1cd5ff1c68d583cb72ed56
Date:   Tue Aug 9 13:41:09 2016 -0600

    layers: Validate Enabled Features in PV

    ParameterValidation was using the hardware supported feature set for
    validation -- these are now replaced with the application-enabled
    features at CreateDevice time.

    Change-Id: If52d3ae3e55c69611e02437d33a651d88ba7043d

PR #839, opened 2016-08-10
Add renderpass tests gh427

Tony-LunarG:
RenderPass tests to cover INVALID_RENDER_AREA, INDEPENDENT_BLEND, and RENDERPASS_INVALID_DS_ATTACHMENT


tests/vktestbinding.h
```diff
@@ -149,7 +149,7 @@ class Device : public internal::Handle<VkDevice> {
     // vkCreateDevice()
     void init(const VkDeviceCreateInfo &info);
     void init(std::vector<const char *>
-                  &extensions); // all queues, all extensions, etc
+                  &extensions, VkPhysicalDeviceFeatures *features = NULL); // all queues, all extensions, etc
```
chrisforbes: nitpick: nullptr

tests/vkrenderframework.cpp
```diff
@@ -585,6 +585,34 @@ void VkDescriptorSetObj::CreateVKDescriptorSet(
     }
 }
 
+VkRenderpassObj::VkRenderpassObj(VkDeviceObj *dev) {
+    // Create a renderPass with a single color attachment
+    VkAttachmentReference attach = {};
+    attach.layout = VK_IMAGE_LAYOUT_GENERAL;
+
+    VkSubpassDescription subpass = {};
+    subpass.pColorAttachments = &attach;
```
chrisforbes: Do you want to set VkSubpassDescription::colorAttachmentCount to 1, so this gets used?

tests/vkrenderframework.cpp
```diff
@@ -585,6 +585,34 @@ void VkDescriptorSetObj::CreateVKDescriptorSet(
...
+
+    VkRenderPassCreateInfo rpci = {};
+    rpci.subpassCount = 1;
+    rpci.pSubpasses = &subpass;
+    rpci.attachmentCount = 1;
+
+    VkAttachmentDescription attach_desc = {};
+    attach_desc.format = VK_FORMAT_B8G8R8A8_UNORM;
```
chrisforbes: So this thing's load op is going to be LOAD, store op is going to be STORE, and both initial and final layouts are going to be UNDEFINED. That seems OK.

tests/layer_validation_tests.cpp
```diff
@@ -4112,6 +4112,102 @@ TEST_F(VkLayerTest, DepthStencilLayoutTransitionForDepthOnlyImageview) {
...
+
+    VkCommandPool cmd_pool;
+    VkCommandPoolCreateInfo pool_create_info{};
+    pool_create_info.sType = VK_STRUCTURE_TYPE_COMMAND_POOL_CREATE_INFO;
+    pool_create_info.queueFamilyIndex = noib_device.graphics_queue_node_index_;
+    pool_create_info.flags = VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT;
+
+    vkCreateCommandPool(noib_device.device(), &pool_create_info, nullptr,
+                        &cmd_pool);
+    VkCommandBufferObj *cmd_buffer =
```
chrisforbes: Don't need to allocate this using `new` and then have to remember to delete it -- on the stack will do?

tests/layer_validation_tests.cpp
```diff
@@ -4112,6 +4112,102 @@ TEST_F(VkLayerTest, DepthStencilLayoutTransitionForDepthOnlyImageview) {
...
+    vkDestroyCommandPool(noib_device.device(), cmd_pool, NULL);
+}
+
+TEST_F(VkLayerTest, RenderPassDepthStencilAttachmentUnused) {
+    TEST_DESCRIPTION("Specify no depth attachement in renderpass then specify "
+                     "depth attachments in subpass");
+
+    m_errorMonitor->SetDesiredFailureMsg(
+        VK_DEBUG_REPORT_ERROR_BIT_EXT,
+        "vkCreateRenderPass has no depth/stencil attachment, yet subpass");
```
chrisforbes: This is actually a pretty weird error -- is it the right thing? Surely you'll get an error anyway for trying to use a `B8G8R8A8_UNORM` image (which won't have the appropriate format features)  as a depth/stencil attachment.

tests/vkrenderframework.cpp
chrisforbes: Actually, this isn't quite OK, per last VU point for `VkAttachmentDescription`:

> finalLayout must not be VK_IMAGE_LAYOUT_UNDEFINED or VK_IMAGE_LAYOUT_PREINITIALIZED

tests/layer_validation_tests.cpp
```diff
@@ -4112,6 +4112,102 @@ TEST_F(VkLayerTest, DepthStencilLayoutTransitionForDepthOnlyImageview) {
...
+    ASSERT_NO_FATAL_FAILURE(InitState());
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                                         "Cannot execute a render pass with renderArea not within the bound of the framebuffer.");
+
+    // Framebuffer for render target is 256x256, exceed that for INVALID_RENDER_AREA
+    m_renderPassBeginInfo.renderArea.extent.width = 257;
+    m_renderPassBeginInfo.renderArea.extent.height = 257;
+    BeginCommandBuffer();
```
chrisforbes: This seems fairly obviously correct, but I think it's an example of a test that has all the interesting bits hidden in the plumbing...

tests/layer_validation_tests.cpp
Tony-LunarG: I needed cmd_buffer to go out of scope so its destructor, which calls FreeCommandBuffers, to be called before I called DestroyCommandPool.  I could use { } to control the scope - would that be preferable?

tests/layer_validation_tests.cpp
Tony-LunarG: I needed a non-depth image to get to the error I was trying to trigger:

    if (!depth_format_present) {
        // No depth attachment is present, validate that subpasses set depthStencilAttachment to VK_ATTACHMENT_UNUSED;
        for (uint32_t i = 0; i < pCreateInfo->subpassCount; i++) {
            if (pCreateInfo->pSubpasses[i].pDepthStencilAttachment &&
                pCreateInfo->pSubpasses[i].pDepthStencilAttachment->attachment != VK_ATTACHMENT_UNUSED)

tests/layer_validation_tests.cpp
```diff
@@ -4112,6 +4112,102 @@ TEST_F(VkLayerTest, DepthStencilLayoutTransitionForDepthOnlyImageview) {
...
+                     "attachements");
+    ASSERT_NO_FATAL_FAILURE(InitState());
+
+    m_errorMonitor->SetDesiredFailureMsg(
+        VK_DEBUG_REPORT_ERROR_BIT_EXT,
+        "Invalid Pipeline CreateInfo: If independent blend feature not "
+        "enabled, all elements of pAttachments must be identical");
+    VkPhysicalDeviceFeatures features = {};
+    features.independentBlend = VK_FALSE;
+    std::vector<const char *> extension_names;
```
MarkMLunarG: When the VkLayerTest test fixture creates a device for you (that's isn't being used, which I think is awkward*) it initializes the features this way:
    // request all supportable features enabled
    auto features = phy().features();
    dev_info.pEnabledFeatures = &features;

Is there a reason that you skipped that part?

* Since you're going to the trouble of improving the framework, have you considered adding a test fixture that doesn't create and initialize a bunch of unused bits? A lot of upcoming tests could use such a thing. Creating 2 devices and all the other baggage isn't in the scope of the testing.

Tony-LunarG: Yes, the framework created a device, commandPool, and commandBuffer that I didn't use, but that didn't seem like a lot in the deluge of Vulkan code that these tests require.  I probably could have plumbed the device features one more level back to InitState and moved device creation there, but so far, of the numerous tests written, the large majority have been fine with the default device, and I don't know if every test that uses the default device calls InitState or not.

MarkMLunarG: g_test provides the text fixture feature which is largely ignored by the layer validation tests. That feature could be used here to add another fixture that inherits from VkRenderFramework the same way the VkLayerTest fixture does. Then a different Setup function can be used that doesn't create "device, commandPool, commandBuffer and all that".  I see a lot more deluge in VkRenderFramework than in most of the tests, which could be eliminated via better use of the text fixtures.

tests/layer_validation_tests.cpp
chrisforbes: Right, I see. Sunk by the pool not being managed using a helper object the same way.

Slight preference for introducing a new scope to do it.

tests/layer_validation_tests.cpp
chrisforbes: Right, your code does what you intended; I'm just questioning whether the check you're targeting is actually sensible.

tests/layer_validation_tests.cpp
MarkMLunarG: We need a guard against a stream of water jokes before we all end up drowned in them

tests/layer_validation_tests.cpp
Tony-LunarG: I think I'll at least look at plumbing it up one more level to see what kind of havoc that would or wouldn't cause.  Mark L says that more feature validation is coming.

Tony-LunarG:
Ok, dynamic allocation changed to scoping - is it good to push?

tests/layer_validation_tests.cpp
chrisforbes: Can address that at a later point... I have no further objections to you pushing this.

chrisforbes:
:+1:

Tony-LunarG:
Cherry-Piked to master

PR #840, opened 2016-08-10
Github issue 790: Validate QueueFamilyIndices across CB creation and submission

mark-lunarg:


layers/core_validation.cpp
```diff
@@ -4682,6 +4682,42 @@ static bool validateCommandBufferState(layer_data *dev_data, GLOBAL_CB_NODE *pCB
...
+// Validate that queueFamilyIndices of primary and any secondary command buffers match this queue
+static bool validateQueueFamilyIndices(layer_data *dev_data, GLOBAL_CB_NODE *pCB, VkQueue queue) {
+    bool skip_call = false;
+    auto pPool = getCommandPoolNode(dev_data, pCB->createInfo.commandPool);
+    auto queue_node = getQueueNode(dev_data, queue);
+
+    if (pPool && queue_node && (pPool->queueFamilyIndex != queue_node->queueFamilyIndex)) {
+        skip_call |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
+            reinterpret_cast<uint64_t>(pCB->commandBuffer), __LINE__, DRAWSTATE_INVALID_QUEUE_FAMILY, "DS",
+            "vkQueueSubmit: Primary ommand buffer 0x%" PRIxLEAST64
```
chrisforbes: s/ommand/command/

layers/core_validation.cpp
```diff
@@ -4682,6 +4682,42 @@ static bool validateCommandBufferState(layer_data *dev_data, GLOBAL_CB_NODE *pCB
...
+            " created in queue family %d is being submitted on queue 0x%" PRIxLEAST64 " from queue family %d.",
+            reinterpret_cast<uint64_t>(pCB->commandBuffer), pPool->queueFamilyIndex,
+            reinterpret_cast<uint64_t>(queue), queue_node->queueFamilyIndex);
+    }
+
+    if (!pCB->secondaryCommandBuffers.empty()) {
```
chrisforbes: Is this the right time to do it, or can/should this be done at `CmdExecuteCommands` time instead?

chrisforbes: Extra check for empty around here seems spurious? Not a big deal though

layers/core_validation.cpp
```diff
@@ -4682,6 +4682,42 @@ static bool validateCommandBufferState(layer_data *dev_data, GLOBAL_CB_NODE *pCB
...
+        for (auto secondaryCmdBuffer : pCB->secondaryCommandBuffers) {
+            GLOBAL_CB_NODE *pSubCB = getCBNode(dev_data, secondaryCmdBuffer);
+            if (pSubCB) {
+                pPool = getCommandPoolNode(dev_data, pSubCB->createInfo.commandPool);
```
chrisforbes: Seems unfortunate that we have to chase this through the pool, but we do lots of crazy stuff at submit time...

chrisforbes:
:+1:

layers/core_validation.cpp
mark-lunarg: There is valid-usage language in CmdExecuteCommands for this, so I'll move the secondary-CB checking into validateSecondaryCommandBufferState and only check the primary at submit time -- great catch!  I'll make sure there's no check-for-empty in the new stuff.

layers/core_validation.cpp
mark-lunarg: Yes, maybe in the future...

mark-lunarg:
Pushed to Master.

PR #842, opened 2016-08-10
Remove Windows SDK powershell script and replace it with a C program

lenny-lunarg:
This pull request removes the ConfigLayersAndVulkanDLL.ps1 script, and replaces it with a new configure_runtime.c program.


windowsRuntimeInstaller/configure_runtime.c
```diff
@@ -0,0 +1,706 @@
...
+#include <string.h>
+#include <windows.h>
+
+// This hack gets Visual Studio 2013 to handle C99 stuff properly
+// If we drop support for 2013, it would be a good idea to remove this
+#if _MSC_VER < 1900
+#define inline __inline
+#define snprintf _snprintf
+#endif
+
```
MarkY-LunarG: One thing I would suggest is that if someone  compiles this accidentally with 64-bit set, it throws an error.

#ifdef WIN64
#error "This executable should always be compiled as 32-bit"
#endif 

That's likely not it exactly, but you get the gist.

windowsRuntimeInstaller/configure_runtime.c
```diff
@@ -0,0 +1,706 @@
...
+{
+    long major;
+    long minor;
+    long patch;
+    long build;
+    char extended[SDK_VERSION_BUFFER_SIZE];
+};
+
+const char* FLAG_ABI_MAJOR = "--abi-major";
+const char* FLAG_OS_SIZE = "--ossize";
```
MarkY-LunarG: I think FLAG_OS_SIZE can be removed.

windowsRuntimeInstaller/configure_runtime.c
```diff
@@ -0,0 +1,706 @@
...
+        fprintf(log, "Updating x86 explicit layers to path: %s\n", install_path);
+        break;
+    }
+
+    HKEY hKey;
+    REGSAM flags = KEY_ALL_ACCESS;
+    if(platform == PLATFORM_X64) {
+        flags |= KEY_WOW64_64KEY;
+    }
+    if(RegOpenKeyEx(HKEY_LOCAL_MACHINE, "SOFTWARE\\Khronos\\Vulkan\\ExplicitLayers",
```
MarkY-LunarG: For 32-bit layers on a 64-bit OS, the registry key is SOFTWARE\WOW6432Node\Khronos\Vulkan\ExplicitLayers.  Since you've disabled 64-redirect, you should have to manually set both SOFTWARE\Khronos\Vulkan\ExplicitLayers (64-bit layers) and SOFTWARE\WOW6432Node\Khronos\Vulkan\ExplicitLayers (32-bit layers)

windowsRuntimeInstaller/configure_runtime.c
```diff
@@ -0,0 +1,706 @@
...
+    }
+
+    bool removed_one;
+    do {
+        HKEY hKey;
+        REGSAM flags = KEY_ALL_ACCESS;
+        if(platform == PLATFORM_X64) {
+            flags |= KEY_WOW64_64KEY;
+        }
+        if(RegOpenKeyEx(HKEY_LOCAL_MACHINE, "SOFTWARE\\Khronos\\Vulkan\\ExplicitLayers",
```
MarkY-LunarG: See note above about 32-bit and 64-bit ExplicitLayer registry locations.

windowsRuntimeInstaller/configure_runtime.c
```diff
@@ -0,0 +1,706 @@
...
+        { fclose(log); });
+
+    // Update the explicit layers that are set in the windows registry
+    CHECK_ERROR_HANDLED(update_registry_layers(log, platform, &latest_runtime_version), { fclose(log); });
+
+    fclose(log);
+    return 0;
+}
+
+int add_explicit_layers(FILE* log, const char* install_path, enum Platform platform)
```
MarkY-LunarG: I think you need to differentiate here between 64-bit/32-bit OS and 64-bit/32-bit layers/registry.  Right now, you're treating them as one in the same.  But 64-bit OS has both 64 and 32-bit layers/registry.

windowsRuntimeInstaller/configure_runtime.c
lenny-lunarg: So, you would think that would be the case, but disabling the 64-redirect was actually disabling only the file system redirection - not the registry redirection. The registries will still redirect, unless you open the key with KEY_WOW64_64KEY, which you see set just above. As a result, when setting the 32-bit registries I let the filesystem redirection move it to WOW6432Node, but for 64-bit registries I disable the redirection.

windowsRuntimeInstaller/configure_runtime.c
lenny-lunarg: Yep, looks like I missed that when I removed that flag.

windowsRuntimeInstaller/configure_runtime.c
MarkY-LunarG: After talking, Lenny agreed to add a comment to clarify this above line 226.

windowsRuntimeInstaller/configure_runtime.c
```diff
@@ -0,0 +1,706 @@
...
+            return -4 + i;
+        } else if(*b_current < *a_current) {
+            return 4 - i;
+        }
+    }
+    
+    return strncmp(a->extended, b->extended, SDK_VERSION_BUFFER_SIZE);
+}
+
+int copy_file(const char* destination, const char* source)
```
MarkY-LunarG: I think it would be better to use a system call or CopyFile call than to manually read and copy contents.

windowsRuntimeInstaller/configure_runtime.c
```diff
@@ -0,0 +1,706 @@
...
+{
+    struct SDKVersion version;
+    unsigned windows_path_size = GetWindowsDirectory(NULL, 0); // Size includes null terminator
+    char* system_path = malloc(windows_path_size +
+        max_s(strlen(PATH_SYSTEM32), strlen(PATH_SYSWOW64)));
+    GetWindowsDirectory(system_path, windows_path_size);
+
+    strcpy(system_path + windows_path_size - 1, PATH_SYSTEM32);
+    fprintf(log, "Updating system directory: %s\n", system_path);
+    CHECK_ERROR_HANDLED(update_system_file(log, "vulkan-1", ".dll", system_path, abi_major, false,
```
MarkY-LunarG: the 1 in vulkan-1 is for the ABI.  If you support multiple ABIs, you might want to only pass in "vulkan-".

windowsRuntimeInstaller/configure_runtime.c
```diff
@@ -0,0 +1,706 @@
...
+        TCHAR name[COPY_BUFFER_SIZE];
+        DWORD nameSize = COPY_BUFFER_SIZE;
+        RegEnumKeyEx(hKey, i, name, &nameSize, NULL, NULL, NULL, NULL);
+        
+        if(strncmp("VulkanSDK", name, 9)) {
+            continue;
+        }
+        
+        HKEY subKey;
+        if(RegOpenKeyEx(hKey, name, 0, KEY_READ | KEY_WOW64_64KEY, &subKey) != ERROR_SUCCESS) {
```
MarkY-LunarG: You're always setting the WOW64 key here.  Is that correct even for a 32-bit system?

lenny-lunarg: On a 32-bit system, that flag gets ignored, since there is no 64 bit system. I believe that this is correct, because I believe that the SDK installs to the 64-bit registries only on a 64-bit system (and obviously the only registries on 32-bit). We should double check with David that this is the case. In any case, it's probably worth another clarifying comment here.

windowsRuntimeInstaller/configure_runtime.c
lenny-lunarg: CopyFile is the way to go. I did not realize that function existed from my 30 second google search.

davidlunarg:
The new lines added to InstallerRT.nsi end with ^M. Existing lines do not.

The functions in configure_runtime.c should have a comment header describing what the function does. I realize these descriptions are in the forward declarations at the top of file. I personally prefer the header with code, and a block of forward declarations at the top of the file, with a simple comment like
"// Forward declarations" for the whole block.

The README.txt file needs to be updated. It still mentions the powershell script.

How is the build of configure_runtime.exe done? Should LoaderAndValidationLayers/CMakeLists.txt file be changed so it is built? Or is it it built from the LunarG SDK build process? It seems all .exe files should be built when LVL is built.

The InstallerRT.nsi script still mentions the powershell script in comments. These comments need to be updated.

windowsRuntimeInstaller/configure_runtime.c
```diff
@@ -0,0 +1,706 @@
...
+        0, flags, &hKey) != ERROR_SUCCESS) {
+        return 20;
+    }
+
+    const char* pattern = platform == PLATFORM_X64 ? "%s\\Bin\\VkLayer*.json" : "%s\\Bin32\\VkLayer*.json";
+    int filter_size = snprintf(NULL, 0, pattern, install_path) + 1;
+    if(filter_size < 0) {
+        return 30;
+    }
+    char* filter = malloc(filter_size);
```
davidlunarg: There are several calls to malloc, return value is never checked.

windowsRuntimeInstaller/configure_runtime.c
```diff
@@ -0,0 +1,706 @@
...
+inline size_t max_s(size_t a, size_t b) { return a > b ? a : b; }
+inline size_t min_s(size_t a, size_t b) { return a > b ? a : b; }
+
+// Add the registry entries for all explicit layers
+//
+// log (input) - Logging file stream
+// install_path (input) - The installation path of the SDK which provides the layers
+// platform (input) - The platform to set the installation for (x64 or x86)
+// Returns: Zero on success, an error code on failure
+int add_explicit_layers(FILE* log, const char* install_path, enum Platform platform);
```
davidlunarg: The FILE *log arg appears in quite a few function param lists. Can it be removed and made a static global? The value of *log never changes.

windowsRuntimeInstaller/configure_runtime.c
```diff
@@ -0,0 +1,706 @@
...
+// This hack gets Visual Studio 2013 to handle C99 stuff properly
+// If we drop support for 2013, it would be a good idea to remove this
+#if _MSC_VER < 1900
+#define inline __inline
+#define snprintf _snprintf
+#endif
+
+#define COPY_BUFFER_SIZE (1024)
+#define CHECK_ERROR(statement) { int error = (statement); if(error) return error; }
+#define CHECK_ERROR_HANDLED(statement, handler) { int error = (statement); if(error) { { handler } return error; } }
```
davidlunarg: It seems the every call to this macro has { } around the last arg. Can the {} be removed?

windowsRuntimeInstaller/configure_runtime.c
```diff
@@ -0,0 +1,706 @@
...
+
+int parse_arguments(FILE* log, int argc, char** argv, long* abi_major)
+{
+    *abi_major = 0;
+
+    // Parse arguments
+    for(int i = 0; i < argc; ++i) {
+        if(!strcmp(argv[i], FLAG_ABI_MAJOR)) {
+            if(i + 1 == argc) {
+                fprintf(log, "ERROR: No value given for flag %s.\n", FLAG_ABI_MAJOR);
```
davidlunarg: Print a usage line of args are bad??

windowsRuntimeInstaller/README.txt
```diff
@@ -78,7 +78,7 @@ Some notes on the behavior of the Windows Vulkan Runtime Installer:
       the Vulkan runtime.  When it is run on a 32-bit version of
       Windows, it will install the 32 bit version of the Vulkan runtime.
 
-   o  The VulkanRT-<version>-Installer.exe created with the above steps
+   o  The VulkanRT-<version>-Installer.exe created with the above ste`
```
davidlunarg: Oops, "ps" was replaced with a '

Issue #785, opened 2016-07-21
polygonMode can not be VK_POLYGON_MODE_POINT or VK_POLYGON_MODE_LINE if fillModeNonSolid is not supported.


cdwfs:
When creating a graphics pipeline, the `parameter_validation` layer should ensure that the `VkPipelineRasterizationStateCreateInfo.polygonMode` does not specify an unsupported polygon mode, based on the value of `VkPhysicalDeviceFeatures.fillModeNonSolid`.

(I've written the new check, and will add a negative test & submit the pull request tomorrow if all goes well)

mark-lunarg:
Resolved by pr #843 .  Thanks all.

PR #800, opened 2016-07-27
Added check for invalid polygonMode when fillModeNonSolid=VK_FALSE

cdwfs:
I also added a new test (`VkLayerTest.PSOPolygonModeInvalid`), and fixed a random typo I spotted in another test's name.

In order to test the new check, I had to create a temporary `VkDevice` with the `fillModeNonSolid` feature forced to false. To this end, I added a new constructor to `VkDeviceObj` to take a set of physical device features to use, instead of always enabling all supported features. Behavior of the existing constructors is unchanged.

(My first PR to this repo; please let me know if I've missed anything!)

PR #843, opened 2016-08-10
Issue #785, Polygon nonSolidFillMode Enable Check

mark-lunarg:
For issue #785, added validation check for this feature, along with a test (note that the test depends on the test framework changes in PR #839, though the validation check itself does not.)

### DIFF
--- a/layers/parameter_validation.cpp
+++ b/layers/parameter_validation.cpp
@@ -2791,6 +2791,16 @@ bool PreCreateGraphicsPipelines(VkDevice device, const VkGraphicsPipelineCreateI
                         "unrecognized enumerator");
                 return false;
             }
+
+            if ((pCreateInfos->pRasterizationState->polygonMode & (VK_POLYGON_MODE_LINE | VK_POLYGON_MODE_POINT)) &&
+                (data->physical_device_features.fillModeNonSolid == false)) {
+                log_msg(
+                    mdd(device), VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0, __LINE__,
+                    DEVICE_FEATURE, LayerName,
+                    "vkCreateGraphicsPipelines parameter, VkPolygonMode pCreateInfos->pRasterizationState->polygonMode cannot be "
+                    "VK_POLYGON_MODE_POINT or VK_POLYGON_MODE_LINE if VkPhysicalDeviceFeatures->fillModeNonSolid is false.");
+                return false;
+            }
         }
 
         int i = 0;
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -1668,6 +1668,95 @@ TEST_F(VkLayerTest, IgnoreUnrelatedDescriptor) {
         vkFreeMemory(m_device->device(), buffer_memory, NULL);
     }
 }
+
+TEST_F(VkLayerTest, PSOPolygonModeInvalid) {
+    VkResult err;
+
+    TEST_DESCRIPTION("Attempt to use a non-solid polygon fill mode in a "
+        "pipeline when this feature is not enabled.");
+
+    ASSERT_NO_FATAL_FAILURE(InitState());
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+
+    std::vector<const char *> device_extension_names;
+    auto features = m_device->phy().features();
+    // Artificially disable support for non-solid fill modes
+    features.fillModeNonSolid = false;
+    // The sacrificial device object
+    VkDeviceObj test_device(0, gpu(), device_extension_names, &features);
+
+    VkRenderpassObj render_pass(&test_device);
+
+    VkPipelineLayoutCreateInfo pipeline_layout_ci = {};
+    pipeline_layout_ci.sType = VK_STRUCTURE_TYPE_PIPELINE_LAYOUT_CREATE_INFO;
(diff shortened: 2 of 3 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -1678,6 +1678,154 @@ TEST_F(VkLayerTest, IgnoreUnrelatedDescriptor) {
...
+
+    m_errorMonitor->SetDesiredFailureMsg(
+        VK_DEBUG_REPORT_ERROR_BIT_EXT,
+        "polygonMode cannot be VK_POLYGON_MODE_POINT or VK_POLYGON_MODE_LINE");
+
+    ASSERT_NO_FATAL_FAILURE(InitState());
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+
+    VkPhysicalDeviceFeatures features = {};
+    features.fillModeNonSolid = false;
```
chrisforbes: This isn't achieving anything, the struct is never used.

tests/layer_validation_tests.cpp
```diff
@@ -1678,6 +1678,154 @@ TEST_F(VkLayerTest, IgnoreUnrelatedDescriptor) {
...
+    dyn_state_ci.pDynamicStates = dynamic_states;
+
+    VkPipelineShaderStageCreateInfo shaderStages[2];
+    memset(&shaderStages, 0, 2 * sizeof(VkPipelineShaderStageCreateInfo));
+
+    VkShaderObj vs(m_device, bindStateVertShaderText,
+        VK_SHADER_STAGE_VERTEX_BIT, this);
+    VkShaderObj fs(m_device, bindStateFragShaderText,
+        VK_SHADER_STAGE_FRAGMENT_BIT,
+        this); // TODO - We shouldn't need a fragment shader
```
chrisforbes: This appears again... is it still true with production drivers?

tests/layer_validation_tests.cpp
```diff
@@ -1678,6 +1678,154 @@ TEST_F(VkLayerTest, IgnoreUnrelatedDescriptor) {
...
+    vi_ci.vertexBindingDescriptionCount = 0;
+    vi_ci.pVertexBindingDescriptions = nullptr;
+    vi_ci.vertexAttributeDescriptionCount = 0;
+    vi_ci.pVertexAttributeDescriptions = nullptr;
+
+    VkPipelineInputAssemblyStateCreateInfo ia_ci = {};
+    ia_ci.sType = VK_STRUCTURE_TYPE_PIPELINE_INPUT_ASSEMBLY_STATE_CREATE_INFO;
+    ia_ci.topology = VK_PRIMITIVE_TOPOLOGY_TRIANGLE_STRIP;
+
+    VkPipelineRasterizationStateCreateInfo rs_ci = {};
```
chrisforbes: This struct is really the only interesting part of the setup -- did you consider using VkPipelineObj::SetRasterization to override it rather than doing everything by hand?

tests/layer_validation_tests.cpp
chrisforbes: Best you can do at the moment is to inspect the actual features exposed, and skip if they're not what you want... not ideal, but the right hook into device creation doesn't exist yet

chrisforbes:
Does this completely supercede #800 ?

mark-lunarg:
@chrisforbes , I didn't notice the #800 until after this stuff was done.  The framework changes have already been done as part of another PR, and unless you have a preference it looks like it will ultimately supercede that PR -- if I can I will adapt one or more of those commits since it is such a treat to get such a complete submission.

mark-lunarg:
@chrisforbes -- Incorporated your feedback, and used @cdwfs commits as much as possible for this change.  Both of you guys should have a look and provide feedback.  Thanks!

layers/parameter_validation.cpp
```diff
@@ -2791,6 +2791,16 @@ bool PreCreateGraphicsPipelines(VkDevice device, const VkGraphicsPipelineCreateI
                         "unrecognized enumerator");
                 return false;
             }
+
+            if ((pCreateInfos->pRasterizationState->polygonMode & (VK_POLYGON_MODE_LINE | VK_POLYGON_MODE_POINT)) &&
```
chrisforbes: This works by accident - these aren't bit flags. Perhaps just `...->polygonMode != VK_POLYGON_MODE_FILL` ?

chrisforbes:
One issue with the layer change, otherwise big :+1: from me.

tests/layer_validation_tests.cpp
```diff
@@ -1668,6 +1668,95 @@ TEST_F(VkLayerTest, IgnoreUnrelatedDescriptor) {
...
+    rs_ci.pNext = nullptr;
+    rs_ci.lineWidth = 1.0f;
+    rs_ci.rasterizerDiscardEnable = true;
+
+    VkShaderObj vs(&test_device, bindStateVertShaderText,
+        VK_SHADER_STAGE_VERTEX_BIT, this);
+    VkShaderObj fs(&test_device, bindStateFragShaderText,
+        VK_SHADER_STAGE_FRAGMENT_BIT, this);
+
+    // Set polygonMode to unsupporte value POINT, should fail
```
cdwfs: nit: "unsupporte" -> "unsupported"

cdwfs:
Looks good to me as well, besides the typo I noted. Much cleaner than my original test as well. Thanks all!

mark-lunarg:
Thanks @chrisforbes and @cdwfs for the help.  Pushed to Master.

PR #845, opened 2016-08-11
Tobin gh605 mem valid per range

### LINKED ISSUE #605: vkCmdResolveImage generates invalid "Cannot read invalid memory..., please fill..." message
I'm converting a multisample image into a single sample one using vkCmdResolveImage, and getting an invalid validation layer error:

> vkCmdResolveImage(): Cannot read invalid memory 0x9, please fill the memory before using. (code: 12)

The previous render pass has two attachments, a multisample image (view) a multisample depth image (view):

```
  std::array<VkAttachmentDescription, 2> attachment_descriptions;
  // color image:
  attachment_descriptions[0].flags = 0;
  attachment_descriptions[0].format = /* some surface format */;
  attachment_descriptions[0].samples = VK_SAMPLE_COUNT_4_BIT;
  attachment_descriptions[0].loadOp = VK_ATTACHMENT_LOAD_OP_LOAD;
  attachment_descriptions[0].storeOp = VK_ATTACHMENT_STORE_OP_STORE;
  attachment_descriptions[0].stencilLoadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
  attachment_descriptions[0].stencilStoreOp = VK_ATTACHMENT_STORE_OP_DONT_CARE;
  attachment_descriptions[0].initialLayout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL;
  attachment_descriptions[0].finalLayout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL;

  // depth image:
  attachment_descriptions[1].flags = 0;
  attachment_descriptions[1].format = /* some depth format */;
  attachment_descriptions[1].samples = VK_SAMPLE_COUNT_4_BIT;
  attachment_descriptions[1].loadOp = VK_ATTACHMENT_LOAD_OP_CLEAR;
  attachment_descriptions[1].storeOp = VK_ATTACHMENT_STORE_OP_DONT_CARE; // <-- CAUSING ERROR
  attachment_descriptions[1].stencilLoadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
  attachment_descriptions[1].stencilStoreOp = VK_ATTACHMENT_STORE_OP_DONT_CARE;
  attachment_descriptions[1].initialLayout = VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL;
  attachment_descriptions[1].finalLayout = VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL;
```

Both views are later attached to a framebuffer and used in a render pass to draw something as usual, filling the whole image. The important part is, that I don't care for the depth image's content once the render pass is finished, so I set the storeOp of attachment 2 (index 1) to VK_ATTACHMENT_STORE_OP_DONT_CARE, while setting the color image's storeOp to VK_ATTACHMENT_STORE_OP_STORE. Once the render pass is done, I call vkCmdResolveImage (ignoring a few barriers in between). This generates the validation layer error from above.

When setting both storeOps to VK_ATTACHMENT_STORE_OP_STORE, the error is gone, even though the depth image's content isn't needed for resolving the multisample color image.

Both the color and the depth image are using the same memory object (0x9), but non-overlapping and well aligned regions within the memory object, respecting size and alignment of what vkGetImageMemoryRequirements says.

tobine:
This moves memory allocation valid tracking to be per-range instead of per allocation. Also improves alias tracking but lots more work to do in that department. It does fix a couple of errors in how invalid alias cases were being flagged.
This should fix #605 with current changes, but still need to fix the failing test.
Noted various holes with TODOs.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5297,8 +5297,8 @@ static bool validateIdleBuffer(const layer_data *my_data, VkBuffer buffer) {
 // pad_ranges bool indicates a linear and non-linear comparison which requires padding
 // In the case where padding is required, if an alias is encountered then a validation error is reported and skip_call
 //  may be set by the callback function so caller should merge in skip_call value if padding case is possible.
-static bool rangesIntersect(layer_data const *dev_data, MEMORY_RANGE const *range1, MEMORY_RANGE const *range2, bool &skip_call) {
-    skip_call = false;
+static bool rangesIntersect(layer_data const *dev_data, MEMORY_RANGE const *range1, MEMORY_RANGE const *range2, bool *skip_call) {
+    *skip_call = false;
     auto r1_start = range1->start;
     auto r1_end = range1->end;
     auto r2_start = range2->start;
@@ -5319,7 +5319,7 @@ static bool rangesIntersect(layer_data const *dev_data, MEMORY_RANGE const *rang
         const char *r2_linear_str = range2->linear ? "linear" : "non-linear";
         const char *r2_type_str = range2->image ? "image" : "buffer";
         auto obj_type = range1->image ? VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT : VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT;
-        skip_call |=
+        *skip_call |=
             log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, obj_type, range1->handle, 0, MEMTRACK_INVALID_ALIASING,
                     "MEM", "%s %s 0x%" PRIx64 " is aliased with %s %s 0x%" PRIx64
                            " which is in violation of the Buffer-Image Granularity section of the Vulkan specification.",
@@ -5337,7 +5337,7 @@ static bool rangesIntersect(layer_data const *dev_data, MEMORY_RANGE const *rang
     range_wrap.start = offset;
     range_wrap.end = end;
     bool tmp_bool;
-    return rangesIntersect(dev_data, range1, &range_wrap, tmp_bool);
+    return rangesIntersect(dev_data, range1, &range_wrap, &tmp_bool);
 }
 // For given mem_info, set all ranges valid that intersect [offset-end] range
 // TODO : For ranges where there is no alias, we may want to create new buffer ranges that are valid
@@ -5348,7 +5348,7 @@ static void SetMemRangesValid(layer_data const *dev_data, DEVICE_MEM_INFO *mem_i
     map_range.start = offset;
     map_range.end = end;
     for (auto &handle_range_pair : mem_info->bound_ranges) {
-        if (rangesIntersect(dev_data, &handle_range_pair.second, &map_range, tmp_bool)) {
+        if (rangesIntersect(dev_data, &handle_range_pair.second, &map_range, &tmp_bool)) {
             // TODO : WARN here if tmp_bool true?
             handle_range_pair.second.valid = true;
         }
@@ -5382,7 +5382,7 @@ static bool InsertMemoryRange(layer_data const *dev_data, uint64_t handle, DEVIC
     for (auto &obj_range_pair : mem_info->bound_ranges) {
         auto check_range = &obj_range_pair.second;
         bool intersection_error = false;
-        if (rangesIntersect(dev_data, &range, check_range, intersection_error)) {
+        if (rangesIntersect(dev_data, &range, check_range, &intersection_error)) {
             skip_call |= intersection_error;
             range.aliases.insert(check_range);
             tmp_alias_ranges.insert(check_range);
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5328,7 +5328,7 @@ static bool rangesIntersect(layer_data const *dev_data, MEMORY_RANGE const *rang
     // Ranges intersect
     return true;
 }
-// Simplified rangesIntersect that calls above function with limited params: base_address to start at and size to consider
+// Simplified rangesIntersect that calls above function to check range1 for intersection with offset & end addresses
 static bool rangesIntersect(layer_data const *dev_data, MEMORY_RANGE const *range1, VkDeviceSize offset, VkDeviceSize end) {
     // Create a local MEMORY_RANGE struct to wrap offset/size
     MEMORY_RANGE range_wrap;
@@ -10196,7 +10196,7 @@ CmdExecuteCommands(VkCommandBuffer commandBuffer, uint32_t commandBuffersCount, 
 
 // For any image objects that overlap mapped memory, verify that their layouts are PREINIT or GENERAL
 static bool ValidateMapImageLayouts(VkDevice device, DEVICE_MEM_INFO const *mem_info, VkDeviceSize offset,
-                                    VkDeviceSize end_address) {
+                                    VkDeviceSize end_offset) {
     bool skip_call = false;
     layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
     // Iterate over all bound image ranges and verify that for any that overlap the
@@ -10205,7 +10205,7 @@ static bool ValidateMapImageLayouts(VkDevice device, DEVICE_MEM_INFO const *mem_
     for (auto image_handle : mem_info->bound_images) {
         auto img_it = mem_info->bound_ranges.find(image_handle);
         if (img_it != mem_info->bound_ranges.end()) {
-            if (rangesIntersect(dev_data, &img_it->second, offset, end_address)) {
+            if (rangesIntersect(dev_data, &img_it->second, offset, end_offset)) {
                 std::vector<VkImageLayout> layouts;
                 if (FindLayouts(dev_data, VkImage(image_handle), layouts)) {
                     for (auto layout : layouts) {
@@ -10235,10 +10235,10 @@ MapMemory(VkDevice device, VkDeviceMemory mem, VkDeviceSize offset, VkDeviceSize
     if (mem_info) {
         // TODO : This could me more fine-grained to track just region that is valid
         mem_info->global_valid = true;
-        auto end_address = (VK_WHOLE_SIZE == size) ? mem_info->alloc_info.allocationSize - 1 : offset + size - 1;
-        skip_call |= ValidateMapImageLayouts(device, mem_info, offset, end_address);
+        auto end_offset = (VK_WHOLE_SIZE == size) ? mem_info->alloc_info.allocationSize - 1 : offset + size - 1;
+        skip_call |= ValidateMapImageLayouts(device, mem_info, offset, end_offset);
         // TODO : Do we need to create new "bound_range" for the mapped range?
-        SetMemRangesValid(dev_data, mem_info, offset, end_address);
+        SetMemRangesValid(dev_data, mem_info, offset, end_offset);
         if ((dev_data->phys_dev_mem_props.memoryTypes[mem_info->alloc_info.memoryTypeIndex].propertyFlags &
              VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT) == 0) {
             skip_call =
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5353,6 +5353,7 @@ static bool InsertMemoryRange(layer_data const *dev_data, uint64_t handle, DEVIC
     range.image = is_image;
     range.handle = handle;
     range.linear = is_linear;
+    range.valid = mem_info->global_valid;
     range.memory = mem_info->mem;
     range.start = memoryOffset;
     range.size = memRequirements.size;
@@ -10230,6 +10231,8 @@ MapMemory(VkDevice device, VkDeviceMemory mem, VkDeviceSize offset, VkDeviceSize
     std::unique_lock<std::mutex> lock(global_lock);
     DEVICE_MEM_INFO *mem_info = getMemObjInfo(dev_data, mem);
     if (mem_info) {
+        // TODO : This could me more fine-grained to track just region that is valid
+        mem_info->global_valid = true;
         auto end_address = (VK_WHOLE_SIZE == size) ? mem_info->alloc_info.allocationSize - 1 : offset + size - 1;
         skip_call |= ValidateMapImageLayouts(device, mem_info, offset, end_address);
         // TODO : Do we need to create new "bound_range" for the mapped range?
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5332,7 +5332,7 @@ static bool rangesIntersect(layer_data const *dev_data, MEMORY_RANGE const *rang
 static bool InsertMemoryRange(layer_data const *dev_data, uint64_t handle, DEVICE_MEM_INFO *mem_info, VkDeviceSize memoryOffset,
                               VkMemoryRequirements memRequirements, bool is_image, bool is_linear) {
     bool skip_call = false;
-    MEMORY_RANGE range; // range to be inserted and returned
+    MEMORY_RANGE range;
 
     range.image = is_image;
     range.handle = handle;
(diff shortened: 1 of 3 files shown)


layers/core_validation.cpp
```diff
@@ -5303,38 +5303,81 @@ static bool print_memory_range_error(layer_data *dev_data, const uint64_t object
...
 static MEMORY_RANGE InsertMemoryRange(uint64_t handle, DEVICE_MEM_INFO *mem_info, VkDeviceSize memoryOffset,
-                                      VkMemoryRequirements memRequirements,
-                                      std::unordered_map<uint64_t, uint32_t> &handle_index_map) {
-    MEMORY_RANGE range;
+                                      VkMemoryRequirements memRequirements, bool is_image) {
+    //                                      std::unordered_map<uint64_t, uint32_t> &handle_index_map) {
+    MEMORY_RANGE range; // range to be inserted and returned
+    std::unordered_map<uint64_t, uint32_t> handle_index_map;
+    if (is_image) {
+        handle_index_map = mem_info->image_to_index_map;
```
chrisforbes: This is a copy...

layers/core_validation.cpp
```diff
@@ -5303,38 +5303,81 @@ static bool print_memory_range_error(layer_data *dev_data, const uint64_t object
...
+        range.image = true;
+    } else {
+        handle_index_map = mem_info->buffer_to_index_map;
```
chrisforbes: ... and so is this ...

layers/core_validation.cpp
```diff
@@ -5303,38 +5303,81 @@ static bool print_memory_range_error(layer_data *dev_data, const uint64_t object
...
+    // Update Memory aliasing prior to inserting new range
+    // TODO : Take into consideration padding for image/buffer aliasing
+    for (auto check_range : mem_info->bound_ranges) {
+        if (rangesIntersect(&range, check_range.start, check_range.end - check_range.start)) {
+            range.aliases.insert(&check_range);
+            check_range.aliases.insert(&range);
+        }
+    }
     mem_info->bound_ranges.push_back(range);
     handle_index_map[handle] = mem_info->bound_ranges.size() - 1;
```
chrisforbes: ... so doing this modifies the local copy ...

layers/core_validation.cpp
```diff
@@ -5303,38 +5303,81 @@ static bool print_memory_range_error(layer_data *dev_data, const uint64_t object
...
     return range;
 }
```
chrisforbes: ... which evaporates here

layers/core_validation.cpp
```diff
@@ -5303,38 +5303,81 @@ static bool print_memory_range_error(layer_data *dev_data, const uint64_t object
...
+    if (is_image) {
+        handle_to_index_map = &mem_info->image_to_index_map;
+    } else {
+        handle_to_index_map = &mem_info->buffer_to_index_map;
+    }
+    auto index = (*handle_to_index_map)[handle];
+    auto erase_range = mem_info->bound_ranges[index];
+    for (auto alias_range : erase_range.aliases) {
+        alias_range->aliases.erase(&erase_range);
+        erase_range.aliases.erase(alias_range);
```
chrisforbes: This method of tracking aliasing doesn't really sit well. I think it needs some discussion

layers/core_validation.cpp
```diff
@@ -5303,38 +5303,81 @@ static bool print_memory_range_error(layer_data *dev_data, const uint64_t object
...
+//  map and clean up any aliases for range being removed.
+static void RemoveMemoryRange(uint64_t handle, DEVICE_MEM_INFO *mem_info, bool is_image) {
+    std::unordered_map<uint64_t, uint32_t> *handle_to_index_map;
+    if (is_image) {
+        handle_to_index_map = &mem_info->image_to_index_map;
+    } else {
+        handle_to_index_map = &mem_info->buffer_to_index_map;
+    }
+    auto index = (*handle_to_index_map)[handle];
+    auto erase_range = mem_info->bound_ranges[index];
```
chrisforbes: This is a copy ...

layers/core_validation.cpp
```diff
@@ -5303,38 +5303,81 @@ static bool print_memory_range_error(layer_data *dev_data, const uint64_t object
...
+    for (auto alias_range : erase_range.aliases) {
+        alias_range->aliases.erase(&erase_range);
```
chrisforbes: ... so taking its address here is the address of an object on the stack, which is unlikely to be useful to remove.

layers/core_validation_types.h
```diff
@@ -208,9 +208,12 @@ struct MemRange {
 
 struct MEMORY_RANGE {
     uint64_t handle;
+    bool image; // True for image, false for buffer
     VkDeviceMemory memory;
     VkDeviceSize start;
     VkDeviceSize end;
+    // Set of ptrs to every range aliased with this one
+    std::unordered_set<MEMORY_RANGE *> aliases;
```
chrisforbes: These are intended to be pointers into `DEVICE_MEM_INFO::bound_ranges`, which is a vector. Unfortunately, later inserts into that same vector invalidate them all...

layers/core_validation.cpp
```diff
@@ -5303,38 +5303,81 @@ static bool print_memory_range_error(layer_data *dev_data, const uint64_t object
...
+        handle_index_map = mem_info->buffer_to_index_map;
+        range.image = false;
+    }
     range.handle = handle;
     range.memory = mem_info->mem;
     range.start = memoryOffset;
     range.end = memoryOffset + memRequirements.size - 1;
+    // Update Memory aliasing prior to inserting new range
+    // TODO : Take into consideration padding for image/buffer aliasing
+    for (auto check_range : mem_info->bound_ranges) {
```
chrisforbes: (2) This is a copy...

layers/core_validation.cpp
```diff
@@ -5303,38 +5303,81 @@ static bool print_memory_range_error(layer_data *dev_data, const uint64_t object
...
+        if (rangesIntersect(&range, check_range.start, check_range.end - check_range.start)) {
+            range.aliases.insert(&check_range);
```
chrisforbes: (2) ... so this is inserting the address of a stack object, which isn't what you want.

layers/core_validation.cpp
```diff
@@ -5303,38 +5303,81 @@ static bool print_memory_range_error(layer_data *dev_data, const uint64_t object
...
+            check_range.aliases.insert(&range);
+        }
+    }
     mem_info->bound_ranges.push_back(range);
```
chrisforbes: (2) ... and here push_back potentially invalidates all the pointers you'd intended to insert into the aliases sets, as it may have to reallocate the backing memory to grow.

layers/core_validation.cpp
```diff
@@ -5303,38 +5303,81 @@ static bool print_memory_range_error(layer_data *dev_data, const uint64_t object
...
+    } else {
+        handle_to_index_map = &mem_info->buffer_to_index_map;
+    }
+    auto index = (*handle_to_index_map)[handle];
+    auto erase_range = mem_info->bound_ranges[index];
+    for (auto alias_range : erase_range.aliases) {
+        alias_range->aliases.erase(&erase_range);
+        erase_range.aliases.erase(alias_range);
+    }
+    mem_info->bound_ranges.erase(mem_info->bound_ranges.begin() + index);
```
chrisforbes: ... same problem as the push_back marked (2) above.

layers/core_validation.cpp
```diff
@@ -5272,86 +5272,99 @@ static bool validateIdleBuffer(const layer_data *my_data, VkBuffer buffer) {
...
+    return true;
 }
-
-static MEMORY_RANGE InsertMemoryRange(uint64_t handle, DEVICE_MEM_INFO *mem_info, VkDeviceSize memoryOffset,
-                                      VkMemoryRequirements memRequirements, bool is_image) {
-    //                                      std::unordered_map<uint64_t, uint32_t> &handle_index_map) {
+// Simplified rangesIntersect that calls above function with limited params
+static bool rangesIntersect(layer_data const *dev_data, MEMORY_RANGE const *range1, VkDeviceSize offset, VkDeviceSize size) {
+    // Create a local MEMORY_RANGE struct to wrap offset/size
+    MEMORY_RANGE range_wrap;
```
chrisforbes: Bit unfortunate that this temporary wrapper now contains an unordered_set... :(

layers/core_validation.cpp
```diff
@@ -5272,86 +5272,99 @@ static bool validateIdleBuffer(const layer_data *my_data, VkBuffer buffer) {
...
     MEMORY_RANGE range; // range to be inserted and returned
     std::unordered_map<uint64_t, uint32_t> handle_index_map;
-    if (is_image) {
-        handle_index_map = mem_info->image_to_index_map;
-        range.image = true;
-    } else {
-        handle_index_map = mem_info->buffer_to_index_map;
-        range.image = false;
-    }
+    handle_index_map = is_image ? mem_info->image_to_index_map : mem_info->buffer_to_index_map;
```
chrisforbes: This doesn't fix the complaint in previous commit -- you need ptr or ref. Converting to a single assignment like this does let you use ref though, so there's a win...

layers/core_validation.cpp
```diff
@@ -5333,8 +5333,6 @@ static bool InsertMemoryRange(layer_data const *dev_data, uint64_t handle, DEVIC
                               VkMemoryRequirements memRequirements, bool is_image, bool is_linear) {
     bool skip_call = false;
```
chrisforbes: Alright, this solves the bulk of the problems noted earlier.

layers/core_validation.cpp
```diff
@@ -5341,17 +5341,24 @@ static bool InsertMemoryRange(layer_data const *dev_data, uint64_t handle, DEVIC
...
+    range.aliases.clear();
+    // Update Memory aliasing
+    // Save aliase ranges so we can copy into final map entry below. Can't do it in loop b/c we don't yet have final ptr. If we
+    // inserted into map before loop to get the final ptr, then we may enter loop when not needed & we check range against itself
+    std::unordered_set<MEMORY_RANGE *> tmp_alias_ranges;
     for (auto &obj_range_pair : mem_info->bound_ranges) {
-        bool intersection_error = false;
         auto check_range = &obj_range_pair.second;
+        bool intersection_error = false;
         if (rangesIntersect(dev_data, &range, check_range, intersection_error)) {
```
chrisforbes: Per style guide, we shouldn't be using ref-to-nonconst as an out parameter. It's not immediately clear at this call site that `intersection_error` is written by `rangesIntersect`. Pass it by ptr instead?

layers/core_validation.cpp
```diff
@@ -5378,8 +5385,8 @@ static void RemoveMemoryRange(uint64_t handle, DEVICE_MEM_INFO *mem_info, bool i
     auto erase_range = &mem_info->bound_ranges[handle];
     for (auto alias_range : erase_range->aliases) {
         alias_range->aliases.erase(erase_range);
-        erase_range->aliases.erase(alias_range);
     }
+    erase_range->aliases.clear();
```
chrisforbes: Don't even need to do this-- it's going away in the next line.

layers/core_validation.cpp
```diff
@@ -5341,17 +5341,24 @@ static bool InsertMemoryRange(layer_data const *dev_data, uint64_t handle, DEVIC
     range.start = memoryOffset;
     range.size = memRequirements.size;
     range.end = memoryOffset + memRequirements.size - 1;
-    // Update Memory aliasing prior to inserting new range
+    range.aliases.clear();
```
chrisforbes: This /ought/ to be spurious, unless `MEMORY_RANGE` doesn't construct properly?

layers/core_validation.cpp
```diff
@@ -5341,17 +5341,24 @@ static bool InsertMemoryRange(layer_data const *dev_data, uint64_t handle, DEVIC
...
+    // Update Memory aliasing
+    // Save aliase ranges so we can copy into final map entry below. Can't do it in loop b/c we don't yet have final ptr. If we
+    // inserted into map before loop to get the final ptr, then we may enter loop when not needed & we check range against itself
+    std::unordered_set<MEMORY_RANGE *> tmp_alias_ranges;
     for (auto &obj_range_pair : mem_info->bound_ranges) {
-        bool intersection_error = false;
         auto check_range = &obj_range_pair.second;
+        bool intersection_error = false;
         if (rangesIntersect(dev_data, &range, check_range, intersection_error)) {
             skip_call |= intersection_error;
```
chrisforbes: It would be nice if the rest of the side effects were skipped if you're going to skip the call...

layers/core_validation.cpp
```diff
@@ -5310,18 +5312,32 @@ static bool rangesIntersect(layer_data const *dev_data, MEMORY_RANGE const *rang
     // Ranges intersect
     return true;
 }
-// Simplified rangesIntersect that calls above function with limited params
-static bool rangesIntersect(layer_data const *dev_data, MEMORY_RANGE const *range1, VkDeviceSize offset, VkDeviceSize size) {
+// Simplified rangesIntersect that calls above function with limited params: base_address to start at and size to consider
```
chrisforbes: These names `base_address`, `size` don't agree with the actual parameters, `offset` and `end`

layers/core_validation.cpp
```diff
@@ -5310,18 +5312,32 @@ static bool rangesIntersect(layer_data const *dev_data, MEMORY_RANGE const *rang
...
-
+// For given mem_info, set all ranges valid that intersect [offset-end] range
+// TODO : For ranges where there is no alias, we may want to create new buffer ranges that are valid
+static void SetMemRangesValid(layer_data const *dev_data, DEVICE_MEM_INFO const *mem_info, VkDeviceSize offset, VkDeviceSize end) {
+    bool tmp_bool = false;
+    MEMORY_RANGE map_range;
+    map_range.linear = true;
+    map_range.start = offset;
+    map_range.end = end;
+    for (auto handle_range_pair : mem_info->bound_ranges) {
```
chrisforbes: Probably `auto &`, this is doing a lot of copying...

layers/core_validation.cpp
```diff
@@ -10200,11 +10228,13 @@ MapMemory(VkDevice device, VkDeviceMemory mem, VkDeviceSize offset, VkDeviceSize
...
     VkResult result = VK_ERROR_VALIDATION_FAILED_EXT;
     std::unique_lock<std::mutex> lock(global_lock);
-#if MTMERGESOURCE
-    DEVICE_MEM_INFO *pMemObj = getMemObjInfo(dev_data, mem);
-    if (pMemObj) {
-        pMemObj->valid = true;
-        if ((dev_data->phys_dev_mem_props.memoryTypes[pMemObj->alloc_info.memoryTypeIndex].propertyFlags &
+    DEVICE_MEM_INFO *mem_info = getMemObjInfo(dev_data, mem);
+    if (mem_info) {
+        auto end_address = (VK_WHOLE_SIZE == size) ? mem_info->alloc_info.allocationSize - 1 : offset + size - 1;
```
chrisforbes: nitpick: end_offset? This isn't an address yet...

layers/core_validation_types.h
```diff
@@ -221,7 +222,8 @@ struct MEMORY_RANGE {
 // Data struct for tracking memory object
 struct DEVICE_MEM_INFO {
     void *object; // Dispatchable object used to create this memory (device of swapchain)
-    bool valid; // Stores if the color/depth/buffer memory has valid data or not
+    // bool valid; // Stores if the color/depth/buffer memory has valid data or not
+    // TODO : What to do with this guy in migrating valid to range?
     bool stencil_valid; // TODO: Stores if the stencil memory has valid data or not. The validity of this memory may ultimately need
```
chrisforbes: This is a bit of a headscratcher.

tobine:
Pushed some changes to address @chrisforbes feedback and also fixed the test issue that was causing crash on one test. This branch is now clean on the farm so I'd like to merge it if I can get an "all-clear" review.

chrisforbes:
@tobine I'd consider squashing some of this down as the intermediate states I made a bunch of noise about early in the series are /really/ busted.

Other than that, :+1: from me.

Issue #822, opened 2016-08-04
Core validation FlushMappedMemoryRanges and InvalidateMappedMemoryRanges don't handle mapped memory where size was VK_WHOLE_SIZE


gary-sweet:
If you map memory with a range size of VK_WHOLE_SIZE, subsequent calls to vkFlushMappedMemoryRanges and vkInvalidateMappedMemoryRanges also with VK_WHOLE_SIZE will overflow the shadow buffer and/or crash.

mark-lunarg:
As for #820 , we have a potential fix for this issue and believe that this memory type should be supported on our shield TV.  After we have a repro case on that HW we'll make sure it works.

mark-lunarg:
See PR #852.

mark-lunarg:
Fix pushed to Master.

commit eefe1df8a93c1cbfb1815488501c61d9e57515ea
Author: Mark Lobodzinski <mark@lunarg.com>
Date:   Mon Aug 15 15:00:12 2016 -0600

    layers: GH822, handle WHOLE_SIZE for mapped memory

    A support routine for FlushMappedMemoryRanges and InvalidateMapped-
    MemoryRanges ignored VK_WHOLE_SIZE, causing invalid memory accesses.

    Change-Id: I28616faa1339a394a3462e72780efd1b852ac489

Issue #820, opened 2016-08-04
Core validation needs to respect alignment constraints for MapMemory shadow copy


gary-sweet:
The core validation layer has a shadow copy of non-coherent mapped memory regions. When it mallocs the shadow region (in initializeAndTrackMemory) it does not respect the alignment constraints that exist for the real device memory (e.g. minMemoryMapAlignment, nonCoherentAtomSize) so the CPU pointer that is returned does not match the alignment constraints that are expected.

This is causing some of our tests (which validate the alignment of the returned pointer) to fail.

mark-lunarg:
Can you give us a summary of what platform/hardware you are on, and what the available memory types are for your setup?  The systems we have at-hand only appear to support HOST_VISIBLE AND COHERENT.

mark-lunarg:
Gary, I have a potential fix for this issue and believe that this memory type should be supported on our shield TV.  I need to work up a test case for it before we continue, but the fix should be coming along shortly.

gary-sweet:
This is our own internal driver development for our set-top-box platforms.

We have memory types which include (HOST_VISIBLE | HOST_COHERENT) but also (HOST_VISIBLE | HOST_CACHED).

Thanks
Gary

mark-lunarg:
See PR #852.

gary-sweet:
This fix doesn't quite work yet. See my comments in PR #852.

mark-lunarg:
Fixed by #852 .  Thanks!

Issue #819, opened 2016-08-04
Core validation broken when using non-coherent memory


gary-sweet:
The core validation layer uses a shadow copy of memory during MapMemory and FlushMappedMemoryRanges. It adds guard regions to the front and end so that it can detect buffer overwrites.

Unfortunately, it doesn't copy data back again during InvalidateMappedMemoryRanges. This means that any memory you read back does not reflect the current state of device memory.

tobine:
@karl-lunarg Assigning these memory issues to you. Re-assign as appropriate.

mark-lunarg:
Gary, can you try #852 ?  Also for issues #820 and #822.

mark-lunarg:
Fix pushed to Master.

commit 9e21e56fc23ade48efd3752d2e2729895eb90349
Author: Mark Lobodzinski <mark@lunarg.com>
Date:   Tue Aug 16 09:06:15 2016 -0600

    layers: GH819, Fix non-coherent memory shadow copy

    InvalidateMappedMemoryRanges was not updating core validation's
    shadow copy of non-coherent memory.

    Change-Id: I7b5e05791405871489014d6b08595c9bae941185

PR #852, opened 2016-08-16
Various non-coherent memory handling fixes

mark-lunarg:
Hopefully for #819 , #820 , and #822 , along with a test case that exposed the crashes on NCM systems.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -4906,34 +4906,53 @@ static bool deleteMemRanges(layer_data *my_data, VkDeviceMemory mem) {
                                 "Unmapping Memory without memory being mapped: mem obj 0x%" PRIxLEAST64, (uint64_t)mem);
         }
         mem_info->mem_range.size = 0;
-        if (mem_info->p_data) {
-            free(mem_info->p_data);
-            mem_info->p_data = 0;
+        if (mem_info->shadow_copy) {
+            free(mem_info->shadow_copy_base);
+            mem_info->shadow_copy_base = 0;
+            mem_info->shadow_copy = 0;
         }
     }
     return skip_call;
 }
 
+// Guard value for pad data
 static char NoncoherentMemoryFillValue = 0xb;
 
-static void initializeAndTrackMemory(layer_data *dev_data, VkDeviceMemory mem, VkDeviceSize size, void **ppData) {
+static void initializeAndTrackMemory(layer_data *dev_data, VkDeviceMemory mem, VkDeviceSize offset, VkDeviceSize size,
+                                     void **ppData) {
     auto mem_info = getMemObjInfo(dev_data, mem);
     if (mem_info) {
         mem_info->p_driver_data = *ppData;
         uint32_t index = mem_info->alloc_info.memoryTypeIndex;
         if (dev_data->phys_dev_mem_props.memoryTypes[index].propertyFlags & VK_MEMORY_PROPERTY_HOST_COHERENT_BIT) {
-            mem_info->p_data = 0;
+            mem_info->shadow_copy = 0;
         } else {
             if (size == VK_WHOLE_SIZE) {
-                size = mem_info->alloc_info.allocationSize;
+                size = mem_info->alloc_info.allocationSize - offset;
             }
-            size_t convSize = (size_t)(size);
-            mem_info->p_data = malloc(2 * convSize);
-            memset(mem_info->p_data, NoncoherentMemoryFillValue, 2 * convSize);
-            *ppData = static_cast<char *>(mem_info->p_data) + (convSize / 2);
+            mem_info->shadow_pad_size = dev_data->phys_dev_properties.properties.limits.minMemoryMapAlignment;
+            assert(vk_safe_modulo(mem_info->shadow_pad_size,
+                                  dev_data->phys_dev_properties.properties.limits.minMemoryMapAlignment) == 0);
+            // Ensure start of mapped region reflects hardware alignment constraints
+            uint64_t map_alignment = dev_data->phys_dev_properties.properties.limits.minMemoryMapAlignment;
+
+            // From spec: (ppData - offset) must be aligned to at least limits::minMemoryMapAlignment.
+            uint64_t start_offset = offset % map_alignment;
+            // Data passed to driver will be wrapped by a guardband of data to detect over- or under-writes.
+            mem_info->shadow_copy_base = malloc(2 * mem_info->shadow_pad_size + size + map_alignment + start_offset);
+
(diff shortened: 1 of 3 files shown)


mark-lunarg:
Gary, if you can, please test these on your setup there.  I hacked up the layers to do some testing but we won't have an actual non-coherent memory system until later this week.

layers/core_validation.cpp
```diff
@@ -5008,10 +5008,15 @@ static void initializeAndTrackMemory(layer_data *dev_data, VkDeviceMemory mem, V
...
-            size_t convSize = (size_t)(size);
-            mem_info->p_data = malloc(2 * convSize);
-            memset(mem_info->p_data, NoncoherentMemoryFillValue, 2 * convSize);
-            *ppData = static_cast<char *>(mem_info->p_data) + (convSize / 2);
+            size_t map_size = (size_t)(size);
+
+            // Ensure start of mapped region reflects hardware alignment constraints
+            size_t map_alignment = dev_data->phys_dev_properties.properties.limits.minMemoryMapAlignment;
+            mem_info->p_data = malloc(2 * map_size + map_alignment);
+            mem_info->alignment_offset = reinterpret_cast<uintptr_t>(mem_info->p_data) % map_alignment;
```
tobine: It's a bit clunky to carry the alignment_offset around to continually increment p_data and I'm afraid it will be error-prone.

I think the only place we'll need the original p_data after allocation is when the memory is freed in deleteMemRanges().
One alternative is to increment p_data from the start (as soon as alignment_offset is calculated). That then avoids all of the future additions of alignment_offset, but instead when the memory is freed p_data will need to be decremented by alignment_offset. That's still not great but I'd prefer that over this current code.

Another option would be to not store the alignment offset, but instead store "p_data" (allocation address plus alignment offset) and "p_data_base" (allocation address). Then p_data is used everywhere except when memory is freed and "p_data_base" would be used. I don't have a preference for this option, just putting it out as another alternative.

tests/layer_validation_tests.cpp
```diff
@@ -2181,6 +2182,21 @@ TEST_F(VkLayerTest, InvalidMemoryMapping) {
     vkFlushMappedMemoryRanges(m_device->device(), 1, &mmr);
     m_errorMonitor->VerifyFound();
 
+    // Map using WHOLE_SIZE, then Flush and Invalidate using same.
```
tobine: I'm not a fan of including this expected passing test as part of "InvalidMemoryMapping" test.
Should be broken out into a separate test (best option IMO) or the test should be renamed to account for increased scope.

layers/core_validation.cpp
mark-lunarg: I prefer the option of saving the pointer and the base address.  I'll do the rework, but Gary can do the testing with what he has, if he is able.

tobine:
With those couple clean-ups I noted 
LGTM :+1:

gary-sweet:
Ok - I've given this a test now.

#819 looks fixed.
#822 looks fixed.

#820 has a bug:

You have aligned the size of the start of the memory block, but you are returning that plus the size of the front guard band (size/2), which then misaligns what you are returning again. You need to make sure the address returned is aligned, not the start of the guard banded block.

If you add "assert(reinterpret_cast<uintptr_t>(*ppData) & (map_alignment - 1) == 0);" at the end of the function, this would trap any mistakes in the arithmetic.

mark-lunarg:
Yeah, we had seen that yesterday and I added a fix along with the changes for Tobin's review suggestions.  I'll add the assert as well and have something for you to try shortly.

mark-lunarg:
I reworked 822 pretty severely, changing the pad size from half the allocation size to just the granularity size, added the assert as Gary suggested.  Please have another look at this commit if you get a chance.

layers/core_validation.cpp
```diff
@@ -5003,18 +5005,30 @@ static void initializeAndTrackMemory(layer_data *dev_data, VkDeviceMemory mem, V
         mem_info->p_driver_data = *ppData;
         uint32_t index = mem_info->alloc_info.memoryTypeIndex;
         if (dev_data->phys_dev_mem_props.memoryTypes[index].propertyFlags & VK_MEMORY_PROPERTY_HOST_COHERENT_BIT) {
-            mem_info->p_data = 0;
+            mem_info->shadow_copy_base = 0;
+            mem_info->shadow_copy = 0;
         } else {
+            uint64_t shadow_pad_size = dev_data->phys_dev_properties.properties.limits.minMemoryMapAlignment;
```
tobine: Confusing that you declare "shadow_pad_size" up here apart from map_alignment below. It appears that this var is more for convenience as it has same value as map_alignment. Would be nice to declare it below with a comment. If it's truly the same as map_alignment throughout it could either be eliminated or assigned directly from map_alignment.

tests/layer_validation_tests.cpp
```diff
@@ -2197,6 +2197,72 @@ TEST_F(VkLayerTest, InvalidMemoryMapping) {
     vkFreeMemory(m_device->device(), mem, NULL);
 }
 
+TEST_F(VkLayerTest, NonCoherentMemoryMapping) {
+
+    TEST_DESCRIPTION("Ensure that validations handling of non-coherent memory "
+                     "mapping while using VK_WHILE_SIZE does not cause access "
```
tobine: WHILE -> WHOLE

gary-sweet:
Almost, but not quite I'm afraid.

If I call vkMapMemory with an offset that is non-zero, I get the start address of the shadow block back. The offset has been ignored.

layers/core_validation.cpp
mark-lunarg: I purposely decoupled them so that we could easily change the pad size, so I'd like to keep them separate but will move the line and add comment.

mark-lunarg:
OK, updated for feedback and hopefully addressed the ignored offsets.  Gary, can you please give this another shot?  Sorry to keep asking, but our Android build guy has been difficult to pin down lately.

gary-sweet:
This is passing all of my tests here now.
Thanks

mark-lunarg:
Thanks again, and keep the issues coming.  Pushed to Master.

commit 5f06b9eba1bf89d4066ca44120dcb72c4cf44104
Date:   Mon Aug 15 14:27:26 2016 -0600

    layers: GH820, MapMemory shadow copy alignment

    For host-visible non-coherent memory, the validation layers use a
    shadow copy of the mapped memory region. Added proper alignment
    for the shadow copy, fixed handling of offsets, and implemented
    resolve of the shadow data on invalidate calls.

    Change-Id: I88815799865d3a30c76e16cf96ace630e1bbcf2d

PR #854, opened 2016-08-16
LX581, Warn if CmdBlitImage Extents are NULL

mark-lunarg:
Extents work differently for this API than others, Id says it bit them and they see it in other places in the wild (apparently, setting the z_extent to zero on Nvidia works while AMD respects the coordinates).

### DIFF
--- a/layers/image.cpp
+++ b/layers/image.cpp
@@ -1033,6 +1033,28 @@ CmdBlitImage(VkCommandBuffer commandBuffer, VkImage srcImage, VkImageLayout srcI
     bool skipCall = false;
     layer_data *device_data = get_my_data_ptr(get_dispatch_key(commandBuffer), layer_data_map);
 
+    // Warn for zero-sized regions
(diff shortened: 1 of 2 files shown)


tobine:
The other thing I thought of is warning in cases like srcOffset[1].x > srcOffset[0].x but spec is imprecise and seems to allow for either array element to be greater.
Minor ? on the missing "&" in casts but otherwise passing on the farm and
LGTM :+1:

mark-lunarg:
Yeah, spec implies (through omission) that src or dstOffsets[1] can be less than [0].  This appears to be the only spot where they pulled this [0]-[1] thing, using extents everywhere else which makes it more confusing.  Also, in other places, Z is ignored for 2-D ops but that doesn't seem to be the case here.  

The & in the casts is only for non-dispatchable handles.  Dispatchable ones are really a pointer.  Thanks for all your patience @tobine .

mark-lunarg:
Pushed to Master.

PR #856, opened 2016-08-17
WIP: Testing hand-coded version of validation error messages update

tobine:
This is initial, simple prototype to demonstrate potential updates to validation error
reporting flow that tie the error messages to the spec.

This is just intended for review and comments very early in the process.

Key Benefits Attempting to add:
1. Unique ids for all validation messages
2. Tie validation messages directly to spec (including spec reference in actual message)
3. Have simple way to track progress toward goal of implementing all validation checks
4. Easily track exactly which checks have tests to verify them

Questions/Issues
1. How to handle validation requirements that aren't valid usage? Do we need new tag in spec or develop solution that augments spec?
2. How to handle validation checks that slightly change?
3. Best codegen strategy? Python parsing of xml?


MarkMLunarG:
FWIW, when I was giving this consideration some time back, I was envisioning using C++ code to generate a requirements database from the spec and then using C++ to generate the associated error text, codes, and related information for both core_validation and validation_later_tests. I've used this technique with some success before for managing register spec information for a plethora of GPUs. Some possible advantages to a database approach include:
1. Easy centralized storage of related information and meta data (knowledge base, common issues, urls, and proper usage examples accessible through LunarX or eq)
2. Simpler concurrent management of multiple khronos versions over time
3. Internationalization could be supported and open source contributors could provide localization
4. Better facilitates closed loop tracking and verification of coverage from spec to test to core validation coverage
5. Basic GUI tools could be developed to maintain and update the stored information
6. Handling of special cases can be handled through proper table design and interface tools
7. Lots of good SQL drivers are available via Qt and other sources

I some experience with this from past projects if there's any interest in using this approach.

spec.py
```diff
@@ -0,0 +1,146 @@
...
+                                #print "dict contents: %s:" % (self.val_error_dict)
+        #print "Validation Error Dict has a total of %d unique errors and contents are:\n%s" % (unique_enum_id, self.val_error_dict)
+    def genHeader(self, header_file):
+        """Generate a header file based on the contents of a parsed spec"""
+        file_contents = []
+        file_contents.append('// Copyright here')
+        file_contents.append('#pragma once')
+        file_contents.append('// Comment on enum declaration')
+        enum_decl = ['enum UNIQUE_VALIDATION_ERROR_CODE {']
+        error_string_map = ['std::unordered_map<int, char const *const> validation_error_map{']
```
chrisforbes: Are these guys ever actually sparse, or could this be an array?

tobine:
After investigating a few possible solutions I've gone ahead and for a first pass I'm just generating the unique error enum values and strings from the xhtml spec that's generated from the core API xml file.

I use the xhtml spec so I can get correct section names and links into the spec URL for each error.

What I've checked in here is still rough, but demonstrates the general contents of the generated header file as well as how the unique error enum and output message can be integrated for a single error.
There are 1634 valid usage checks captured, which should be all that are in the current spec.
Error enums strings are now of the format "VALIDATION_ERROR_<SPEC_SECTION>_<UNIQUE_ID> = <UNIQUE_ID>".
Each of those unique enum values is then used as the key in a map to lookup a unique error string that includes the valid usage text along with a link into the spec section containing that text. The idea is that this error string should be appended to the end of any error messages for that unique error enum.

There's still a ton of ?s about how to best use/maintain this flow. Here's a few

1. What's the best way to integrate all of the new error codes and messages into the existing checks? We can certainly start doing them by hand but for repeated generated checks like in object_tracker and parameter_checker is there an nice way to automate the updates?
2. How to capture the checks that don't currently have "valid usage" line items? Should all validation errors have a valid usage line item?
3. How to best maintain the error codes in the long-term? As new error codes are added, we can simply append them to the end which will work ok, but in that case the codegen needs to be much smarter. Best solution is potentially to embed them into the spec so codegen can stay simple and source will never change.

spec.py
tobine: Could definitely be an array right now. Assuming we define unique ids as I have here and then just keep growing list from the end, array is fine permanent solution.
That makes sense to me but I can also see potential solution where unique ids are sparse and partitioned into "categories" where each covers a range of numbers and new errors are added into separate ranges at different rates. Categories could cover not only types of checks or spec sections, but things like warnings if/when we pull them into the same framework.

mark-lunarg:
Hey @tobine, the enums have the form 'VALIDATION_ERROR_3_1_0' where the 3_1 refers to chapter 3 section 1.  The enums themselves should probably not refer to any location, just an ID (which, ideally, should get generated by and embedded in vk.xml or the plain-language valid usage sections of the spec).   Here, if anything at all gets moved these chapter/section will be incorrect forever since the enums cannot change.

tobine:
Pushed a number of updates to the scripting stuff. Layer use of enums is broken for a bit on this branch as I'm only focusing on scripts.
Changes error enum format to just be VALIDATION_ERROR_ID
Can now generate a simple database file that contains the error enum, error string, as well as fields for testname and whether or not check is in the code
From a given database file I can parse a spec file and make sure that errors that have moved around in the spec get assigned to the same ID. In instances where it seems like an error message may have changed, I flag those cases for manual verification.

Tested this out between older spec and spec based off of branch 1600 which has a number of updates. Appears to work well with this limited dataset.

tobine:
Replaced by #1019

Issue #862, opened 2016-08-20
Build errors of demos ( cube, vulkaninfo) on the wayland only window system.


elongbug:
we need to fix build errors of demos ( cube, vulkaninfo) on the wayland only window system.

cube and vulkaninfo  check only  "__linux__" macro for " #include <X11/Xutil.h> "

so we need to change the checking of this macro as "VK_USE_PLATFORM_XLIB_KHR" and "VK_USE_PLATFORM_XCB_KHR".

sl1pkn07:
In linux:

```
Scanning dependencies of target cube
[ 70%] Building C object demos/CMakeFiles/cube.dir/cube.c.o
/tmp/makepkg/vulkan-git/src/Vulkan-LoaderAndValidationLayers/demos/cube.c: In function 'demo_create_window':
/tmp/makepkg/vulkan-git/src/Vulkan-LoaderAndValidationLayers/demos/cube.c:2288:9: error: 'struct demo' has no member named 'window'; did you mean 'width'?
     demo->window = wl_compositor_create_surface(demo->compositor);
         ^~
/tmp/makepkg/vulkan-git/src/Vulkan-LoaderAndValidationLayers/demos/cube.c:2288:53: error: 'struct demo' has no member named 'compositor'
     demo->window = wl_compositor_create_surface(demo->compositor);
                                                     ^~
/tmp/makepkg/vulkan-git/src/Vulkan-LoaderAndValidationLayers/demos/cube.c:2289:14: error: 'struct demo' has no member named 'window'; did you mean 'width'?
     if (!demo->window) {
              ^~
/tmp/makepkg/vulkan-git/src/Vulkan-LoaderAndValidationLayers/demos/cube.c:2295:9: error: 'struct demo' has no member named 'shell_surface'; did you mean 'surface'?
     demo->shell_surface = wl_shell_get_shell_surface(demo->shell, demo->window);
         ^~
/tmp/makepkg/vulkan-git/src/Vulkan-LoaderAndValidationLayers/demos/cube.c:2295:58: error: 'struct demo' has no member named 'shell'
     demo->shell_surface = wl_shell_get_shell_surface(demo->shell, demo->window);
                                                          ^~
/tmp/makepkg/vulkan-git/src/Vulkan-LoaderAndValidationLayers/demos/cube.c:2295:71: error: 'struct demo' has no member named 'window'; did you mean 'width'?
[...]

PR #863, opened 2016-08-20
demos: GH862 Fix build errors of demos (cube, vulkaninfo) on the wayl…

elongbug:
…and only window system

Signed-off-by: Mun Gwan-gyeong <elongbug@gmail.com>

### DIFF
--- a/demos/cube.c
+++ b/demos/cube.c
@@ -29,7 +29,7 @@
 #include <stdbool.h>
 #include <assert.h>
 #include <signal.h>
-#if defined(__linux__) && !defined(ANDROID)
+#if defined(VK_USE_PLATFORM_XLIB_KHR) || defined(VK_USE_PLATFORM_XCB_KHR)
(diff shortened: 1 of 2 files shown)


elongbug:
this patch fixes #862

demos/cube.c
```diff
@@ -29,7 +29,7 @@
 #include <stdbool.h>
 #include <assert.h>
 #include <signal.h>
-#if defined(__linux__) && !defined(ANDROID)
+#if defined(VK_USE_PLATFORM_XLIB_KHR) | defined(VK_USE_PLATFORM_XCB_KHR)
```
MarkY-LunarG: Both these should use the boolean or || not the bitwise or |.

elongbug: Thanks for your comments. I have updated with your guide.

Tony-LunarG:
Cherry picked to master - thanks

PR #866, opened 2016-08-22
Add validation for shader use of input attachments

chrisforbes:


layers/vk_validation_layer_details.md
```diff
@@ -137,6 +137,8 @@ It flags errors when inconsistencies are found across interfaces between shader
 | Descriptor type mismatch | Flags error if a descriptor type does not match the shader resource type. | DESCRIPTOR_TYPE_MISMATCH | vkCreateGraphicsPipelines | TODO | NA |
 | Feature not enabled | Flags error if a capability declared by the shader requires a feature not enabled on the device | FEATURE_NOT_ENABLED | vkCreateGraphicsPipelines | TODO | NA |
 | Bad capability | Flags error if a capability declared by the shader is not supported by Vulkan shaders | BAD_CAPABILITY | vkCreateGraphicsPipelines | TODO | NA |
+| Missing input attachment | Flags error if shader consumes an input attachment which is not provided in the subpass in which a pipeline will be used | MISSING_INPUT_ATTACHMENT | vkCreateGraphicsPipelines | CreatePipelineInputAttachment* | NA |
```
tobine: Wow, I double-checked doc validator and see that it actually has very simplistic wildcard handling. Yay.

layers/core_validation.cpp
```diff
@@ -1550,6 +1550,38 @@ static void collect_interface_by_location(shader_module const *src, spirv_inst_i
     }
 }
 
+static void collect_interface_by_input_attachment_index(debug_report_data *report_data, shader_module const *src,
+                                                        std::unordered_set<uint32_t> const &accessible_ids,
+                                                        std::vector<std::pair<uint32_t, interface_var>> &out) {
```
tobine: Non-const parameter passed by ref against style guide. I see that "collect_interface_by_descriptor_slot" below also does this but we should update these to ptrs to be consistent w/ style guide.

tobine:
Nice tests.
Just minor nit on non-const param by ref, otherwise
LGTM :+1:

chrisforbes:
Alright, there's a handful of other cases of that in SC. I'll fix them all together in a followup (they can actually just return the container)

chrisforbes:
Picked to master; followup filed as #871

PR #868, opened 2016-08-22
layers: Update descriptor binding to use an ordered map

tobine:
When dealing with dynamic offsets, the binding numbers, and array indices within
a binding, are all sequential.
This is according to the spec "Descriptor Set Binding" section which states,
"...entries are ordered by the binding numbers in the descriptor set layouts; and
within a binding array, elements are in order."

This means when processing the binding numbers to validate the dynamic offsets at
draw time, we should do so in order to make sure that the sequential dynamic offset
index correctly correlates with the dynamic offset being processed.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -2900,8 +2900,9 @@ cvdescriptorset::DescriptorSet *getSetNode(const layer_data *my_data, VkDescript
 //  3. Grow updateBuffers for pCB to include buffers from STORAGE*_BUFFER descriptor buffers
 static bool validate_and_update_drawtime_descriptor_state(
     layer_data *dev_data, GLOBAL_CB_NODE *pCB,
-    const vector<std::tuple<cvdescriptorset::DescriptorSet *, unordered_map<uint32_t, descriptor_req>,
-                            std::vector<uint32_t> const *>> &activeSetBindingsPairs, const char *function) {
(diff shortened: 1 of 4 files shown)


layers/core_validation.cpp
```diff
@@ -2840,8 +2840,8 @@ cvdescriptorset::DescriptorSet *getSetNode(const layer_data *my_data, VkDescript
 //  3. Grow updateBuffers for pCB to include buffers from STORAGE*_BUFFER descriptor buffers
 static bool validate_and_update_drawtime_descriptor_state(
     layer_data *dev_data, GLOBAL_CB_NODE *pCB,
-    const vector<std::tuple<cvdescriptorset::DescriptorSet *, unordered_map<uint32_t, descriptor_req>,
-                            std::vector<uint32_t> const *>> &activeSetBindingsPairs) {
+    const vector<std::tuple<cvdescriptorset::DescriptorSet *, std::map<uint32_t, descriptor_req>, std::vector<uint32_t> const *>>
```
chrisforbes: Given we build this structure once (at pipeline creation time) and then subsequent use is all traversals, how about we use `std::vector<std::pair<uint32_t, descriptor_req>>`? Can sort it /once/ after we're done building it.

chrisforbes:
Needs a test to demonstrate the bad behavior.

layers/core_validation.cpp
tobine: Yeah good idea

PR #869, opened 2016-08-22
Fix to not lose track of whether an index buffer is bound

Tony-LunarG:
Fixes false positive in several CTS tests

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -3952,7 +3952,7 @@ static void set_cb_pso_status(GLOBAL_CB_NODE *pCB, const PIPELINE_NODE *pPipe) {
     // Account for any dynamic state not set via this PSO
     if (!pPipe->graphicsPipelineCI.pDynamicState ||
         !pPipe->graphicsPipelineCI.pDynamicState->dynamicStateCount) { // All state is static
-        pCB->status = CBSTATUS_ALL;
+        pCB->status |= CBSTATUS_ALL;
     } else {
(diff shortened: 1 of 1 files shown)


layers/core_validation.cpp
```diff
@@ -3952,7 +3952,7 @@ static void set_cb_pso_status(GLOBAL_CB_NODE *pCB, const PIPELINE_NODE *pPipe) {
     // Account for any dynamic state not set via this PSO
     if (!pPipe->graphicsPipelineCI.pDynamicState ||
         !pPipe->graphicsPipelineCI.pDynamicState->dynamicStateCount) { // All state is static
-        pCB->status = CBSTATUS_ALL;
+        pCB->status |= CBSTATUS_ALL;
```
chrisforbes: Looks correct. Do we want to rename `CBSTATUS_ALL` to something less misleading?

chrisforbes:
:+1:

layers/core_validation.cpp
tobine: @chrisforbes Yeah we should do something here. My thought was that we may want to keep a separate bitmask that's clearly just the dynamic state, and keep those as the lowermost bits of pCB->status. We could still track pCB->status as a single uint64_t, and there is probably more non-dynamic state that it makes to track there at some point. This fix is :+1: for the immediate problem.
I'll file separate issue for cleanup that needs to occur at some point.

Tony-LunarG:
Cherry picked to master

PR #871, opened 2016-08-23
layers: return collect_* result containers rather than using & out

chrisforbes:


tobine:
LGTM :+1:

chrisforbes:
Pushed to master

Issue #873, opened 2016-08-23
Wrong vk_format_table sizes for ETC2/EAC RGBA8888 formats?


sg-brcm:
The format table entries in vk_layer_utils.cpp set the size for the two ETC2/EAC RGBA8888 formats to 8 bytes (64bits). However the khronos dataformat specification states they are 128bit blocks (to accommodate the full alpha), so I was expecting this table to say 16 bytes as for ASTC.

This is causing a validation failure when doing an image copy to an uncompressed 128bit class color format, which I expected to be compatible with these compressed formats. Have I misunderstood or is the table wrong?

Specifically the two entries in question are:
    {8, 4, VK_FORMAT_COMPATIBILITY_CLASS_ETC2_EAC_RGBA_BIT}, //  [VK_FORMAT_ETC2_R8G8B8A8_UNORM_BLOCK]
    {8, 4, VK_FORMAT_COMPATIBILITY_CLASS_ETC2_EAC_RGBA_BIT}, // [VK_FORMAT_ETC2_R8G8B8A8_SRGB_BLOCK]

mark-lunarg:
See PR #875 .

mark-lunarg:
Pushed to Master.  Thanks!

sg-brcm:
Thanks for quick response.

PR #875, opened 2016-08-23
layers: Fix ETC2 RGBA format class size

mark-lunarg:
In reference to github issue #873 .

These formats were incorrectly specified as using an 8-byte storage
class, but each 4x4 block is actually compressed to 128 bits.

Change-Id: I33353cbe3574f0c640b68c8c3c544a286c79ed49


chrisforbes:
:+1: looks correct.

There's a typo in the spec, though: the description for `VK_FORMAT_ETC2_R8G8B8A8_SRGB_BLOCK` in the 1.0.24 spec says it's 64 bits, but then describes two 64bit halves.

mark-lunarg:
Thanks @chrisforbes , I created an issue in VulkanDocs for that error.  Pushed to Master.

commit 46d9632baa0e447686de5f9d5ea02bd7a30043e7
Date:   Tue Aug 23 08:49:36 2016 -0600

    layers: GH873, Fix ETC2 RGBA format class size

    These formats were incorrectly specified as using an 8-byte storage
    class, but each 4x4 block is actually compressed to 128 bits.

    Change-Id: I33353cbe3574f0c640b68c8c3c544a286c79ed49

PR #881, opened 2016-08-24
Add Valid Usage checks for samplecounts in image transfers

mark-lunarg:
For CopyBufferToImage: dstImage must have a sample count equal to VK_SAMPLE_COUNT_1_BIT
For CopyImageToBuffer: srcImage must have a sample count equal to VK_SAMPLE_COUNT_1_BIT
For BlitImage: The sample count of srcImage and dstImage must both be equal to VK_SAMPLE_COUNT_1_BIT


lentinem:
Now that the levels code is checked in can we start using that for new checks?

mark-lunarg:
@lentinem , can you point me to an example?  I'm not seeing anything level-related.

lentinem:
For some reason I had thought tobin had checked the CL in but it looks like I may be mistaken as it's still open here: https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/782. The files that are not core validation have been checked in but we may be waiting on that. I'll let Tobin comment but ideally we would use that new flag - enabledLevels - to guard this check (something like if (enabledLevels | LAYER_IMAGE_TRANSFER_SAMPLE_COUNTS)).

layers/vk_validation_layer_details.md
```diff
@@ -57,7 +57,7 @@ The Draw State portion of the core validation layer tracks state leading into Dr
...
 | DS Copy Update | DS copy update with invalid state such as bad object state, mis-matched descriptor types or a bad binding index. | INVALID_COPY_UPDATE | vkUpdateDescriptorSets | InvalidDSUpdateIndex CopyDescriptorUpdateErrors DSUsageBitsErrors DSAspectBitsErrors | NA |
 | DS Update Type | Verifies that structs in DS Update tree are properly created, currently valid, and of the right type | INVALID_UPDATE_STRUCT | vkUpdateDescriptorSets | InvalidDSUpdateStruct | NA |
-| MSAA Sample Count | Verifies that Pipeline, RenderPass, and Subpass sample counts are consistent | NUM_SAMPLES_MISMATCH | vkCmdBindPipeline vkCmdBeginRenderPass vkCmdNextSubpass | NumSamplesMismatch | NA |
+| MSAA Sample Count | Verifies that Draw-time, Pipeline, RenderPass, and Subpass sample counts are consistent | NUM_SAMPLES_MISMATCH | vkCmdBindPipeline vkCmdBeginRenderPass vkCmdNextSubpass vkCopyImage kCopyImageToBuffer vkCopyBufferToImage| NumSamplesMismatch ImageSampleCounts | NA |
```
tobine: Function names are incorrect.
Should be vkCmdBlitImage, vkCmdCopyBufferToImage, vkCmdCopyImageToBuffer

You added "vkCopyImage" but the new check is not used in "vkCmdCopyImage" should it be?

tobine:
Regarding #782, it's not yet merged. We're moving it to its own extension. Also, we only have 64 bits to work with so we don't have enough for every individual check. We need to group them in some fashion which I wanted to put a bit more thought into before adding the flag checks.

mark-lunarg:
Made @tobine 's corrections and pushed to Master.

PR #882, opened 2016-08-24
layers: GH849: Defend against poorly-written applications

ianelliottus:
A test application was written that didn't properly call some count-based query
commands.  In this particular case, they always hard-coded the count variable
to 1 the 2nd time the query was called.  The swapchain layer was only recording
the 1st count value in this case, and this caused a seg fault when the
vkCreateSwapchainKHR command validation code was run.

This commit better defends against such poorly-written applications, so that
the vkCreateSwapchainKHR command validation code won't seg fault.  This
includes looking for VK_INCOMPLETE as a successful return code from the ICD.


ianelliottus:
Cherry-picked and pushed.

PR #889, opened 2016-08-25
Gh428

MarkMLunarG:
GH428 Addition of tests etc for sync validation cases

This branch adds tests that challenge the validation layer on
various synchronization validation cases including:

1) INVALID_CB_SIMULTANEOUS_USE - 2 additional test conditions
    related to VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT
    SimultaneousUse

2) OBJECT_INUSE - Attempt to delete an object while in use
    InUseDestroyedSignaled
    InvalidCmdBufferDescriptorSetImageSamplerDestroyed

3) INVALID_EVENT - call vkQueueSubmit with an event that has
    been deleted
    InUseDestroyedSignaled

4) QUEUE_FORWARD_PROGRESS - call vkQueueSubmit with a semaphore
    that has already been signaled
    QueueForwardProgressFenceWait

5) INVALID_FENCE_STATE - call vkWaitFence with a fence
    that has not been submitted on a queue
    QueueForwardProgressFenceWait

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -569,6 +569,7 @@ public:
         eInvalidMemoryOffset,
         eBindNullBuffer,
         eFreeInvalidHandle,
+        eNone,
     };
 
     enum eTestConditions {
@@ -592,7 +593,7 @@ public:
 
             vkCreateBuffer(aVulkanDevice->device(), &buffer_create_info, nullptr,
                            &vulkanBuffer);
-            VkMemoryRequirements memory_reqs = {0};
+            VkMemoryRequirements memory_reqs = {};
 
             vkGetBufferMemoryRequirements(aVulkanDevice->device(),
                                           vulkanBuffer, &memory_reqs);
@@ -619,7 +620,7 @@ public:
     // A constructor which performs validation tests within construction.
     VkBufferTest(VkDeviceObj *aVulkanDevice,
                  VkBufferUsageFlags aBufferUsage,
-                 eTestEnFlags aTestFlag)
+                 eTestEnFlags aTestFlag = eNone)
         : AllocateCurrent(false), BoundCurrent(false),
         CreateCurrent(false), VulkanDevice(aVulkanDevice->device()) {
 
@@ -13858,10 +13859,10 @@ TEST_F(VkLayerTest, SimultaneousUse) {
     ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
 
     const char *simultaneous_use_message1 =
+            "w/o VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT set!";
+    const char *simultaneous_use_message2 =
             "does not have VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT set and "
             "will cause primary command buffer";
-    const char *simultaneous_use_message2 =
-            "w/o VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT set!";
 
     VkCommandBufferAllocateInfo command_buffer_allocate_info = {};
     command_buffer_allocate_info.sType =
@@ -13886,36 +13887,45 @@ TEST_F(VkLayerTest, SimultaneousUse) {
     command_buffer_begin_info.pInheritanceInfo = &command_buffer_inheritance_info;
 
     vkBeginCommandBuffer(secondary_command_buffer, &command_buffer_begin_info);
+    vkCmdBeginRenderPass(m_commandBuffer->handle(), &renderPassBeginInfo(),
+                         VK_SUBPASS_CONTENTS_SECONDARY_COMMAND_BUFFERS);
+    vkCmdExecuteCommands(m_commandBuffer->handle(), 1,
+                         &secondary_command_buffer);
+    vkCmdEndRenderPass(m_commandBuffer->GetBufferHandle());
     vkEndCommandBuffer(secondary_command_buffer);
-
-    command_buffer_begin_info.flags = VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT;
-    vkBeginCommandBuffer(m_commandBuffer->handle(), &command_buffer_begin_info);
-
-    vkEndCommandBuffer(m_commandBuffer->handle());
 
     VkSubmitInfo submit_info = {};
     submit_info.sType = VK_STRUCTURE_TYPE_SUBMIT_INFO;
     submit_info.commandBufferCount = 1;
     submit_info.pCommandBuffers =&m_commandBuffer->handle();
+    vkQueueSubmit(m_device->m_queue, 1, &submit_info, VK_NULL_HANDLE);
 
-    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_WARNING_BIT_EXT,
+    vkBeginCommandBuffer(m_commandBuffer->handle(), &command_buffer_begin_info);
+    vkCmdBeginRenderPass(m_commandBuffer->handle(), &renderPassBeginInfo(),
+                         VK_SUBPASS_CONTENTS_SECONDARY_COMMAND_BUFFERS);
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT,
                                          simultaneous_use_message1);
     vkCmdExecuteCommands(m_commandBuffer->handle(), 1,
                          &secondary_command_buffer);
     m_errorMonitor->VerifyFound();
+    vkCmdEndRenderPass(m_commandBuffer->GetBufferHandle());
+    vkEndCommandBuffer(m_commandBuffer->handle());
 
     m_errorMonitor->SetDesiredFailureMsg(0, "");
-    vkResetCommandBuffer(m_commandBuffer->handle(), 0);
-
-    submit_info.pCommandBuffers = &secondary_command_buffer;
     vkQueueSubmit(m_device->m_queue, 1, &submit_info, VK_NULL_HANDLE);
 
+    command_buffer_begin_info.flags = VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT;
     vkBeginCommandBuffer(m_commandBuffer->handle(), &command_buffer_begin_info);
-    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT,
+    vkCmdBeginRenderPass(m_commandBuffer->handle(), &renderPassBeginInfo(),
+                         VK_SUBPASS_CONTENTS_SECONDARY_COMMAND_BUFFERS);
+
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_WARNING_BIT_EXT,
                                          simultaneous_use_message2);
     vkCmdExecuteCommands(m_commandBuffer->handle(), 1,
                          &secondary_command_buffer);
     m_errorMonitor->VerifyFound();
+    vkCmdEndRenderPass(m_commandBuffer->GetBufferHandle());
+    vkEndCommandBuffer(m_commandBuffer->handle());
 }
 
 TEST_F(VkLayerTest, InUseDestroyedSignaled) {
@@ -13974,7 +13984,7 @@ TEST_F(VkLayerTest, InUseDestroyedSignaled) {
                                     nullptr, &fence));
 
     VkDescriptorPoolSize descriptor_pool_type_count = {};
-    descriptor_pool_type_count.type = VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE;
+    descriptor_pool_type_count.type = VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER;
     descriptor_pool_type_count.descriptorCount = 1;
 
     VkDescriptorPoolCreateInfo descriptor_pool_create_info = {};
@@ -13991,7 +14001,7 @@ TEST_F(VkLayerTest, InUseDestroyedSignaled) {
                                              nullptr, &descriptorset_pool));
 
     VkDescriptorSetLayoutBinding descriptorset_layout_binding = {};
-    descriptorset_layout_binding.descriptorType = VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE;
+    descriptorset_layout_binding.descriptorType = VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER;
     descriptorset_layout_binding.descriptorCount = 1;
     descriptorset_layout_binding.stageFlags = VK_SHADER_STAGE_ALL;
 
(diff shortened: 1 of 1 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -13486,6 +13486,242 @@ TEST_F(VkLayerTest, ValidRenderPassAttachmentLayoutWithLoadOp) {
...
+
+    vkCmdBindPipeline(m_commandBuffer->GetBufferHandle(),
+                      VK_PIPELINE_BIND_POINT_GRAPHICS, pipe.handle());
+    vkCmdBindDescriptorSets(m_commandBuffer->GetBufferHandle(),
+                            VK_PIPELINE_BIND_POINT_GRAPHICS, pipeline_layout, 0,
+                            1, &descriptorset, 0, NULL);
+
+    vkCmdDraw(m_commandBuffer->handle(), 3, 1, 0, 0);
+
+    vkEndCommandBuffer(m_commandBuffer->handle());
```
tobine: This line is crashing on a couple of android devices. I suspect b/c there's a vkCmdDraw() issued w/o a Renderpass.

MarkMLunarG: Thanks, that's a side effect to me being new to the pull request process. Since the branch was already on github and the pull request didn't do a force push - the pull request was created with an older version.

MarkMLunarG:
The latest PR update brings back the OBJECT_INUSE test for free descriptor

tests/layer_validation_tests.cpp
```diff
@@ -7741,7 +7741,8 @@ TEST_F(VkLayerTest, InvalidCmdBufferPipelineDestroyed) {
 TEST_F(VkLayerTest, InvalidCmdBufferDescriptorSetBufferDestroyed) {
     TEST_DESCRIPTION("Attempt to draw with a command buffer that is invalid "
                      "due to a bound descriptor set with a buffer dependency "
-                     "being destroyed.");
+                     "being destroyed. Attempt to destroy a DescriptorSet"
```
tobine: missing space at end of line before quote

tests/layer_validation_tests.cpp
```diff
@@ -13486,6 +13491,233 @@ TEST_F(VkLayerTest, ValidRenderPassAttachmentLayoutWithLoadOp) {
     vkDestroyRenderPass(m_device->device(), rp, NULL);
 }
 
+TEST_F(VkLayerTest, SimultaneousUseInUseDestroyedSignaled) {
+    TEST_DESCRIPTION("Use vkCmdExecuteCommands with invalid state"
```
tobine: Needs spaces at the ends of these lines or txt will run together.

tests/layer_validation_tests.cpp
```diff
@@ -13486,6 +13491,233 @@ TEST_F(VkLayerTest, ValidRenderPassAttachmentLayoutWithLoadOp) {
     vkDestroyRenderPass(m_device->device(), rp, NULL);
 }
 
+TEST_F(VkLayerTest, SimultaneousUseInUseDestroyedSignaled) {
```
tobine: This is too many unrelated checks in a single test.
I'm fine with the lifetime checks (in-use & early destroy) being in a single test.
The invalid_fence_wait_message & queue_forward_progress_message test are also related and could be together in a single test although I'm fine with them being separate as well.
Finally the simultaneous_use_message1 & 2 tests can be in their own separate test.

Note that GitHub issues that captured multiple missing checks were not meant to imply that all of those checks should be handled in a single test.

It's fine to combine multiple test cases into a single test, but those cases should have a clear relationship where they're hitting a series of variations on the same types of checks.

tests/layer_validation_tests.cpp
```diff
@@ -13634,6 +13639,292 @@ TEST_F(VkLayerTest, ValidRenderPassAttachmentLayoutWithLoadOp) {
...
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+
+    const char *submit_with_deleted_event_message =
+            "Cannot submit cmd buffer using deleted event 0x";
+    const char *cannot_delete_event_message =
+            "Cannot delete event 0x";
+    const char *cannot_delete_semaphore_message =
+            "Cannot delete semaphore 0x";
+    const char *cannot_destroy_fence_message =
+            "Fence 0x";
```
tobine: This may set a new record for shortest error string to match :)

tests/layer_validation_tests.cpp
```diff
@@ -13634,6 +13639,292 @@ TEST_F(VkLayerTest, ValidRenderPassAttachmentLayoutWithLoadOp) {
...
+                                         cannot_delete_semaphore_message);
+    vkDestroySemaphore(m_device->device(), semaphore, nullptr);
+    m_errorMonitor->VerifyFound();
+
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                                         cannot_destroy_fence_message);
+    vkDestroyFence(m_device->device(), fence, nullptr);
+    m_errorMonitor->VerifyFound();
+
+    m_errorMonitor->SetDesiredFailureMsg(0, "");
```
tobine: what is this line doing?

tobine:
LGTM :+1:
Passes on our farm as well.
You've got a few calls to `m_errorMonitor->SetDesiredFailureMsg(0, "");` which I haven't seen used before. Is that resetting some state in the error monitor? Rather than having a bunch of those calls peppered throughout the code I'd prefer that it was built into the error monitor so that after VerifyFound() or VerifyNotFound() is called it can just reset any state there.

tests/layer_validation_tests.cpp
```diff
@@ -13634,6 +13639,292 @@ TEST_F(VkLayerTest, ValidRenderPassAttachmentLayoutWithLoadOp) {
...
+                                        nullptr, &semaphore));
+    VkSubmitInfo submit_info = {};
+    submit_info.sType = VK_STRUCTURE_TYPE_SUBMIT_INFO;
+    submit_info.commandBufferCount = 1;
+    submit_info.pCommandBuffers =&m_commandBuffer->handle();
+    submit_info.signalSemaphoreCount = 1;
+    submit_info.pSignalSemaphores = &semaphore;
+    vkQueueSubmit(m_device->m_queue, 1, &submit_info, VK_NULL_HANDLE);
+    m_errorMonitor->SetDesiredFailureMsg(0, "");
+    vkResetCommandBuffer(m_commandBuffer->handle(), 0);
```
chrisforbes: This command buffer is pending execution here, you can't reset it. Saying you don't care about the error on the previous line isn't good enough -- the callback will return VK_FALSE, the call will still happen, and so you get UB.

tests/layer_validation_tests.cpp
chrisforbes: `vk*WaitIdle` before this, so you can tear everything down without errors, rather than masking them away?

tests/layer_validation_tests.cpp
```diff
@@ -13634,6 +13639,292 @@ TEST_F(VkLayerTest, ValidRenderPassAttachmentLayoutWithLoadOp) {
...
+
+    command_buffer_begin_info.flags = VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT;
+    vkBeginCommandBuffer(m_commandBuffer->handle(), &command_buffer_begin_info);
+
+    vkEndCommandBuffer(m_commandBuffer->handle());
+
+    VkSubmitInfo submit_info = {};
+    submit_info.sType = VK_STRUCTURE_TYPE_SUBMIT_INFO;
+    submit_info.commandBufferCount = 1;
+    submit_info.pCommandBuffers =&m_commandBuffer->handle();
```
chrisforbes: This is written here, but then clobbered on :13699 before being submitted.

tests/layer_validation_tests.cpp
```diff
@@ -13634,6 +13639,292 @@ TEST_F(VkLayerTest, ValidRenderPassAttachmentLayoutWithLoadOp) {
...
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_WARNING_BIT_EXT,
+                                         simultaneous_use_message1);
+    vkCmdExecuteCommands(m_commandBuffer->handle(), 1,
+                         &secondary_command_buffer);
+    m_errorMonitor->VerifyFound();
+
+    m_errorMonitor->SetDesiredFailureMsg(0, "");
+    vkResetCommandBuffer(m_commandBuffer->handle(), 0);
+
+    submit_info.pCommandBuffers = &secondary_command_buffer;
```
chrisforbes: Can't submit a secondary command buffer directly to a queue-- and doesn't look like what you were trying to do?

tests/layer_validation_tests.cpp
MarkMLunarG: What I've discovered in the process of writing tests is that a test can easily leave core_validation in a grumpy mood. Setting the BIT_EXT to 0 basically guarantees that any log_msg call will immediately returns false, thus making it possible to call something like vkResetCommandBuffer without any complaints from core_validation. It makes things less noisy and prevents something being skipped.

tests/layer_validation_tests.cpp
```diff
@@ -13634,6 +13639,292 @@ TEST_F(VkLayerTest, ValidRenderPassAttachmentLayoutWithLoadOp) {
...
+    vkEndCommandBuffer(m_commandBuffer->handle());
+
+    VkSubmitInfo submit_info = {};
+    submit_info.sType = VK_STRUCTURE_TYPE_SUBMIT_INFO;
+    submit_info.commandBufferCount = 1;
+    submit_info.pCommandBuffers =&m_commandBuffer->handle();
+
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_WARNING_BIT_EXT,
+                                         simultaneous_use_message1);
+    vkCmdExecuteCommands(m_commandBuffer->handle(), 1,
```
chrisforbes: m_commandBuffer is not in the recording state here.

tests/layer_validation_tests.cpp
```diff
@@ -13634,6 +13639,292 @@ TEST_F(VkLayerTest, ValidRenderPassAttachmentLayoutWithLoadOp) {
...
+    m_errorMonitor->SetDesiredFailureMsg(0, "");
+    vkResetCommandBuffer(m_commandBuffer->handle(), 0);
+
+    submit_info.pCommandBuffers = &secondary_command_buffer;
+    vkQueueSubmit(m_device->m_queue, 1, &submit_info, VK_NULL_HANDLE);
+
+    vkBeginCommandBuffer(m_commandBuffer->handle(), &command_buffer_begin_info);
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                                         simultaneous_use_message2);
+    vkCmdExecuteCommands(m_commandBuffer->handle(), 1,
```
chrisforbes: This secondary command buffer expects to be executed within a renderpass -- but m_commandBuffer's recording is currently not.

tests/layer_validation_tests.cpp
chrisforbes: Right, but returning false from an /error/ is your quickest path to leaving the layers and/or the driver in a bad state.

This is the common theme to most of my comments on this PR -- if you do invalid things /other/ than what you're looking for (and so returning VK_TRUE for), you get UB and all bets are off. The solution to this is to /not do/ other invalid things, not to try to sweep them under the rug.

tests/layer_validation_tests.cpp
```diff
@@ -8309,7 +8310,11 @@ TEST_F(VkLayerTest, InvalidCmdBufferDescriptorSetImageSamplerDestroyed) {
     Draw(1, 0, 0, 0);
     EndCommandBuffer();
     // Destroy descriptor set invalidates the cb, causing error on submit
+    m_errorMonitor->SetDesiredFailureMsg(
+        VK_DEBUG_REPORT_ERROR_BIT_EXT,
```
chrisforbes: I'm confused about this. Freeing a descriptor set used by a recorded but /not in-flight/ command buffer invalidates that command buffer, rather than producing an error from in `vkFreeDescriptorSets`. If we get and expect an error here, then the free never actually happens, and so the submit-time check a bit further down won't work...

chrisforbes: I've poked around a bit to try to understand why this error might be triggered in this case, and it looks like a /freshly recorded/ command buffer to be considered in-flight, we've got to have screwed something up in the state tracking in one of the (expected to fail with no side effects) `QueueSubmit` calls above. This /should/ have manifested as an error at `BeginCommandBuffer` on :8304, but nobody is looking for that.

tests/layer_validation_tests.cpp
MarkMLunarG: Probably a more important example where I used this is to quiet all messages and avoid skip_call being set before destroying all the resources at the end of the test. This will do that.

tests/layer_validation_tests.cpp
MarkMLunarG: I didn't see your comments @chrisforbes until after I pushed. I'll plan to do a follow push after I take care of some other issues.

tests/layer_validation_tests.cpp
MarkMLunarG: I see things a bit different because I see my role as creating scenarios that simulate common invalid practices that an inexperienced developer would do. That's not to say that there are some things that can be cleaned up, as you pointed out

tests/layer_validation_tests.cpp
MarkMLunarG: OK. I copied this from another test, another drawback to the copy paste method of writing tests. I've put together a significant re-haul and cleanup of this test.

tests/layer_validation_tests.cpp
MarkMLunarG: I simply added the message sentinel to test code that was already written. I think that you are saying that the fact that it worked is a problem with core_validation, which is beyond the test work that I'm chartered to complete. I spent a great deal of time attempting getting this test to work by normal means but because active_slots never is assigned any values, descriptor sets never get bound to the command buffer. I don't see any means to solve this outside of resolving issues in core_validation

tests/layer_validation_tests.cpp
```diff
@@ -13838,6 +13847,32 @@ TEST_F(VkLayerTest, InUseDestroyedSignaled) {
...
+    buffer_info.range = 1024;
+
+    VkWriteDescriptorSet write_descriptor_set = {};
+    write_descriptor_set.sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
+    write_descriptor_set.dstSet = descriptorset;
+    write_descriptor_set.descriptorCount = 1;
+    write_descriptor_set.descriptorType = VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER;
+    write_descriptor_set.pBufferInfo = &buffer_info;
+
+    vkUpdateDescriptorSets(m_device->device(), 1, &write_descriptor_set, 0,
```
chrisforbes: This isn't valid, per 11.6. `buffer` has no memory bound to it at this point -- you can't update a descriptor set with it in this state.

MarkMLunarG: Thanks for the thorough review. I can see the value in doing the code right. At times it's easy to loose sight of that.

chrisforbes:
:+1:

PR #893, opened 2016-08-26
Fix, that display only build under Linux works.

FslNopper:
At the moment, when configured for Linux, XCB and/or XLIB are assumed. If only the display extension wants to be used, the current build fails.

The current changes fixes these issues. Also, it is also possible to cross compile the code e.g. using the following additional flags now works:
-DBUILD_WSI_XCB_SUPPORT=OFF -DBUILD_WSI_XLIB_SUPPORT=OFF -DCMAKE_SYSTEM_PROCESSOR=arm

If cross compiled, the tri, cube and smoke demo are not build.


stroyan:
This was applied via other pull requests but never closed.

PR #894, opened 2016-08-26
Validation Layer Extension white-list

mark-lunarg:


layers/unique_objects.h
```diff
@@ -101,34 +124,60 @@ static void checkInstanceRegisterExtensions(const VkInstanceCreateInfo *pCreateI
     instanceExtMap[pDisp] = {};
 
     for (i = 0; i < pCreateInfo->enabledExtensionCount; i++) {
-        if (strcmp(pCreateInfo->ppEnabledExtensionNames[i], VK_KHR_SURFACE_EXTENSION_NAME) == 0)
+        bool extension_valid = false;
+
+        if (strcmp(pCreateInfo->ppEnabledExtensionNames[i], VK_KHR_SURFACE_EXTENSION_NAME) == 0) {
             instanceExtMap[pDisp].wsi_enabled = true;
```
courtney-g: Why don't we set extension_valid = true here?

layers/unique_objects.h
```diff
@@ -101,34 +124,60 @@ static void checkInstanceRegisterExtensions(const VkInstanceCreateInfo *pCreateI
...
-        if (strcmp(pCreateInfo->ppEnabledExtensionNames[i], VK_KHR_DISPLAY_EXTENSION_NAME) == 0)
+        }
+        if (strcmp(pCreateInfo->ppEnabledExtensionNames[i], VK_KHR_DISPLAY_EXTENSION_NAME) == 0) {
             instanceExtMap[pDisp].display_enabled = true;
```
courtney-g: And here?

layers/unique_objects.h
mark-lunarg: Only the ifdef'd checks use the extension-valid flag since those extension names might not be defined.  All others are handled in the loop below -- this seemed better than checking them twice.

courtney-g:
Other than two comments above, LGTM.

layers/unique_objects.h
courtney-g: Okay, so we could set extension_valid here, but don't need to as it will be checked below regardless.

layers/unique_objects.h
```diff
@@ -42,12 +42,31 @@
 
 namespace unique_objects {
 
+// The display-server-specific WSI extensions are handled explicitly
+static const char *kUniqueObjectsSupportedInstanceExtensions = VK_EXT_DEBUG_MARKER_EXTENSION_NAME VK_EXT_DEBUG_REPORT_EXTENSION_NAME
+    VK_KHR_DISPLAY_EXTENSION_NAME VK_KHR_SURFACE_EXTENSION_NAME;
```
ianelliottus: The VK_KHR_*_surface extensions are missing from this string.

BTW, this seems like just a concatenated list of extension names.  True?

mark-lunarg: Yes, they'd have to be excluded as most of them are not defined at any given time.  This is a concatenated list of extension names.  The extension names guarded by defines are checked  just above this in the ifdef-guarded sections.

layers/unique_objects.h
ianelliottus: It took me a while to figure out what this function was doing w.r.t. checking the whitelist and why "extension_valid" wasn't set for some.  Perhaps a comment would help--that "extension_valid" is used for platform-specific extensions that may or may not be valid to specify for a given loader-build.

ianelliottus: Thinking about this some more, it does seem strange that the WSI platform extensions (VK_KHR_*_surface) don't show up in the list.

mark-lunarg:
Updated the lists (with the hated ifdefs) to contain the complete list of white-listed extensions, including the VK_KHR_*_surface ones, pushed to Master.  Thanks guys.

PR #900, opened 2016-08-30
Fix Dota2 crash in Windows

mark-lunarg:
Two issues, in_use was not initialized in SAMPLER_NODE and an invalid access during resource tracking for same. This occurs during GetFenceStatus after a call to DestroySampler.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -4601,7 +4601,9 @@ static void DecrementBoundResources(layer_data const *dev_data, GLOBAL_CB_NODE c
         }
         case VK_DEBUG_REPORT_OBJECT_TYPE_SAMPLER_EXT: {
             auto sampler_node = getSamplerNode(dev_data, reinterpret_cast<VkSampler &>(obj.handle));
-            sampler_node->in_use.fetch_sub(1);
+            if (sampler_node) {
+                sampler_node->in_use.fetch_sub(1);
+            }
             break;
         }
         default:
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -4601,7 +4601,9 @@ static void DecrementBoundResources(layer_data const *dev_data, GLOBAL_CB_NODE c
         }
         case VK_DEBUG_REPORT_OBJECT_TYPE_SAMPLER_EXT: {
             auto sampler_node = getSamplerNode(dev_data, reinterpret_cast<VkSampler &>(obj.handle));
-            sampler_node->in_use.fetch_sub(1);
+            if (sampler_node) {
```
tobine: same situation could arise for descriptor set "set_node" in the code above so would be good to add guard there as well.

tobine:
LGTM :+1:
I've got the sampler in-use init fix and a few other related fixes in tobin_add_in_use_checks_for_objects_on_delete branch but haven't had time to write tests for it yet so it's languishing.

mark-lunarg:
Updated per comments and pushed to Master.

commit c72f92cc2c217e46f843449a05370129aeac5596
Date:   Tue Aug 30 11:30:04 2016 -0600

    layers: Fix Dota2 crash in Windows

    Two issues, in_use was not initialized in SAMPLER_NODE and an invalid
    access during resource tracking for SAMPLER_NODE and SET_NODE. This
    occured during GetFenceStatus after a call to DestroySampler.

    Change-Id: I7654a7c4a2677450097643422f44290be5dbac2a

PR #905, opened 2016-08-31
WIP - loader: gh851 wrap debug report callbacks

MarkY-LunarG:
Allow layers to wrap the debug report callbacks so they can
enable more messaging than the application, but also filter
the items returned.

Change-Id: I3fe8feecfa1838869de8a7338ff610e5ebca2e61

### DIFF
--- a/loader/debug_report.c
+++ b/loader/debug_report.c
@@ -65,22 +65,27 @@ util_CreateDebugReportCallback(struct loader_instance *inst,
                                VkDebugReportCallbackCreateInfoEXT *pCreateInfo,
                                const VkAllocationCallbacks *pAllocator,
                                VkDebugReportCallbackEXT callback) {
-    VkLayerDbgFunctionNode *pNewDbgFuncNode;
+    VkLayerDbgFunctionNode *pNewDbgFuncNode = NULL;
+
 #if (DEBUG_DISABLE_APP_ALLOCATORS == 1)
     {
 #else
     if (pAllocator != NULL) {
-        pNewDbgFuncNode = (VkLayerDbgFunctionNode *)pAllocator->pfnAllocation(
-            pAllocator->pUserData, sizeof(VkLayerDbgFunctionNode),
-            sizeof(int *), VK_SYSTEM_ALLOCATION_SCOPE_OBJECT);
+        pNewDbgFuncNode =
+            (VkLayerDbgFunctionNode *)pAllocator->pfnAllocation(
+                pAllocator->pUserData, sizeof(VkLayerDbgFunctionNode),
+                sizeof(int *), VK_SYSTEM_ALLOCATION_SCOPE_OBJECT);
     } else {
 #endif
-        pNewDbgFuncNode = (VkLayerDbgFunctionNode *)loader_instance_heap_alloc(
-            inst, sizeof(VkLayerDbgFunctionNode),
-            VK_SYSTEM_ALLOCATION_SCOPE_OBJECT);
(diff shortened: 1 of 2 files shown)


MarkY-LunarG:
Baldur has verified this fixes his issue.  But I wanted to verify I wasn't breaking anything else in the process.

loader/debug_report.c
```diff
@@ -346,6 +353,41 @@ VKAPI_ATTR VkResult VKAPI_CALL terminator_CreateDebugReportCallback(
         storage_idx++;
     }
 
+    // Setup the debug report callback in the terminator since a layer may want
+    // to grab the information itself (RenderDoc) and then return back to the
+    // user callback a sub-set of the messages.
+    if (NULL == inst->DbgFunctionHead) {
```
courtney-g: It's not clear to me why adding the callback to the list is dependent on the list being empty.
What happens if an app tries to register multiple callbacks?

loader/debug_report.c
```diff
@@ -311,6 +317,7 @@ VKAPI_ATTR VkResult VKAPI_CALL terminator_CreateDebugReportCallback(
     struct loader_instance *inst = (struct loader_instance *)instance;
     VkResult res = VK_SUCCESS;
     uint32_t storage_idx;
+    VkLayerDbgFunctionNode *pNewDbgFuncNode;
```
courtney-g: Initialize to NULL.

loader/debug_report.c
```diff
@@ -346,6 +353,41 @@ VKAPI_ATTR VkResult VKAPI_CALL terminator_CreateDebugReportCallback(
...
+        }
+        if (!pNewDbgFuncNode) {
+            return VK_ERROR_OUT_OF_HOST_MEMORY;
+        }
+        memset(pNewDbgFuncNode, 0, sizeof(VkLayerDbgFunctionNode));
+    }
+    if (NULL == inst->DbgFunctionHead ||
+        *pCallback != inst->DbgFunctionHead->msgCallback ||
+        pCreateInfo->pfnCallback != inst->DbgFunctionHead->pfnMsgCallback) {
+        pNewDbgFuncNode->msgCallback = *pCallback;
```
courtney-g: pNewDbgFuncNode is only allocated when inst->DbgFunctionHead is NULL.
What happens if DbgFunctionHead is not NULL but the second part of the if is true? Seems that pNewDbgFuncNode will not have been allocated in that case and we'll get segfault here.

MarkY-LunarG: Good point.  I didn't think about that, but it makes sense now, I should have noticed that with the linked list being set up.  I've locally removed all the "if (NULL == ....)" checks and no longer do any checks.  I just do it.  Anything else other than that?

loader/debug_report.c
MarkY-LunarG: Removed

loader/debug_report.c
```diff
@@ -66,28 +66,38 @@ util_CreateDebugReportCallback(struct loader_instance *inst,
                                const VkAllocationCallbacks *pAllocator,
                                VkDebugReportCallbackEXT callback) {
     VkLayerDbgFunctionNode *pNewDbgFuncNode;
+
+    if (NULL == inst->DbgFunctionHead ||
```
MarkY-LunarG: I removed this if check also.

loader/debug_report.c
MarkY-LunarG: Done in both places.

courtney-g:
:+1:

PR #911, opened 2016-09-02
Fix VK_NV_external_memory_capabilities support

jjulianoatnv:
-Need to init the instance function pointer

-The function should gracefully fall back on
 drivers that don't support it.


jjulianoatnv:
I tested this using both an ICD that does not exposes the extension and an ICD that does not expose the extension.

Without this fix, when an ICD exposes the extension then

> icd->GetPhysicalDeviceExternalImageFormatPropertiesNV

is always NULL, causing the terminator function to return VK_ERROR_INITIALIZATION_FAILED.

MarkY-LunarG:
Good catch.  I'll get this in today.

MarkY-LunarG:
Committed to master with commit: ede46175c78a66f359f7eadd5d8b906101d001e5

PR #916, opened 2016-09-07
layers: Clarify invalid memory message

### LINKED ISSUE #906: Invalid memory/please fill warning is not granular enough
We have very few memory allocations, so when the error shows
_vkCmdBeginRenderPass(): Cannot read invalid memory 0x75, please fill the memory before using._
We can potentially have hundreds of resources in that allocation. As the layer knows what framebuffer/attachment/resource is uninitialized, it can show also that information.

tobine:
Fixes #906

When reporting an invalid memory region, include the object type and
handle that is bound to that region.


chrisforbes:
Layers change LGTM. Do we actually have no tests that hit this case, though?

tobine:
@chrisforbes indeed no tests. I pushed a follow-on commit to add new enum for this particular error and updated documentation to note lack of tests.

PR #922, opened 2016-09-07
Tobin add in use checks for objects on delete

tobine:
This is a bunch of improvements for cmd buffer & object lifetime tracking.

Flag errors if QueryPool, Pipeline, or Sampler objects are destroyed while in-use and add tests for those cases. Update the destroy functions for all of those objects as well as events and semaphores to only remove the object from the map if the call down the destroy chain is NOT being skipped. For all 5 of those object types make them use the unified increment/decrement functions.

There are a bunch more of these cases that need to be handled but this is a good start.


layers/core_validation_types.h
```diff
@@ -507,7 +507,7 @@ class PIPELINE_NODE : public BASE_NODE {
     PIPELINE_NODE()
         : pipeline{}, graphicsPipelineCI{}, computePipelineCI{}, active_shaders(0), duplicate_shaders(0), active_slots(),
           vertexBindingDescriptions(), vertexAttributeDescriptions(), attachments(), blendConstantsEnabled(false), render_pass_ci(),
-          pipeline_layout() {}
+          pipeline_layout() { in_use.store(0); }
```
chrisforbes: Can we just push this up into BASE_NODE::BASE_NODE() ?

tests/layer_validation_tests.cpp
```diff
@@ -14166,6 +14166,177 @@ TEST_F(VkLayerTest, PipelineInUseDestroyedSignaled) {
...
+    pipeline_layout_ci.setLayoutCount = 1;
+    pipeline_layout_ci.pSetLayouts = &ds_layout;
+
+    VkPipelineLayout pipeline_layout;
+    err = vkCreatePipelineLayout(m_device->device(), &pipeline_layout_ci, NULL,
+                                 &pipeline_layout);
+    ASSERT_VK_SUCCESS(err);
+
+    VkImageObj image(m_device);
+    // Initialize image with USAGE_TRANSIENT_ATTACHMENT
```
chrisforbes: Dead comment

chrisforbes:
:+1:

layers/core_validation_types.h
tobine: Using a base class to perform common work for all derived classes?
INGENIOUS!
Done

tests/layer_validation_tests.cpp
tobine: Killed, so that makes it double-dead

tobine:

Issue #919, opened 2016-09-07
Core validation incorrect for vkCmdClearAttachments


gary-sweet:
When the core validation layer is validating that an attachment is in the reference list of active subpasses it doesn't take into account the extra indirection required by the subpass description array.

The following code fragment contains the issue:
```
for (uint32_t attachment_idx = 0; attachment_idx < attachmentCount; attachment_idx++) {
    const VkClearAttachment *attachment = &pAttachments[attachment_idx];
    ...
    for (uint32_t i = 0; i < pSD->colorAttachmentCount; i++) {
       if (attachment->colorAttachment == pSD->pColorAttachments[i].attachment) {
          found = true;
```

This is testing that the value of the attachment->colorAttachment field of the VkClearAttachment exists in the subpass description list for the subpass.

The spec states that the colorAttachment field `is an index to the pColorAttachments array in the VkSubpassDescription structure of the current subpass`.  The correct tests here would be that attachment->colorAttachment was smaller than the size of the pSD->pColorAttachments array. You might also then test that the attachment referenced by the pSD->pColorAttachment at that index exists in the render pass attachment list.

My tests which found this problem make a render pass with many attachments (up to 20). Each subpass uses a small subset of those, and crucially, not in contiguous order. So the first subpass might use attachments 1, 4, 6 & 9 for example. The validation code as it is today complains about this usage, which I believe to be valid.

chrisforbes:
I agree with your analysis.

chrisforbes:
We should also be downgrading these errors to warnings-- the spec defines the behavior for out of range attachment indices, and use of VK_ATTACHMENT_UNUSED. I will fix this at the same time.

chrisforbes:
Fixed in #923

gary-sweet:
Great, thanks. I've pulled the fix and it fixes my issue locally.

PR #923, opened 2016-09-08
CmdClearAttachments fixes for #919

chrisforbes:


chrisforbes:
Is clean on the internal farm.

layers/core_validation.cpp
```diff
@@ -8016,18 +8016,11 @@ VKAPI_ATTR void VKAPI_CALL CmdClearAttachments(VkCommandBuffer commandBuffer, ui
...
-                    if (attachment->colorAttachment == pSD->pColorAttachments[i].attachment) {
-                        found = true;
-                        break;
-                    }
-                }
-                if (!found) {
+                if (attachment->colorAttachment >= pSD->colorAttachmentCount) {
                     skip_call |= log_msg(
-                        dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
+                        dev_data->report_data, VK_DEBUG_REPORT_WARNING_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
```
tobine: Should this still be an error based on this valid usage language for vkCmdClearAttachments in 17.2?

`If the aspectMask member of any given element of pAttachments contains VK_IMAGE_ASPECT_COLOR_BIT, the colorAttachment member of those elements must refer to a valid color attachment in the current subpass`

This is potentially contradictory to the statement further down:

`If colorAttachment is VK_ATTACHMENT_UNUSED or is greater than or equal to VkSubpassDescription::colorAttachmentCount, then the clear has no effect.`

However, that statement could be construed to only be when VK_IMAGE_ASPECT_COLOT_BIT is NOT set. For now I'd go with the valid usage statement as higher authority and perhaps file spec ? for clarification on the potential contradiction.

tests/layer_validation_tests.cpp
```diff
@@ -13069,9 +13069,8 @@ TEST_F(VkLayerTest, MissingClearAttachment) {
     TEST_DESCRIPTION("Points to a wrong colorAttachment index in a VkClearAttachment "
                      "structure passed to vkCmdClearAttachments");
     ASSERT_NO_FATAL_FAILURE(InitState());
-    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT,
-                                         "vkCmdClearAttachments() attachment index 1 not found in attachment "
-                                         "reference array of active subpass 0");
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_WARNING_BIT_EXT,
```
tobine: WARN vs. ERROR will depend on my other comment above.

tobine:
Change looks good but I'd like to verify the WARN vs ERROR comment.
I filed spec issue [488](https://gitlab.khronos.org/vulkan/vulkan/issues/488) for clarity.

Can we keep it ERROR for this push and then update based on the outcome of that issue?

chrisforbes:
It's actually really weird that the spec goes out of its way to define some invalid operations as having no effect -- Vulkan doesn't do this in many other places...

chrisforbes:
OK, changing back to ERROR for now.

chrisforbes:
If we have some input to the resolution of 488, it would be nice to make this just UB, and get rid of the  "has no effect" language.

chrisforbes:
Pushed to master with warning->error change.

PR #924, opened 2016-09-08
layers: Require VkGraphicsPipelineCreateInfo::pDepthStencilState

chrisforbes:
If the subpass uses a depth/stencil attachment and rasterization is not
disabled, then this structure must be present.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -3425,6 +3425,20 @@ static bool verifyPipelineCreateState(layer_data *my_data, const VkDevice device
                 }
             }
         }
+
+        // If rasterization is not disabled, and subpass uses a depth/stencil
+        // attachment, pDepthStencilState must be a pointer to a valid structure
+        auto subpass_desc = renderPass ? &renderPass->pCreateInfo->pSubpasses[pPipeline->graphicsPipelineCI.subpass] : nullptr;
+        if (subpass_desc && subpass_desc->pDepthStencilAttachment &&
+            subpass_desc->pDepthStencilAttachment->attachment != VK_ATTACHMENT_UNUSED) {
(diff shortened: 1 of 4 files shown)


mark-lunarg:
+1.  Can you also add a test for this guy?

layers/core_validation.cpp
```diff
@@ -3420,6 +3420,20 @@ static bool verifyPipelineCreateState(layer_data *my_data, const VkDevice device
...
+        // If rasterization is not disabled, and subpass uses a depth/stencil
+        // attachment, pDepthStencilState must be a pointer to a valid structure
+        auto subpass_desc = renderPass ? &renderPass->pCreateInfo->pSubpasses[pPipeline->graphicsPipelineCI.subpass] : nullptr;
+        if (subpass_desc && subpass_desc->pDepthStencilAttachment &&
+            subpass_desc->pDepthStencilAttachment->attachment != VK_ATTACHMENT_UNUSED) {
+            if (!pPipeline->graphicsPipelineCI.pDepthStencilState) {
+                skip_call |= log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0,
+                                     __LINE__, DRAWSTATE_INVALID_PIPELINE_CREATE_STATE, "DS",
+                                     "Invalid Pipeline CreateInfo State: "
+                                     "pDepthStencilState is NULL when rasterization is enabled and subpass uses a "
```
tobine: Can include subpass number and renderpass handle in the error message to remind user of what they set for the pipeline. I could see this error arising as a result of copying pipeline state but forgetting to change renderpass and/or subpass.

chrisforbes:
Added test; clean on farm; merged

PR #925, opened 2016-09-08
Fix, that display only build under Linux works for master branch.

McNopper:
At the moment, when configured for Linux, XCB and/or XLIB are assumed. If only the display extension wants to be used, the current build fails.

The current changes fixes these issues. Also, it is also possible to cross compile the code e.g. using the following additional flags now works:
-DBUILD_WSI_XCB_SUPPORT=OFF -DBUILD_WSI_XLIB_SUPPORT=OFF -DCMAKE_SYSTEM_PROCESSOR=arm

If cross compiled, the tri, cube and smoke demo are not build.

Similar to https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/893, but for master branch.


karl-lunarg:
Pushed with 886977f32594cb0155af4e0d044f97ff813db60b.

Note that I had to fix vulkaninfo so that it would compile cleanly with XLIB and XCB off and fix the cross compile checks in the demos CMake file.

karl-lunarg:
This change was a little sticky.  The original pull request used EQUAL when testing for not cross-compiling (${CMAKE_SYSTEM_PROCESSOR} EQUAL ${CMAKE_HOST_SYSTEM_PROCESSOR}).  When using EQUAL, the test always failed, which made it do the intended thing of not building cube.  The trouble was that it didn't build cube when run with XLIB and XCB left ON and NOT specifying the CMAKE SYSTEM_PROCESSOR.  IOW, cube didn't build with the default build settings.

I can only guess that this wasn't tested with the default build settings. 

I changed the EQUAL to STREQUAL and cube started building again with the default settings.  But it still attempted to build cube with 
    -DBUILD_WSI_XCB_SUPPORT=OFF
    -DBUILD_WSI_XLIB_SUPPORT=OFF
    -DCMAKE_SYSTEM_PROCESSOR=arm.

It turns out that just setting CMAKE_SYSTEM_PROCESSOR isn't sufficient.  The variable inside the CMake file stayed the same (x86_64) even when set as above (to arm).   It seems that one needs to also set CMAKE_SYSTEM_NAME=Linux in order for CMake to pay attention to CMAKE_SYSTEM_PROCESSOR.  A better approach would be to create a toolchain file and include it in the main CMake file, if present.  If toolchain files are desirable, we could add an include to the top CMake file someday.  It would be better to do that if/when full cross-compile is working.

Bottom line, now in order to avoid building cube (and the others), you need to specify:
    -DBUILD_WSI_XCB_SUPPORT=OFF
    -DBUILD_WSI_XLIB_SUPPORT=OFF
    -DCMAKE_SYSTEM_NAME=Linux
    -DCMAKE_SYSTEM_PROCESSOR=arm

McNopper:
Sorry, that this all did no go so well as thought.

Finally, I tested the change and it builds from our side.

Thank you very much!

PR #927, opened 2016-09-08
layers: Remove old/unnecessary FIXME from swapchain layer

ianelliottus:
This FIXME comment does not apply.  It may have applied, if the line being
referred to once copied a pointer to user data, but at this point, the actual
data is being copied into a struct owned by the swapchain layer, which is what
the comment was about.


ianelliottus:
This was a no-brainer, and didn't need review.  I just cherry-picked and pushed this to master.

PR #928, opened 2016-09-08
WIP: loader: Add checks for usage of wsi extensions

MarkY-LunarG:
The loader really should validate that the WSI extensions are
enabled before being called.  Additionally, I needed to add
more checks for the KHR_display_swapchain extension in the
parameter_validation and object_tracker layers.

Change-Id: I3d07d46baf551be6f5f07e5374d6c683e3f52e7e

### DIFF
--- a/layers/object_tracker.cpp
+++ b/layers/object_tracker.cpp
@@ -2940,6 +2940,40 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateAndroidSurfaceKHR(VkInstance instance, cons
 }
 #endif // VK_USE_PLATFORM_ANDROID_KHR
 
+VKAPI_ATTR VkResult VKAPI_CALL CreateSharedSwapchainsKHR(VkDevice device, uint32_t swapchainCount,
+                                                         const VkSwapchainCreateInfoKHR *pCreateInfos,
+                                                         const VkAllocationCallbacks *pAllocator, VkSwapchainKHR *pSwapchains) {
+    bool skip_call = false;
+    uint32_t i = 0;
+    {
+        std::lock_guard<std::mutex> lock(global_lock);
+        skip_call |= ValidateDispatchableObject(device, device, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_EXT, false);
+        if (NULL != pCreateInfos) {
+            for (i = 0; i < swapchainCount; i++) {
+                skip_call |= ValidateNonDispatchableObject(device, pCreateInfos[i].oldSwapchain,
+                                                           VK_DEBUG_REPORT_OBJECT_TYPE_SWAPCHAIN_KHR_EXT, true);
+                layer_data *device_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
+                skip_call |= ValidateNonDispatchableObject(device_data->physical_device, pCreateInfos[i].surface,
+                                                           VK_DEBUG_REPORT_OBJECT_TYPE_SURFACE_KHR_EXT, false);
+            }
+        }
+    }
+    if (skip_call) {
+        return VK_ERROR_VALIDATION_FAILED_EXT;
+    }
+    VkResult result =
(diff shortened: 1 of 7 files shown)


layers/object_tracker.h
```diff
@@ -92,6 +92,7 @@ struct layer_data {
     debug_report_data *report_data;
     std::vector<VkDebugReportCallbackEXT> logging_callback;
     bool wsi_enabled;
+    bool wsi_display_swapchain_enabled;
```
tobine: Add constructor initializer for this new element in layer_data() init list below

tobine:
Validation changes look good. Just adding that one constructor initializer would be good.
Loader changes look ok to me but I'm less familiar with that code.
Changes are clean on our farm.

layers/object_tracker.cpp
```diff
@@ -3066,11 +3100,16 @@ static inline PFN_vkVoidFunction InterceptWsiEnabledCommand(const char *name, Vk
 static void CheckDeviceRegisterExtensions(const VkDeviceCreateInfo *pCreateInfo, VkDevice device) {
     layer_data *device_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
     device_data->wsi_enabled = false;
```
ianelliottus: While you aren't technically changing "wsi_enabled", I see that it's being overloaded in this layer.  For an instance, it's being used for VK_KHR_surface, and for a device, it's being used for VK_KHR_swapchain.  That could cause confusion/bugs down the road.  It might be best to use more-explicit names, like the other names.

MarkY-LunarG: I totally agree.  But we're trying to rush out the new extensions for Nvidia because they want to include them in the next driver upload.  I plan on creating a bug to come back and correct the other WSI extension checks later.  I think it's messy and could easily return incorrect data.  Someone mentioned to me that it was originally done this way because you had to have the other WSI extensions enabled for these to even work.  But I agree, I think it's a pretty lame excuse.

layers/object_tracker.h
MarkY-LunarG: Done

loader/wsi.c
```diff
@@ -158,6 +158,20 @@ void wsi_create_instance(struct loader_instance *ptr_instance,
...
 
+void wsi_create_device(struct loader_device *dev,
+    const VkDeviceCreateInfo *pCreateInfo) {
+    dev->loader_dispatch.enabled_known_extensions.khr_display_swapchain = 0;
+
+    for (uint32_t i = 0; i < pCreateInfo->enabledExtensionCount; i++) {
+        if (strcmp(pCreateInfo->ppEnabledExtensionNames[i],
+                   VK_KHR_DISPLAY_SWAPCHAIN_EXTENSION_NAME) == 0) {
+            dev->loader_dispatch.enabled_known_extensions
+                .khr_display_swapchain = 1;
```
ianelliottus: I'm not sure what this is (it's new to me), but I find it interesting that this is the only WSI extension in the struct.

ianelliottus:
I've looked at the changes, and I don't see any show-stoppers.  My recollection is that the VK_KHR_display extension wasn't properly validated in the swapchain layer, and this change isn't touching that layer either.  From a quick glance, I can see that the swapchain layer isn't doing much with validating vkCreateSharedSwapchainsKHR().

Is additional validation planned by someone?

loader/wsi.c
MarkY-LunarG: We currently check the extension enables in CreateInstance, but some of the WSI functions are device level functions and should be enabled in CreateDevice.  That's what this function is for.  It loops through all the enabled extensions in the CreateDevice DeviceCreateInfo struct, and if it finds one of interest (in this case 'KHR_display_swapchain') then it does its work.

MarkY-LunarG:
That's being tracked by another bug [actually I think it's a JIRA right now].

layers/object_tracker.cpp
ianelliottus: Okay, thanks!

loader/wsi.c
MarkY-LunarG: BTW, the KHR_swapchain extension should also be checked here, but as you pointed out earlier, that is also currently done incorrectly.

PR #936, opened 2016-09-12
Cube:  Add separate present queue and ownership barriers

Tony-LunarG:
Here's my best shot at using separate present and graphics queues with barriers to transfer image ownership back and forth between graphics and present queues.  I'm flat out guessing at some of this so feel free to correct code where you see issues.

### DIFF
--- a/demos/cube.c
+++ b/demos/cube.c
@@ -403,7 +403,7 @@ struct demo {
     PFN_vkDebugReportMessageEXT DebugReportMessage;
 
     uint32_t current_buffer;
-    uint32_t queue_count;
+    uint32_t queue_family_count;
 };
 
 VKAPI_ATTR VkBool32 VKAPI_CALL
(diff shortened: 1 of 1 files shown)


demos/cube.c
```diff
@@ -653,10 +675,75 @@ static void demo_draw_build_cmd(struct demo *demo, VkCommandBuffer cmd_buf) {
     // COLOR_ATTACHEMENT_OPTIMAL to PRESENT_SRC_KHR
     vkCmdEndRenderPass(cmd_buf);
 
+    if (demo->separate_present_queue) {
+        VkImageMemoryBarrier image_ownership_barrier = {
+            .sType = VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER,
+            .pNext = NULL,
+            .srcAccessMask = 0,
+            .dstAccessMask = VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT,
+            .oldLayout = VK_IMAGE_LAYOUT_PRESENT_SRC_KHR,
```
ianelliottus: Isn't this VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL (from about 40 lines above)?

ianelliottus:
@cnorthrop and @stroyan, I'm wondering if some of this pull request addresses issues that you were seeing in cube?

@cubanismo and @jjulianoatnv, this pull request is trying to update the cube demo to show how to use either separate or shared graphics/present queues (per spec issue 459).  Can you have a look please?

stroyan:
This adds one more semaphore for synchronization between vkAcquireNextImageKHR, vkQueueSubmit, and vkQueuePresentKHR calls.  That handles the additional vkQueueSubmit.  But it does not improve an existing problem with reuse of the semaphores if the demo_draw calls get ahead of asynchronous vkQueuePresentKHR calls. demos/smoke/Shell.cpp does avoid that problem by using a vkWaitForFences before vkAcquireNextImage.  Issue #937 was just opened for improving validation to catch that kind of problem.

demos/cube.c
Tony-LunarG: prepare_buffers sets the initial layout to PRESENT_SRC_KHR (line 1040), the barrier below takes it from PRESENT_SRC_KHR to COLOR_ATTACHMENT _OPTIMAL, and then the end of the renderpass sets it back to PRESENT_SRC_KHR.  It's only in COLOR_ATTACHMENT_OPTIMAL for the duration of the graphics command buffer

ianelliottus:
@Tony-LunarG, this looks pretty good to me, but I'd like somebody with a better understanding of barriers to also take a look.

demos/cube.c
ianelliottus: I think I may have commented on the wrong line, in the review tool.  When I was looking at the checked-out code with an editor, my recollection is that it looked like it was MOSLTY doing what you described.  At the end of the function, it didn't take it from COLOR_ATTACHMENT_OPTIMAL to PRESENT, but from PRESENT to PRESENT.

Tony-LunarG: There isn't a specific barrier going from C_A_O to Present, it's using the layout transition available at the end of the renderpass.  
The PRESENT to PRESENT barriers are the barriers that release and acquire ownership of the images between queues.  We had an issue about layout transitions and change of ownership and the resolution was that you cannot transition layouts and ownership in the same barrier (because you have to submit exactly the same barrier to both queues).

ianelliottus: Oh.  Got it.  Thanks!

ianelliottus:
+1. LGTM

ianelliottus:
@Tony-LunarG, do you want to merge this now?  I can if you want me to.  Thanks!

Tony-LunarG:
Changes pushed to master

PR #939, opened 2016-09-13
demos: Give Android versions permissions for internet & read/write files

ianelliottus:
This is in support of RenderDoc (and other tracing libraries).  On Android,
these "native" applications must have permission to read/write files on the
Android filesystem so that a trace-capture library can write a trace file.
After building and installing the application, the following commands should be
run from a Linux shell, in order to truly grant those permissions (not normally
required for an Android Java-based app, with a GUI; but is required for these
shell-launched, native apps):

adb shell pm grant com.example.Cube android.permission.READ_EXTERNAL_STORAGE
adb shell pm grant com.example.Cube android.permission.WRITE_EXTERNAL_STORAGE

### DIFF
--- a/demos/android/AndroidManifest.xml.cube
+++ b/demos/android/AndroidManifest.xml.cube
@@ -1,6 +1,11 @@
 <?xml version="1.0"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.example.Cube" android:versionCode="1" android:versionName="1.0">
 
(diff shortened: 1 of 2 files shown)


ianelliottus:
@cnorthrop, I was thinking that it would be good to add this.  Do you agree?

cnorthrop:
Yes, I'm running with these locally.

ianelliottus:
Cherry-picked and pushed to master.

PR #940, opened 2016-09-13
layers: Cleanup unexpected push_constants msgs

karl-lunarg:
Fix to core validation CmdPushConstants to avoid logging
multiple messages.
Fix to PushConstants layer tests to avoid generating
multiple errors in a single test.

Change-Id: I618d6e5aff55fa918da5bd81c47092174b266da1

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -6513,7 +6513,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreatePipelineLayout(VkDevice device, const VkPip
     bool skip_call = false;
     layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
     // Push Constant Range checks
-    uint32_t i = 0;
+    uint32_t i, j;
     for (i = 0; i < pCreateInfo->pushConstantRangeCount; ++i) {
         skip_call |= validatePushConstantRange(dev_data, pCreateInfo->pPushConstantRanges[i].offset,
                                                pCreateInfo->pPushConstantRanges[i].size, "vkCreatePipelineLayout()", i);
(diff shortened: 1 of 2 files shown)


tobine:
This one needs some discussion.
I see how this cleans up the particular unexpected errors for the given test, but don't know if I like the precedent it sets going fwd.
I'd prefer not to have multiple "if (!skip_call)" checks. My preference is to do early skip_call checks primarily in cases where we're avoiding a crash.

For the test framework I think the real concern is unexpected errors that we get BEFORE the intended error. I think it's fine to hit some additional errors after the expected error, but before the call is actually skipped.

Perhaps we should tweak the unexpected error reporting to shut off once the first expected error is encountered for a given test? Just one thought.

mark-lunarg:
Let's talk about it tomorrow.

karl-lunarg:
Changes made as discussed today.  I had to make one minor change to core_validation to change a log_msg() call to ERROR from WARNING, which was probably an earlier mistake.
I was able to rework the test itself so that it now runs quiet.

layers/core_validation.cpp
```diff
@@ -6533,7 +6533,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreatePipelineLayout(VkDevice device, const VkPip
                 const uint32_t maxB = minB + pCreateInfo->pPushConstantRanges[j].size;
                 if ((minA <= minB && maxA > minB) || (minB <= minA && maxB > minA)) {
                     skip_call |=
-                        log_msg(dev_data->report_data, VK_DEBUG_REPORT_WARNING_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
+                        log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
```
chrisforbes: I don't see valid usage language to back this up, but it's a weird thing for an app to do...

karl-lunarg: That may be why it was originally a warning.  It may be OK to define overlapping ranges as long as the pushes stay within the correct ranges wrt stage flags.

I've changed it back to a warning and fixed up the core_validation code to not return a VkResult error if this warning fires.   I also changed the call to log_msg() for the warning case to NOT set skip_call, because it is probably OK to still call the driver in this case.  I checked other cases where log_msg() is called for a warning and it most often modifies skip_call.  I sort of expected skip_call to NOT be (potentially) modified for warnings.

chrisforbes: Right, it's always OK to continue past a warning -- the app is just doing something weird, not something invalid, so there's no UB. I'm not so sure about ignoring the app's request to skip -- but there are precedents elsewhere in the code for warning and below.

karl-lunarg: I went ahead and added back the modification to set skip_call and committed via 07a398f7c8da45c8286cba4ef33239646ee87dc9.   Thanks for the help and sorry for the churn.

karl-lunarg:
Pushed 07a398f7c8da45c8286cba4ef33239646ee87dc9

PR #944, opened 2016-09-15
WIP: Marky gh181 surface khr

MarkY-LunarG:
This includes changes needed to allow the ICDs to create and destroy their own KHR_surface objects instead of having the loader generate a generic object they use.  Piers has validated that with his changes, everything appears to work.

### DIFF
--- a/layers/unique_objects.h
+++ b/layers/unique_objects.h
@@ -275,13 +275,6 @@ void explicit_DestroyInstance(VkInstance instance, const VkAllocationCallbacks *
 // Handle CreateDevice
 static void createDeviceRegisterExtensions(const VkDeviceCreateInfo *pCreateInfo, VkDevice device) {
     layer_data *my_device_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
-    VkLayerDispatchTable *pDisp = get_dispatch_table(unique_objects_device_table_map, device);
-    PFN_vkGetDeviceProcAddr gpa = pDisp->GetDeviceProcAddr;
-    pDisp->CreateSwapchainKHR = (PFN_vkCreateSwapchainKHR)gpa(device, "vkCreateSwapchainKHR");
-    pDisp->DestroySwapchainKHR = (PFN_vkDestroySwapchainKHR)gpa(device, "vkDestroySwapchainKHR");
-    pDisp->GetSwapchainImagesKHR = (PFN_vkGetSwapchainImagesKHR)gpa(device, "vkGetSwapchainImagesKHR");
-    pDisp->AcquireNextImageKHR = (PFN_vkAcquireNextImageKHR)gpa(device, "vkAcquireNextImageKHR");
-    pDisp->QueuePresentKHR = (PFN_vkQueuePresentKHR)gpa(device, "vkQueuePresentKHR");
     my_device_data->wsi_enabled = false;
 
     for (uint32_t i = 0; i < pCreateInfo->enabledExtensionCount; i++) {
--- a/loader/loader.c
+++ b/loader/loader.c
@@ -412,7 +412,7 @@ vkSetInstanceDispatch(VkInstance instance, void *object) {
 VKAPI_ATTR VkResult VKAPI_CALL
 vkSetDeviceDispatch(VkDevice device, void *object) {
     struct loader_device *dev;
-    struct loader_icd *icd = loader_get_icd_and_device(device, &dev);
+    struct loader_icd *icd = loader_get_icd_and_device(device, &dev, NULL);
 
     if (!icd) {
         return VK_ERROR_INITIALIZATION_FAILED;
@@ -1243,8 +1243,10 @@ out:
 }
 
 struct loader_icd *loader_get_icd_and_device(const VkDevice device,
-                                             struct loader_device **found_dev) {
+                                             struct loader_device **found_dev,
+                                             uint32_t *icd_index) {
     *found_dev = NULL;
+    uint32_t index = 0;
     for (struct loader_instance *inst = loader.instances; inst;
          inst = inst->next) {
         for (struct loader_icd *icd = inst->icds; icd; icd = icd->next) {
@@ -1255,8 +1257,12 @@ struct loader_icd *loader_get_icd_and_device(const VkDevice device,
                 if (loader_get_dispatch(dev->device) ==
                     loader_get_dispatch(device)) {
                     *found_dev = dev;
+                    if (NULL != icd_index) {
+                        *icd_index = index;
+                    }
                     return icd;
                 }
+            index++;
         }
     }
     return NULL;
@@ -1634,16 +1640,21 @@ static bool loader_icd_init_entrys(struct loader_icd *icd, VkInstance inst,
     LOOKUP_GIPA(CreateDisplayModeKHR, false);
     LOOKUP_GIPA(GetDisplayPlaneCapabilitiesKHR, false);
     LOOKUP_GIPA(DestroySurfaceKHR, false);
+    LOOKUP_GIPA(CreateSwapchainKHR, false);
 #ifdef VK_USE_PLATFORM_WIN32_KHR
+    LOOKUP_GIPA(CreateWin32SurfaceKHR, false);
     LOOKUP_GIPA(GetPhysicalDeviceWin32PresentationSupportKHR, false);
 #endif
 #ifdef VK_USE_PLATFORM_XCB_KHR
+    LOOKUP_GIPA(CreateXcbSurfaceKHR, false);
     LOOKUP_GIPA(GetPhysicalDeviceXcbPresentationSupportKHR, false);
 #endif
 #ifdef VK_USE_PLATFORM_XLIB_KHR
+    LOOKUP_GIPA(CreateXlibSurfaceKHR, false);
     LOOKUP_GIPA(GetPhysicalDeviceXlibPresentationSupportKHR, false);
 #endif
 #ifdef VK_USE_PLATFORM_WAYLAND_KHR
+    LOOKUP_GIPA(CreateWaylandSurfaceKHR, false);
     LOOKUP_GIPA(GetPhysicalDeviceWaylandPresentationSupportKHR, false);
 #endif
     LOOKUP_GIPA(GetPhysicalDeviceExternalImageFormatPropertiesNV, false);
@@ -3232,10 +3243,20 @@ loader_gpa_instance_internal(VkInstance inst, const char *pName) {
     return NULL;
 }
 
+
 static VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL
 loader_gpa_device_internal(VkDevice device, const char *pName) {
     struct loader_device *dev;
-    struct loader_icd *icd = loader_get_icd_and_device(device, &dev);
+    struct loader_icd *icd = loader_get_icd_and_device(device, &dev, NULL);
+
+    // Certain device entry-points still need to go through a terminator before
+    // hitting the ICD.  This could be for several reasons, but the main one
+    // is currently unwrapping an object before passing the appropriate info
+    // along to the ICD.
+    if (!strcmp(pName, "vkCreateSwapchainKHR")) {
+        return (PFN_vkVoidFunction)terminator_vkCreateSwapchainKHR;
+    }
+
     return icd->GetDeviceProcAddr(device, pName);
 }
 
@@ -4429,6 +4450,7 @@ terminator_EnumeratePhysicalDevices(VkInstance instance,
         for (j = 0; j < phys_devs[i].count && idx < copy_count; j++) {
             loader_set_dispatch((void *)&inst->phys_devs_term[idx], inst->disp);
             inst->phys_devs_term[idx].this_icd = phys_devs[i].this_icd;
+            inst->phys_devs_term[idx].icd_index = i;
             inst->phys_devs_term[idx].phys_dev = phys_devs[i].phys_devs[j];
             pPhysicalDevices[idx] =
                 (VkPhysicalDevice)&inst->phys_devs_term[idx];
--- a/loader/loader.c
+++ b/loader/loader.c
@@ -3243,6 +3243,22 @@ loader_gpa_instance_internal(VkInstance inst, const char *pName) {
     return NULL;
 }
 
+VKAPI_ATTR void VKAPI_CALL loader_override_terminating_device_proc(
+    VkDevice device, struct loader_dev_dispatch_table *disp_table) {
+    struct loader_device *dev;
+    struct loader_icd *icd = loader_get_icd_and_device(device, &dev, NULL);
+
+    // Certain device entry-points still need to go through a terminator before
+    // hitting the ICD.  This could be for several reasons, but the main one
+    // is currently unwrapping an object before passing the appropriate info
+    // along to the ICD.
+    if ((PFN_vkVoidFunction)disp_table->core_dispatch.CreateSwapchainKHR ==
+        (PFN_vkVoidFunction)icd->GetDeviceProcAddr(device,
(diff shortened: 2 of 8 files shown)


MarkY-LunarG:
BTW, what is the official process for informing IHVs of these kind of changes to the loader/icd interface.  Is this something brought up in the TSG working group, or did you all directly talk to individuals at those companies (Nvidia, AMD, Intel)?

courtney-g:
I would encourage you to contact key stakeholders that are not already involved with the development to make sure they know about it and the potential impact. Other than that we have ChangeLogs for the spec and SDK. The change is backward compatible so it's not absolutely necessary for folks to do anything unless they need that functionality.

layers/unique_objects.h
```diff
@@ -275,13 +275,6 @@ void explicit_DestroyInstance(VkInstance instance, const VkAllocationCallbacks *
 // Handle CreateDevice
 static void createDeviceRegisterExtensions(const VkDeviceCreateInfo *pCreateInfo, VkDevice device) {
     layer_data *my_device_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
-    VkLayerDispatchTable *pDisp = get_dispatch_table(unique_objects_device_table_map, device);
-    PFN_vkGetDeviceProcAddr gpa = pDisp->GetDeviceProcAddr;
-    pDisp->CreateSwapchainKHR = (PFN_vkCreateSwapchainKHR)gpa(device, "vkCreateSwapchainKHR");
```
MarkY-LunarG: Setting up the function pointers is all taken care of in explicit_CreateDevice() -> initDeviceTable() -> layer_init_device_dispatch_table().  It's duplicate code.

loader/wsi.c
```diff
@@ -207,6 +211,32 @@ terminator_DestroySurfaceKHR(VkInstance instance, VkSurfaceKHR surface,
                              const VkAllocationCallbacks *pAllocator) {
     struct loader_instance *ptr_instance = loader_get_instance(instance);
 
+// Android doesn't have to worry about multiple ICD scenario, but the rest do.
+#ifndef VK_USE_PLATFORM_ANDROID_KHR
```
MarkY-LunarG: I'm fine with removing all these.  I didn't want to be premature.

loader/wsi.c
```diff
@@ -468,32 +583,83 @@ LOADER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkCreateWin32SurfaceKHR(
...
 
-    pIcdSurface->base.platform = VK_ICD_WSI_PLATFORM_WIN32;
-    pIcdSurface->hinstance = pCreateInfo->hinstance;
-    pIcdSurface->hwnd = pCreateInfo->hwnd;
+    pIcdSurface->win_surf.base.platform = VK_ICD_WSI_PLATFORM_WIN32;
+    pIcdSurface->win_surf.hinstance = pCreateInfo->hinstance;
+    pIcdSurface->win_surf.hwnd = pCreateInfo->hwnd;
+
+    // Setup the new sizes and offsets so we can grow the structures in the
+    // future withouth having problems
```
MarkY-LunarG: <facepalm>

loader/wsi.c
```diff
@@ -659,32 +874,83 @@ vkCreateWaylandSurfaceKHR(VkInstance instance,
...
     if (pIcdSurface == NULL) {
-        return VK_ERROR_OUT_OF_HOST_MEMORY;
+        vkRes = VK_ERROR_OUT_OF_HOST_MEMORY;
+        goto out;
     }
 
-    pIcdSurface->base.platform = VK_ICD_WSI_PLATFORM_WAYLAND;
-    pIcdSurface->display = pCreateInfo->display;
-    pIcdSurface->surface = pCreateInfo->surface;
+    pIcdSurface->wayland_surf.base.platform = VK_ICD_WSI_PLATFORM_WAYLAND;
```
MarkY-LunarG: Seems reasonable.  I'll make a utility function for what I can.

loader/wsi.c
```diff
@@ -468,32 +583,83 @@ LOADER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkCreateWin32SurfaceKHR(
...
+                if (VK_SUCCESS != vkRes) {
+                    goto out;
+                }
+            }
+        }
+    }
 
-    *pSurface = (VkSurfaceKHR)pIcdSurface;
+    *pSurface = (VkSurfaceKHR)(pIcdSurface);
+
```
MarkY-LunarG: I'm a fan of single entry single exit.  Plus with a lot of the functions, there's so much common clean-up code throughout the loader that's skipped in many cases which has caused a number of leaks, I'm trying to remedy.  I can see your concern, but I think with branch prediction in the CPUs, the logic will be skipped, and this isn't a call that's triggered very often.

loader/table_ops.h
```diff
@@ -540,6 +540,20 @@ loader_lookup_device_dispatch_table(const VkLayerDispatchTable *table,
     if (!strcmp(name, "CmdExecuteCommands"))
         return (void *)table->CmdExecuteCommands;
 
+    if (!strcmp(name, "CreateSwapchainKHR")) {
+        // For CreateSwapChainKHR we need to use the entry and terminator
+        // functions to properly unwrap the SurfaceKHR object.
+        return (void *)vkCreateSwapchainKHR;
```
courtney-g: It seems odd that that specific function doesn't match the pattern of all the others. (Using table->XXX)
Why wouldn't that work here? Is the unwrapping done inside the trampoline code?

loader/wsi.c
courtney-g: I think it's safe to remove them.

loader/wsi.c
courtney-g: I'll leave this to your call.

loader/table_ops.h
MarkY-LunarG: Because table->XXX always points past the loader for these functions.  It never enters the loader (which goes against what Jon originally told me, but I found through manually debugging).  It's also the only device extension we have to perform any form of unwrapping on.  The others we typically have work to do are instance extensions which already have loader entry/terminator calls.  Remember, I had to do that because the pCreateInfo->surface has to be unwrapped to pass along the ICD version of the surface handle?

layers/unique_objects.h
ianelliottus: Just to clarify, is this a change that's not directly related to the topic of this pull request?

loader/loader.c
```diff
@@ -1634,16 +1640,21 @@ static bool loader_icd_init_entrys(struct loader_icd *icd, VkInstance inst,
...
 #ifdef VK_USE_PLATFORM_XCB_KHR
+    LOOKUP_GIPA(CreateXcbSurfaceKHR, false);
     LOOKUP_GIPA(GetPhysicalDeviceXcbPresentationSupportKHR, false);
 #endif
 #ifdef VK_USE_PLATFORM_XLIB_KHR
+    LOOKUP_GIPA(CreateXlibSurfaceKHR, false);
     LOOKUP_GIPA(GetPhysicalDeviceXlibPresentationSupportKHR, false);
 #endif
 #ifdef VK_USE_PLATFORM_WAYLAND_KHR
+    LOOKUP_GIPA(CreateWaylandSurfaceKHR, false);
```
ianelliottus: Code is missing here for the MIR entrypoints (both old and new).

loader/table_ops.h
```diff
@@ -259,7 +259,7 @@ static inline void loader_init_device_extension_dispatch_table(
     VkLayerDispatchTable *table = &dev_table->core_dispatch;
     table->AcquireNextImageKHR =
         (PFN_vkAcquireNextImageKHR)gpa(dev, "vkAcquireNextImageKHR");
-    table->CreateSwapchainKHR =
+    table->CreateSwapchainKHR =
```
ianelliottus: Nit: Probably good to remove this trailing-space at the end of a line:-)

loader/table_ops.h
```diff
@@ -540,6 +540,20 @@ loader_lookup_device_dispatch_table(const VkLayerDispatchTable *table,
     if (!strcmp(name, "CmdExecuteCommands"))
         return (void *)table->CmdExecuteCommands;
 
+    if (!strcmp(name, "CreateSwapchainKHR")) {
+        // For CreateSwapChainKHR we need to use the entry and terminator
```
ianelliottus: Nit: Is "entry" the same as "trampoline"?  If so, please change to "trampoline".

loader/wsi.c
```diff
@@ -446,6 +542,40 @@ vkQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR *pPresentInfo) {
...
+        pIcdSurface->platform_size = platform_size;
+        pIcdSurface->non_platform_offset = (uint32_t)(
+            (uint8_t *)(&pIcdSurface->base_size) - (uint8_t *)pIcdSurface);
+        pIcdSurface->entire_size = sizeof(VkIcdSurface);
+
+        if (create_icd_surfs) {
+            pIcdSurface->real_icd_surfaces = loader_instance_heap_alloc(
+                instance, sizeof(VkSurfaceKHR) * instance->total_icd_count,
+                VK_SYSTEM_ALLOCATION_SCOPE_OBJECT);
+            if (pIcdSurface == NULL) {
```
ianelliottus: This should be:

if (pIcdSurface->real_icd_surfaces == NULL) {

loader/wsi.c
```diff
@@ -1238,20 +1530,21 @@ VKAPI_ATTR VkResult VKAPI_CALL terminator_CreateDisplayPlaneSurfaceKHR(
         return VK_ERROR_EXTENSION_NOT_PRESENT;
     }
 
-    pIcdSurface = loader_instance_heap_alloc(
-        inst, sizeof(VkIcdSurfaceDisplay), VK_SYSTEM_ALLOCATION_SCOPE_INSTANCE);
+    pIcdSurface =
+        AllocateIcdSurfaceStruct(inst, sizeof(pIcdSurface->display_surf.base),
+                                 sizeof(pIcdSurface->display_surf), false);
```
ianelliottus: Does the "false" parameter imply that the VK_KHR_display extension will continue to support the old method into the foreseeable future?  If so, please add an explicit comment.  If not, please tell me more.

loader/wsi.c
```diff
@@ -468,32 +598,66 @@ LOADER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkCreateWin32SurfaceKHR(
...
+                    &pIcdSurface->real_icd_surfaces[i]);
+                if (VK_SUCCESS != vkRes) {
+                    goto out;
+                }
+            }
+        }
+    }
 
-    *pSurface = (VkSurfaceKHR)pIcdSurface;
+    *pSurface = (VkSurfaceKHR)(pIcdSurface);
```
ianelliottus: Should we initialize pSurface to VK_NULL_HANDLE up above?

loader/table_ops.h
MarkY-LunarG: Technically, a trampoline is a function which can be compiled out to optimize a pass-through.  In my terminology, an entry will never go away.

loader/wsi.c
```diff
@@ -956,11 +1254,9 @@ VKAPI_ATTR VkResult VKAPI_CALL terminator_CreateAndroidSurfaceKHR(
     }
 
     // Next, if so, proceed with the implementation of this function:
-    VkIcdSurfaceAndroid *pIcdSurface = NULL;
-
-    pIcdSurface =
+    VkIcdSurfaceAndroid *pIcdSurface =
         loader_instance_heap_alloc(ptr_instance, sizeof(VkIcdSurfaceAndroid),
-                                   VK_SYSTEM_ALLOCATION_SCOPE_INSTANCE);
+                                   VK_SYSTEM_ALLOCATION_SCOPE_OBJECT);
```
ianelliottus: Sigh.  Since Android has its own loader, perhaps we should remove the Android-specific code.  Doesn't need to be part of this pull request, but we should do it sometime.

courtney-g: Yeah, I had a similar request earlier.

loader/table_ops.h
ianelliottus: Sorry, but I'm confused after reading your reply.  Are you using "entry" for what Jon Ashburn, Courtney, et.al. have been calling a "trampoline" throughout the loader code, or is this "entry" function a different type of function?  I'd like to avoid mixing terminology, but am fine with adding a  new term for something different.

loader/wsi.h
```diff
@@ -19,16 +19,49 @@
...
+#endif // VK_USE_PLATFORM_XCB_KHR
+#ifdef VK_USE_PLATFORM_XLIB_KHR
+        VkIcdSurfaceXlib xlib_surf;
+#endif // VK_USE_PLATFORM_XLIB_KHR
+        VkIcdSurfaceDisplay display_surf;
+    };
+    uint32_t base_size; // Size of VkIcdSurfaceBase
+    uint32_t platform_size; // Size of corresponding VkIcdSurfaceXXX
+    uint32_t non_platform_offset; // Start offset to base_size
+    uint32_t entire_size; // Size of entire VkIcdSurface
```
ianelliottus: I'm not understanding the purpose of these 4 members.  What are their purpose?

ianelliottus:
I've finished my initial review.  The changes look pretty good!  I see only one serious bug and some minor issues and nits; and some questions.

loader/wsi.c
MarkY-LunarG: I will, but according to the vulkan spec, in general the values can only be expected to be valid if the return is successful.

loader/wsi.c
MarkY-LunarG: That's something we can do in the long run.  I've already go a ton of changes in this commit, so I don't want to get too messy.

loader/wsi.h
MarkY-LunarG: The individual VkIcdSurface structures are all being used by ICDs, currently.  This new code creates the union for these structures, but I added size checks here because I want to allow the structures to change in size without breaking.  This way, if new items are added, we would still know how to skip to get to pertinent data.

loader/loader.c
MarkY-LunarG: Good catch.  I'll fix that.

loader/table_ops.h
MarkY-LunarG: My understanding was that a "trampoline" is designed to go away.  But an "entry" is designed to perform initial functionality on entering a library/module/etc.  If I'm wrong, I'm fine with changing the name.  Here's a brief [article](http://www.datchley.name/recursion-tail-calls-and-trampolines/) talking about trampolines.

layers/unique_objects.h
MarkY-LunarG: Good point.  I'll pull it out.

loader/wsi.h
ianelliottus: Okay, thanks for the clarification!

loader/wsi.c
ianelliottus: Agreed--best to do this in another pull request.

loader/table_ops.h
ianelliottus: I did a little searching, and found different uses for "trampoline" functions (including [).](https://books.google.com/books?id=aJ1av7UFBPwC&pg=PA280&lpg=PA280&dq=thunk+trampoline&source=web&ots=YQcG0nOf7Z&sig=PZA-Pbx19C9Mnw_1myezK4x3QWQ&hl=en&sa=X&oi=book_result&ct=result#v=onepage&q&f=false)  In the end, I'm pretty convinced that you're using "entry" in the way that the loader folks (and Graham Sellers from AMD, and John McDonald from Valve) have used "trampoline", and that you should switch to that in order to avoid confusion:-).

MarkY-LunarG: Will do.

PR #955, opened 2016-09-21
Fix broken tracking of fence retirement for WSI fences

### LINKED ISSUE #954: Layers: False positive waiting on fence submitted to vkAcquireNextImageKHR
I'm changing cube to follow the recommended present synchronization from the example in the WSI spec, and I'm seeing a false positive from DS.  The call sequence is vkAquireNextImageKHR with a given fence, vkWaitForFence with the same fence, and then calling vkResetFences with the fence gives me:

ERROR: [DS] Code 15 : Fence 0xa is in use.

In the vkWaitForFences validation, pFence->signaler.first is NULL, so RetireWorkOnQueue is never called, because this is a fence that doesn't go into any queue, it's just passed to vkAcquireNextImage.  I'm not sure of the specifics, but vkWaitForFence should mark the fence as no longer in use.

chrisforbes:
Fixes #954.


chrisforbes:
Is clean on the farm.

tobine:
LGTM :+1:

PR #956, opened 2016-09-21
demos: Change cube to use FIFO present mode

Tony-LunarG:
And throttle using fences

WIP from suggestions from WSI working group.


courtney-g:
👍

courtney-g:
Ian, please take a look and make sure I didn't miss anything regarding use of fences. I'm not as familiar with that.

demos/cube.c
```diff
@@ -1981,6 +1987,9 @@ static void demo_cleanup(struct demo *demo) {
                              &demo->buffers[i].cmd);
     }
     free(demo->buffers);
+    for (i = 0; i < FRAME_LAG; i++) {
+        vkDestroyFence(demo->device, demo->fences[i], NULL);
```
chrisforbes: This is producing two validation errors on exit (with the fix from #954)

Need to ensure that these fences are not in flight before destroying them. Since they delimit WSI operations rather than queue operations, `DeviceWaitIdle` above doesn't guarantee they have retired.

The `DestroySwapchainKHR` call on :1975 should also not be done until the WSI fences have retired, so maybe fix this up at `DeviceWaitIdle`.

ianelliottus: I believe this is correct.  Just in case, I'm doing a little research to check this.

ianelliottus: I think that line 1991 should be replaced with the following:
```
    if (demo->fencesInited[i])
    {
        vkWaitForFences(demo->device, 1, &demo->fences[i], VK_TRUE, UINT64_MAX);
    }
   vkDestroyFence(demo->device, 1, &demo->fences[i], NULL);
```

ianelliottus: I'm not following the logic for why @chrisforbes suggests that the fences must be signaled before calling vkDestroySwapchainKHR, but I may be missing something:-).

demos/cube.c
```diff
@@ -872,21 +889,11 @@ static void demo_prepare_buffers(struct demo *demo) {
...
     }
 
-    // If mailbox mode is available, use it, as is the lowest-latency non-
-    // tearing mode.  If not, try IMMEDIATE which will usually be available,
-    // and is fastest (though it tears).  If not, fall back to FIFO which is
-    // always available.
+    // The FIFO present mode is guaranteed by the spec to be supported
+    // and to have no tearing.  MAILBOX is another option, but for fast
+    // rendering apps we could use a lot of gpu power rendering frames
+    // that are discarded and never displayed
```
ianelliottus: The request to change this was discussed during today's system integration group meeting.  We agreed to the change, short-term.  However, there is concern about this (without the ability for app's to correctly throttle themselves, which somebody pointed out isn't fully realizable with the fence approach used in this pull request, for all platforms) leading apps to rely on FIFO mode to throttle themselves.  However, this is the best we have for now.

I agreed to do another pull request where I'll add in some more comments, and example code for how to search for the other modes.  I'll get to this when I can.

ianelliottus:
There's one change that I suggested, but otherwise this looks good.  Thanks!

demos/cube.c
chrisforbes: @ianelliottus my argument there is the swapchain should not be destroyed while there are WSI operations in-flight. It might not actually be required -- spec just says `All uses of presentable images acquired from swapchain must have completed execution` --, but you get it for free, from ensuring that the fences have retired.

ianelliottus: I see what you're saying.  I don't think it works with this particular fence, which just says that the PE has finished with the image.  If the app has already submitted command buffers that use an image, this particular fence won't help with that.  Another fence would be needed to know when the queued command buffers are done.

chrisforbes: @ianelliottus You already know all the queued work is done -- you've called DWI. The only thing that can be in flight at this point is whatever work the PE is doing between ANI and its fence retiring.

Tony-LunarG:
Updated to wait for fences before destroying and to use multiple semphores for all roles

Tony-LunarG:
Pushed to master

PR #958, opened 2016-09-22
Tobin gh663 image bound for ds

### LINKED ISSUE #663: VkBuffer objects not bound to VkDeviceMemory objects are not detected on use

tobine:
Added new check to make sure memory is bound to any image used to update descriptor set.

Fixes #663 

Also added test to hit this case. Some related work that came up along the way and is also in this PR:

- Updated test framework to handle multiple expected errors
- Added some code to clear memory bindings for objects bound at the time that vkFreeMemory() is called
- Fixed broken aliasing test that was re-binding image to new memory
- Updated a bunch of tests to account for modified error message


layers/descriptor_sets.cpp
```diff
@@ -688,6 +688,11 @@ bool cvdescriptorset::ValidateImageUpdate(VkImageView image_view, VkImageLayout
     if (image_node) {
         format = image_node->createInfo.format;
         usage = image_node->createInfo.usage;
+        // Validate that memory is bound to image
+        if (ValidateMemoryIsBoundToImage(dev_data, image_node, "vkUpdateDescriptorSets()")) {
```
chrisforbes: I don't think this is right. You can't get here without first continuing past an error in `CreateImageView`.

tests/layer_validation_tests.cpp
```diff
@@ -7557,6 +7570,168 @@ TEST_F(VkLayerTest, InvalidCmdBufferDescriptorSetImageSamplerDestroyed) {
...
+    descriptor_write.sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
+    descriptor_write.dstSet = descriptorSet;
+    descriptor_write.dstBinding = 0;
+    descriptor_write.descriptorCount = 1;
+    descriptor_write.descriptorType = VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER;
+    descriptor_write.pImageInfo = &img_info;
+    // Break memory binding and attempt update
+    vkFreeMemory(m_device->device(), image_memory, nullptr);
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                                         " used with no memory bound. Memory should be bound by calling vkBindImageMemory() and ");
```
chrisforbes: I don't think this is actually the error you want. The app /can't/ fix it by calling BindImageMemory -- the only valid thing they can do with the image is to destroy it.

tests/layer_validation_tests.cpp
```diff
@@ -146,14 +147,22 @@ class ErrorMonitor {
             *m_bailout = true;
         }
         string errorString(msgString);
-        if (m_desiredMsgSet && errorString.find(m_desiredMsg) != string::npos) {
-            if (m_msgFound) { // If multiple matches, don't lose all but the last!
-                m_otherMsgs.push_back(m_failureMsg);
+        bool unexpected = true;
+        if (m_desiredMsgSet) {
```
chrisforbes: If you're going to do this, the m_desiredMsgSet flag is no longer needed.

tests/layer_validation_tests.cpp
```diff
@@ -146,14 +147,22 @@ class ErrorMonitor {
             *m_bailout = true;
         }
         string errorString(msgString);
-        if (m_desiredMsgSet && errorString.find(m_desiredMsg) != string::npos) {
-            if (m_msgFound) { // If multiple matches, don't lose all but the last!
-                m_otherMsgs.push_back(m_failureMsg);
+        bool unexpected = true;
```
chrisforbes: This control flow has got really messy :(

tests/layer_validation_tests.cpp
```diff
@@ -189,27 +196,29 @@ class ErrorMonitor {
...
+                FAIL() << "Expected to succeed but got error: " << msg;
+            }
         }
     }
 
   private:
     VkFlags m_msgFlags;
-    string m_desiredMsg;
-    string m_failureMsg;
+    std::unordered_set<string> m_desiredMsgs;
```
chrisforbes: I need to think really carefully about this... I can see it's the easy option, but are multiple messages /really/ unavoidable?

tests/layer_validation_tests.cpp
```diff
@@ -130,11 +131,11 @@ class ErrorMonitor {
...
         // also discard all collected messages to this point
         test_platform_thread_lock_mutex(&m_mutex);
-        m_failureMsg.clear();
+        m_failureMsgs.clear();
         m_otherMsgs.clear();
-        m_desiredMsg = msgString;
+        m_desiredMsgs.insert(msgString);
         m_msgFound = VK_FALSE;
-        m_msgFlags = msgFlags;
+        m_msgFlags |= msgFlags;
```
chrisforbes: This doesn't work for subtests which use different severity levels.

tests/layer_validation_tests.cpp
```diff
@@ -7557,6 +7570,168 @@ TEST_F(VkLayerTest, InvalidCmdBufferDescriptorSetImageSamplerDestroyed) {
     vkDestroyDescriptorPool(m_device->device(), ds_pool, NULL);
 }
 
+TEST_F(VkLayerTest, DescriptorImageUpdateNoMemoryBound) {
+    TEST_DESCRIPTION("Attempt an image descriptor set update without any memory bound to image.");
```
chrisforbes: This could be clearer-- it's an image descriptor set update with an image whose memory has been freed.

layers/core_validation.cpp
```diff
@@ -637,6 +637,24 @@ static bool reportMemReferencesAndCleanUp(layer_data *dev_data, DEVICE_MEM_INFO
             log_msg(dev_data->report_data, VK_DEBUG_REPORT_INFORMATION_BIT_EXT, obj.type, obj.handle, __LINE__,
                     MEMTRACK_FREED_MEM_REF, "MEM", "VK Object 0x%" PRIxLEAST64 " still has a reference to mem obj 0x%" PRIxLEAST64,
                     obj.handle, (uint64_t)pMemObjInfo->mem);
+            // Clear mem binding for bound objects
+            switch (obj.type) {
+            case VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT: {
+                auto image_node = getImageNode(dev_data, reinterpret_cast<VkImage &>(obj.handle));
+                assert(image_node); // Any destroyed images should already be removed from bindings
+                image_node->mem = VK_NULL_HANDLE;
```
chrisforbes: This isn't going to do what you want. CV will be quite happy to accept BindImageMemory again after this, if all you do is null out this handle.

layers/core_validation.cpp
```diff
@@ -637,6 +637,24 @@ static bool reportMemReferencesAndCleanUp(layer_data *dev_data, DEVICE_MEM_INFO
...
+                break;
+            }
+            case VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT: {
+                auto buff_node = getBufferNode(dev_data, reinterpret_cast<VkBuffer &>(obj.handle));
+                assert(buff_node); // Any destroyed buffers should already be removed from bindings
+                buff_node->mem = VK_NULL_HANDLE;
```
chrisforbes: Same here -- you dont want to accept BindBufferMemory after this.

tests/layer_validation_tests.cpp
chrisforbes: [I suppose you could hack this back to mostly working by having Verify* set it back to ERROR_BIT, but that's wobbly...]

layers/core_validation.cpp
```diff
@@ -731,8 +750,10 @@ bool ValidateMemoryIsBoundToBuffer(const layer_data *dev_data, const BUFFER_NODE
         if (0 == buffer_node->mem) {
             result = log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT,
                              reinterpret_cast<const uint64_t &>(buffer_node->buffer), __LINE__, MEMTRACK_OBJECT_NOT_BOUND, "MEM",
-                             "%s: VkBuffer object 0x%" PRIxLEAST64 " used without first calling vkBindBufferMemory.", api_name,
-                             reinterpret_cast<const uint64_t &>(buffer_node->buffer));
+                             "%s: VkBuffer object 0x%" PRIxLEAST64 " used with no memory bound. Memory should be bound by calling "
+                                                                   "vkBindImageMemory() and then the bound memory must not be "
```
tobine: Yeah, not sure what happened with these two but I updated formatting manually and clang-format likes it now.

tests/layer_validation_tests.cpp
tobine: I'm not seeing the issue here. When callback gets errors it &'s error bit with msgFlags to see if it's interested in the error. Is your concern that a check for non-error msg would pick up noise of all other msgs at that level?
I'd also be find just saying that multiple messages have to be at same severity level and keeping assignment. That probably makes more sense initially so I'll just keep it as assignment. Do you see some issue in that case?

tests/layer_validation_tests.cpp
tobine: Renamed bool to make less confusing.

tests/layer_validation_tests.cpp
tobine: Yeah, gone.

tests/layer_validation_tests.cpp
tobine: Yeah, agree that splitting the two cases is nicer.

layers/core_validation.cpp
tobine: Yeah, need to protect against re-binding.

tests/layer_validation_tests.cpp
tobine: Yeah, updated. Wrote initial description before I saw that vkCreateImageView would always catch the never-bound error.

tobine:
Pushed some updates but I'll also split the never-bound vs. bound-then-freed error messages to be unique for clarity and add handling to prevent the re-bind case. Thanks @chrisforbes

tests/layer_validation_tests.cpp
chrisforbes: Suppose you have two subtests, one looking for a warning, the other looking for an error. The second subtest sees the warnings too, with this change.

Keeping it as assignment as you suggest pretty much solves it.

tests/layer_validation_tests.cpp
```diff
@@ -146,14 +147,20 @@ class ErrorMonitor {
             *m_bailout = true;
         }
         string errorString(msgString);
-        if (m_desiredMsgSet && errorString.find(m_desiredMsg) != string::npos) {
-            if (m_msgFound) { // If multiple matches, don't lose all but the last!
-                m_otherMsgs.push_back(m_failureMsg);
+        bool found_expected = false;
+        for (auto desired_msg : m_desiredMsgs) {
```
chrisforbes: Still not a huge fan of this in principle, but :+1: for making it clearer.

tobine:

Also opened #964 to split never-bound vs. re-bound errors and make sure we correctly catch re-bound case. I'll work on that next.

PR #959, opened 2016-09-22
layers: Remove spurious using BASE_NODE::*.

chrisforbes:
Inheritance is all public (except for FRAMEBUFFER_NODE, which ought to
have been), and there's no template inheritance involved, so these
didn't do anything interesting.

Fix FRAMEBUFFER_NODE to publicly inherit BASE_NODE, and drop the noise
everywhere.


tobine:
LGTM :+1:

chrisforbes:
Pushed to master

PR #961, opened 2016-09-22
layers: Explicitly delete IMAGE_NODE, BUFFER_NODE copy constructors

chrisforbes:
Previously we defined copy constructors that would almost certainly lead
to bugs if these objects were copied (they copied the atomic values,
etc). The layer doesn't copy these, just disallow it.


chrisforbes:
Pushed to master

PR #966, opened 2016-09-22
layers: Distinguish never-bound from un-bound memory cases

### LINKED ISSUE #964: Catch re-bind memory case
Tracking issue to make sure core_validation correctly catches case where memory is re-bound to image/buffer.
Also want to split error messages so you get unique error for never-bound vs. re-bound cases, which are currently one message.

tobine:
Fixes #964

Added special MEMORY_UNBOUND handle (0xF..FE) that indicates that memory
bound to an object has been freed. When attempting to bind memory or
checking for bound memory, distinguish the never-bound case from the
memory un-bound case.

Update tests to account for new error messages.

There's a sliver of exposure here if an actual memory handle is MEMORY_UNBOUND.
We could remove that exposure by never having unique_objects return MEMORY_UNBOUND
as a handle. I believe the exposure is small enough that we don't need to do that,
but am open to other opinions.


layers/core_validation.cpp
```diff
@@ -787,28 +799,36 @@ bool ValidateMemoryIsBoundToBuffer(const layer_data *dev_data, const BUFFER_NODE
...
+        if (mem_info) {
+            DEVICE_MEM_INFO *prev_binding = getMemObjInfo(dev_data, *mem_binding);
+            if (prev_binding) {
+                skip_call |=
+                    log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_MEMORY_EXT,
+                            reinterpret_cast<uint64_t &>(mem), __LINE__, MEMTRACK_REBIND_OBJECT, "MEM",
+                            "In %s, attempting to bind memory (0x%" PRIxLEAST64 ") to object (0x%" PRIxLEAST64
+                            ") which has already been bound to mem object 0x%" PRIxLEAST64,
+                            apiName, reinterpret_cast<uint64_t &>(mem), handle, reinterpret_cast<uint64_t &>(prev_binding->mem));
+            } else if (*mem_binding == MEMORY_UNBOUND) {
```
courtney-g: What about sparse memory?

tobine: Yeah, good point. Sparse is a giant hole in validation in general, but no need to make it worse. Pushed update to allow rebind in sparse case.

tobine:

PR #967, opened 2016-09-22
Android smoke

cnorthrop:
This series gets Smoke working for Android.


cnorthrop:
cherry-picked to master

PR #968, opened 2016-09-22
android: Add Android and C++ support to a couple files

cnorthrop:
These are needed downstream in VulkanTools for Android support.


cnorthrop:
cherry-picked to master

PR #975, opened 2016-09-23
Stroyan cmake fixes

stroyan:
These changes get windows builds working with recent ImageMagick versiosn and Visual Studio.
I noted that the ImageMagick was not found on my system.  The current use of FIND_PATH for headers
trips over cached variable values and fails on either first cmake or repeated use of cmake.

Using visual studio to build tries to use parallel builds.  That was turned off in the msbuild script.
I tracked down the places that parallel builds failed.  I then enabled them in build_windows_targets.bat.


stroyan:
Pushed to master

PR #977, opened 2016-09-26
layers: Trim out some more old MT renderpass junk

chrisforbes:
- RENDER_PASS_NODE::attachments[i].attachment was always i.
- This whole structure is noise; we can use the VkAttachmentDescription
  directly.


layers/core_validation.cpp
```diff
@@ -10247,36 +10236,36 @@ CmdBeginRenderPass(VkCommandBuffer commandBuffer, const VkRenderPassBeginInfo *p
         if (renderPass) {
             uint32_t clear_op_size = 0; // Make sure pClearValues is at least as large as last LOAD_OP_CLEAR
             cb_node->activeFramebuffer = pRenderPassBegin->framebuffer;
-            for (size_t i = 0; i < renderPass->attachments.size(); ++i) {
+            for (size_t i = 0; i < renderPass->pCreateInfo->attachmentCount; ++i) {
                 MT_FB_ATTACHMENT_INFO &fb_info = framebuffer->attachments[i];
-                VkFormat format = renderPass->pCreateInfo->pAttachments[renderPass->attachments[i].attachment].format;
```
tobine: Wow, this was slightly confusing. Good riddance.

layers/core_validation.cpp
```diff
@@ -10026,17 +10026,6 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateRenderPass(VkDevice device, const VkRenderP
         render_pass->hasSelfDependency = has_self_dependency;
         render_pass->subpassToNode = subpass_to_node;
 #if MTMERGESOURCE
```
tobine: Any need to keep this #define?

chrisforbes: No, I'll remove it.

chrisforbes:
Pushed to master.

PR #984, opened 2016-09-27
layers: Indicate correct source for validateCommandBufferState() error

### LINKED ISSUE #982: Misleading error message from vkCmdExecuteCommands() with non-executable CB
If you forget to call vkEndCommandBuffer(), the message in validateCommandBufferState() assumes the caller is vkQueueSubmit(), which sent me on a bit of a wild goose chase. Would it be possible to detect if the caller is vkCmdExecuteCommands() and change its message accordingly?

tobine:
Fixes #982

Instead of hard-coding vkQueueSubmit, take a parameter to clarify if the error originates
from vkQueueSubmit() or vkCmdExecuteCommands()

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -4852,7 +4852,7 @@ static bool validateCommandBufferSimultaneousUse(layer_data *dev_data, GLOBAL_CB
     return skip_call;
 }
 
-static bool validateCommandBufferState(layer_data *dev_data, GLOBAL_CB_NODE *pCB) {
+static bool validateCommandBufferState(layer_data *dev_data, GLOBAL_CB_NODE *pCB, const char *call_source) {
     bool skip_call = false;
     // Validate ONE_TIME_SUBMIT_BIT CB is not being submitted more than once
     if ((pCB->beginInfo.flags & VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT) && (pCB->submitCount > 1)) {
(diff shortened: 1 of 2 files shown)


lentinem:
I know this is a small change but it would be good if we could add the flag guard for these checks.

mark-lunarg:
👍

tobine:
@lentinem I pushed an update that guards these checks with bools. Always on by default until the validation flags system is tied in and we define enum value(s) that can disable these checks.
Does this match with your thinking?
I initially put two checks around the function call to do vkValidateCommandBufferState() rather than just a single early return in the function. That saves the function call but can also be more error prone if people add new calls to that function without the guard.
I think long term we'll want to merge these checks into log_msg() and/or look-up the bools from an array (or map) based on the unique enum value of each check. That will minimize extra noise in the code, but for now we don't yet have the unique enum values.

chrisforbes:
@tobine I don't think you want to merge the flag check into `log_msg()` if you hope to get performance out  of disabling checks. By the time you've decided to call `log_msg()`, the damage/work is already done.

tobine:
@chrisforbes log_msg() integration would be for convenience of being able to simply disable individual checks. Still definitely want some higher-level bools to turn off large swaths of checks and related logic.

mark-lunarg:
If each message ends up having a unique identifier (instead of the coarse enum we have now) we can disable individual checks on the back-end, in log_msg().  For swath control it seems like the 'pre-call' validation will need tests for related groupings of validation checks.  Or, are we thinking that _each validation check_ will have its own test?  Both ways are workable though this still seems a bit squirrely from an implementation-guideline standpoint.

lentinem:
This looks reasonable for me although I would also like to stick it in the validation function instead of conditionally call the function. That we we can have

validateA(...) {
  if (disabled) return;
  ...
}

validateCall(...) {
  validateA(...);
  valdiateB(...);
  ...
}

tobine:
Pushed update with bool check inside of function. That should be cleaner and easier to maintain going fwd. One more thumb up on this before I push?

layers/core_validation.h
```diff
@@ -60,6 +60,16 @@
 #include <list>
 #include <deque>
 
+/*
```
mark-lunarg: The /* */ comments seem to be popping up occasionally, I think we decided upon the // style comments except for the copyright messages.

layers/core_validation.cpp
```diff
@@ -4852,15 +4852,17 @@ static bool validateCommandBufferSimultaneousUse(layer_data *dev_data, GLOBAL_CB
     return skip_call;
 }
 
-static bool validateCommandBufferState(layer_data *dev_data, GLOBAL_CB_NODE *pCB) {
-    bool skip_call = false;
+static bool validateCommandBufferState(layer_data *dev_data, GLOBAL_CB_NODE *pCB, const char *call_source) {
+    bool skip = false;
+    if (dev_data->instance_state->disabled.command_buffer_state)
+        return skip;
```
lentinem: I would reverse this order so we just return false at the top; don't care too much though.

lentinem:
:+1:

tobine:

PR #985, opened 2016-09-27
Build fixes

rib:
A couple of minor cmake issues I found when building on Linux


karl-lunarg:
This is a strange one because the problem fixed by the second change (to tests/layers/CMakeLists.txt) only shows up when NOT using make -j (non-parallel) build.

karl-lunarg:
Closing without CLA, since these are just build fixes.

Thanks!

PR #986, opened 2016-09-27
layers: Fix state tracking for vkFlushMappedMemoryRanges()

tobine:
WIP

MarkL, please check this out. I don't think it's quite right yet, but I got a bug report related to this code and I think there are definitely some issues.

Checks for underflow and overflow when flushing mapped memory weren't properly handling
offset. Underflow check should go from the base of the shadow memory to the range offset.
The overflow check should go from the end of the mapped memory range to the end of the
shadow memory.
Finally, when copying the region, the start address of the copy should include offset.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -10872,7 +10872,7 @@ static bool ValidateAndCopyNoncoherentMemoryToDriver(layer_data *dev_data, uint3
             if (mem_info->shadow_copy) {
                 VkDeviceSize size = (mem_info->mem_range.size != VK_WHOLE_SIZE)
                                         ? mem_info->mem_range.size
-                                        : (mem_info->alloc_info.allocationSize - pMemRanges[i].offset);
+                                        : (mem_info->alloc_info.allocationSize - mem_info->mem_range.offset);
                 char *data = static_cast<char *>(mem_info->shadow_copy);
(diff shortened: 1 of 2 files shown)


mark-lunarg:
Yeah, it looks OK.  Do you have a test case that I can try this with?  The issue that prompted the original changes was that the shadow-copy needed to follow the same restrictions as the device copy -- I'd like to double-check the loop that checks for overflow in the debugger, but if it's passing your repro case then it seems OK to me.

tobine:
No test case, just complaints from a user of errors that he thought were invalid. Upon inspection I agree that there's something fishy so this was my attempt to fix it up.
With these changes I'm seeing some new errors on some android devices w/ LVL tests so if you don't see anything clearly wrong with my changes I'll dig into it a bit further.

tobine:
@chrisforbes if you have some time take a look at this and let me know if you see any issues. Seeing new unexpected errors on one vendor with these changes that I'm investigating further but wanted to get insight from you if you're able.

chrisforbes:
From offline discussion: not convinced this is the right thing.

chrisforbes:
Tests?

tobine:
@chrisforbes I added an additional case to NonCoherentMemoryMapping test to demonstrate that the previous bug with the flushed offset is no longer present. Force pushed update.

PR #988, opened 2016-09-28
Validate current subpass against pipeline subpass

### LINKED ISSUE #933: Layers does not check if proper pipeline is bound after vkCmdNextSubpass
Consider following scenario:

```
vkCmdBeginRenderpass()
vkCmdBindPipeline(A); 
vkCmdDraw();
vkCmdNextSubpass();
vkCmdDraw();   <- Wrong, because pipeline A subpass index is 0, while current subass index is 1
```

Validation layers should validate pipeline subpass index wrt subpass index in each drawcall - not only when binding the pipeline.

chrisforbes:
Fixes #933.


mark-lunarg:
LGTM.

layers/core_validation.cpp
```diff
@@ -3167,6 +3167,14 @@ static bool validatePipelineDrawtimeState(layer_data const *my_data,
                         reinterpret_cast<uint64_t &>(pCB->activeRenderPass->renderPass), reinterpret_cast<uint64_t &>(pPipeline),
                         reinterpret_cast<const uint64_t &>(pPipeline->graphicsPipelineCI.renderPass), err_string.c_str());
         }
+
+        if (pPipeline->graphicsPipelineCI.subpass != pCB->activeSubpass) {
```
tobine: Is there explicit spec language for this requirement? The check looks good and makes sense but in ~5min I couldn't find spec language enforcing this so we should add it if it's not there.

chrisforbes: No, there isn't -- it's really vague about this. The closest you get is the first para of 7.2.

chrisforbes: See KhronosGroup/Vulkan-Docs#375

chrisforbes:
Clean on the farm; pushed to master.

PR #989, opened 2016-09-28
Fix incorrect array index in inner loop of instance extension validation

janharaldfredriksen-arm:
Change-Id: I27606fb32049bb531131ca29357d79491e3f96a7

The loop counter of the enabled layers is "j" but the array indexing is done using "i".


MarkY-LunarG:
Committed 0be3cba1f21ace8b2db84accf5125adca2c91b46

Issue #990, opened 2016-09-28
Refactor QueueSubmit to only update state when submit occurs


tobine:
Need to partition QueueSubmit core_validation work into "PreValidate" and "PostUpdate" sections so that state is only updated if the call isn't skipped.

chrisforbes:
Dupe of #945

PR #991, opened 2016-09-28
Tobin buffer view in use check

tobine:
Incorporate bufferView into existing in-use and command buffer invalid framework along with tests for both cases.

The cmd buffer invalid test has some unexpected errors that highlight a deficiency in QueueSubmit state management that I've noted in #990

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5404,6 +5404,8 @@ VKAPI_ATTR void VKAPI_CALL DestroyFence(VkDevice device, VkFence fence, const Vk
 
 // For given obj node, if it is use, flag a validation error and return callback result, else return false
 bool ValidateObjectNotInUse(const layer_data *dev_data, BASE_NODE *obj_node, VK_OBJECT obj_struct) {
+    if (dev_data->instance_state->disabled.object_in_use)
+        return false;
     bool skip = false;
     if (obj_node->in_use.load()) {
         skip |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, obj_struct.type, obj_struct.handle, __LINE__,
@@ -5723,6 +5725,8 @@ VKAPI_ATTR void VKAPI_CALL DestroyBuffer(VkDevice device, VkBuffer buffer,
 
 static bool PreCallValidateDestroyBufferView(layer_data *dev_data, VkBufferView buffer_view, BUFFER_VIEW_STATE **buffer_view_state,
                                              VK_OBJECT *obj_struct) {
+    if (dev_data->instance_state->disabled.destroy_buffer_view)
+        return false;
     bool skip = false;
     *buffer_view_state = getBufferViewState(dev_data, buffer_view);
     if (buffer_view_state) {
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5721,22 +5721,39 @@ VKAPI_ATTR void VKAPI_CALL DestroyBuffer(VkDevice device, VkBuffer buffer,
     }
 }
 
+static bool PreCallValidateDestroyBufferView(layer_data *dev_data, VkBufferView buffer_view, BUFFER_VIEW_STATE **buffer_view_state,
+                                             VK_OBJECT *obj_struct) {
+    bool skip = false;
+    *buffer_view_state = getBufferViewState(dev_data, buffer_view);
+    if (buffer_view_state) {
+        *obj_struct = {reinterpret_cast<uint64_t &>(buffer_view), VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_VIEW_EXT};
+        skip |= ValidateObjectNotInUse(dev_data, *buffer_view_state, *obj_struct);
+    }
+    return skip;
+}
+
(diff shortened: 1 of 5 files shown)


layers/descriptor_sets.cpp
```diff
@@ -975,9 +975,7 @@ void cvdescriptorset::TexelDescriptor::CopyUpdate(const Descriptor *src) {
 void cvdescriptorset::TexelDescriptor::BindCommandBuffer(const core_validation::layer_data *dev_data, GLOBAL_CB_NODE *cb_node) {
     auto bv_state = getBufferViewState(dev_data, buffer_view_);
```
lentinem: Add a flag check here.

layers/core_validation.cpp
```diff
@@ -5710,14 +5723,20 @@ VKAPI_ATTR void VKAPI_CALL DestroyBuffer(VkDevice device, VkBuffer buffer,
...
     layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
-
+    bool skip = false;
     std::unique_lock<std::mutex> lock(global_lock);
     auto view_state = getBufferViewState(dev_data, bufferView);
     if (view_state) {
+        VK_OBJECT obj_struct = {reinterpret_cast<uint64_t &>(bufferView), VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_VIEW_EXT};
+        skip |= ValidateObjectNotInUse(dev_data, view_state, obj_struct);
+        // Any bound cmd buffers are now invalid
+        invalidateCommandBuffers(view_state->cb_bindings, obj_struct);
```
lentinem: Add flag checks to these functions and make this follow the validate, call, update (invalidate) paradigm.

layers/core_validation.cpp
```diff
@@ -621,6 +621,19 @@ void AddCommandBufferBindingBuffer(const layer_data *dev_data, GLOBAL_CB_NODE *c
     buff_node->cb_bindings.insert(cb_node);
 }
 
+// Create binding link between given buffer view node and its buffer with command buffer node
+void AddCommandBufferBindingBufferView(const layer_data *dev_data, GLOBAL_CB_NODE *cb_node, BUFFER_VIEW_STATE *view_state) {
+    // First add bindings for bufferView
```
lentinem: Add flag.

tobine:
I refactored this to use Pre/Post architecture and added flags for the validation checks.

layers/core_validation.cpp
```diff
@@ -5708,17 +5723,41 @@ VKAPI_ATTR void VKAPI_CALL DestroyBuffer(VkDevice device, VkBuffer buffer,
...
-    if (view_state) {
-        dev_data->bufferViewMap.erase(bufferView);
+    // Common data objects use pre & post call
+    BUFFER_VIEW_STATE *buffer_view_state = nullptr;
+    VK_OBJECT obj_struct;
+    // Validate state before calling down chain, update common data if we'll be calling down chain
+    bool skip = PreCallValidateDestroyBufferView(dev_data, bufferView, &buffer_view_state, &obj_struct);
+    if (!skip) {
+        // We're making call so update state
+        PostCallRecordDestroyBufferView(dev_data, bufferView, buffer_view_state, obj_struct);
```
lentinem: I have mixed feelings in this case. We do not know if destroy failed so there isn't much we can do but if we are establishing precedent for other checks then the "driver" call should be before this.

layers/core_validation.cpp
```diff
@@ -5708,17 +5723,41 @@ VKAPI_ATTR void VKAPI_CALL DestroyBuffer(VkDevice device, VkBuffer buffer,
...
+        skip |= ValidateObjectNotInUse(dev_data, *buffer_view_state, *obj_struct);
+    }
+    return skip;
+}
+
+static void PostCallRecordDestroyBufferView(layer_data *dev_data, VkBufferView buffer_view, BUFFER_VIEW_STATE *buffer_view_state,
+                                            VK_OBJECT obj_struct) {
+    dev_data->bufferViewMap.erase(buffer_view);
+    // Any bound cmd buffers are now invalid
+    invalidateCommandBuffers(buffer_view_state->cb_bindings, obj_struct);
```
lentinem: This function should also have a flag that only invalidates the bindings if we are doing cb tracking.

layers/core_validation.cpp
tobine: Since this is a void function I put the PostCall* before the call down the chain to save having to unlock and then re-lock.
For non-void functions I agree that PostCall* should always go after call down chain such as existing "PostCallRecordCreateImageView()".

lentinem: Right; but for consistency it seems like doing the driver call first makes sense.

layers/core_validation.cpp
tobine: If you look at the logic I don't believe this is needed. invalidateCommandBuffers() only acts on binding that have been added for a given object so if no bindings are added then invalidateCommandBuffers() is virtually a no-op other than the function call overhead which would still be incurred with the flag.
I can add a flag separately that disables all cmd buffer<->object binding but that's beyond the scope of the PR and I'd like to hold off on that until some further refactoring is done.

lentinem:
👍

tobine:

PR #993, opened 2016-09-29
Add environment variable for alternate vk_layer_settings.txt

### LINKED ISSUE #983: support placing vk_layer_settings.txt in a directory other than CWD
The validation layers can be configured through a file named vk_layer_settings.txt. In a webinar it was claimed that this file must reside in the application's CWD at the time the layers are initialized.

Three issues:

1. Sometimes the app's CWD is a location that is not writable. I.e., there is no way to create a new file in that directory.

2. CWD at the time the layers are initialized may be different from CWD at app start, and/or different from CWD after the layers are read. Mandating that the file be in CWD at "some time" during init seems fragile for apps that change their CWD.

3. Driver and tools authors might want to configure layers when running apps that were authored by 3rd parties. Dropping a vk_layer_settings.txt file into each app's CWD (and discovering each app's CWD) could be tedious for a driver or tool author who tests against 100 different apps.

Proposed solution: Permit an optional environment variable to override the pathname to the settings file.

Since enabling of layers is already controlled through an environment variable, adding another environment variable for the location of the settings file seems no less secure than the status quo.

lenny-lunarg:
Add the VK_LAYER_SETTINGS_FILE environment variable, which specifies an a file path to be used in the place of vk_layer_settings.txt. If the environment variable isn't found, the layers revert to the old behavior (vk_layer_settings.txt in the working directory).

This fixes #983

### DIFF
--- a/layers/vk_layer_config.cpp
+++ b/layers/vk_layer_config.cpp
@@ -21,14 +21,15 @@
  * Author: Tobin Ehlis <tobin@lunarg.com>
  * Author: Mark Lobodzinski <mark@lunarg.com>
  **************************************************************************/
-#include <fstream>
-#include <string>
-#include <map>
-#include <string.h>
-#include <vulkan/vk_layer.h>
-#include <iostream>
 #include "vk_layer_config.h"
 #include "vulkan/vk_sdk_platform.h"
+#include <fstream>
+#include <iostream>
+#include <map>
+#include <string.h>
+#include <string>
+#include <sys/stat.h>
+#include <vulkan/vk_layer.h>
 
 #define MAX_CHARS_PER_LINE 4096
 
@@ -48,6 +49,25 @@ class ConfigFile {
 };
 
 static ConfigFile g_configFileObj;
+
+std::string getEnvironment(const char *variable) {
+#if !defined(__ANDROID__) && !defined(_WIN32)
+    const char *output = getenv(variable);
+    return output == NULL ? "" : output;
+#elif defined(_WIN32)
+    int size = GetEnvironmentVariable(variable, NULL, 0);
+    if (size == 0) {
+        return "";
+    }
+    char *buffer = new char[size];
+    GetEnvironmentVariable(variable, buffer, size);
+    std::string output = buffer;
+    delete[] buffer;
+    return output;
+#else
+    return "";
+#endif
+}
 
 const char *getLayerOption(const char *_option) { return g_configFileObj.getOption(_option); }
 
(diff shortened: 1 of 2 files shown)


courtney-g:
:+1:

layers/vk_layer_config.cpp
```diff
@@ -49,6 +49,25 @@ class ConfigFile {
 
 static ConfigFile g_configFileObj;
 
+std::string getEnvironment(const char* variable)
```
mark-lunarg: Lenny, we need to clang-format this source.  Your C is all ahoo!

layers/vk_layer_config.cpp
```diff
@@ -155,7 +174,8 @@ ConfigFile::~ConfigFile() {}
 const char *ConfigFile::getOption(const std::string &_option) {
     std::map<std::string, std::string>::const_iterator it;
     if (!m_fileIsParsed) {
-        parseFile("vk_layer_settings.txt");
+        std::string envFile = getEnvironment("VK_LAYER_SETTINGS_FILE");
```
mark-lunarg: Karl had the fine suggestion that the env var could either be a directory or a path to a specific filename.  If it was a directory, you'd look there for the standard filename instead of the normal spot, else if it's a filename we'd use that.

lenny-lunarg:
The path suggestion is a good idea, I'll add the ability to check for a path and then run it through clang-format.

layers/vk_layer_config.cpp
karl-lunarg: And if you go this route, it may be better to make the environment variable

`VK_LAYER_SETTINGS_PATH`

courtney-g: I was thinking of making the same suggestion. However, having the env be a full filename would make it clear which vk_settings file is being used. Felt a little more specific. Though I'm find with search path as well.

MarkY-LunarG:
Shouldn't we update the appropriate Markdown at the same time?

layers/vk_layer_config.cpp
karl-lunarg: I'm OK with the @courtney-g thinking as well.  It is simpler, satisfies the requirement, and is just as flexible.  So I cast a medium-energy vote in that direction.

karl-lunarg:
> Shouldn't we update the appropriate Markdown at the same time?

I'm not sure that there's one in this repo.   Certainly the LunarXchange docs need updating.

MarkY-LunarG:
In this repo, layers\README.md mentions the vk_layer_settings.txt file.  That should be updated.  The other file (outside of this repo) is LunarHub-Doc/layers.md.  We should make sure to update both.

PR #994, opened 2016-09-29
Update LVL to 1.0.28 spec version

mark-lunarg:


MarkY-LunarG:
LGTM

courtney-g:
:+1:

generator.py
```diff
@@ -3142,6 +3143,8 @@ def genCmd(self, cmdinfo, name):
             for param in params:
                 len = self.getLen(param)
                 if len:
+                    if '::' in len:
```
stroyan: This appears redundant because getLen already does this replacement.

generator.py
```diff
@@ -2580,6 +2580,7 @@ def makeThreadUseBlock(self, cmd, functionprefix):
                     else:
                         # externsync can list members to synchronize
                         for member in externsync.split(","):
+                            member = str(member).replace("::", "->", 1)
```
stroyan: If there is more than one '::' in the string we will probably want to replace them all.  So we won't want the max replacement count of 1.

generator.py
```diff
@@ -3256,6 +3259,8 @@ def getLen(self, param):
                 result = len.split(',')[0]
             else:
                 result = len
+            if '::' in result:
+                result = str(result).replace('::', '->', 1)
```
stroyan: The 'if' test is not needed.  The replace will only affect when that is true.
This replacement count of 1 could also be removed.

generator.py
mark-lunarg: Done

generator.py
mark-lunarg: Thanks Mike!

generator.py
mark-lunarg: Done.

mark-lunarg:
Pushed to Master.  Thanks for the review help.

PR #995, opened 2016-09-29
Check srcAccessMask in transitions from PRESENT_SRC_KHR

Tony-LunarG:
From the vkAcquireNextImageKHR section of the spec:

The vkCmdWaitEvents or vkCmdPipelineBarrier used to transition the image away from VK_IMAGE_LAYOUT_PRESENT_SRC_KHR layout must have dstStageMask and dstAccessMask parameters set based on the next use of the image. The srcAccessMask must include VK_ACCESS_MEMORY_READ_BIT to ensure that all prior reads by the presentation engine are complete before the image layout transition occurs.


tests/layer_validation_tests.cpp
```diff
@@ -10701,6 +10701,45 @@ TEST_F(VkLayerTest, InvalidBarriers) {
...
+    range.baseArrayLayer = 0;
+    range.layerCount = 1;
+    barrier.subresourceRange = range;
+    VkCommandBufferObj cmdbuf(m_device, m_commandPool);
+    cmdbuf.BeginCommandBuffer();
+    cmdbuf.PipelineBarrier(VK_PIPELINE_STAGE_ALL_COMMANDS_BIT, VK_PIPELINE_STAGE_ALL_COMMANDS_BIT, 0, 0, nullptr, 0, nullptr, 1,
+                           &barrier);
+    barrier.oldLayout = VK_IMAGE_LAYOUT_PRESENT_SRC_KHR;
+    barrier.newLayout = VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL;
+    cmdbuf.PipelineBarrier(VK_PIPELINE_STAGE_ALL_COMMANDS_BIT, VK_PIPELINE_STAGE_ALL_COMMANDS_BIT, 0, 0, nullptr, 0, nullptr, 1,
```
chrisforbes: Error is generated at this point. No need to actually submit the command buffer.

layers/core_validation.cpp
```diff
@@ -8723,6 +8723,15 @@ static bool ValidateBarriers(const char *funcName, VkCommandBuffer cmdBuffer, ui
                                          dst_q_f_index, dev_data->phys_dev_properties.queue_family_properties.size());
                 }
             }
+            if ((mem_barrier->oldLayout == VK_IMAGE_LAYOUT_PRESENT_SRC_KHR) &&
```
chrisforbes: This should really be done in `ValidateMaskBitsFromLayouts`, like all the other similar constraints.

Tony-LunarG:
Fixed

chrisforbes:
@Tony-LunarG what do you want to do about that edge case you raised in email, where the layout doesn't change?

Tony-LunarG:
I wasn't sure enough about the optimization to do it.  Are the src and dst access masks meaningless if the layout isn't changing?  If so, I'll add the optimization to this request

Tony-LunarG:
Cherry picked to master

PR #1000, opened 2016-09-30
loader: gh971 Fix double free on physical devices

yavn:
I'm getting a crash on Windows 7 x64 when running `dEQP-VK.api.info.instance.physical_devices` test.

The pointer is not reset to NULL after a free, which makes it possible to take that branch again on the next call and free again.


tobine:
LGTM :+1:

MarkY-LunarG:
Committed: 0387853c4e919ff9ea53a62df378f91036dbfe9c

PR #1002, opened 2016-09-30
Stroyan symbol exports

stroyan:
This reduces the number of exported symbols in layers.
Only necessary symbols are exported by the layers on linux and android.
It also changes android layers to favor their own symbols for collisions with other sources.
It also gets the wrap_objects layer test working again with the new "Unexpected:" lines output.


cnorthrop:
Android changes LGTM

stroyan:
Cherry-picked to master.

Issue #52, opened 2016-02-29
Ability to install to a prefix using make/cmake


salamanderrake:
Could you set it up to where we could install the libraries(layers and etc...) and headers under a prefix defaulting to /usr/local so we could install a local copy of the files where we could point all the tools and other sdk stuff to, it would make it so much simpler for us. Also it would help with packaging systems that don't use CPack or deb files. As it is now CMAKE_INSTALL_PREFIX and make install are ignored.

karl-lunarg:
Committed with #1003.

PR #1003, opened 2016-09-30
build: gh52-CMake changes to enable make install on Linux

karl-lunarg:
See BUILD.md for details.

Change-Id: Ide1f635a57b57af384d9d1baac20c2256629f812


karl-lunarg:
Meant to address https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/issues/52

stroyan:
You could point out in the BUILD.md that cube and smoketest continue to use the local build libraries because of there rpath.  That could catch someone off guard.  It looks good apart from that.

karl-lunarg:
Good idea about mentioning rpath - done.  Committed with 63af15292d25b7bb24d89601469946226c087bd0.

PR #1004, opened 2016-10-01
Fix CMake modules not being added correctly

tambry:
Assume the following folder structure:
```
my_project
 - Vulkan-LoaderAndValidationLayers
 - cmake
 - src
 - CMakeLists.txt
```

Let's say I add Vulkan-LoaderAndValidationLayers as a sub-project that needs to be built in my CMakeLists.txt. In Vulkan-LoaderAndValidationLayers's CMakeLists.txt the `"${CMAKE_SOURCE_DIR}" == "my_project"`. This means that the cmake directory from my_project will be added as a CMake module path instead of the my_project/Vulkan-LoaderAndValidationLayers/cmake directory.

`"${CMAKE_CURRENT_SOURCE_DIR}" == "my_project/Vulkan-LoaderAndValidations"` if the CMakeLists.txt in Vulkan-LoaderAndValidationLayers is being executed. This ensures that the correct CMake modules directory gets always added and the correct modules are found.


courtney-g:
👍

karl-lunarg:
Fixed with cc2786d849bc767e140df472f5604785272791ae.

PR #1005, opened 2016-10-03
More assorted prep for merging swapchain bits to CV

chrisforbes:
There's a few things happening here. The first third or so are opportunistic fixes for bad Vulkan in the tests. The rest are untangling instance/device tracking in CV, in prep for tracking VkSurfaceKHR (instance level object). Some of this (phys device handling) was really bogus and would have failed horribly on systems with more than one.


chrisforbes:
Pushed to master.

PR #1007, opened 2016-10-03
1.0.29 Header Updates

mark-lunarg:


include/vulkan/vulkan.hpp
```diff
@@ -4114,35 +4114,6 @@ namespace vk
 
   struct DisplayPlanePropertiesKHR
   {
-    DisplayPlanePropertiesKHR( DisplayKHR currentDisplay_ = DisplayKHR(), uint32_t currentStackIndex_ = 0 )
```
courtney-g: I'm not certain where the vulkan.hpp comes from, but it seems suspicious that so much is changing in an otherwise minor header update.

lenny-lunarg: Fyi, vulkan.hpp is generated from https://github.com/KhronosGroup/Vulkan-Hpp. As for the changes, it looks like the new vk.xml added an option `returnedonly="true"` to both `VkDisplayModePropertiesKHR` and `VkSurfaceFormatKHR`. I believe what's happening is the hpp is just removing functionality that isn't needed if the object is returned only. So I think this looks fine, but I'd have to try using it to be sure.

mark-lunarg: There are (as is usual) large changes to vk.xml, including in this case removing the 'hand-written' validation stuff.  Some of what came out of vulkan.hpp was from these code sections.

mark-lunarg:
Pushed to Master.

PR #1008, opened 2016-10-03
cube: Demonstrate how to select other VkPresentModeKHR's

ianelliottus:
This comes from the Khronos-internal discussion of
    https://github.com/LunarG/VulkanSamples/issues/98.  Cube will, by default,
    choose VK_PRESENT_FIFO_KHR mode.  However, when that change was made, the old
    (example) code was lost that showed how to search for the other present modes.
    It was also discussed that there isn't good guidance on why an application
    might want to use a mode other than FIFO.  This change adds both of those, but
    only selects FIFO by default.  #ifdef's can be used to select other modes.
    With that code is a reason why an application might wish to use that present
    mode.

### DIFF
--- a/demos/cube.c
+++ b/demos/cube.c
@@ -925,10 +925,58 @@ static void demo_prepare_buffers(struct demo *demo) {
     }
 
     // The FIFO present mode is guaranteed by the spec to be supported
-    // and to have no tearing.  MAILBOX is another option, but for fast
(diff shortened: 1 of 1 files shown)


Tony-LunarG:
LGTM too, without the printf

ianelliottus:
Cherry-picked and pushed to master.

PR #1009, opened 2016-10-03
layers: Be careful about casts. ptr -> uint64_t& isnt right on 32bit.

chrisforbes:
Unbreaks the 32bit world. Sorry for the noise.


mark-lunarg:
std::ispointer -- my thing for today.  LGTM!

chrisforbes:
Pushed to master.

chrisforbes:
@mark-lunarg maybe we should use this to do good helpers to get rid of all the reinterpret_cast noise in the other layers?

mark-lunarg:
Yeah, that sounds great!  Maybe put the templates in layer utils or something.  Any ideas for the ultra-noisy log_msg calls?  Blick.

PR #1010, opened 2016-10-03
Fix about 80 or so warnings on Windows

mark-lunarg:


chrisforbes:
Right, I'd forgotten msvc was so cranky about conversion to book. LGTM.

demos/cube.c
```diff
@@ -3027,7 +3027,7 @@ static void demo_init_vk_swapchain(struct demo *demo) {
     if (presentQueueFamilyIndex == UINT32_MAX) {
         // If didn't find a queue that supports both graphics and present, then
         // find a separate present queue.
-        for (size_t i = 0; i < demo->queue_family_count; ++i) {
+        for (uint32_t i = 0; i < demo->queue_family_count; ++i) {
```
stroyan: This one doesn't really need the "size_t" or "uint32_t".  There is already a "uint32_t i;" at the top level of demo_init_vk_swapchain.

layers/core_validation.cpp
```diff
@@ -10194,7 +10194,7 @@ CmdBeginRenderPass(VkCommandBuffer commandBuffer, const VkRenderPassBeginInfo *p
         if (renderPass) {
             uint32_t clear_op_size = 0; // Make sure pClearValues is at least as large as last LOAD_OP_CLEAR
             cb_node->activeFramebuffer = pRenderPassBegin->framebuffer;
-            for (size_t i = 0; i < renderPass->createInfo.attachmentCount; ++i) {
+            for (uint32_t i = 0; i < renderPass->createInfo.attachmentCount; ++i) {
```
stroyan: While you are fixing warnings in core_validation.cpp, there are calls in initializeAndTrackMemory to malloc and memset that would like the size parameters to be size_t instead of uint64_t.  A "static_cast <size_t>(...)" would be good.

demos/cube.c
mark-lunarg: Sheesh Mike, you don't miss a thing.  Done.

layers/core_validation.cpp
mark-lunarg: I didn't see any warnings related to this in VS2015 or Linux but made the changes as requested.

mark-lunarg:
Updated and pushed to Master.

PR #1011, opened 2016-10-04
Fix object_tracker for 32bit properly

chrisforbes:
The previous approach doesn't actually work; not sure what I was thinking there...


cnorthrop:
Changes LGTM - passes on the farm now.

chrisforbes:
Picked to master.

PR #1012, opened 2016-10-04
layer: Add tracking of surface extensions (and surface objects) to CV

chrisforbes:
This is based on what Swapchain currently does.


layers/core_validation.cpp
```diff
@@ -104,6 +104,28 @@ struct instance_layer_data {
     VkLayerInstanceDispatchTable dispatch_table;
 
     unordered_map<VkPhysicalDevice, PHYSICAL_DEVICE_STATE> physical_device_map;
+    unordered_map<VkSurfaceKHR, SURFACE_STATE> surface_map;
+
+    bool surfaceExtensionEnabled = false;
+    bool displayExtensionEnabled = false;
+#ifdef VK_USE_PLATFORM_ANDROID_KHR
```
mark-lunarg: It's more consistent this way, but what about removing the ifdefs from these bools?  They're creeping into CV like cancer.

layers/core_validation.cpp
```diff
@@ -11755,6 +11904,81 @@ intercept_khr_swapchain_command(const char *name, VkDevice dev) {
...
+        bool instance_layer_data::*enable;
+    } khr_surface_commands[] = {
+#ifdef VK_USE_PLATFORM_ANDROID_KHR
+        {"vkCreateAndroidSurfaceKHR", reinterpret_cast<PFN_vkVoidFunction>(CreateAndroidSurfaceKHR),
+            &instance_layer_data::androidSurfaceExtensionEnabled},
+#endif // VK_USE_PLATFORM_ANDROID_KHR
+#ifdef VK_USE_PLATFORM_MIR_KHR
+        {"vkCreateMirSurfaceKHR", reinterpret_cast<PFN_vkVoidFunction>(CreateMirSurfaceKHR),
+            &instance_layer_data::mirSurfaceExtensionEnabled},
+    //    {"vkGetPhysicalDeviceMirPresentationSupportKHR",
```
mark-lunarg: Are these gonna get uncommented later on?

layers/core_validation.h
```diff
@@ -215,3 +215,11 @@ struct PHYSICAL_DEVICE_STATE {
     VkPhysicalDevice phys_device = VK_NULL_HANDLE;
     std::vector<VkQueueFamilyProperties> queue_family_properties;
 };
+
+struct SURFACE_STATE {
+    VkSurfaceKHR surface;
```
tobine: Do we want to struct init to VK_NULL_HANDLE here?

layers/core_validation.cpp
chrisforbes: Yes, we could have the bools always exist. The other blocks are annoying though -- various things they use only exist if VK_USE_* is defined.

layers/core_validation.cpp
chrisforbes: Yes.

chrisforbes: This is pretty messy, I'll just remove all the commented noise until the later commits that enable these functions.

layers/core_validation.cpp
```diff
@@ -11755,6 +11904,81 @@ intercept_khr_swapchain_command(const char *name, VkDevice dev) {
...
+    };
+
+    instance_layer_data *instance_data = nullptr;
+    if (instance) {
+        instance_data = get_my_data_ptr(get_dispatch_key(instance), instance_layer_data_map);
+    }
+
+    for (size_t i = 0; i < ARRAY_SIZE(khr_surface_commands); i++) {
+        if (!strcmp(khr_surface_commands[i].name, name)) {
+            if (instance_data && !(instance_data->*(khr_surface_commands[i].enable)))
```
tobine: This syntax confuses me. Is this pulling the dereferenced enable bit for that appropriate surface type?

layers/core_validation.cpp
```diff
@@ -11452,6 +11516,86 @@ GetPhysicalDeviceQueueFamilyProperties(VkPhysicalDevice physicalDevice, uint32_t
...
 
+template<typename TCreateInfo, typename FPtr>
+static VkResult CreateSurface(VkInstance instance, TCreateInfo const *pCreateInfo,
+                              VkAllocationCallbacks const *pAllocator, VkSurfaceKHR *pSurface,
+                              FPtr fptr)
+{
+    instance_layer_data *instance_data = get_my_data_ptr(get_dispatch_key(instance), instance_layer_data_map);
+
+    // Call down the call chain:
+    VkResult result = (instance_data->dispatch_table.*fptr)(instance, pCreateInfo, pAllocator, pSurface);
```
tobine: Is this dereferencing fptr to do a lookup into dispatch_table?

layers/core_validation.h
chrisforbes: Either that, or tweak things so this type doesn't have to be default-constructible.

layers/core_validation.cpp
chrisforbes: Exactly.

layers/core_validation.cpp
chrisforbes: Yes-- fptr isnt really a pointer at all; you can think of pointer-to-member as a properly-typed equivalent of C offsetof(struct, member) which is then applied at the point of use.

chrisforbes:
Done.

chrisforbes:
Pushed to master.

layers/core_validation.cpp
ianelliottus: @chrisforbes , this is using a static data structure for a per-instance function.  Is that going to work?  For example, if an application creates a VkInstance with no instance extensions enabled, and this function gets called, and then the application creates another VkInstance with several instance extensions enabled, and this function gets called, aren't the results for both instances going to be the same (i.e. based on the first-created instance)?

chrisforbes: We read some bool out of the instance_data here-- khr_surface_commands[i].enable just chooses which of them.

Using pointer-to-member here seems to be confusing everyone-- should I think about a different way to do it?

PR #1016, opened 2016-10-05
tests: Get run_wrap_objects_tests.sh working.

stroyan:
Fail from run_loader_tests.sh if run_wrap_objects_tests.sh fails.

Don't run tests in parallel to avoid hangs on some devices.

Filter out new "Unexpected:" lines from vk_layer_validation_tests.
Those lines include varying details that never pass a diff.


stroyan:
Cherry-picked to master.

PR #1019, opened 2016-10-05
layers: Initial codegen of unique error enums

### LINKED ISSUE #780: Identifying layer debug report messages reliably is currently impossible
At the moment there's no sane reliable way to identify a specific message or a narrow category of similar messages, for muting in the application's debug report callback if they are spammy and useless.

Calls to `log_msg` from the layers all use `__LINE__` as the `location` field, which makes it useless for identifying messages outside of a particular build of the layers. You could use it in a debugger session for conditional breakpoints, or add a temporary if, but it's not something you could commit.

I was using the `messageCode` field which is predefined with e.g. `DRAWSTATE_INVALID_BARRIER` or `DRAWSTATE_INVALID_BUFFER` in [core_validation_error_enums.h](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/core_validation_error_enums.h) and other headers. Note that these aren't immediately convenient to use, since core_validation_error_enums.h isn't shipped with the SDK, and other layers use enums defined within their layer headers which come with a number of other `#include` dependencies so they can't just be dropped into a project.

I found today though that items are removed and/or added in the middle of the list - changing all of the values and meaning that `messageCode` is not any more portable between layer versions than `location` is.

I'd recommend that `messageCode` should be treated the same as any ABI guarantee and those enum values should *never* change.

I've also recommended in the past eliminating use of `__LINE__` and replacing it with a immutable globally unique ID across all layers. This would be very simple to implement in any external scripting language, and since python is already required to build many of the layers it would be trivial to add as a pre-build step. An outline:

1. Create a file `layers/log_msg_id.txt` which contains a number that always increments.
2. Add `#define LAYER_LOG_MSG_ID(id) id` to `vk_layer_utils.h`
3. When a new log_msg is added, for the location field use `NEXT_LAYER_LOG_MSG_ID`.
4. Add a python script that runs pre-build step. It will look at every instance of `LAYER_LOG_MSG_ID(id)` to make sure they are unique (this will prevent merge conflicts creating duplicate IDs, just needs to be manually replaced with `NEXT_LAYER_LOG_MSG_ID` and re-run). Then for each instance of `NEXT_LAYER_LOG_MSG_ID` it replaces it with `LAYER_LOG_MSG_ID(contents of log_msg_id.txt)` and increments the txt file value.

You then immediately have unique IDs that never change no matter what, and conflicts are automatically detected if multiple people use the same ID allocation. This is much more useful than `__LINE__` because it's equally usable in a debugger or localised use, but it can also be committed or used in perpetuity.

tobine:
Fixes #780 
Replaces #856 

This is the initial commit of spec.py script that's used to generate a header
and database file that contain unique error enums per valid usage line in the
spec.

The basic usage is just "python spec.py" which will write out a header file to
"layers/vk_validation_error_messages.h" that contains unique error enum values
for each valid usage line in the current spec that's online at
https://www.khronos.org/registry/vulkan/specs/1.0/xhtml/vkspec.html

The header also contains validation_error_map with unique enum values as keys
that map to customized output messages for each validation error.

Updated one check to use the new model in core_validation.

The database file contains an enum value and error message on each line, as well
as fields indicating if the enum has been implemented in code and the testname
for a layer validation test that verifies that particular check.

When updating the header as the spec changes, the "-update" option to spec.py
will pull the database file and use it as a baseline reference to compare against
the updated spec. The script will attempt to keep errors with same message to have
the same error enum. When it's unsure if messages are updated or new, the script
will flag errors for manual review.


mark-lunarg:
+1.  <nit>Wish "VALIDATION_ERROR_*" was a bit shorter</nit>

karl-lunarg:
For the error database check_implemented field, can we define a value 'U' which means Unknown and then set all the entries to U?  When we go through the list, we would then change the entries the either Y or N.  This would avoid having multiple people re-checking the ones marked N to see if they were implemented or not.

U's would not be in the database for long.  We would quickly resolve them to Y or N.

karl-lunarg:
Would you consider moving spec.py to the layers directory?

I'm trying to think of ways to de-clutter the top-level dir in LVL and adding a file doesn't help.  I think that the scope/purpose of spec.py is restricted enough to layers to make putting it in layers reasonable.

I dunno, maybe all the *.py files in the TLD need to be in a directory of their own anyway.

mark-lunarg:
@karl-lunarg +1.  Let's make a scripts directory, or a tools directory or something.  We'd talked about moving all of the .py, .bat and .sh files to there anyhow.

karl-lunarg:
Looks pretty good.  If you can commit soon, I'll take a big chunk of them to look at.

Oh, when you put a 'Y' in check_implemented, does that mean that it uses the new error enums and we can't put a 'Y' in until it does?    (I think the answer is yes)  If so, then 'N' is a good starting value since all are 'N' and we may not need my suggestion of 'U'.

tobine:
I updated this commit to put spec.py file in layers directory. I updated default for check_implemented column to be 'U' and testname to be 'Unknown'. Those should be updated to 'N' and 'None' respectively for checks are not implemented and don't have tests.

I like keeping the 'U' as a way to flag which ones have been given any attention. 'N' will indicate that at least someone looked at it and most likely some amount of work has to be done to start using the new enum since the person didn't just start using it and update to 'Y'.

'Y' does indicate that the related check is implemented and using the new error enum & associated message string.

I'll sleep on this for tonight and plan to push early tomorrow if nothing comes up.

mark-lunarg:
If columns are cheap, maybe we should add one for the existence of a validation test covering that particular validation error.

tobine:
Fix was cherry-picked and pushed to master

PR #1021, opened 2016-10-06
Swapchain merge part3

chrisforbes:


layers/core_validation.cpp
```diff
@@ -11156,17 +11156,55 @@ CreateEvent(VkDevice device, const VkEventCreateInfo *pCreateInfo, const VkAlloc
...
+        auto swapchain_state = unique_ptr<SWAPCHAIN_NODE>(new SWAPCHAIN_NODE(pCreateInfo, *pSwapchain));
+        surface_state->swapchain = swapchain_state.get();
+        dev_data->device_extensions.swapchainMap[*pSwapchain] = std::move(swapchain_state);
+    }
+    else {
+        surface_state->swapchain = nullptr;
     }
 
+    // Spec requires that even if CreateSwapchainKHR fails, oldSwapchain behaves
+    // as replaced.
```
tobine: Did you clang-format this? Comment can all fit on one line

layers/vk_validation_layer_details.md
```diff
@@ -122,6 +122,8 @@ The Draw State portion of the core validation layer tracks state leading into Dr
 | Proper synchronization of acquired images | vkAcquireNextImageKHR should be called with a valid semaphore and/or fence | SWAPCHAIN_NO_SYNC_FOR_ACQUIRE | vkAcquireNextImageKHR | TODO | None |
 | Swapchain image index too large | Validates that an image index is within the number of images in a swapchain | SWAPCHAIN_INVALID_IMAGE | vkQueuePresentKHR | VkWsiEnabledLayerTest.TestEnabledWsi | None |
 | Can't present a non-owned image | Validates that application only presents images that it owns | SWAPCHAIN_IMAGE_NOT_ACQUIRED | vkQueuePresentKHR | TODO | None |
+| Surface already has swapchain | Cannot create a swapchain for a surface that already has one [other than the swapchain being replaced] | SWAPCHAIN_ALREADY_EXISTS | vkCreateSwapchainKHR | TODO | None |
```
tobine: Does TestEnabledWsi test hit either of these cases?

chrisforbes: No- TestEnabledWsi hits very little. We need some real tests.

layers/core_validation.cpp
chrisforbes: clang-format leaves this as-is (and /also/ leaves the single-line version as-is)

chrisforbes:
Updated for formatting (including comment unsplit, which clang-format has no opinion on)

chrisforbes:
Pushed to master.

layers/core_validation.cpp
```diff
@@ -11156,17 +11156,50 @@ CreateEvent(VkDevice device, const VkEventCreateInfo *pCreateInfo, const VkAlloc
...
+    }
+
+    return false;
+}
+
 VKAPI_ATTR VkResult VKAPI_CALL CreateSwapchainKHR(VkDevice device, const VkSwapchainCreateInfoKHR *pCreateInfo,
                                                   const VkAllocationCallbacks *pAllocator,
                                                   VkSwapchainKHR *pSwapchain) {
     layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
+    auto surface_state = getSurfaceState(dev_data->instance_data, pCreateInfo->surface);
```
ianelliottus: You are not checking whether surface_state is NULL, but then PreCallValidateCreateSwapchainKHR() is dereferencing it.  This can cause a crash.

layers/core_validation.cpp
```diff
@@ -11156,17 +11156,50 @@ CreateEvent(VkDevice device, const VkEventCreateInfo *pCreateInfo, const VkAlloc
...
+    auto old_swapchain_state = getSwapchainNode(dev_data, pCreateInfo->oldSwapchain);
+
+    if (PreCallValidateCreateSwapchainKHR(dev_data, pCreateInfo, surface_state, old_swapchain_state))
+        return VK_ERROR_VALIDATION_FAILED_EXT;
+
```
ianelliottus: This is not using the "skip_call" model, but is just returning early if there was any error from PrecallValidateCreateSwapchainKHR().  How come?

tobine: The "PreCallValidate*" model is the preferred model that we'd like to migrate all code to. If you search core_validation for "PreCallValidate" you'll see a few similar examples. The idea is to minimize code in the top-level intercept function and then have any "skip_call" flags contained in PreCallValidate* function so that's the only function API entry points need to call and then either exit early or continue making a call down the chain.

layers/core_validation.cpp
```diff
@@ -11156,17 +11156,50 @@ CreateEvent(VkDevice device, const VkEventCreateInfo *pCreateInfo, const VkAlloc
...
 }
 
+static bool PreCallValidateCreateSwapchainKHR(layer_data *dev_data, VkSwapchainCreateInfoKHR const *pCreateInfo,
+                                              SURFACE_STATE *surface_state, SWAPCHAIN_NODE *old_swapchain_state) {
+    auto most_recent_swapchain = surface_state->swapchain ? surface_state->swapchain : surface_state->old_swapchain;
+
+    if (most_recent_swapchain != old_swapchain_state || (surface_state->old_swapchain && surface_state->swapchain)) {
+        if (log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_EXT,
+                    reinterpret_cast<uint64_t>(dev_data->device), __LINE__, DRAWSTATE_SWAPCHAIN_ALREADY_EXISTS, "DS",
+                    "vkCreateSwapchainKHR(): surface has an existing swapchain other than oldSwapchain"))
```
ianelliottus: I like the check/error, but (FYI) the relevant valid usage statement is going away (see spec issue 533, and public spec issue 387).

layers/swapchain.cpp
```diff
@@ -1941,12 +1918,6 @@ VKAPI_ATTR void VKAPI_CALL DestroySwapchainKHR(VkDevice device, VkSwapchainKHR s
         // Delete the SwpSwapchain associated with this swapchain:
         if (pSwapchain->pDevice) {
             pSwapchain->pDevice->swapchains.erase(swapchain);
-            if (device != pSwapchain->pDevice->device) {
-                skip_call |=
-                    log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_EXT,
-                            reinterpret_cast<uint64_t>(device), __LINE__, SWAPCHAIN_DESTROY_SWAP_DIFF_DEVICE, swapchain_layer_name,
-                            "vkDestroySwapchainKHR() called with a different VkDevice than the VkSwapchainKHR was created with.");
```
ianelliottus: Ditto

layers/core_validation.cpp
ianelliottus: I guess my point is that this is not doing proper "skip_call", as I understand it; but is assuming that skip_call should automatically be set, and that early exit is always desired.

Other checks will also be added to this "PreCallValidate" function (as Chris continues to migrate functionality from the swapchain layer), and early-exit isn't automatically called for.

tobine: PreCallValidateCreateSwapchainKHR() will only return true if one of the callbacks returns true. This is consistent with previously skip_call behavior. New checks will be managed in PreCallValidate, as you said, but any time PreCallValidate returns true, then early exit is called for in CreateSwapchainKHR() and the call down the chain should not be made.

tobine: @ianelliottus I don't understand where you're drawing your conclusion from so perhaps you're misunderstanding what I've said or the code.
Please look at PreCallValidateCreateSwapchainKHR(). It only returns true when log_msg() returns true, which means the app has requested a bailout. When this occurs, we skip the call down the chain in CreateSwapchainKHR(). This is consistent with all other skip_call behavior throughout the code. I don't see any change from the norm here or going forward. The PreCallValidate* pattern just shields the top-level API intercept from a bunch of skip_call setting and checking noise.

layers/swapchain.cpp
chrisforbes: object_tracker already picks this up. To get here in that case, you have to have continued past the error from object_tracker not finding the object for the swapchain handle.

layers/core_validation.cpp
ianelliottus: You're correct--I was misunderstanding the code, which behaves as you said.  Sorry!

PR #1029, opened 2016-10-06
layers: Updating some CreateBuffer checks to use unique enums

karl-lunarg:
Small batch of updates to use unique validation error enums and
messages for a few checks.  Updated the database file and
flagged some other checks that are missing with the error
enums and TODO notes.

Change-Id: I2020a88bbdcf7ea6d725f667523e89b160fe7ab8


karl-lunarg:
A big part of this is the codegen'd code in parameter_validation_vkCreateBuffer() and figuring out how to come up with the error enums and messages there.  parameter_validation_vkCreateBuffer() has several calls to helpers that call log_msg().  Each of these helpers can be instrumented to take extra error enum and error_msg parms.  But I wonder if anyone has some ideas how to get codegen to add these to each call to the helpers in parameter_validation_vkCreateBuffer().

karl-lunarg:
@tobine Any thoughts on these (just updated) changes?

With the change to ValidateObject() in object_tracker, I'm thinking that my next step might be to add the validation error code to all calls to ValidateObject() in this file, which will knock quite a few off the list pretty easily.

(We can defer my earlier comment about the codegen param validation)

tobine:
LGTM :+1:

PR #1032, opened 2016-10-08
cube.c: Remove excess tests from presentation fences

szdarkhack:
Instead of using a bool array and both checking and writing to it every frame for no reason, just create the presentation fences with the VK_FENCE_CREATE_SIGNALED_BIT flag.


Tony-LunarG:
Cherry picked to master

PR #1036, opened 2016-10-10
Tobin descriptor error code passing

tobine:
Updated descriptor update validation checks to use unique error enums and added validation flag for vkUpdateDescriptorSets().
Needed to update some functions to pass the error enum around as a parameter.

Also updated database file for 24 enums that were updated.

Changed some "error" var names to "error_msg" to clearly distinguish it from "error_code" that is now passed around as well.


lentinem:
:+1:

tobine:

PR #1038, opened 2016-10-11
Tobin gh1034 storage image layout check

### LINKED ISSUE #1034: Layers: warn if STORAGE_IMAGE descriptor is updated with layout other than GENERAL
Seems (as of today) there is a bug in dEQP-VK.compute.basic.copy_image_to_ssbo vulkanCTS tests not being caught by Validation Layers: 

`vkUpdateDescriptorSets` is called with `VkWriteDescriptorSet` slot, where 
```
descriptorType = VK_DESCRIPTOR_TYPE_STORAGE_IMAGE (3)
pImageInfo->imageLayout = VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL (5)
```

This is invalid and layers should emit an ERROR in such case. Spec says:
```
Load and store operations on storage images can only be done on images in VK_IMAGE_LAYOUT_GENERAL layout.
```

tobine:
Fixes #1034 

Adds new check to make sure STORAGE_IMAGE descriptors are only updated with GENERAL image layout along with associated test.

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -12985,6 +12985,105 @@ TEST_F(VkLayerTest, InvalidImageLayout) {
     vkDestroyImage(m_device->device(), dst_image, NULL);
 }
 
+TEST_F(VkLayerTest, InvalidStorageImageLayout) {
+    TEST_DESCRIPTION("Attempt to update a STORAGE_IMAGE descriptor w/o GENERAL layout.");
+    VkResult err;
+
+    ASSERT_NO_FATAL_FAILURE(InitState());
+
+    const VkFormat tex_format = VK_FORMAT_R8G8B8A8_UNORM;
+    VkImageTiling tiling;
+    VkFormatProperties format_properties;
+    vkGetPhysicalDeviceFormatProperties(gpu(), tex_format, &format_properties);
+    if (format_properties.linearTilingFeatures & VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT) {
+        tiling = VK_IMAGE_TILING_LINEAR;
+    } else if (format_properties.optimalTilingFeatures & VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT) {
+        tiling = VK_IMAGE_TILING_OPTIMAL;
+    } else {
+        printf("Device does not support VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT; "
+               "skipped.\n");
+        return;
+    }
+
+    VkDescriptorPoolSize ds_type = {};
+    ds_type.type = VK_DESCRIPTOR_TYPE_STORAGE_IMAGE;
+    ds_type.descriptorCount = 1;
+
+    VkDescriptorPoolCreateInfo ds_pool_ci = {};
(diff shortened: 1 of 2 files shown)


lentinem:
:+1:

tobine:
@mark-lunarg If those commit messages are wrong, I don't wanna be right...

I don't recall which draconian regulations are limiting our commit message line lengths to less than postage stamp size, but I propose a rebellion.

Seriously, though, if 7-word 63-char initial lines and 75 char description lines are breaking some system, then we should fix/update/replace that system.

mark-lunarg:
@tobine, we went through this (including a rebellion!) months ago -- I'm kinda fuzzy, but it's basically standard open-source procedure.  E.g., see [The Adroid guidelines,](https://source.android.com/source/submit-patches.html) [among](https://wiki.openstack.org/wiki/GitCommitMessages) [many](https://git-scm.com/book/ch5-2.html) [others.](http://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting)

tobine:
@mark-lunarg Alright, so people may look down on me for my insanely long 63-char initial line, but nothing will actually break? I can live with that. I'll do my best to be more char-frugal on initial lines and quicker to carriage return on description lines in the future. Since I have been losing 8 chars of every first line for "layers: " I plan to now use "v:" as my prefix for validation layer changes. The "v" stands for "Victory!"

tests/layer_validation_tests.cpp
```diff
@@ -8907,6 +8907,7 @@ TEST_F(VkLayerTest, DescriptorSetCompatibility) {
     ds_pool_ci.maxSets = 5;
     ds_pool_ci.poolSizeCount = NUM_DESCRIPTOR_TYPES;
     ds_pool_ci.pPoolSizes = ds_type_count;
+    ds_pool_ci.flags = VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT;
```
chrisforbes: I don't see any freeing in the middle of the test happening, an alternative would be to just remove the FreeDescriptorSets calls in the cleanup, and let the pool destruction take care of it?

Happy either way.

tests/layer_validation_tests.cpp
```diff
@@ -12988,6 +12988,153 @@ TEST_F(VkLayerTest, InvalidImageLayout) {
...
+    VkImage image;
+    const int32_t tex_width = 32;
+    const int32_t tex_height = 32;
+    VkImageCreateInfo image_create_info = {};
+    image_create_info.sType = VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO;
+    image_create_info.pNext = NULL;
+    image_create_info.imageType = VK_IMAGE_TYPE_2D;
+    image_create_info.format = tex_format;
+    image_create_info.extent.width = tex_width;
+    image_create_info.extent.height = tex_height;
```
chrisforbes: Use VkImageObj (and its dubiously-named targetView member function) to shorten this by about a page?

mark-lunarg:
@tobine, nothing should break, but there are Linux tools that work best when the 50/72 thingy is observed.  I *hope* you're joking about the 'v:' (please?).  The first line is supposed to be more of a title than a description which is what the 72-char lines are intended for which can be a bit of an adjustment.  We want the logs to look as consistent and professional as the source does, and both are improving rapidly!

Issue #1033, opened 2016-10-08
Cube demo WSI barriers


szdarkhack:
According to my understanding of the synchronization operations, the image barriers for the acquisition and release of the swapchain images by the application could be improved.

For reference, the issues discussed appear in the function demo_draw_build_cmd().

For starters, and this is not even a sync issue, the acquisition barrier specifies the oldLayout as PRESENT_SRC, while it does not care for preserving the contents (as mentioned in a comment in a subsequent barrier, and evidenced by the lack of ownership transfer). It should, therefore, be set as UNDEFINED, so that the implementation skip the conversion.

Now on to sync. The spec contains the following text (section 6.2 Semaphores):

> This semaphore signal operation defines the first half of a memory dependency, guaranteeing that all memory accesses defined by the submitted queue operations in the batch are made available, and that those queue operations have completed execution.

> Semaphore signal operations for vkQueueSubmit additionally include all queue operations previously submitted via vkQueueSubmit in their half of a memory dependency, and all batches that are stored at a lower index in the same pSubmits array.

> Signaling of semaphores can be waited on by similarly including them in a batch, defining a queue operation to wait for a signal. A semaphore wait operation defines the second half of a memory dependency for the semaphores being waited on. This half of the memory dependency guarantees that the first half has completed execution, and also guarantees that all available memory accesses are made visible to the queue operations in the batch.

> Semaphore wait operations for vkQueueSubmit additionally include all queue operations subsequently submitted via vkQueueSubmit in their half of a memory dependency, and all batches that are stored at a higher index in the same pSubmits array.

That, to me, indicates that the acquisition and release semaphores take care of the memory sync themselves. It seems, therefore, that it would be unnecessary to include anything in srcAccessMask and dstAccessMask in either barrier (of course the barriers themselves are still required for layout transitions and ownership transfer). Of course, it is unclear whether WSI operations count as "queue operations submitted via vkQueueSubmit", so including MEMORY_READ in the dstAccessMask of the present barrier might be necessary (see below). This really needs some clarification in the spec.

Additionally, this (section 6.5.2 Pipeline Stage Flags):

> The VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT is useful for accomplishing memory barriers and layout transitions when the next accesses will be done in a different queue or by a presentation engine; in these cases subsequent commands in the same queue do not need to wait, but the barrier or transition must complete before semaphores associated with the batch signal.

[...]

PR #1041, opened 2016-10-11
Cube barrier changes from GH #1033

Tony-LunarG:
I left in the ACCESS_MEMORY_READ bit but can easily remove it if it's decided it's wrong.  I do get validation warnings specifying a srcAccessMask with LAYOUT_UNDEFINED.


szdarkhack:
The acquire barrier should not need any access masks at all, since it does not need to create a memory dependency anyway (the contents of the image need not be preserved, hence the UNDEFINED layout). I'm still unclear about the presentation barrier though.

Tony-LunarG:
So we don't need/want to transition the image's layout to COLOR_ATTACHMENT_OPTIMAL?

szdarkhack:
The renderpass can do the layout transitions (by setting initial layout UNDEFINED, first subpass layout COLOR_ATTACHMENT_OPTIMAL, final layout PRESENT_SRC) as per the spec in 7.1, page 126:

> The implementation will automatically perform layout transitions as needed between subpasses to make each subpass use the requested layouts.

Since we ONLY need to do layout transition after the image is acquired (i.e. no memory dependency and no ownership transfer), the renderpass can take care of that by itself and should require no extra barrier. Do note that this is not the case for presentation, as we (potentially) need to pass the image over to the presentation queue, which can only be done via a pipeline barrier.

So, to summarize, the acquire barrier should be dropped and the initial layout for both attachments in the renderpass description should be switched to UNDEFINED since we don't care about the initial contents.

Tony-LunarG:
Ok - changes pushed.

szdarkhack:
Looks good to me. Since this is supposed to be sample code, perhaps it would be beneficial to add a comment or two explaining why there is no barrier and why the initial layout is UNDEFINED, to serve as a better tutorial.

Tony-LunarG:
Cherry picked to master

PR #1043, opened 2016-10-11
tests: Remove document validation test

cnorthrop:
This test requires files outside of the build directory, which
doesn't work when distributing build results.  It will soon be
superseded by a new test for valid enums.  It is still covered
by Linux run_all_tests.sh


tobine:

PR #1044, opened 2016-10-12
layers: Add error enums to object tracker

karl-lunarg:
- Add error enum to many calls to ValidateObject().
- Specify null_allowed=true when VU allows it.
  (mostly Destroy* cases)
- Fix error handler to work properly on positive tests
  in layer validation tests.

Change-Id: Ie3ec19655fab1b31c37a4089cb1c4c83a2194ef5


tests/layer_validation_tests.cpp
```diff
@@ -149,7 +149,12 @@ class ErrorMonitor {
         string errorString(msgString);
         bool found_expected = false;
         for (auto desired_msg : m_desiredMsgs) {
-            if (errorString.find(desired_msg) != string::npos) {
+            if (desired_msg.length() == 0) {
```
tobine: This change should be in a separate CL

tobine:
Great work overall, just a couple things I'd like you to check out in review comments.

layers/object_tracker.cpp
```diff
@@ -363,7 +363,7 @@ VKAPI_ATTR void VKAPI_CALL DestroyInstance(VkInstance instance, const VkAllocati
         }
     }
 
-    ValidateObject(instance, instance, VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, false);
+    ValidateObject(instance, instance, VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, false, VALIDATION_ERROR_00021);
```
karl-lunarg: My search used VK_NULL_HANDLE instead of NULL.  One could argue that the spec needs changing here.  Fixed.

layers/object_tracker.cpp
```diff
@@ -434,7 +434,7 @@ VKAPI_ATTR void VKAPI_CALL DestroyInstance(VkInstance instance, const VkAllocati
 VKAPI_ATTR void VKAPI_CALL DestroyDevice(VkDevice device, const VkAllocationCallbacks *pAllocator) {
 
     std::unique_lock<std::mutex> lock(global_lock);
-    ValidateObject(device, device, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_EXT, false);
+    ValidateObject(device, device, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_EXT, false, VALIDATION_ERROR_00052);
```
karl-lunarg: My search used VK_NULL_HANDLE instead of NULL.  One could argue that the spec needs changing here.  Fixed.

tests/layer_validation_tests.cpp
karl-lunarg: Will do.  I was on the fence about this, but this change was needed to keep the tests from crashing with the other changes applied.  (I think I added a missing ValidateObject call in FreeMemory that caused multiple errors in positive tests.)  I like to avoid "broken" commits.

karl-lunarg:
@tobine, If it helps any, I semi-mechanically did some verification.  I extracted the error enums for all *changed* lines in the database file.  I also extracted all the error enums that I added to object_tracker, and then compared the two.  I found a couple of errors that I fixed, but otherwise was able to reconcile everything.  So, I'm pretty certain that the db changes match the code.   The thing that could go wrong is simply using the wrong error enum in any of the calls, which is possible.  Of course we don't want these, but the worst case is that they point to the wrong place in the spec.

layers/object_tracker.cpp
tobine: I suspect that since dispatchable objects are ptrs, they intentionally use "NULL" instead of "VK_NULL_HANDLE" in those cases. I see that the FreeCommandBuffers language also has "NULL." I suspect you're still working your way through object_tracker, but keep that in mind when you get to ValidateCommandBuffer() call in FreeCommandBuffers().

karl-lunarg: OK.  Yes, I intend to make another pass through OT to spend more time on resolving the unresolved.

karl-lunarg:
pushed to master

PR #1047, opened 2016-10-12
Click vk.xml code gen structre/files with 1.0.30 VulkanDocs repo

mark-lunarg:
LVL had diverged from the codegen used in VulkanDocs and we have not been picking up updates (except to vk.xml).  Old generator.py classes are now each in their own files.  Updated the LVL files (reg.py, generator.py, genvk.py, etc) to the 1.0.30 verisons, modified our vkxml codegen for the new files and split the generators into their own files.  

The python scripts will be moving into a subdirectory in the near future.


mark-lunarg:
Ran clean on our farm.

cnorthrop:
Clean here too, other than things that have already been fixed on master by 62fb8720548cc9270398460ce0d2b6a11ac446e7

We don't automatically build/test smoke yet (or cube) but hoping to change that soon.

lvl_genvk.py
```diff
@@ -0,0 +1,380 @@
...
+        '*/',
+        ''
+    ]
+
+    # Defaults for generating re-inclusion protection wrappers (or not)
+    protectFile = protect
+    protectFeature = protect
+    protectProto = protect
+
+# Comment out for LVL
```
mark-lunarg: Thanks @stroyan, I missed the commented lines and will remove them, and will refactor/squash the commits as you suggest.

mark-lunarg:
Pushed to Master.

PR #1049, opened 2016-10-12
Tobin in use destroy unification

tobine:
Adding in-use checks led to various other clean-up as well including:

1. Rename various "node" structs to "state"
2. Add tests for a couple missing in-use cases
3. Add validation flags to various Destroy functions
4. Refactor Destroy functions to honor Pre/Post pattern
5. Update errors to use unique error enums

Still more of this work to do, but this seemed like plenty for one PR


layers/core_validation.cpp
```diff
@@ -2341,7 +2341,7 @@ cvdescriptorset::DescriptorSetLayout const *getDescriptorSetLayout(layer_data co
     return it->second;
 }
 
-static PIPELINE_LAYOUT_NODE const *getPipelineLayout(layer_data const *my_data, VkPipelineLayout pipeLayout) {
+static PIPELINE_LAYOUT_NODE const *getPipelineStateLayout(layer_data const *my_data, VkPipelineLayout pipeLayout) {
```
chrisforbes: No, this isn't right.. overzealous find&replace.

tobine:

PR #1050, opened 2016-10-13
layers: Suppress warning generation after app acknowledges an error

chrisforbes:
The app already knows about the problem-- it has requested that the call
be skipped. Producing another warning afterward is just noise.


chrisforbes:
Pushed to master

PR #1061, opened 2016-10-17
tests: Add CmdBlitImage format tests

stroyan:
Test two cases of mismatch of signed or unsigned int formats
in CmdBlitImage calls.


mark-lunarg:
Hey Mike, as we go along we've been adding these tests to the ValError database and updating the log messages with the new ID as we go.  If you want you can assign it to me and I'll add that part.  Thanks for checking out the bug, BTW.

stroyan:
Cherry-picked to master.

PR #1062, opened 2016-10-17
Tobin gh1053 empty descriptors

### LINKED ISSUE #1053: vkUpdateDescriptorSets validation fails when using sparse bindings
I'm seeing a (possibly) incorrect validation error being triggered when writing to a descriptor set which has sparse bindings:

> "vkUpdateDescriptorsSets() failed write update validation for Descriptor Set 0x39 with error: Attempting write update to descriptor set 0000000000000039 binding #2 with 2 total descriptors but update of 1 descriptors starting at binding offset of 2 combined with update array element offset of 0 oversteps the size of this descriptor set."

My descriptor set has two used bindings, one dynamicUbo at binding 0 and one dynamicUbo at binding 2. Binding 1 is marked as an unused binding (descriptorCount==0)

When creating the VkDescriptorSetLayout, binding 1 was defined as an unused binding. I used a VkDescriptorSetLayoutbinding with descriptorCount == 0, following the spec recommendation:

> descriptorCount is the number of descriptors contained in the binding, accessed in a shader as an array. If descriptorCount is zero this binding entry is reserved and the resource must not be accessed from any stage via this binding within any pipeline using the set layout. (section 13.2.1, following "Valid Usage")

When running *without* validation layers, all appears fine and jolly, and when I inspect the frame using NSight, it looks as i would expect it to do.

Specifically [this line](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/descriptor_sets.cpp#L1138) in the validation layers may be a source for this phenomenon:  

    if ((start_idx + update->descriptorCount) > p_layout_->GetTotalDescriptorCount()) {`

With `start_idx` being calculated a few lines earlier as: 

    auto start_idx = 
    p_layout_->GetGlobalStartIndexFromBinding(update->dstBinding) 
    + update->dstArrayElement;

and `dstBinding`, in my use-case, being `2`.

Here, the total number of descriptors (`2`) is compared against the current binding (`2`) + arrayIndexOffset (`0`) + descriptorCount (`1`) being written, which gives a sum total of (`3`), which fails the check. 

If the count was calculated by summing all descriptors*descriptorCount, the sum total would omit an empty, "reserved" binding, which could explain the issue, but unfortunately I don't know enough about the internal workings of the validation layers to feel confident about this.

tobine:
Fixes #1053 

Fix a bug related to empty descriptor bindings potentially causing errors when updating descriptors that fall after the empty bindings.

### DIFF
--- a/layers/descriptor_sets.cpp
+++ b/layers/descriptor_sets.cpp
@@ -40,7 +40,7 @@ cvdescriptorset::DescriptorSetLayout::DescriptorSetLayout(debug_report_data *rep
         binding_to_global_start_index_map_[p_create_info->pBindings[i].binding] = global_index;
         global_index += p_create_info->pBindings[i].descriptorCount ? p_create_info->pBindings[i].descriptorCount - 1 : 0;
         binding_to_global_end_index_map_[p_create_info->pBindings[i].binding] = global_index;
-        global_index++;
+        global_index += p_create_info->pBindings[i].descriptorCount ? 1 : 0;
         bindings_.push_back(safe_VkDescriptorSetLayoutBinding(&p_create_info->pBindings[i]));
         // In cases where we should ignore pImmutableSamplers make sure it's NULL
         if ((p_create_info->pBindings[i].pImmutableSamplers) &&
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -1142,7 +1142,7 @@ TEST_F(VkLayerTest, FillBufferAlignment) {
 
 // This is a positive test. No failures are expected.
 TEST_F(VkLayerTest, IgnoreUnrelatedDescriptor) {
-    TEST_DESCRIPTION("Ensure that the vkUpdateDescriptorSet validation code "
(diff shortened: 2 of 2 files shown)


layers/descriptor_sets.cpp
```diff
@@ -40,7 +40,7 @@ cvdescriptorset::DescriptorSetLayout::DescriptorSetLayout(debug_report_data *rep
         binding_to_global_start_index_map_[p_create_info->pBindings[i].binding] = global_index;
         global_index += p_create_info->pBindings[i].descriptorCount ? p_create_info->pBindings[i].descriptorCount - 1 : 0;
         binding_to_global_end_index_map_[p_create_info->pBindings[i].binding] = global_index;
-        global_index++;
+        global_index += p_create_info->pBindings[i].descriptorCount ? 1 : 0;
```
chrisforbes: This is still more tangled up than it needs to be, but fixes the bug.

mark-lunarg:
:+1: I'm guessing positive LVTs won't need to go in the ValError database.

tobine:

PR #1063, opened 2016-10-17
Tobin gh1037 sparse as separate bindings

tobine:
Adds very limited initial understanding of sparse memory bindings to core_validation.

Ultimately correct support for sparse memory binding may require an overhaul of memory tracking, but this is enough to allow some current crashes on CTS tests to pass.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -403,14 +403,14 @@ static VkDeviceMemory *GetObjectMemBinding(layer_data *my_data, uint64_t handle,
         auto image_state = getImageState(my_data, VkImage(handle));
         *sparse = image_state->createInfo.flags & VK_IMAGE_CREATE_SPARSE_BINDING_BIT;
         if (image_state)
-            return &image_state->mem;
+            return &image_state->binding.mem;
         break;
     }
     case VK_DEBUG_REPORT_OBJECT_TYPE_BUFFER_EXT: {
         auto buff_node = getBufferNode(my_data, VkBuffer(handle));
         *sparse = buff_node->createInfo.flags & VK_BUFFER_CREATE_SPARSE_BINDING_BIT;
         if (buff_node)
-            return &buff_node->mem;
+            return &buff_node->binding.mem;
         break;
     }
     default:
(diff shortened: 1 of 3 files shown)


Tony-LunarG:
I'm not able to build this branch on Windows - complaints of undefined external FENCE_NODE::FENCE_NODE building core_validation.  We've tried 2 different Windows systems. I deleted my build directory, did a git clean -fdx and still can't build.

tobine:
Yep, I see build error as well. Just force pushed update.

tests/layer_validation_tests.cpp
```diff
@@ -3331,6 +3331,94 @@ TEST_F(VkLayerTest, PipelineNotBound) {
...
+    VkSparseMemoryBind binds[2];
+    binds[0].flags = 0;
+    binds[0].memory = memory_one;
+    binds[0].memoryOffset = 0;
+    binds[0].resourceOffset = 0;
+    binds[0].size = memory_info.allocationSize;
+    binds[1].flags = 0;
+    binds[1].memory = memory_two;
+    binds[1].memoryOffset = 0;
+    binds[1].resourceOffset = 0;
```
karl-lunarg: I talked with Tony about this, but I don't think that this can be zero, since it would cause the two allocations to be mapped on top of each other in the resource at offset zero.  I think that `memory_reqs.size / 2 ` or `memory_info.allocationSize` would likely work, even considering alignment constraints.

tests/layer_validation_tests.cpp
```diff
@@ -3331,6 +3331,94 @@ TEST_F(VkLayerTest, PipelineNotBound) {
     vkDestroyDescriptorPool(m_device->device(), ds_pool, NULL);
 }
 
+// This is a positive test. No failures are expected.
+TEST_F(VkLayerTest, BindSparse) {
+    TEST_DESCRIPTION("Bind 2 memory ranges to one image, destroy the image"
```
karl-lunarg: Maybe get the word "sparse" in the description?

Tony-LunarG: As we discussed, I want to get the alignment right rather than hope that the memory can be split in half.  Can we divorce the test from the rest of the code so I don't hold thing up?

Tony-LunarG:
It looks like this also needs the fix to AddCommandBufferBindingBuffer to put the buffer in the command buffer's object_bindings

tests/layer_validation_tests.cpp
```diff
@@ -3331,6 +3331,94 @@ TEST_F(VkLayerTest, PipelineNotBound) {
...
+    binds[1].size = memory_info.allocationSize;
+
+    VkSparseImageOpaqueMemoryBindInfo opaqueBindInfo;
+    opaqueBindInfo.image = image;
+    opaqueBindInfo.bindCount = 2;
+    opaqueBindInfo.pBinds = binds;
+
+    VkFence fence = VK_NULL_HANDLE;
+
+    VkBindSparseInfo bindSparseInfo;
```
chrisforbes: In the interests of readability, perhaps `= {}`, and then just set the members that are interesting?

tests/layer_validation_tests.cpp
chrisforbes: This needs some care. A 64x64 32bpp image is 16KiB of texels. At least one vendor (nvidia) has a sparse binding granularity of 64KiB [both memory_reqs.size and memory_reqs.alignment are 64KiB for this resource], so there is no way of chopping an image this small into two bindings.

tobine:
Added one additional fix and pulled out the test which @Tony-LunarG will fix up independently. I can either push these changes independently, which will fix some issues when running CTS with validation, or Tony can rebase them onto his test when it's fixed.

Tony-LunarG:
I'd say go ahead and push your changes and I'll push the test later today.

tobine:

PR #1066, opened 2016-10-18
Update to V1.0.31 header

mark-lunarg:
Wish they were all like this.

### DIFF
--- a/include/vulkan/vulkan.h
+++ b/include/vulkan/vulkan.h
@@ -43,7 +43,7 @@ extern "C" {
 #define VK_VERSION_MINOR(version) (((uint32_t)(version) >> 12) & 0x3ff)
 #define VK_VERSION_PATCH(version) ((uint32_t)(version) & 0xfff)
 // Version of this file
-#define VK_HEADER_VERSION 30
+#define VK_HEADER_VERSION 31
 
 
 #define VK_NULL_HANDLE 0
--- a/include/vulkan/vulkan.hpp
+++ b/include/vulkan/vulkan.hpp
@@ -40,7 +40,7 @@
 # include <vector>
 #endif /*VULKAN_HPP_DISABLE_ENHANCED_MODE*/
 
-static_assert( VK_HEADER_VERSION ==  30 , "Wrong VK_HEADER_VERSION!" );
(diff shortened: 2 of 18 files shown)


courtney-g:
Starting to make me wonder about how to simplify such updates.
:+1:

mark-lunarg:
Pushed to Master.

karl-lunarg:
@courtney-g   Stop updating the spec as often? :-)

TobiasHector:
@karl-lunarg I previously suggested we didn't need to update the header and xml along with the spec if the only changes were in the spec text, which would remove the need to do this kind of mostly pointless update to the validation layers... At the time nobody thought it would be an issue, but if it's something that's causing you needless busy work, I suggest you raise an issue against the spec :).

karl-lunarg:
Hi @TobiasHector ,  Versions 30 and 31 didn't update the xml or headers, so we could have skipped these.  We've done that before.  But we will pick up the latest, even when there are no changes, just before we release an SDK in order to get the latest in the SDK.  30 was a good example of that.

I think that it is probably globally better if the version of the public header matches the version of the public spec in order to avoid questions about that, if nothing else.  I know that it causes some unneeded version churn of the xml/header file, but we can always pass on pulling in a version that had no changes.

@courtney-g may been suggesting to make the changes simpler.  It could be done in one commit, for example, and probably should be so we don't have version inconsistency across commits.  @mark-lunarg: can we do that in the future?   There are probably ways to automate making the changes to make things easier on us, and we can do that in places where it makes sense.

Also, in the past, we haven't always done pull requests for updates of the version, especially when the changes were only in the version.   I think that pull requests for version updates that add significant features may be valuable so that others can see the additions and what code has been done to cope with the new features.  But perhaps we can refrain from doing pull requests when only the version is different.

mark-lunarg:
Just a side note -- LoaderAndValidationLayers now has dependencies on components outside of the header and xml files.  It's probably a good idea to pick up revisions that modify any valid usage text (which did occur in 30 and 31).  We also check for changes in the python worker-scripts that used for codegen in LVL.  If none of these guys change then sure, it's a no-op.

courtney-g:
@karl-lunarg I was thinking automation. Having to remember to change 16 json files is annoying.

PR #1069, opened 2016-10-18
Add Validation Error Id Support to ErrorMonitor

mark-lunarg:
SetDesiredMessage is now overloaded to take a string or a unique
error enum ID (VALIDATION_ERROR_xxxxx) to identify a particular
message. Passing in VALIDATION_ERROR_MAX_ENUM will match all
messages for positive tests.  Also updated a couple of tests to use
error enums instead of strings.

### DIFF
--- a/layers/spec.py
+++ b/layers/spec.py
@@ -3,7 +3,6 @@
 import sys
 import xml.etree.ElementTree as etree
 import urllib2
-#import codecs
 
 #############################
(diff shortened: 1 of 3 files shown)


tobine:
This looks pretty good to me but wanted to have @chrisforbes give check it out as well.

mark-lunarg:
Pushed to Master.

PR #1070, opened 2016-10-18
layers: More err enum work in obj_tracker

karl-lunarg:
Change-Id: I299b68ec2e66d77812ce5dff0d457e328cdd6c5c


layers/object_tracker.cpp
```diff
@@ -1578,31 +1583,30 @@ VKAPI_ATTR void VKAPI_CALL UpdateDescriptorSets(VkDevice device, uint32_t descri
...
                     for (uint32_t idx3 = 0; idx3 < pDescriptorWrites[idx1].descriptorCount; ++idx3) {
                         if (pDescriptorWrites[idx1].pImageInfo[idx3].imageView) {
                             skip_call |= ValidateObject(device, pDescriptorWrites[idx1].pImageInfo[idx3].imageView,
-                                                                       VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_VIEW_EXT, false);
+                                                        VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_VIEW_EXT, false, VALIDATION_ERROR_00943);
                         }
-                        if (pDescriptorWrites[idx1].pImageInfo[idx3].sampler) {
-                            skip_call |= ValidateObject(device, pDescriptorWrites[idx1].pImageInfo[idx3].sampler,
-                                                                       VK_DEBUG_REPORT_OBJECT_TYPE_SAMPLER_EXT, false);
+                        if (pDescriptorWrites[idx1].pImageInfo[idx3].imageLayout) {
```
mark-lunarg: This looks funny.  VkImageLayouts aren't objects -- was this supposed to get changed from a VkSampler check?

layers/object_tracker.cpp
```diff
@@ -675,8 +675,8 @@ VKAPI_ATTR VkResult VKAPI_CALL InvalidateMappedMemoryRanges(VkDevice device, uin
         if (pMemoryRanges) {
             for (uint32_t idx0 = 0; idx0 < memoryRangeCount; ++idx0) {
                 if (pMemoryRanges[idx0].memory) {
-                    skip_call |= ValidateObject(device, pMemoryRanges[idx0].memory,
-                                                               VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_MEMORY_EXT, false);
+                    skip_call |= ValidateObject(device, pMemoryRanges[idx0].memory, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_MEMORY_EXT,
```
tobine: I believe this error is for a valid struct type, which will be checked by parameter_checker, should this be 00648 instead?

layers/object_tracker.cpp
```diff
@@ -652,8 +652,8 @@ VKAPI_ATTR VkResult VKAPI_CALL FlushMappedMemoryRanges(VkDevice device, uint32_t
...
```
tobine: I believe this error is for a valid struct type, which will be checked by parameter_checker, should this be 00648 instead?

layers/object_tracker.cpp
tobine: Yes, shouldn't parameter_checker check the layout?

layers/object_tracker.cpp
karl-lunarg: Yes.   Thanks, will fix.

layers/object_tracker.cpp
karl-lunarg: Yup

PR #1071, opened 2016-10-19
scripts:Fix valid usage parsing

### LINKED ISSUE #1067: Fix missing unique error enums
I noticed a block of 16 valid use cases that are missing unique error ids. Could be caused by spec.py parsing bug and/or an error in spec xhtml. Missing checks that I saw are in spec section "18.5. Image Copies with Scaling" for API call vkCmdBlitImage().

First missing check is `srcImage must have been created with VK_IMAGE_USAGE_TRANSFER_SRC_BIT usage flag`

With all checks after that in the "Valid Usage" block missing.

tobine:
Fixes #1067

Original valid usage parsing was naive and in how it used "strong" tags
to disable valid usage parsing which led to missing some checks and
including some spec language that wasn't valid usage.

This change isolates valid usage parsing to sidebar blocks which will
always contain the valid usage language. This prevents anything outside
of a sidebar from being incorrectly caught as valid usage and it also
catched a number of valid usage cases that were missed with previous
parser.

There are about 50 valid usage cases removed and about 300 added.


tobine:

PR #1073, opened 2016-10-19
LVTs: Create new VkPositiveLayerTest class

mark-lunarg:
Created new test class (VkPositiveLayerTest) for positive tests,
moved existing positive test into this class and split tests having
a mix of positive and negative tests. This should make things a bit
cleaner when adding valid usage IDs and adding test names into the
VU database.  In the future the class can be extended to do error
checking automatically.

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -565,6 +565,11 @@ void VkLayerTest::GenericDrawPreparation(VkCommandBufferObj *commandBuffer, VkPi
     commandBuffer->BindDescriptorSet(descriptorSet);
 }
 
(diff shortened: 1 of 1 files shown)


tobine:
Running this on our farm as lots of code was touched

tobine:
We're going to have to change some of our test exceptions for this PR to work correctly. Let us clean it up on our side and if everything looks good I'll push it.

tobine:
This is now clean on our farm. Will wait until tomorrow to merge in case anyone else had input or feedback.

mark-lunarg:
Pushed to Master.  Thanks @tobine.

PR #1082, opened 2016-10-21
Build Linux version with Travis CI.

petmac:
This could be useful for those who don't have a Linux box at hand.

I tried to keep the file as simple as I could, but perhaps in the future it could be expanded to specify more environments.


karl-lunarg:
This is a good thing to add, even though we just went through an exercise to clean out the top-level directory in the repo.  And I would like to look into hooking this repo up to Travis-ci.  The KhronosGroup/Vulkan-Samples repo does this.  But I might experiment with a fork first, though.

@petmac Were you able to actually try this to verify that it works as provided?

> This could be useful for those who don't have a Linux box at hand.

Just so I understand the use case, I imagine that a developer might be working on Windows with a fork of this repo and prepares a pull request.  Before submitting the PR, they would hook up their fork to Travis and have Travis compile for Linux to make sure it builds.  Is that right?

petmac:
Hi @karl-lunarg 

Yes, I ran this on a fork and built the Linux version OK. I'm not a Travis CI expert by any means, so it's likely that the script can be improved.

Your use case is exactly right. For what it's worth, I was experimenting with OS X, and wanted to ensure that I didn't break the Linux version along the way.

karl-lunarg:
Thanks!   I went ahead and pulled in the file as-is.  We can enhance the file later to add more build environments.   And at some point, we can hook up Travis CI to this repo so that people can see the build status.   We do run a bunch of our own CI machines, so this would be pretty redundant.  But there's some value in seeing the status on the GitHub page, like https://github.com/KhronosGroup/Vulkan-Samples.

Issue #1083, opened 2016-10-22
The loader doesn't handle a few Linux filesystem structures correctly


doraskayo:
The code is question is located in [vk_loader_platform.h](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/2d0b3724d6f7038d6cea5608b7c3be9637034b84/loader/vk_loader_platform.h#L60-L88), and has a few issues:
- When an absolute path is set to `CMAKE_INSTALL_DATADIR`, the implementation expects the ICD and layer manifests to be in a wrong location. For example, when `CMAKE_INSTALL_DATADIR=/usr/share` is defined, the ICD and layer manifests are expected to be in either `/usr/usr/share/vulkan/...` or `/usr/share/vulkan/...`, while they should _only_ be in `/usr/share/vulkan/...`.
- As explained in the ["Special Cases"](https://cmake.org/cmake/help/v3.4/module/GNUInstallDirs.html#special-cases) section of CMake's GNUInstallDirs documentation, there are also **three** cases that need extra attention when a prefix is defined using `CMAKE_INSTALL_PREFIX`. Only **one** of these special cases has a different code in the current implementation, and that is when `CMAKE_INSTALL_PREFIX=/usr` is defined. Unfortunately, that code also handles that case a bit poorly, as it expects the ICD and layer manifests to be in `/usr/etc/vulkan/...`, while they should actually be in `/etc/vulkan/...`.

Instead of doing a lot of work to get each of these cases handled correctly, it's possible to simply use CMake's built-in `CMAKE_INSTALL_FULL_<var>` instead of the currently used `CMAKE_INSTALL_<var>` in [CMakeLists.txt](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/2d0b3724d6f7038d6cea5608b7c3be9637034b84/CMakeLists.txt#L276-L277), which provides a standards-compliant absolute path to the aforementioned directories.

I've already fixed this issue and I'm currently preparing a pull request which I'll submit later today.

Issue #1084, opened 2016-10-22
Regression: the loader detects every ICD and layer twice in Linux when the filesystem has a prefix


doraskayo:
When `CMAKE_INSTALL_PREFIX` is set to `/usr` (such as in most Linux distributions), the code in [vk_loader_platform.h](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/2d0b3724d6f7038d6cea5608b7c3be9637034b84/loader/vk_loader_platform.h#L60-L88) causes the loader to search for ICD and layer manifests in the same directories twice, thus detecting them twice. This is quite bad, since it causes the loader to list every physical device twice for Vulkan clients that query for them, potentially causing undefined behavior when clients are dealing with multiple devices. (I've found a few tests that fail due to that.)

I'm not sure if filesystem prefixes were ever handled correctly, but at least in Arch Linux, this behavior is a regression; previous versions handled that case correctly. Unfortunately, I'm not sure in which commit the regression was introduced, but it doesn't really matter since it's simple enough to fix.

I've already fixed this issue and I'm currently preparing a pull request which I'll submit later today.

PR #1086, opened 2016-10-22
GH-1083, GH-1084: fix the loader's detection of manifests in Linux

doraskayo:
This pull request fixes GH-1083 and GH-1084.

I've combined the two commits into a single pull request because they pretty much depend on one another. In short, this pull request fixes quite a few outstanding issues with the loader's path-handling in Linux when built with various `CMAKE_INSTALL_<var>` defines while making use of the useful `CMAKE_INSTALL_FULL_<var>` define to resolve the absolute paths of the manifest directories.

It's important to note that the pull request introduces an additional requirement for packagers to use the `CMAKE_INSTALL_PREFIX` define when building for prefixed filesystems, but I highly doubt they don't already do that. Previously, `/usr/share/vulkan/...` and `/etc/vulkan/...` were hardcoded, even if the filesystem didn't have these directories at all, or if they weren't expected to be loaded. (e.g., when the manifests were in `/usr/local/share/vulkan/...` and `/usr/local/etc/vulkan/...`.)

Please refer to the issues and commit messages for a more detailed explanation.

I've tested this using vulkaninfo, demos and other Vulkan applications, and also ran strace on a few of them to compare and look for possible issues with filesystem access, but found none. I've only managed to run a few of the built-in tests, probably because I'm using an Intel Haswell GPU on Linux, which still has an incomplete Vulkan driver that keeps crashing. However, the tests I did manage to run had identical results before and after applying this pull request. I only tested on Linux, but as far as I can tell, it's also the only platform affected. Feel free to run additional tests.


loader/vk_loader_platform.h
```diff
@@ -59,14 +59,14 @@
 
 #if defined(LOCALPREFIX)
 #define LOCAL_DRIVERS_INFO                                                     \
-    LOCALPREFIX "/" SYSCONFDIR VULKAN_DIR VULKAN_ICDCONF_DIR ":"               \
-    LOCALPREFIX "/" DATADIR    VULKAN_DIR VULKAN_ICDCONF_DIR ":"
+    SYSCONFDIR VULKAN_DIR VULKAN_ICDCONF_DIR ":"                               \
```
MarkY-LunarG: This change will break XDG support added with commit ID 227a2c8eae5448646ae7ca14c62945c3ab3f4ccb.  Is there a way you can make your changes co-exist with that?

karl-lunarg: I'll check into this, but I think that the change in the way SYSCONFDIR and DATADIR are defined in CMakeLists.txt  may still allow the XDG support to work.

karl-lunarg:
@doraskayo:
I've pulled in this pull request and started testing it.

One issue is that building the repo without setting CMAKE_INSTALL_PREFIX results in a loader that doesn't look in the right places.  In this case:

CMAKE_INSTALL_SYSCONFDIR = etc
CMAKE_INSTALL_DATADIR = share
CMAKE_INSTALL_FULL_SYSCONFDIR = /usr/local/etc
CMAKE_INSTALL_FULL_SYSCONFDIR = /usr/local/share

Since your changes use the FULL variant, the loader searches in /usr/local/etc/vulkan/icd.d and /usr/local/share/vulkan/icd.d.  It misses searching in /etc/vulkan/icd.d.

I was able to work around this by specifying CMAKE_INSTALL_PREFIX=/ and I did note that you said:

> It's important to note that the pull request introduces an additional requirement for packagers to use the CMAKE_INSTALL_PREFIX define when building for prefixed filesystems

But I'd really like this to work as before so that builds used by non-packagers work without having to specify CMAKE_INSTALL_PREFIX.

One option would be to add this sort of thing to the CMakeLists.txt file:

```
IF(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
  SET(CMAKE_INSTALL_PREFIX
    "/" CACHE PATH "VULKAN install prefix" FORCE
    )
ENDIF(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
```

but this changes the CMake default of /usr/local.  I suppose that this can be handled with some docs.  (Note that if the user specifies CMAKE_INSTALL_PREFIX=/usr/local (re-setting the default), it will still make the test above fail since just touching the var, even with the default, will make CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT false.)

There was also a question raised about XDG_DATA_HOME, but my tests show that is still working OK with these changes.

Otherwise, this patch seems OK.  Do you have any better alternative or thoughts about handling the case of not specifying CMAKE_INSTALL_PREFIX better?

doraskayo:
@MarkY-LunarG, as Karl pointed out, this change should not affect XDG support, as it only changes the way system paths are handled, not user paths.

@karl-lunarg, just to make sure I understand the requirement here, would you like `/etc/vulkan/...` to always be searched, regardless of where Vulkan is built/installed by the user? If so, it's possible to check whether `CMAKE_INSTALL_FULL_DATADIR` is set to `/etc`, and if not, use the previous logic of adding `/etc` as an additional path to be searched by concatenating it with a colon separator. That way, `/etc` won't be searched twice in case, for example, `CMAKE_INSTALL_PREFIX=/usr` is defined by the user/packager (as is in most Linux distros) or in case `CMAKE_INSTALL_DATADIR=/etc` was defined. The same logic should also be applied to `/usr/share/vulkan/...` which was also always searched in the previous logic.

However, note that doing that would set us back to hardcoding system paths in the loader. That means that if a user wishes to install the loader and layers under an "isolated" path for testing software, such as the traditional `/usr/local/...`, they wouldn't be able to prevent the loader from loading ICD and layer manifests from `/etc/vulkan/...` and `/usr/share/vulkan/...` without explicitly setting both VK_LAYER_PATH and VK_ICD_FILENAMES to some value or commenting out lines in `CMakeLists.txt` before building.

Anyhow, I can update the pull request to either document the new behavior or hardcode the required paths. Let me know your thoughts.

karl-lunarg:
@doraskayo If you have not already, you can review https://vulkan.lunarg.com/doc/sdk/1.0.30.0/linux/LoaderAndLayerInterface.html which is our documentation for the loader, where it describes the search paths used by the loader.

Probably one key element here is that the LVL repo user doesn't build or install actual Vulkan drivers.  The Vulkan drivers come from IHV's and their install packages install the driver SO and the ICD JSON manifest file.  For example, installing the nvidia driver installs /etc/vulkan/icd.d/nvidia_icd.json.  This implies that if I build for an "isolated" system path, I may still want the loader to search /etc/vulkan for ICDs.

A very common workflow for the LVL repo is just building it with no special CMake settings and then setting LD_LIBRARY_PATH (or use rpath) to find the loader, and VK_LAYER_PATH to find the layers, where both the loader and layers are in the build directory configured by CMake.  The loader finds the drivers by looking for manifest files in the documented locations.  So it is important to keep this workflow working.

I don't think that very many people were using "make install" to put the binaries out into system search paths like /usr/local because make install was very broken until recently, to the point of being unusable.  We've only been trying to make it useful lately as part of finally getting around to it, and partly due to a few people showing interest, including me :-). 

Another key workflow is for package maintainers.  Sadly, the CMake files probably were not initially built for this workflow and obviously still are not set up for it, so we appreciate your help.

I think a good next step might be for me to take you up on your offer to update the pull request to "hardcode the paths to retain the previous behavior", with something additional in place to avoid searching the same place twice.   This sounds good to me in light of the documentation that I referenced above which is pretty clear and rigid in the area of search paths.

Thanks!

doraskayo:
@karl-lunarg, just as I thought ;)

I've updated the pull request with an additional commit, and changed the wording of the previous commits a bit to better describe their contents. The solution I've come up with is not as elegant as I would have hoped, mostly due to limited functionality of the C preprocessor. However, it seems to work very nicely. I'd appreciate another round of tests, as I'm still unable to fully run all of the tests due to the limited stability of my driver.

karl-lunarg:
I tested the latest changes by building with no specific CMAKE vars set and found that the search paths are not the same as they were before any of the changes.

With the changes, the search paths include /usr/local/etc, /usr/local/share, /etc, /usr/share.

Without the changes, the search paths include /etc, /usr/share, which is what is documented and desired.

Note that you can do:
`VK_LOADER_DEBUG=all ./vulkaninfo`
to spew loader debug info to stderr.  This info includes the search paths.

You can also discern that this would be the case by inspecting the code:
```
#define DEFAULT_VK_DRIVERS_INFO                                                \
    SYSCONFDIR   VULKAN_DIR VULKAN_ICDCONF_DIR ":"                             \
    DATADIR      VULKAN_DIR VULKAN_ICDCONF_DIR                                 \
    EXTRA_DRIVERS_SYSCONFDIR_INFO                                              \
    EXTRA_DRIVERS_DATADIR_INFO
```
since SYSCONFDIR and DATADIR come from CMAKE_INSTALL_FULL_SYSCONFDIR and CMAKE_INSTALL_FULL_DATADIR, respectively, and default to /usr/local/etc and /usr/local/share.  (These are not wanted)

The /etc and /usr/share paths come from the latest code addition that sets EXTRASYSCONFDIR and EXTRADATADIR. (These are wanted)

I looked for a bit and didn't see a good way to get around this problem.

Then, I started thinking that putting the two /usr/local paths in front might be desirable behavior.  For example, if I just build normally with no CMAKE vars defined and I do a "sudo make install", the "install" targets should go to /usr/local, with the (layer) manifest files going to /usr/local/etc.  (Currently, the manifest files go to /etc which is likely a bug.  See #1100.)

I hesitate to make this change to add the paths since our docs already specify something different.  But adding the paths seem to make sense and it seems that adding /usr/local isn't very risky since users should be aware of what is in their /usr/local.

Thoughts?

doraskayo:
This is indeed the intended behavior, exactly for the reasons you stated. The idea is that when a user defines certain paths in `CMAKE_INSTALL_PREFIX`, `CMAKE_INSTALL_SYSCONFDIR` or `CMAKE_INSTALL_DATADIR`, their intent is to make building the library and optionally installing it  (with the install command) work for their specific system, by telling us about the structure of their filesystem. This is how almost every Linux distribution installs its packages. In the loader's case, we can interpret these defines as where to look for manifests.

The "pure" way of dealing with that use case is to only listen to what is stated in these defines (or not stated, when some of them are left to their defaults) so that the loader works exactly as the user expects. For example, if `CMAKE_INSTALL_SYSCONFDIR=/etc2` is defined, the user tries to tell us that their system configuration is in `/etc2` and not anymore else. The previous version of this pull request did exactly that, and "purely" relied on these defines.

Then we agreed that `/etc/vulkan` and `/usr/share/vulkan` were far too standard to simply forget about, so they are now always searched. This introduces a slightly unexpected behavior for a user who tells us their system configuration is in `/etc2`, because we'll be looking in `/etc` as well, which they may not want us to do. It's important to note here that even though the documentation explicitly states some paths as the "right" paths to place certain things (such as the manifests), these paths may simply not be right for some filesystem structures. For example, if a user's distribution puts data files inside `/share` and not `/usr/share` because that's what they think is best, then the ICD manifests are going to be located in `/share/vulkan/icd.d`, and not `/usr/share/vulkan/icd.d`. If that user builds the loader from source without specifying `CMAKE_INSTALL_DATADIR=/share`, we won't be able to find their ICD manifests. It's basically impossible to make it "just work" for everyone when building without the right `CMAKE_INSTALL` defines. It is possible, however, to respect both `$XDG_DATA_DIRS` and `$XDG_CONFIG_DIRS` environment variables to try to do a better job at guessing the paths and only fall back to `/usr/share` and `/etc` in case these variables aren't available. This should of course be in addition to searching `CMAKE_INSTALL_FULL_DATADIR` and `CMAKE_INSTALL_FULL_SYSCONFDIR` (while making sure there is no duplication), to properly support building and installing on non-XDG compliant systems while still supporting different filesystem structures. However, I feel this should be an additional feature and separate from this pull request, as the intent of this pull request is, after all, to first fix the bugs in the existing implementation (as outlined in GH-1083 and GH-1084.)

Finally, for the specific issue of `/usr/local/share` and `/usr/local/etc`, the reason these paths are searched is simply because the default value of `CMAKE_INSTALL_PREFIX` is `/usr/local` by standard. We can't tell when a user builds the repository for the intent of installing it afterwards (in `/usr/local/...`) or when they build for testing purposes. Regardless, searching `/usr/local/...` as well in that case would be the right thing to do, and is not unexpected.

karl-lunarg:
Thank you very much for the help and the info!  We appreciate the help in understanding the needs of package maintainers and want to help make their jobs easier.

> Finally, for the specific issue of /usr/local/share and /usr/local/etc, the reason these paths are searched is simply because the default value of CMAKE_INSTALL_PREFIX is /usr/local by standard.

Yes, I thought about the argument that it would be strange to let the default value of a build tool dictate the search paths, but as you say, it is not unexpected and seems to be norm for things like configure.

I've pushed these commits with only some slight edits to the commit messages.
I'll get the docs caught up here where needed, and as well as on the LunarXchange web site.

PR #1088, opened 2016-10-24
Tobin continuing in use updates

tobine:
Update a bunch of Destroy functions to:

1. Follow Pre/Post pattern
2. Make sure they have unique error enum
3. Add a validation flag
4. Have correct support for in_use check
5. Correctly flag cmd buffer invalid for any bound cmd buffer

Also fixed up an object_tracker bug where skip wasn't being set and updated database file.
Finally, did a bunch of renaming to get rid of "node" and use more descriptive "state" instead.


layers/core_validation.cpp
```diff
@@ -6061,22 +6061,38 @@ DestroyPipelineLayout(VkDevice device, VkPipelineLayout pipelineLayout, const Vk
     dev_data->dispatch_table.DestroyPipelineLayout(device, pipelineLayout, pAllocator);
 }
 
+static bool PreCallValidateDestroySampler(layer_data *dev_data, VkSampler sampler, SAMPLER_NODE **sampler_state,
+                                          VK_OBJECT *obj_struct) {
+    if (dev_data->instance_data->disabled.destroy_sampler)
+        return false;
```
chrisforbes: `*sampler_state` is left as nullptr in this case, but `PostCallRecordDestroySampler` will unconditionally dereference it.

layers/core_validation.cpp
```diff
@@ -6247,21 +6247,22 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateQueryPool(VkDevice device, const VkQueryPoo
...
-DestroyCommandPool(VkDevice device, VkCommandPool commandPool, const VkAllocationCallbacks *pAllocator) {
-    layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
-    bool skip_call = false;
-    std::unique_lock<std::mutex> lock(global_lock);
-    // Verify that command buffers in pool are complete (not in-flight)
-    auto pPool = getCommandPoolNode(dev_data, commandPool);
-    skip_call |= checkCommandBuffersInFlight(dev_data, pPool, "destroy command pool with", VALIDATION_ERROR_00077);
+static bool PreCallValidateDestroyCommandPool(layer_data *dev_data, VkCommandPool pool, COMMAND_POOL_NODE **cp_state) {
+    if (dev_data->instance_data->disabled.destroy_command_pool)
+        return false;
```
chrisforbes: Same problem here. If this is disabled, `*cp_state` gets left as nullptr...

layers/core_validation.cpp
```diff
@@ -6247,21 +6247,22 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateQueryPool(VkDevice device, const VkQueryPoo
...
+}
 
-    if (skip_call)
-        return;
+static void PostCallRecordDestroyCommandPool(layer_data *dev_data, VkCommandPool pool, COMMAND_POOL_NODE *cp_state) {
     // Must remove cmdpool from cmdpoolmap, after removing all cmdbuffers in its list from the commandBufferMap
-    clearCommandBuffersInFlight(dev_data, pPool);
-    for (auto cb : pPool->commandBuffers) {
+    clearCommandBuffersInFlight(dev_data, cp_state);
+    for (auto cb : cp_state->commandBuffers) {
```
chrisforbes: .. and then you dereference it unconditionally here.

layers/core_validation.cpp
```diff
@@ -742,16 +742,16 @@ static bool freeMemObjInfo(layer_data *dev_data, void *object, VkDeviceMemory me
...
+        // pInfo->cb_bindings -- this copy not needed in c++14,
         // and probably not needed in practice in c++11
-        auto bindings = pInfo->command_buffer_bindings;
-        for (auto cb : bindings) {
-            if (!dev_data->globalInFlightCmdBuffers.count(cb)) {
-                clear_cmd_buf_and_mem_references(dev_data, cb);
+        auto bindings = pInfo->cb_bindings;
+        for (auto cb_node : bindings) {
+            if (!dev_data->globalInFlightCmdBuffers.count(cb_node->commandBuffer)) {
+                clear_cmd_buf_and_mem_references(dev_data, cb_node->commandBuffer);
```
chrisforbes: What's going on here? `clear_cmd_buf_and_mem_references` still takes a `GLOBAL_CB_NODE *`, but `cb_node->commandBuffer` is a `VkCommandBuffer`

layers/core_validation.cpp
```diff
@@ -742,16 +742,16 @@ static bool freeMemObjInfo(layer_data *dev_data, void *object, VkDeviceMemory me
 
         assert(pInfo->object != VK_NULL_HANDLE);
         // clear_cmd_buf_and_mem_references removes elements from
-        // pInfo->command_buffer_bindings -- this copy not needed in c++14,
+        // pInfo->cb_bindings -- this copy not needed in c++14,
         // and probably not needed in practice in c++11
-        auto bindings = pInfo->command_buffer_bindings;
-        for (auto cb : bindings) {
-            if (!dev_data->globalInFlightCmdBuffers.count(cb)) {
```
chrisforbes: This isn't changing but is generally murky. Perhaps it would be a good idea to inline `clear_cmd_buf_and_mem_references` here so the reasoning can be local.

layers/core_validation.cpp
```diff
@@ -678,34 +678,34 @@ static void clear_cmd_buf_and_mem_references(layer_data *dev_data, const VkComma
...
+                            (uint64_t)mem_info->mem, __LINE__, MEMTRACK_FREED_MEM_REF, "MEM",
                             "Attempting to free memory object 0x%" PRIxLEAST64 " which still contains " PRINTF_SIZE_T_SPECIFIER
                             " references",
-                            (uint64_t)pMemObjInfo->mem, (cmdBufRefCount + objRefCount));
+                            (uint64_t)mem_info->mem, (cb_ref_count + obj_ref_count));
     }
 
-    if (cmdBufRefCount > 0 && pMemObjInfo->cb_bindings.size() > 0) {
-        for (auto cb : pMemObjInfo->cb_bindings) {
+    if (cb_ref_count > 0 && mem_info->cb_bindings.size() > 0) {
```
chrisforbes: This `&&` is nonsense. Both sides are the same expression.

layers/core_validation.cpp
```diff
@@ -678,34 +678,34 @@ static void clear_cmd_buf_and_mem_references(layer_data *dev_data, const VkComma
...
+                    "Command Buffer 0x%p still has a reference to mem obj 0x%" PRIxLEAST64, cb, (uint64_t)mem_info->mem);
         }
         // Clear the list of hanging references
-        pMemObjInfo->cb_bindings.clear();
+        mem_info->cb_bindings.clear();
     }
 
-    if (objRefCount > 0 && pMemObjInfo->obj_bindings.size() > 0) {
-        for (auto obj : pMemObjInfo->obj_bindings) {
+    if (obj_ref_count > 0 && mem_info->obj_bindings.size() > 0) {
```
chrisforbes: Ditto

layers/core_validation.cpp
```diff
@@ -5172,24 +5080,76 @@ VKAPI_ATTR VkResult VKAPI_CALL AllocateMemory(VkDevice device, const VkMemoryAll
...
 
-    // From spec : A memory object is freed by calling vkFreeMemory() when it is no longer needed.
-    // Before freeing a memory object, an application must ensure the memory object is no longer
-    // in use by the device—for example by command buffers queued for execution. The memory need
-    // not yet be unbound from all images and buffers, but any further use of those images or
-    // buffers (on host or device) for anything other than destroying those objects will result in
-    // undefined behavior.
+static bool PreCallValidateFreeMemory(layer_data *dev_data, VkDeviceMemory mem, DEVICE_MEM_INFO **mem_info, VK_OBJECT *obj_struct) {
+    if (dev_data->instance_data->disabled.free_memory)
+        return false;
```
chrisforbes: On this path, `*mem_info` remains `nullptr`...

layers/core_validation.cpp
```diff
@@ -5172,24 +5080,76 @@ VKAPI_ATTR VkResult VKAPI_CALL AllocateMemory(VkDevice device, const VkMemoryAll
...
+    if (*mem_info) {
+        *obj_struct = {reinterpret_cast<uint64_t &>(mem), VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_MEMORY_EXT};
+        skip |= ValidateObjectNotInUse(dev_data, *mem_info, *obj_struct, VALIDATION_ERROR_00620);
+    }
+    return skip;
+}
 
+static void PostCallRecordFreeMemory(layer_data *dev_data, VkDeviceMemory mem, DEVICE_MEM_INFO *mem_info, VK_OBJECT obj_struct) {
+    // Clear mem binding for any bound objects
+    if (mem_info->obj_bindings.size() > 0) {
```
chrisforbes: ... then boom.

layers/core_validation.cpp
```diff
@@ -5172,24 +5080,76 @@ VKAPI_ATTR VkResult VKAPI_CALL AllocateMemory(VkDevice device, const VkMemoryAll
...
+                buff_node->binding.mem = MEMORY_UNBOUND;
+                break;
+            }
+            default:
+                // Should only have buffer or image objects bound to memory
+                assert(0);
+            }
+        }
+        // Clear the list of hanging references
+        mem_info->obj_bindings.clear();
```
chrisforbes: Not sure why you bother with this. It's about to be deallocated.

layers/core_validation.cpp
chrisforbes: This guard around the loop is also spurious.

layers/core_validation.cpp
```diff
@@ -5172,24 +5080,76 @@ VKAPI_ATTR VkResult VKAPI_CALL AllocateMemory(VkDevice device, const VkMemoryAll
...
+}
+
+VKAPI_ATTR void VKAPI_CALL FreeMemory(VkDevice device, VkDeviceMemory mem, const VkAllocationCallbacks *pAllocator) {
+    layer_data *dev_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
+    DEVICE_MEM_INFO *mem_info = nullptr;
+    VK_OBJECT obj_struct;
     std::unique_lock<std::mutex> lock(global_lock);
-    bool skip_call = freeMemObjInfo(my_data, device, mem, false);
-    print_mem_list(my_data);
-    printCBList(my_data);
```
chrisforbes: Good riddance to this... can the implementation go away too?

layers/core_validation.cpp
```diff
@@ -5436,21 +5436,34 @@ DestroySemaphore(VkDevice device, VkSemaphore semaphore, const VkAllocationCallb
     }
 }
 
+static bool PreCallValidateDestroyEvent(layer_data *dev_data, VkEvent event, EVENT_NODE **event_state, VK_OBJECT *obj_struct) {
+    if (dev_data->instance_data->disabled.destroy_event)
+        return false;
```
chrisforbes: Same pattern...

layers/core_validation.cpp
```diff
@@ -5436,21 +5436,34 @@ DestroySemaphore(VkDevice device, VkSemaphore semaphore, const VkAllocationCallb
...
+    *event_state = getEventNode(dev_data, event);
+    if (*event_state) {
+        *obj_struct = {reinterpret_cast<uint64_t &>(event), VK_DEBUG_REPORT_OBJECT_TYPE_EVENT_EXT};
+        skip |= ValidateObjectNotInUse(dev_data, *event_state, *obj_struct, VALIDATION_ERROR_00213);
+    }
+    return skip;
+}
+
+static void PostCallRecordDestroyEvent(layer_data *dev_data, VkEvent event, EVENT_NODE *event_state, VK_OBJECT obj_struct) {
+    invalidateCommandBuffers(event_state->cb_bindings, obj_struct);
```
chrisforbes: [null ptr deref here]

layers/core_validation.cpp
```diff
@@ -8671,11 +8671,11 @@ CmdResetEvent(VkCommandBuffer commandBuffer, VkEvent event, VkPipelineStageFlags
     if (pCB) {
         skip_call |= addCmd(dev_data, pCB, CMD_RESETEVENT, "vkCmdResetEvent()");
         skip_call |= insideRenderPass(dev_data, pCB, "vkCmdResetEvent");
-        auto event_node = getEventNode(dev_data, event);
```
chrisforbes: This is all identical to CmdSetEvent except for some strings... maybe we should sort this out later.

layers/core_validation.cpp
tobine: Potentially but there are still some uses of it so keeping around for this PR

tobine:
Added a couple of additional CLs to address @chrisforbes feedback

PR #1096, opened 2016-10-25
layers:Rename cmd buffer flags for clarity

### LINKED ISSUE #870: More clearly separate dynamic state flags in GLOBAL_CB_NODE
Right now we have status flags that track dynamic state as well as index buffer updated state. We should change CBSTATUS_ALL so that it's clear that it only deals with dynamic state which we can keep in the lowermost bits of status mask. Other state can then be tracked in the status mask in the same fashion as index buffer update when/if appropriate.

tobine:
Fixes #870

CBSTATUS_ALL was not clear so changed to more descriptive
CBSTATUS_ALL_STATE_SET. This mask is used to flag all of the pipeline
state for a cmd buffer as "set." The setting can then be adjusted based
on which state was flagged as dynamic in the pipeline state object.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -4205,12 +4205,12 @@ static void set_cb_pso_status(GLOBAL_CB_NODE *pCB, const PIPELINE_STATE *pPipe) 
     // Account for any dynamic state not set via this PSO
     if (!pPipe->graphicsPipelineCI.pDynamicState ||
         !pPipe->graphicsPipelineCI.pDynamicState->dynamicStateCount) { // All state is static
-        pCB->status |= CBSTATUS_ALL;
+        pCB->status |= CBSTATUS_ALL_STATE_SET;
     } else {
         // First consider all state on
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -4205,12 +4205,12 @@ static void set_cb_pso_status(GLOBAL_CB_NODE *pCB, const PIPELINE_STATE *pPipe)
     // Account for any dynamic state not set via this PSO
     if (!pPipe->graphicsPipelineCI.pDynamicState ||
         !pPipe->graphicsPipelineCI.pDynamicState->dynamicStateCount) { // All state is static
-        pCB->status |= CBSTATUS_ALL;
+        pCB->status |= CBSTATUS_ALL_STATE_SET;
     } else {
         // First consider all state on
         // Then unset any state that's noted as dynamic in PSO
```
chrisforbes: I still think this whole thing is wobbly, but :+1:

tobine:

PR #1098, opened 2016-10-26
layers: Don't put self-dep edges in the subpass DAG, or any other junk

### LINKED ISSUE #1094: Stack overflow due to recursion in core_validation running CTS renderpass test
Running dEQP-VK.renderpass.formats.d16_unorm.input.clear.store.clear with validation enabled, I see a call to CmdBeginRenderpass which calls ValidateDependencies and then recursive calls to CheckPreserved that result in stack overflow.  index alternates between 0 and 1 from successive calls and depth increases by one each time, but no exit condition is ever reached. api_dump available on request.

chrisforbes:
We'd already excluded edges to/from VK_SUBPASS_EXTERNAL, but we don't
want any of the other weird cases either; if we attempted to trace use
of an attachment, and found a subpass which /didn't/ use it, but had a
self-dependency, we'd get stuck chasing that dependency recursively
until we exhausted the stack.

Fixes #1094


Tony-LunarG:
Works for me and fixes the test that was stack overflowing.

chrisforbes:
Clean on the farm, pushed to master.

PR #1099, opened 2016-10-26
Fix codegen issue in dispatch-table-helper

mark-lunarg:
If a dispatch table entry needed to be protected by a #ifdef and was
the last prototype in the list, the corresponding #endif was not
generated causing compilation failures.

Change-Id: I8760bf546bc38d6fc7d80cd3a095e52977dc554e


mark-lunarg:
Pushed to Master.

PR #1102, opened 2016-10-26
Tobin gh1079 barrier aspect mask

### LINKED ISSUE #1079: Transitioning a depth- or stencil-only image with a vkCmdPipelineBarrier() calls confuse core validation layer
Say my image X uses a DEPTH16_UNORM format and my command buffer wants to transition the image to layout 1. As per the following spec language for valid usage of VkImageMemoryBarrier:

```
If image has a depth/stencil format with both depth and stencil components, then aspectMask member of subresourceRange must include both VK_IMAGE_ASPECT_DEPTH_BIT and VK_IMAGE_ASPECT_STENCIL_BIT
```

I specify both DS aspects when defining the subresource range. This, however, has a detrimental effect on the validation layer which outputs an error saying:

image

This is with the latest build of the layers, as obtained from the repo.

I can provide a repro for this issue if necessary.

tobine:
Fixes #1079 

Pull CreateImageView validation from image into core validation and use its aspect mask validation in ValidateBarrier as well.
Update tests, add unique error enums, and update database file.


layers/core_validation.cpp
```diff
@@ -9029,20 +9029,20 @@ static bool ValidateBarriers(const char *funcName, VkCommandBuffer cmdBuffer, ui
...
-                                         __LINE__, DRAWSTATE_INVALID_QUEUE_INDEX, "DS",
-                                         "%s: Image Barrier for image 0x%" PRIx64 " was created with sharingMode of "
-                                         "VK_SHARING_MODE_CONCURRENT.  Src and dst "
-                                         " queueFamilyIndices must be VK_QUEUE_FAMILY_IGNORED.",
-                                         funcName, reinterpret_cast<const uint64_t &>(mem_barrier->image));
+                    skip |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0,
+                                    __LINE__, DRAWSTATE_INVALID_QUEUE_INDEX, "DS",
+                                    "%s: Image Barrier for image 0x%" PRIx64 " was created with sharingMode of "
+                                    "VK_SHARING_MODE_CONCURRENT.  Src and dst "
+                                    " queueFamilyIndices must be VK_QUEUE_FAMILY_IGNORED.",
```
chrisforbes: There's some double spaces in this string

layers/core_validation.cpp
```diff
@@ -6494,6 +6494,68 @@ static void ResolveRemainingLevelsLayers(layer_data *dev_data, uint32_t *levels,
...
+                                    const char *func_name) {
+    bool skip = false;
+    if (vk_format_is_color(format)) {
+        if ((aspect_mask & VK_IMAGE_ASPECT_COLOR_BIT) != VK_IMAGE_ASPECT_COLOR_BIT) {
+            skip |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT,
+                            (uint64_t)image, __LINE__, VALIDATION_ERROR_00741, "IMAGE",
+                            "%s: Color image formats must have the VK_IMAGE_ASPECT_COLOR_BIT set. %s", func_name,
+                            validation_error_map[VALIDATION_ERROR_00741]);
+        }
+        if ((aspect_mask & VK_IMAGE_ASPECT_COLOR_BIT) != aspect_mask) {
```
chrisforbes: I think there's a win to be had here someday by rearranging this as required-bits/ allowed-bits like we do in some other cases -- that might also allow you to avoid producing two errors in each case where an invalid bit is the only bit set. [I see your test updates are currently having to account for that]

tobine:

PR #1104, opened 2016-10-27
vulkan.py-related rework and cleanup

mark-lunarg:
Moving API dependencies out of specific scripts and into vk.xml.  Getting ready for a follow-on series which will autogen vulkan.py.


mark-lunarg:
Pushed to Master.

PR #1106, opened 2016-10-27
Update to cmake files to use variables as destinations used in make install

johnzupin:
CMake files had destinations for install hard-coded. This request updated them to use variables that the user can define.


karl-lunarg:
This looks good.  Thanks!  I need to make one other change to the install command for the layer json files.  It makes sense to make that change along with these.  I'll apply this soon.

karl-lunarg:
pushed to master

PR #1107, opened 2016-10-27
Misc cleanup in cube

Tony-LunarG:
Make cube more straightforward by moving the initialization command buffer creation out of the set_image_layout utility, saving the staging image in the demo struct to get rid of premature command buffer submission, and make present mode a runtime option to get rid of ifdefs and multiple search loops


demos/cube.c
```diff
@@ -3232,13 +3213,20 @@ static void demo_init(struct demo *demo, int argc, char **argv) {
...
     demo->frameCount = INT32_MAX;
 
     for (int i = 1; i < argc; i++) {
         if (strcmp(argv[i], "--use_staging") == 0) {
             demo->use_staging_buffer = true;
             continue;
         }
+        if ((strcmp(argv[i], "--present_mode") == 0) &&
+                (i < argc - 1)) {
+            demo->presentMode = atoi(argv[i+1]);
```
ianelliottus: This sent alarm bells from my security training (HP days).  The atoi() function may not give you a valid value for demo->presentMode.  This isn't a mission-critical app, but it might still be good to do some checking to make sure this is a valid value:-).

ianelliottus:
I restricted myself to the present-mode changes.  Other than one minor issue, LGTM

demos/cube.c
Tony-LunarG: The search code will take care of unsupported (or non-existent) inputs.  If it's not in the list of supported present modes returned from the driver, the code will exit with an error.

PR #1108, opened 2016-10-28
layers: Add shader output variables assignment check

ghost:
According to spec it is invalid to use input variable in consumer
stage, when no value was assigned to it in previous stages.

This change add related validation checks to core layer.


layers/core_validation.cpp
```diff
@@ -1662,6 +1662,102 @@ static std::vector<std::pair<descriptor_slot_t, interface_var>> collect_interfac
...
+                    for (auto operation : operations) {
+                        /* check for opcode and parameter count */
+                        if (insn.opcode() == operation.first && operation.second < insn.len()) {
+                            uint32_t used_id = insn.word(operation.second);
+
+                            /* try to match any interface id with the desired operation parameter */
+                            for (auto it_interface = interface_variables.begin(); it_interface != interface_variables.end(); ++it_interface){
+                                if (it_interface->second.id == used_id){
+                                    if (!ids_usage.count(it_interface->first))
+                                        ids_usage[it_interface->first] = unordered_set<uint32_t>();
```
chrisforbes: You don't need to do this explicitly-- operator[] invoked on the next line default-constructs a missing entry.

layers/core_validation.cpp
```diff
@@ -1662,6 +1662,102 @@ static std::vector<std::pair<descriptor_slot_t, interface_var>> collect_interfac
...
+            producer_pointer_ids[producer_pointer_val.first] = interface_var{ pointer_id };
+        }
+    }
+
+    auto producer_saved_pointers = collect_interface_usage(producer, producer_pointer_ids, producer_entrypoint, unordered_map<spv::Op, unsigned> {{ spv::OpStore, 1 }});
+
+    /* for every location loaded by consumer, check if it was stored by producer */
+    for (auto consumer_loaded_val : consumer_loaded_vals) {
+        if (!producer_saved_vals.count(consumer_loaded_val.first) && !producer_saved_pointers.count(consumer_loaded_val.first)) {
+            if (log_msg(report_data, VK_DEBUG_REPORT_PERFORMANCE_WARNING_BIT_EXT, VkDebugReportObjectTypeEXT(0), 0,
```
chrisforbes: Is perf warning the right strength?

chrisforbes (changes requested):

chrisforbes:
First wave of comments is just surface stuff from a really brief look. I need to think about this a bit more.

scyganINTEL:
Improved performance by eliminating interface variables' copy, unnecessary iteration and copying of redundant arguments. Also changed the report flag.

Removed atomics used against pointers.

scygan:
@chrisforbes  - requested changes have been implemented. 

Can we have another round of review?

layers/core_validation.cpp
```diff
@@ -1662,6 +1662,104 @@ static std::vector<std::pair<descriptor_slot_t, interface_var>> collect_interfac
     return out;
 }
 
+/* Returns locations mapped to offsets in entrypoint where they were used */
+map<location_t , unordered_set<uint32_t>> collect_interface_usage(shader_module const *src, map<location_t, interface_var>& interface_variables,
```
chrisforbes: `interface_variables` should be ref to const.

chrisforbes (commented):
This looks fairly sensible to me. It needs tests, though.

tobine (changes requested):
The code looks reasonable but it's currently only building for windows. Large streams of error messages for both Linux and Android that look to be centered around the use of spv::Op in unordered_map. Perhaps a hash and equality functions definitions are required?

mark-lunarg:
Labeled 'inactive' and closing -- please reopen if necessary.

PR #1112, opened 2016-10-30
Re-added AllPlatforms as a wsi option to vk-generate.py

mre4ce:


mark-lunarg:
Pushed the vk-generate.py part of the change to Master.  Still working on some follow-on changes -- @mre4ce, let me know if this is not sufficient for you.

PR #1113, opened 2016-10-31
scripts:Update unique error enums as of 10/31/16

tobine:
No new checks this week, updated script to handle blank lines and also
to not output the API name debug info by default.


tobine:

tobine:
@karl-lunarg 253 just handles empty lines, including if the last line is an empty line. In a previous update I added [245](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/spec.py#L245) which is intended to always include a newline at the end of the file. Although now that I examine the file there doesn't appear to be a newline after the last line. If I put a second '\n' char on that line then I get a newline at the end of the file...

After a bit of experimentation it seems that the cleanest way to always have a newline is to append a single newline to the db_lines list prior to writing it out to the file. I'll submit that as its own PR as I'd like to always have the newline.

PR #1116, opened 2016-11-01
layers: Change to including vk_safe_struct code.

stroyan:
Cmake on windows cannot handle multiple projects using generated library code.
This happens with vk_safe_struct.cpp between core_validation and unique_objects.
Having both layers list it in add_vk_layer and the resulting add_library operation causes
their vcxproj files to create vk_safe_struct.cpp for build and remove vk_safe_struct.cpp for rebuild.
That breaks parallel builds when the creation collides.
It breaks rebuild when the recently created file is removed by another project.
Changing to using #include of the cpp file avoids that bad build behavior.


MarkY-LunarG:
Thanks Mike.  The changes look good, but let me test it out on my VS 2015 machine tomorrow to make sure it works.  I believe it should.

MarkY-LunarG:
It works for me!  Thanks again Mike

stroyan:
Cherry-picked to master

Issue #934, opened 2016-09-09
Layer doesn't check validity of scissor rectangles for vkCmdSetScissor


C0lumbo:
Section 25.2 of the spec says the following about scissor rectangles:
- The x and y members of offset must be greater than or equal to 0
- Evaluation of (offset.x + extent.width) must not cause a signed integer addition overflow
- Evaluation of (offset.y + extent.height) must not cause a signed integer addition overflow

parameter_validation_vkCmdSetScissor could do with checking these. Particularly the first of the three, which can cause some peculiar behaviour on some tiling hardware.

karl-lunarg:
Should do viewport at the same time.

mikew-lunarg:
partially fixed by 7572b6f915a6b3f8b5992a42d04ad20788c44b1d

mikew-lunarg:
SetViewport part is fixed by 79a99540d19bd9a809eed9af86463eabe79261eb

PR #1118, opened 2016-11-01
GH934 validate scissor parameters

mikew-lunarg:
Partially addresses [issue 934](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/issues/934), specifically the scissor parameters.


tests/layer_validation_tests.cpp
```diff
@@ -15745,6 +15746,68 @@ TEST_F(VkLayerTest, DuplicateDescriptorBinding) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(VkLayerTest, ViewportBoundsChecking) {
```
chrisforbes: ScissorBoundsChecking, perhaps, since the only misuses here are for the scissor, not the viewport ?

mikew-lunarg: Understood.  This is only half the resolution of the issue, the remainder actually is for Viewport, so when I submit the rest, the name will be more appropriate.

mikew-lunarg:
pushed to master at 7572b6f915a6b3f8b5992a42d04ad20788c44b1d

PR #1122, opened 2016-11-02
smoke: Add testing features

joey-lunarg:
smoketest will now output # of frames rendered and elapsed ms.
Command line argument added to limit the rendered frames,
usage: 'smoketest --c <framecount>'

Change-Id: Id6bc0186169b6511823c4406be655f6d8ca18324

### DIFF
--- a/loader/loader.c
+++ b/loader/loader.c
@@ -3305,8 +3305,8 @@ static void loader_init_dispatch_dev_ext_entry(struct loader_instance *inst,
             dev->loader_dispatch.ext_dispatch.dev_ext[idx] =
                 (PFN_vkDevExt)gdpa_value;
     } else {
-        for (uint32_t i = 0; i < inst->total_icd_count; i++) {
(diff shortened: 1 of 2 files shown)


demos/smoke/Game.h
```diff
@@ -119,6 +146,9 @@ class Game {
                 settings_.no_render = true;
             } else if (*it == "-np") {
                 settings_.no_present = true;
+            } else if (*it == "--c") {
```
mark-lunarg: Ugh.  8 of the options have one '-', while the validate and c options now take two.  Change them all to '--' and update docs in a separate commit?

cnorthrop:
This still works on Android although it would be nice to use shell_->log() instead of cout so it would work on all platforms.

olvaffe:
We could track number of frames in Shell and when the maximum number of frames is reached, Shell could ask Game for debug stats and quit.  We do not need any of the "#ifndef ANDROID" that way.

PR #1125, opened 2016-11-02
Replace vk-loader-generate with vk-generate

mark-lunarg:
- Fixed up  'all' parameter functionality in vk-generate.py
- Switched loader to use vk-generate.py instead of vk-loader-generate.py
- Removed vk-loader-generate.py


mark-lunarg:
Pushed to Master.

PR #1127, opened 2016-11-03
Tobin gh1124 secondary cb layout transitions

### LINKED ISSUE #1124: Secondary cmd buffer image layout transitions aren't reflected in primary cmd buffer
Consider a case where a secondary command buffer includes a PipelineBarrier that transitions an image from VK_IMAGE_LAYOUT_UNDEFINED to VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL.

This secondary command buffer is then included in a primary command buffer.
Next, a PipelineBarrier is bound to the primary cmd buffer that transitions the same image from VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL to VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL.

When this primary cmd buffer is submitted, validation will report something similar to `VULKAN DEBUG: ERROR:  [DS, 6]: Cannot submit cmd buffer using image (0xfe) [sub-resource: aspectMask 0x4 array layer 0, mip level 0], with layout VK_IMAGE_LAYOUT_UNDEFINED when first use is VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL.`

I believe this is incorrect, however, as validation didn't correctly take into account the image layout transitions from the secondary cmd buffer that was executed as part of the primary cmd buffer.

tobine:
Fixes #1124 

Propagate image layout transitions from secondary cmd buffer to primary cmd buffer at submit time.

Add a positive test to hit this case which fails prior to the fix.

Also fix a separate minor bug where activeFramebuffer wasn't correctly set from inheritance info.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -11008,6 +11008,10 @@ CmdExecuteCommands(VkCommandBuffer commandBuffer, uint32_t commandBuffersCount, 
                             "supported on this device.",
                             reinterpret_cast<uint64_t>(pCommandBuffers[i]));
             }
+            // Propagate layout transitions to the primary cmd buffer
+            for (auto ilm_entry : pSubCB->imageLayoutMap) {
+                SetLayout(pCB, ilm_entry.first, ilm_entry.second);
+            }
             pSubCB->primaryCommandBuffer = pCB->commandBuffer;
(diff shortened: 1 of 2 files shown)


tobine:

PR #1132, opened 2016-11-04
tests: Fix ResetEventThenSet to be a negative test

Tony-LunarG:
This addresses GH#1126
I think the original intent of this test was to set an event from the CPU that had a reset in an in flight command buffer without synchronization, getting a validation error.  I also removed the wait, which if done properly, would hang the GPU waiting for an event that never gets set.


chrisforbes:
Alright, this makes the test consistent with the current layer behavior -- the deeper question though is whether this error is justified. The spec doesn't say anything about it.

chrisforbes:
If we're deciding that this error is correct, then this change gets my :+1:.

Tony-LunarG:
Cherry picked to master.

PR #1136, opened 2016-11-08
Android update toolchain

cnorthrop:
In the downstream VulkanTools repo I need to build desktop and Android toolchains at the same time for vktrace.  They currently differ on revisions which makes for inefficient rebuilds every time, and the desktop revisions don't build for Android.

This series avoids that by separating the toolchain locations and changing how Android build decides which revisions it uses, puling from master AOSP instead of hard coded revisions.  Desktop build may also be able to use this strategy in the future.


build-android/jni/Android.mk
```diff
@@ -139,42 +140,47 @@ include $(BUILD_SHARED_LIBRARY)
...
 include $(PREBUILT_STATIC_LIBRARY)
 
 include $(CLEAR_VARS)
 LOCAL_MODULE := SPIRV-Tools-prebuilt
-LOCAL_SRC_FILES := $(SRC_DIR)/external/shaderc/android_test/obj/local/$(TARGET_ARCH_ABI)/libSPIRV-Tools.a
+LOCAL_SRC_FILES := $(ANDROID_DIR)/external/shaderc/android_test/obj/local/$(TARGET_ARCH_ABI)/libSPIRV-Tools.a
+include $(PREBUILT_STATIC_LIBRARY)
+
+include $(CLEAR_VARS)
+LOCAL_MODULE := SPIRV-Tools-opt-prebuilt
```
courtney-g: This seems unrelated to the move to an Android specific external tools.
Should this be removed? Part of a separate CL?
Update the comment to indicate why this is being added?

cnorthrop: It is a new requirement in shaderc build, so the inclusion of it has to coincide with the updated revision.  I can update the commit notes to mention it.

cnorthrop:
Comment updated, cherry-picked to master

PR #1140, opened 2016-11-10
GH1129, Fix for sticky DebugReportCallbacks flags

### LINKED ISSUE #1129: Debug report flags cant be disabled after being enabled.
When creating a debug report callback, with vkCreateDebugReportCallbackEXT, the flags parameter in the VkDebugReportCallbackCreateInfoEXT structure is used to select which types of messages to return. (Errors / Warnings / Info / ...)

However, if you then destroy the debug report, using vkDestroyDebugReportCallbackEXT, and re-create it with different flag settings, any previously enabled message types will still be enabled.  
It seems destroying debug report does not reset the flags.

mark-lunarg:
Fixes #1129.  When a debug report callback was created, destroyed, and re-created using different flags, the default callback flags controlling enabled message types would not get unset. 

Also fixed an issue with some debugreport debug messages not getting printed.

### DIFF
--- a/layers/vk_layer_logging.h
+++ b/layers/vk_layer_logging.h
@@ -188,10 +188,11 @@ static inline VkResult layer_create_msg_callback(debug_report_data *debug_data, 
 
     if (default_callback) {
         AddDebugMessageCallback(debug_data, &debug_data->default_debug_callback_list, pNewDbgFuncNode);
+        debug_data->active_flags |= pCreateInfo->flags;
     } else {
         AddDebugMessageCallback(debug_data, &debug_data->debug_callback_list, pNewDbgFuncNode);
+        debug_data->active_flags = pCreateInfo->flags;
     }
-    debug_data->active_flags |= pCreateInfo->flags;
 
     debug_report_log_msg(debug_data, VK_DEBUG_REPORT_DEBUG_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEBUG_REPORT_EXT,
                          (uint64_t)*pCallback, 0, VK_DEBUG_REPORT_ERROR_CALLBACK_REF_EXT, "DebugReport", "Added callback");
--- a/layers/vk_layer_logging.h
+++ b/layers/vk_layer_logging.h
@@ -64,8 +64,8 @@ static inline void RemoveDebugMessageCallback(debug_report_data *debug_data, VkL
     VkLayerDbgFunctionNode *cur_callback = *list_head;
     VkLayerDbgFunctionNode *prev_callback = cur_callback;
     bool matched = false;
+    VkFlags local_flags = 0;
 
-    debug_data->active_flags = 0;
     while (cur_callback) {
         if (cur_callback->msgCallback == callback) {
             matched = true;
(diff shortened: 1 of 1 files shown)


layers/vk_layer_logging.h
```diff
@@ -65,17 +65,18 @@ static inline void RemoveDebugMessageCallback(debug_report_data *debug_data, VkL
     VkLayerDbgFunctionNode *prev_callback = cur_callback;
     bool matched = false;
 
-    debug_data->active_flags = 0;
```
courtney-g: I think this needs to be a local variable. Initialized to 0.

layers/vk_layer_logging.h
```diff
@@ -65,17 +65,18 @@ static inline void RemoveDebugMessageCallback(debug_report_data *debug_data, VkL
...
     while (cur_callback) {
         if (cur_callback->msgCallback == callback) {
             matched = true;
             prev_callback->pNext = cur_callback->pNext;
             if (*list_head == cur_callback) {
                 *list_head = cur_callback->pNext;
             }
+            debug_data->active_flags = cur_callback->msgFlags;
```
courtney-g: Not needed with local variable.
Since we are not changing debug_data->active_flags, the debug_report_log_msg will report based on the current flags.

layers/vk_layer_logging.h
```diff
@@ -65,17 +65,18 @@ static inline void RemoveDebugMessageCallback(debug_report_data *debug_data, VkL
...
             debug_report_log_msg(debug_data, VK_DEBUG_REPORT_DEBUG_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEBUG_REPORT_EXT,
                                  reinterpret_cast<uint64_t &>(cur_callback->msgCallback), 0, VK_DEBUG_REPORT_ERROR_CALLBACK_REF_EXT,
-                                 "DebugReport", "Destroyed callback");
+                                 "DebugReport", "Destroyed callback\n");
+            debug_data->active_flags = 0;
```
courtney-g: Not needed.

layers/vk_layer_logging.h
```diff
@@ -65,17 +65,18 @@ static inline void RemoveDebugMessageCallback(debug_report_data *debug_data, VkL
...
         } else {
             matched = false;
             debug_data->active_flags |= cur_callback->msgFlags;
```
courtney-g: This would add the flags to the local flag variable.

Then, outside the while loop, we set debug_data->active_flags to our local variable where we computed the new active flag list.

layers/vk_layer_logging.h
```diff
@@ -188,10 +189,11 @@ static inline VkResult layer_create_msg_callback(debug_report_data *debug_data,
 
     if (default_callback) {
         AddDebugMessageCallback(debug_data, &debug_data->default_debug_callback_list, pNewDbgFuncNode);
+        debug_data->active_flags |= pCreateInfo->flags;
     } else {
         AddDebugMessageCallback(debug_data, &debug_data->debug_callback_list, pNewDbgFuncNode);
+        debug_data->active_flags = pCreateInfo->flags;
```
courtney-g: Not sure what this is doing here.
Why is it needed?

karl-lunarg: The problem was that when a debug report callback was destroyed, the active flags would get set to the flags used for the default callback.  If a callback was created again, the old OR statement (at old line 194) would OR in the requested flags, which also picks up the flags from the default callback.

The OR works for the first time a callback is created because the active flags were set to zero when the debug_data structure was first created.

The problem report may be a bit misleading.  I think that the actual observed behavior is that when creating a callback after one was destroyed, the flags for the new callback would already be set to the flags used by the default callback.  And since they were OR'd in, they appeared to be stuck on.  The only flags that would appear "stuck" would be the ones set by the default callback.   So for example, if the default callback sets DEBUG and ERROR, if the app created the first callback with INFO only, then deleted that callback, and then created another callback with ERROR set, the second callback would have DEBUG and ERROR set, and not INFO.  (I think - just from reading the code)

This is the primary problem this issue is addressing.   I didn't dig into the other changes about the Destroy path, and so Mark would have to explain those.

mark-lunarg:
Updated for Courtney's comments and pushed to Master.

renelindsay:
That matches the behavior we observed... DEBUG was sticky, but INFO was not.

PR #1145, opened 2016-11-11
vulkaninfo: fix segfault when DISPLAY is not set

abelal:
Both xlib and xcb interfaces expect the DISPLAY environment
variable to be set before creation of a window and the
display creation mechanism would segfault if that is
not the case and won't provide the user with details on
what has to be done to correct the problem.
We now handle such scenarios and exit cleanly after
providing the user with some details.

Signed-off-by: Awais Belal <awais_belal@mentor.com>

### DIFF
--- a/demos/vulkaninfo.c
+++ b/demos/vulkaninfo.c
@@ -911,6 +911,9 @@ static void app_create_xlib_window(struct app_instance *inst) {
     long visualMask = VisualScreenMask;
     int numberOfVisuals;
 
+    if (inst->xlib_display == NULL)
+        return;
+
     XVisualInfo vInfoTemplate={};
     vInfoTemplate.screen = DefaultScreen(inst->xlib_display);
     XVisualInfo *visualInfo = XGetVisualInfo(inst->xlib_display, visualMask,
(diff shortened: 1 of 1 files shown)


karl-lunarg:
Changes look useful, but on my system (Ubuntu LTS 16.04 nVidia GPU), the instance creation in app_create_instance() fails, which happens way before your changes to check the DISPLAY, etc.

I didn't see a segfault.  Can you identify in what module the segfault occurred for you?  What driver are you using?

For me, app_create_instance() just says it can't create the instance and doesn't really know why, so it doesn't say much more.  The loader does throw some messages about not being able to get a proc addr for CreateInstance from the nvidia ICD, but I don't know if it can determine if the failure was due to a bad or missing DISPLAY env var or not.  So, the user really isn't being told that the problem is due to the DISPLAY variable.

It might be better to move your check for a NULL DISPLAY towards the top of main(), but this still won't help the case where DISPLAY is set, but the app can't open the display.  Perhaps try to open the display there, near the top of main(), and then bail if that fails?

karl-lunarg:
Sort of a side comment (for future reference) here is a [post](https://devtalk.nvidia.com/default/topic/974876/vulkan/getting-vk_error_incompatible_driver/) that asserts that the nvidia vulkan driver needs to have X running and DISPLAY set to just create an instance.  This could be considered odd behavior and perhaps not worth working around as I suggested above.

abelal:
@karl-lunarg, is this good to go then in your opinion?

karl-lunarg:
I suppose.

Could you please sign/execute the CLA?

abelal:
@karl-lunarg, it might be obvious but can you please guide me on how to do that? This is my first contribution here...

abelal:
@karl-lunarg, CLA signed!

karl-lunarg:
Merged https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/commit/f231a17a49b329c79ef12bc55fde9bdc893372dc

PR #1147, opened 2016-11-11
WIP:  loader: gh1120 - Unwrap phys dev handles

MarkY-LunarG:
Unwrap both the physical device handles and the KHR_surface handles
in the loader during both the trampoline and terminator calls for
DebugMarker commands.  This has to be done since the values given
to a application are the loader trampoline versions, and the values
given to the layers are the loader terminator versions.

Thanks to Baldur (Mr. Renderdoc) for discovering this, testing my
fixes, and resolving a bug.

Change-Id: I7618d71ffee6c53d9842758210a9261f6b3a1797

### DIFF
--- a/loader/extensions.c
+++ b/loader/extensions.c
@@ -26,6 +26,100 @@
 #include "loader.h"
 #include "extensions.h"
 #include <vulkan/vk_icd.h>
+#include "wsi.h"
+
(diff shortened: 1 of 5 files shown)


MarkY-LunarG:
Closing as this caused other issues.

PR #1149, opened 2016-11-11
Fix core validation crashes

mark-lunarg:
Invalid LVT caused a crash in core-validation (RetireWorkOnQueue).  Fixed the test to prevent the bad result that LVT passed back, and added some null-checks to keep the layer from crashing.  Also reverted Tony's commit that undid a change which uncovered this crash.


mark-lunarg:
Pushed to Master.

Issue #125, opened 2016-03-14
Validate maxMemoryAllocationCount


jeffbolznv:
mem_tracker should validate that no more than VkPhysicalDeviceLimits::maxMemoryAllocationCount allocations exist at a time.

lentinem:
This might belong in device limits which generally has these types of checks. I'm ok with either but mem tracker is probably easier as it already stores the relevant information.

PR #1150, opened 2016-11-14
GH321, vkAllocateMemory return code handling

mark-lunarg:
Throwback Monday.  If a custom allocator is used and fails, validation will continue and use a null pointer, leading to invalid/unuseful validation errors. Added a check for a good return code.


karl-lunarg:
I'm in the middle of changing this function for #125 in branch karl_gh125_max_mem_count.  I'm adding a PostCallValidateAllocateMemory() which runs the code you've moved inside the if.  So I'll have to change mine to only call the Post function if result is success.

+1

I'll deal with the conflict.

mark-lunarg:
Pushed to Master.

PR #1156, opened 2016-11-15
demos: Enable building combinations of xcb and xlib

Tony-LunarG:
Allow LVL to build and demos to run with xcb, xlib, or xcb+xlib.


Tony-LunarG:
Cherry picked to master

Issue #329, opened 2016-04-13
Validation layer should check that vk(Invalidate|Flush)MappedRange() rounds ranges to VkPhysicalDeviceLimits::nonCoherentAtomSize


misojarv:
The spec section "10.2.1. Host Access to Device Memory Objects" has error condition for VkMappedMemoryRange arguments:
"offset must be a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize" and
"If size is not equal to VK_WHOLE_SIZE, size must be a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize"

The CTS tests in dEQP-VK.memory.mapping.\* currently don't check this and there are probably other tests too.

lentinem:
This is a bug with CTS? Are you saying validation doesn't check this?

misojarv:
This is a feature request.

The CTS does have some known issues with this, but the validation layer doesn't seem to be catching them.

mark-lunarg:
Addressed by PR #1157 .

mikew-lunarg:
There is a discussion about this at GitLab issue 611.
Meanwhile, PR #1157 has been disabled by commits 2fabc52a37efad97b70642932652a7129dcce79e and 82dbf60aef6769c5556deae2f8482ea9ce72c7c3

mark-lunarg:
Gitlab issue #611 has been closed, the PR was accepted, and LVL was updated.

PR #1157, opened 2016-11-15
Github Issue #329, validate against nonCoherentAtomSize device limit

mark-lunarg:
The spec section "10.2.1. Host Access to Device Memory Objects" has error condition for vkMappedMemoryRange arguments: "offset must be a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize" and "If size is not equal to VK_WHOLE_SIZE, size must be a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize"


layers/core_validation.cpp
```diff
@@ -11220,35 +11220,53 @@ static void CopyNoncoherentMemoryFromDriver(layer_data *dev_data, uint32_t mem_r
     }
 }
 
+static bool PreCallValidateFlushMappedMemoryRanges(layer_data *dev_data, uint32_t mem_range_count,
+                                                   const VkMappedMemoryRange *mem_ranges) {
+    bool skip = false;
+    std::unique_lock<std::mutex> lock(global_lock);
```
tobine: Should use a lock_guard here instead of unique_lock and kill the unlock call.

layers/core_validation.cpp
```diff
@@ -11220,35 +11220,53 @@ static void CopyNoncoherentMemoryFromDriver(layer_data *dev_data, uint32_t mem_r
...
+    if (!PreCallValidateFlushMappedMemoryRanges(dev_data, memRangeCount, pMemRanges)) {
         result = dev_data->dispatch_table.FlushMappedMemoryRanges(device, memRangeCount, pMemRanges);
     }
     return result;
 }
 
+static bool PreCallValidateInvalidateMappedMemoryRanges(layer_data *dev_data, uint32_t mem_range_count,
+                                                        const VkMappedMemoryRange *mem_ranges) {
+    bool skip = false;
+    std::unique_lock<std::mutex> lock(global_lock);
```
tobine: Should use a lock_guard here instead of unique_lock and kill the unlock call.

layers/core_validation.cpp
```diff
@@ -11220,35 +11220,53 @@ static void CopyNoncoherentMemoryFromDriver(layer_data *dev_data, uint32_t mem_r
...
+    skip |= validateMemoryIsMapped(dev_data, "vkInvalidateMappedMemoryRanges", mem_range_count, mem_ranges);
+    lock.unlock();
+    return skip;
+}
+
+static void PostCallRecordInvalidateMappedMemoryRanges(layer_data *dev_data, uint32_t mem_range_count,
+                                                       const VkMappedMemoryRange *mem_ranges) {
+    std::unique_lock<std::mutex> lock(global_lock);
```
tobine: Favor lock_guard over unique_lock whenever possible.

mark-lunarg:
Addressed @tobine's comments and pushed to Master.

mikew-lunarg:
There is a discussion about this at GitLab issue 611.
Meanwhile, PR #1157 has been disabled by commits 2fabc52 and 82dbf60

cforfang:
Hi @mikew-lunarg, now that the GitLab issue is resolved it seems like some of these checks should be re-enabled.

For example, the current spec has language like this which doesn't currently seem to be validated: 
```
If size is not equal to VK_WHOLE_SIZE, size must either be a multiple of
VkPhysicalDeviceLimits::nonCoherentAtomSize, or offset plus size must equal the size of
memory.```

PR #1158, opened 2016-11-15
Github Issue 1143 

mark-lunarg:
VertexInputAttributeDescription.format must be allowed as a vertex buffer format, as specified by the VK_FORMAT_FEATURE_VERTEX_BUFFER_BIT flag in VkFormatProperties::bufferFeatures returned by vkGetPhysicalDeviceFormatProperties


layers/image.cpp
```diff
@@ -1368,6 +1406,7 @@ intercept_core_device_command(const char *name) {
         {"vkCmdPipelineBarrier", reinterpret_cast<PFN_vkVoidFunction>(CmdPipelineBarrier)},
         {"vkCmdResolveImage", reinterpret_cast<PFN_vkVoidFunction>(CmdResolveImage)},
         {"vkGetImageSubresourceLayout", reinterpret_cast<PFN_vkVoidFunction>(GetImageSubresourceLayout)},
+        { "vkCreateGraphicsPipelines", reinterpret_cast<PFN_vkVoidFunction>(CreateGraphicsPipelines) },
```
karl-lunarg: nit: inconsistent white space.  If clang-format did this, then maybe change the lines above this one to match?

tests/layer_validation_tests.cpp
```diff
@@ -13527,6 +13527,64 @@ TEST_F(VkLayerTest, CreateImageLimitsViolationMinWidth) {
...
+    memset(&input_binding, 0, sizeof(input_binding));
+
+    VkVertexInputAttributeDescription input_attribs;
+    memset(&input_attribs, 0, sizeof(input_attribs));
+
+    // Pick a really bad format for this purpose and make sure it should fail
+    input_attribs.format = VK_FORMAT_BC2_UNORM_BLOCK;
+    VkFormatProperties format_props = m_device->format_properties(input_attribs.format);
+    if ((format_props.bufferFeatures & VK_FORMAT_FEATURE_VERTEX_BUFFER_BIT) != 0) {
+        return;
```
karl-lunarg: maybe an assert here, since this would be a test case logic error?

mark-lunarg: It's not a test failure -- this'll happen on hardware where this test makes no sense.  an early return is used in many tests that have to check for specific caps/props.

layers/image.cpp
mark-lunarg: Will fix.

tests/layer_validation_tests.cpp
chrisforbes: @mark-lunarg Consider adding a skip message to stdout in this case, like we do in many other cases where tests don't make sense?

mark-lunarg: Yes, thanks.

layers/image.cpp
```diff
@@ -215,6 +215,44 @@ static const VkLayerProperties global_layer = {
 
 // Start of the Image layer proper
```
mark-lunarg: Ugh, yeah.  When I first looked at CV I noticed formatproperties weren't used at all (whereas they are used all over the image layer), but since I'm just grabbing them as I go it doesn't matter.  I'll move it into CV.

layers/core_validation.cpp
```diff
@@ -6769,49 +6771,74 @@ void set_pipeline_state(PIPELINE_STATE *pPipe) {
...
+                VkFormat format = create_infos[i].pVertexInputState->pVertexAttributeDescriptions[j].format;
+                // Internal call to get format info.  Still goes through layers, could potentially go directly to ICD.
+                VkFormatProperties properties;
+                instance_data->dispatch_table.GetPhysicalDeviceFormatProperties(device_data->physical_device, format, &properties);
+                if ((properties.bufferFeatures & VK_FORMAT_FEATURE_VERTEX_BUFFER_BIT) == 0) {
+                    skip |= log_msg(
+                        device_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0,
+                        __LINE__, VALIDATION_ERROR_01413, "IMAGE",
+                        "vkCreateGraphicsPipelines: pCreateInfo[%d].pVertexInputState->vertexAttributeDescriptions[%d].format "
+                        "(%s) is not a supported vertex buffer format, %s.",
```
tobine: Itty bitty nit but for extra error message I've always been concluding the custom error message w/ a "." where you use a comma here, and then not adding an extra "." after valid-usage message.

So, I'm ending error string as:

`. %s"`

Where this code has:

`, %s."`

mark-lunarg: Ah, OK.  I'll change both of the PRs and do this in the future.

mark-lunarg:
Updated and pushed to Master.

PR #1159, opened 2016-11-15
Tobin gh1030 sparse binding size

tobine:
A couple of updates to how sparse bindings are handled.

Need to get size of sparse bindings and also account for separate memory bound in sparse cases when we're dealing with the memory object itself.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -144,7 +144,7 @@ struct layer_data {
     unordered_map<VkImageView, unique_ptr<IMAGE_VIEW_STATE>> imageViewMap;
     unordered_map<VkImage, unique_ptr<IMAGE_STATE>> imageMap;
     unordered_map<VkBufferView, unique_ptr<BUFFER_VIEW_STATE>> bufferViewMap;
-    unordered_map<VkBuffer, unique_ptr<BUFFER_NODE>> bufferMap;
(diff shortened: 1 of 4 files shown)


layers/core_validation.cpp
```diff
@@ -9214,7 +9214,7 @@ static bool ValidateBarriers(const char *funcName, VkCommandBuffer cmdBuffer, ui
 
         auto buffer_node = getBufferNode(dev_data, mem_barrier->buffer);
         if (buffer_node) {
-            auto buffer_size = buffer_node->binding.size;
+            auto buffer_size = buffer_node->GetMemorySize();
             if (mem_barrier->offset >= buffer_size) {
```
chrisforbes: Sum-of-binding-sizes doesn't make any sense for this check if you have sparse residency... for the sparse binding only level, it's OK I suppose.

tobine: Yeah, on further thought I don't like this.
I think instead we should just track size returned from Get*MemoryRequirements and check against that. Reworking.

tobine:
Pushed an update to this branch using size from memory requirements.
Also added new CL to rename BUFFER_NODE->BUFFER_STATE which I missed in my previous rename sweep.

tobine:
Replaced by #1168

PR #1163, opened 2016-11-16
layers: GH1131, Add error enums to object tracker

mikes-lunarg:
Added error enums to object_tracker.cpp and updated
vk_validation_error_database.txt check_implemented column to match

Change-Id: Icdd08ef50316be023b218b0b36a997a8a61e2a1e


layers/object_tracker.cpp
```diff
@@ -2944,14 +2949,14 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateSharedSwapchainsKHR(VkDevice device, uint32
...
         if (NULL != pCreateInfos) {
             for (i = 0; i < swapchainCount; i++) {
-                skip_call |= ValidateObject(device, pCreateInfos[i].oldSwapchain,
-                                                           VK_DEBUG_REPORT_OBJECT_TYPE_SWAPCHAIN_KHR_EXT, true);
+                skip_call |= ValidateObject(device, pCreateInfos[i].oldSwapchain, VK_DEBUG_REPORT_OBJECT_TYPE_SWAPCHAIN_KHR_EXT,
+                                            true, VALIDATION_ERROR_01935);
                 layer_data *device_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
                 skip_call |= ValidateObject(device_data->physical_device, pCreateInfos[i].surface,
-                                                           VK_DEBUG_REPORT_OBJECT_TYPE_SURFACE_KHR_EXT, false);
+                                            VK_DEBUG_REPORT_OBJECT_TYPE_SURFACE_KHR_EXT, false, VALIDATION_ERROR_01926);
```
tobine: This is an interesting case as same CreateInfo struct is shared for vkCreateSwapchainKHR and vkCreateSharedSwapchainsKHR. I'm fine with how it's implemented here, but would like to see some txt in the "note" column of the database file indicating that this error is dual-purposed for both of those functions.

layers/object_tracker.cpp
```diff
@@ -3555,24 +3561,24 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
     if (pCreateInfos) {
         for (uint32_t idx0 = 0; idx0 < createInfoCount; ++idx0) {
             if (pCreateInfos[idx0].basePipelineHandle) {
-                skip_call |= ValidateObject(device, pCreateInfos[idx0].basePipelineHandle,
-                                                           VK_DEBUG_REPORT_OBJECT_TYPE_PIPELINE_EXT, true);
+                skip_call |= ValidateObject(device, pCreateInfos[idx0].basePipelineHandle, VK_DEBUG_REPORT_OBJECT_TYPE_PIPELINE_EXT,
+                                            true, VALIDATION_ERROR_00529);
```
tobine: database file not updated for this error enum

layers/object_tracker.cpp
```diff
@@ -3555,24 +3561,24 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
...
-                                                           VK_DEBUG_REPORT_OBJECT_TYPE_PIPELINE_LAYOUT_EXT, false);
+                skip_call |= ValidateObject(device, pCreateInfos[idx0].layout, VK_DEBUG_REPORT_OBJECT_TYPE_PIPELINE_LAYOUT_EXT,
+                                            false, VALIDATION_ERROR_00546);
             }
             if (pCreateInfos[idx0].pStages) {
                 for (uint32_t idx1 = 0; idx1 < pCreateInfos[idx0].stageCount; ++idx1) {
                     if (pCreateInfos[idx0].pStages[idx1].module) {
                         skip_call |= ValidateObject(device, pCreateInfos[idx0].pStages[idx1].module,
-                                                                   VK_DEBUG_REPORT_OBJECT_TYPE_SHADER_MODULE_EXT, false);
+                                                    VK_DEBUG_REPORT_OBJECT_TYPE_SHADER_MODULE_EXT, false);
```
tobine: No valid usage case and corresponding error enum here? If not then should add a "TODO" comment and file spec bug as needed to get case added.

layers/object_tracker.cpp
```diff
@@ -3605,16 +3611,16 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateComputePipelines(VkDevice device, VkPipelin
...
-                skip_call |= ValidateObject(device, pCreateInfos[idx0].layout,
-                                                           VK_DEBUG_REPORT_OBJECT_TYPE_PIPELINE_LAYOUT_EXT, false);
+                skip_call |= ValidateObject(device, pCreateInfos[idx0].layout, VK_DEBUG_REPORT_OBJECT_TYPE_PIPELINE_LAYOUT_EXT,
+                                            false, VALIDATION_ERROR_00505);
             }
             if (pCreateInfos[idx0].stage.module) {
-                skip_call |= ValidateObject(device, pCreateInfos[idx0].stage.module,
-                                                           VK_DEBUG_REPORT_OBJECT_TYPE_SHADER_MODULE_EXT, false);
+                skip_call |=
+                    ValidateObject(device, pCreateInfos[idx0].stage.module, VK_DEBUG_REPORT_OBJECT_TYPE_SHADER_MODULE_EXT, false);
```
tobine: Do we need valid usage language added to spec for this one?

layers/object_tracker.cpp
```diff
@@ -2697,17 +2702,17 @@ VKAPI_ATTR VkResult VKAPI_CALL QueuePresentKHR(VkQueue queue, const VkPresentInf
             if (pPresentInfo->pSwapchains) {
                 for (uint32_t idx0 = 0; idx0 < pPresentInfo->swapchainCount; ++idx0) {
                     skip_call |= ValidateObject(queue, pPresentInfo->pSwapchains[idx0],
-                                                               VK_DEBUG_REPORT_OBJECT_TYPE_SWAPCHAIN_KHR_EXT, false);
+                                                VK_DEBUG_REPORT_OBJECT_TYPE_SWAPCHAIN_KHR_EXT, false, VALIDATION_ERROR_01969);
```
mikes-lunarg: corrected

layers/object_tracker.cpp
mikes-lunarg: Added comment to 515, 648, 1926, 1935 indicating multiple APIs

layers/object_tracker.cpp
mikes-lunarg: Added enum 515

layers/object_tracker.cpp
mikes-lunarg: Added enum 515

layers/object_tracker.cpp
mikes-lunarg: corrected

layers/object_tracker.cpp
```diff
@@ -314,9 +314,11 @@ static void DestroyObject(T1 dispatchable_object, T2 object, VkDebugReportObject
     }
 }
 
+static const int VALIDATION_ERROR_UNDEFINED = -1;
```
tobine: I'd like to see this declaration in the vk_validation_error_messages.h header. That means the code to declare it needs to be added into the "genHeader" function of the spec.py script. If you'd like to add that to this PR, that's great. Otherwise, I'm fine if this lands first and I can make that change in a follow-on commit.

karl-lunarg:
I double-checked the DB against the code; looks good.

> Also, I suspect that for nearly all of the cases touched here there is no test. In such cases the "testname" column in the database can be updated to "None". This is not a requirement but would be nice for bookkeeping and so they don't all have to be scrubbed in the future.

I agree that there are likely no tests for these and probably for all of object_tracker checks.   @mikes-lunarg, can you put this on your list to do soon, as a separate commit?  Note that this is just saying that there are no tests, and not to actually write them.

karl-lunarg:
Actually, there are quite a few (about 25) tests that trigger object tracker reports.  These are easy to detect by putting a breakpoint on the call to log_msg() in ValidateObject().

Many of these are due to a double destroy of an object in a test.  For example, a test might destroy a Buffer object to intentionally cause an error, and then later on try to destroy the object again as part of the test clean-up.   The second destroy seems like a test case coding error to me and should be cleaned up.

But there are a few tests that legitimately cause object_tracker to report an error.  Some of them even have a comment "object tracker should find this".  These tests could be marked as present in the database.  Anyway, just need to remember make a note to come back and do this.

layers/object_tracker.cpp
mikes-lunarg: I would like to move this change to master first because of some others that are waiting on it

PR #1166, opened 2016-11-17
Tobin gh1162 inactive dynamic bindings

### LINKED ISSUE #1162: Layers: Inactive dynamic bindings may interfere with dyamic_offset validation
Performing dynamic_offset checks against bindings which include inactive bindings returns unexpected results. 

It appears that the check in [descriptor_sets.cpp](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/descriptor_sets.cpp#L429) fails because the index-based mapping between setBindings and dynamic_offsets can get broken if there are inactive bindings. 

---
Example: 

I'm using 2 dynamic UBOs which bind to memory buffer A (binding 0, binding 1), and another dynamic UBO (binding 2) bound to a smaller buffer B. upon draw each binding is bound with an offset. The Ubo for Binding 1 is not used by the shader, and might have been optimised away. 

The error message - using current github master validation layers compiled from source is the following: 

`DS 0x3e encountered the following validation error at vkCmdDrawIndexed() time: Dynamic descriptor in binding #2 at global descriptor index 2 uses buffer 0000000000000023 with dynamic offset 16777472 combined with offset 0 and range 48 that oversteps the buffer size of 16777216.`

---

When I check my data, it appears that offset, range, and buffer size are correct, and the error message prints out correct and expected values for these. But `dynamic_offset` (the value '16777472' above) refers to the buffer offset previous to the buffer that needs to be checked.

My hypothesis is that whilst only active binding slots are checked for `dynamic_offset` correctness, the vector of `dynamic_offsets` passed to `cvdescriptorset::DescriptorSet::ValidateDrawState` may contain offsets for active and inactive slots alike.  

Printf-style debugging when compiling validation layers locally showed me that in this particular case there were indeed 2 bindings being checked (the descriptorSetLayout for this set contains 3, of which one inactive) against 3 `dynamic_offsets`. 

In [core_validation.cpp](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/core_validation.cpp#L3167) where this validation is prepared, the bindings appear to be grabbed only for `pPipe->active_slots`. 

Might be relevant for #880

tobine:
Fixes #1162 

Added a mapping between dynamic descriptor bindings and their associated index into the dynamic offset array.

Also added a positive test that fails without the fix but now works

### DIFF
--- a/layers/descriptor_sets.cpp
+++ b/layers/descriptor_sets.cpp
@@ -28,6 +28,9 @@ cvdescriptorset::DescriptorSetLayout::DescriptorSetLayout(const VkDescriptorSetL
                                                           const VkDescriptorSetLayout layout)
     : layout_(layout), binding_count_(p_create_info->bindingCount), descriptor_count_(0), dynamic_descriptor_count_(0) {
     uint32_t global_index = 0;
+    // Dyn array indicies are ordered by binding # and array index of any array within the binding
+    //  so we store up bindings w/ count in ordered map in order to create dyn array mappings below
+    std::map<uint32_t, uint32_t> binding_to_dyn_count;
     for (uint32_t i = 0; i < binding_count_; ++i) {
         descriptor_count_ += p_create_info->pBindings[i].descriptorCount;
         binding_to_index_map_[p_create_info->pBindings[i].binding] = i;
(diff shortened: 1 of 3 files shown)


tobine:

PR #1167, opened 2016-11-17
GH949, Fix object_tracker to accept NULL handles in destroy/free calls

### LINKED ISSUE #949: Layers: Don't treat vkFree/vkDestroy* on VK_NULL_HANDLE as an error
According to the vulkan spec, passing VK_NULL_HANDLE as the object parameter for vkFree/vkDestroy is perfectly valid and should generate no error, yet the layers seem to do so. (Specifically, ObjectTracker does)

This is rather inconvenient to me since re-freeing something that doesn't exist is a normal occurrence in my code base. (e.g. consider acquiring 5 resources: If the 5th acquisition fails, I can call my normal `destroy` routine to destroy anything that _did_ end up getting acquired. For another example, consider a routine that updates the swapchain to the a size, destroying and recreating stale resources as necessary. On the first frame, there are no old resources to destroy)

The work-around for now is to have an extra condition like `if (obj) vkDestroyThing(dev, obj, NULL);` in front of every free call, but it shouldn't be necessary.

mark-lunarg:
ObjectTracker emitted errors when a destroy/free call was made with VK_NULL_HANDLE, which *is* allowed by the spec.

No positive test for this since it happens to crash in many drivers.  Fixes #949.

### DIFF
--- a/layers/object_tracker.cpp
+++ b/layers/object_tracker.cpp
@@ -280,37 +280,40 @@ static void DestroyObject(T1 dispatchable_object, T2 object, VkDebugReportObject
     auto object_handle = handle_value(object);
     bool custom_allocator = pAllocator != nullptr;
 
-    auto item = device_data->object_map[object_type].find(object_handle);
-    if (item != device_data->object_map[object_type].end()) {
+    if (object_handle != VK_NULL_HANDLE) {
+        auto item = device_data->object_map[object_type].find(object_handle);
+        if (item != device_data->object_map[object_type].end()) {
 
-        OBJTRACK_NODE *pNode = item->second;
-        assert(device_data->num_total_objects > 0);
-        device_data->num_total_objects--;
-        assert(device_data->num_objects[pNode->object_type] > 0);
-        device_data->num_objects[pNode->object_type]--;
+            OBJTRACK_NODE *pNode = item->second;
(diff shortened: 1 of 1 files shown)


layers/object_tracker.cpp
```diff
@@ -280,37 +280,40 @@ static void DestroyObject(T1 dispatchable_object, T2 object, VkDebugReportObject
...
-        device_data->num_objects[pNode->object_type]--;
+            OBJTRACK_NODE *pNode = item->second;
+            assert(device_data->num_total_objects > 0);
+            device_data->num_total_objects--;
+            assert(device_data->num_objects[pNode->object_type] > 0);
+            device_data->num_objects[pNode->object_type]--;
 
-        log_msg(device_data->report_data, VK_DEBUG_REPORT_INFORMATION_BIT_EXT, pNode->object_type, object_handle, __LINE__,
+            log_msg(device_data->report_data, VK_DEBUG_REPORT_INFORMATION_BIT_EXT, pNode->object_type, object_handle, __LINE__,
                 OBJTRACK_NONE, LayerName,
```
tobine: Did you clang-format this? If so I guess it let the mis-alignment slide b/c line didn't change. Same thing on lines 302 & 313.

mark-lunarg:
Fixed and pushed to Master.

PR #1168, opened 2016-11-17
Tobin gh1030 refinement

### LINKED ISSUE #1030: Layers: sparse buffer size not tracked
I'm getting the following error:

> DS(ERROR): object: 0x0 type: 0 location: 8916 msgCode: 10: vkCmdPipelineBarrier: Buffer Barrier 0x1 has offset 0x0 which is not less than total size 0x0.

This happens because `buffer_node->memSize` is 0 in `ValidateBarriers()`. It could be that size isn't updated correctly when buffer is bound in `QueueBindSparse()`. The same probably happens for sparse images as well.

tobine:
Replaces #1159 

Fixes #1030 

I found that those initial changes were causing new errors on smoketest. Upon further investigation, smoketest binds memory to a buffer without ever calling vkGetBufferMemoryRequirements().
There's an implication in the spec that this is invalid, although I don't see that it's explicitly stated anywhere.
To deal with this issue initially, I've added warnings in the case where memory is bound to a buffer/image without first checking the requirements and then validation will make the requirements call itself to fill its state.

There is still more work to be done to refactor BindImage/BufferMemory functions to Pre/Post and add proper checks for all of the VUs along with tests. Also need to fix smoketest. I'll open separate issues for these, but this PR is focused on getting the fix in for 1030.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -9257,7 +9257,7 @@ static bool ValidateBarriers(const char *funcName, VkCommandBuffer cmdBuffer, ui
 
         auto buffer_node = getBufferNode(dev_data, mem_barrier->buffer);
         if (buffer_node) {
-            auto buffer_size = buffer_node->binding.size;
+            auto buffer_size = buffer_node->requirements.size;
             if (mem_barrier->offset >= buffer_size) {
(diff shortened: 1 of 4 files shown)


tobine:

PR #1174, opened 2016-11-18
Move glslang, spirv-tools and spirv-headers revision files into a subdirectory

mark-lunarg:
Unclutters the root dir a bit.


update_external_sources.sh
```diff
@@ -3,7 +3,8 @@
 
 set -e
 
-REVISION_DIR=$PWD/external_revisions
+CURRENT_DIR="$(dirname `readlink -f $0`)"
+REVISION_DIR="$CURRENT_DIR/external_revisions"
 
 GLSLANG_REVISION=$(cat ${REVISION_DIR}/glslang_revision)
```
karl-lunarg: needs quotes (like the next two)

mark-lunarg:
Updated, tested, and pushed to Master.

PR #1175, opened 2016-11-18
Move python scripts into subdirectory

mark-lunarg:
Move the codegen and utility python scripts and vk.xml into a scripts dir.  

Also, vk-layer-introspect.py is removed in this patch series -- if anyone speaks up I'll happily keep it.

@ardaLunarG , I'll work on a VT companion series before this lands in LVL.


tobine:
👎 Just want to make sure this is held up until we can run it through our farm which I don't suspect will be until Mon as we can't currently access it remotely. Will give it a try then and thumb up if clean.

mark-lunarg:
@tobine, bump.

tobine:
Verified clean on our farm. :+1:

mark-lunarg:
Pushed to Master.  Thanks @tobine.

PR #1176, opened 2016-11-19
scripts:Check image.cpp for unique error enums

tobine:
Stats script wasn't checking image.cpp when looking for implemented
unique error enums.

Also updated script to include number of enums that it finds which are
claiming to be implemented but not located in source.


tobine:

PR #1177, opened 2016-11-19
Karl linux32

karl-lunarg:
Turns out that the repo builds and the tests pass on 32-bit Linux.
Cleaning up a few warnings that have to do with copying 64-bit Vulkan handles into pointers, and update docs.


karl-lunarg:
pushed to master

PR #1178, opened 2016-11-21
scripts:Update unique error enums as of 11/21/16

tobine:
4 Updated checks and 1 new check.

Cmd line was:
python spec.py -update -remap 918-768,2:1096-936:2170-1964 > output.txt


layers/vk_validation_error_database.txt
```diff
@@ -758,8 +758,8 @@ VALIDATION_ERROR_00764~^~U~^~Unknown~^~vkCreateImageView~^~For more information
...
-VALIDATION_ERROR_00769~^~Y~^~ImageLayerViewTests~^~vkCreateImageView~^~For more information refer to Vulkan Spec Section '11.5. Image Views' which states 'If layerCount is not VK_REMAINING_ARRAY_LAYERS, (baseArrayLayer + layerCount) must be less than or equal to the arrayLayers specified in VkImageCreateInfo when the image was created' (https://www.khronos.org/registry/vulkan/specs/1.0-extensions/xhtml/vkspec.html#VkImageAspectFlagBits)~^~Dual purposing this for when layerCount is 0, need new valid usage language for that case
+VALIDATION_ERROR_00768~^~Y~^~InvalidImageView,ImageLayerViewTests~^~vkCreateImageView~^~For more information refer to Vulkan Spec Section '11.5. Image Views' which states 'If levelCount is not VK_REMAINING_MIP_LEVELS, levelCount must be non-zero and (baseMipLevel levelCount) must be less than or equal to the mipLevels specified in VkImageCreateInfo when the image was created' (https://www.khronos.org/registry/vulkan/specs/1.0-extensions/xhtml/vkspec.html#VkImageAspectFlagBits)~^~Dual purposing this for when levelCount is 0, need new valid usage language for that case
```
tobine: Yes, I agree that "C" for "Changed" is a good idea. This update also requires some re-evaluation of a couple descriptor checks that I'm following up on, but "C" would make that explicit.

I'm thinking by default when IDs are remapped via the "-remap" option I'll update the script to automatically put "C" in the implemented column and then manual follow-up commits will be required to change back to "Y" and make any appropriate updates in the checking code.

karl-lunarg: +1

PR #1180, opened 2016-11-21
layers:Warn on cmd buffer invalidate

tobine:
From Mikko Strandborg

When a resource gets deleted, any command buffer that references it
gets invalidated, and an error gets thrown whenever that command buffer
is submitted. However, for a developer, it's quite hard to deduce what
exactly caused that to happen. This patch emits a warning whenever a
command buffer that's in CB_RECORDING state (meaning that it has had
its vkBeginCommandBuffer called but not End). This way the developer
can set a breakpoint on that to see exactly what causes the buffer to
be invalidated.

Destroying a resource that's being used in an open command buffer is
almost certainly a bug, but there might still be cases where it might
be valid behavior (for example, when the application decides to abort
an operation), therefore it's not an error but a warning instead.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5181,7 +5181,7 @@ static void PostCallRecordFreeMemory(layer_data *dev_data, VkDeviceMemory mem, D
         }
     }
     // Any bound cmd buffers are now invalid
-    invalidateCommandBuffers(mem_info->cb_bindings, obj_struct);
+    invalidateCommandBuffers(dev_data, mem_info->cb_bindings, obj_struct);
     dev_data->memObjMap.erase(mem);
 }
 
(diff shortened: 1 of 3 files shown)


layers/core_validation.cpp
```diff
@@ -6384,8 +6384,13 @@ VKAPI_ATTR VkResult VKAPI_CALL ResetFences(VkDevice device, uint32_t fenceCount,
 }
 
 // For given cb_nodes, invalidate them and track object causing invalidation
-void invalidateCommandBuffers(std::unordered_set<GLOBAL_CB_NODE *> cb_nodes, VK_OBJECT obj) {
+void invalidateCommandBuffers(const layer_data *dev_data, std::unordered_set<GLOBAL_CB_NODE *> cb_nodes, VK_OBJECT obj) {
     for (auto cb_node : cb_nodes) {
+        if (cb_node->state == CB_RECORDING) {
+            log_msg(dev_data->report_data, VK_DEBUG_REPORT_WARNING_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
+                    (uint64_t)(cb_node), __LINE__, DRAWSTATE_INVALID_COMMAND_BUFFER, "DS",
```
stroyan: This log_msg should use cb_node->commandBuffer instead of cb_node.

tobine:

PR #1184, opened 2016-11-21
demos: Set attachment description flag bits in cube

Tony-LunarG:
Fixes a cubepp crash running on AMD


Tony-LunarG:
Cherry picked to master and sdk-1.0.33

PR #1185, opened 2016-11-21
Tobin fix invalidate cmd buffers

tobine:
Fixed a bug that @stroyan and @chrisforbes noted in a code review after I'd already pushed to master: use cmd buffer object, not ptr to the wrapping node struct.

Also clean up invalidateCommandBuffers() interface to pass ptr instead of copying unordered_set.

### DIFF
--- a/layers/core_validation_types.h
+++ b/layers/core_validation_types.h
@@ -656,7 +656,7 @@ SAMPLER_STATE *getSamplerState(const layer_data *, VkSampler);
 IMAGE_VIEW_STATE *getImageViewState(const layer_data *, VkImageView);
 VkSwapchainKHR getSwapchainFromImage(const layer_data *, VkImage);
 SWAPCHAIN_NODE *getSwapchainNode(const layer_data *, VkSwapchainKHR);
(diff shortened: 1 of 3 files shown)


layers/core_validation_types.h
```diff
@@ -656,7 +656,7 @@ SAMPLER_STATE *getSamplerState(const layer_data *, VkSampler);
 IMAGE_VIEW_STATE *getImageViewState(const layer_data *, VkImageView);
 VkSwapchainKHR getSwapchainFromImage(const layer_data *, VkImage);
 SWAPCHAIN_NODE *getSwapchainNode(const layer_data *, VkSwapchainKHR);
-void invalidateCommandBuffers(const layer_data *, std::unordered_set<GLOBAL_CB_NODE *>, VK_OBJECT);
+void invalidateCommandBuffers(const layer_data *, std::unordered_set<GLOBAL_CB_NODE *> *, VK_OBJECT);
```
chrisforbes: Why not `const &` ? The callee doesn't change this.

tobine:

PR #1191, opened 2016-11-23
layers: Handle partial failure in vkCreate*Pipelines

yavn:
Hi, this is driven by CTS and deals with an edge case of creating graphics and compute pipelines in bulk.

As per the spec, it's possible to create multiple pipelines with a single call. In case of a failure invalid handles will be set to VK_NULL_HANDLE and the remaining non-NULL handles are valid and must be dealt with correctly.

Currently layers check for an error code and ignore the result completely if there was an error. As a result returned handles aren't sanitized to global ids and are not known to the layers. Attempting to free them is treated as an error by the layers, leading to a memory leak.

### DIFF
--- a/layers/object_tracker.cpp
+++ b/layers/object_tracker.cpp
@@ -3589,8 +3589,8 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
     VkResult result = get_dispatch_table(ot_device_table_map, device)
                           ->CreateGraphicsPipelines(device, pipelineCache, createInfoCount, pCreateInfos, pAllocator, pPipelines);
     lock.lock();
-    if (result == VK_SUCCESS) {
-        for (uint32_t idx2 = 0; idx2 < createInfoCount; ++idx2) {
+    for (uint32_t idx2 = 0; idx2 < createInfoCount; ++idx2) {
+        if (pPipelines[idx2] != VK_NULL_HANDLE) {
             CreateObject(device, pPipelines[idx2], VK_DEBUG_REPORT_OBJECT_TYPE_PIPELINE_EXT, pAllocator);
         }
     }
@@ -3631,8 +3631,8 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateComputePipelines(VkDevice device, VkPipelin
     VkResult result = get_dispatch_table(ot_device_table_map, device)
                           ->CreateComputePipelines(device, pipelineCache, createInfoCount, pCreateInfos, pAllocator, pPipelines);
     lock.lock();
-    if (result == VK_SUCCESS) {
-        for (uint32_t idx1 = 0; idx1 < createInfoCount; ++idx1) {
+    for (uint32_t idx1 = 0; idx1 < createInfoCount; ++idx1) {
+        if (pPipelines[idx1] != VK_NULL_HANDLE) {
             CreateObject(device, pPipelines[idx1], VK_DEBUG_REPORT_OBJECT_TYPE_PIPELINE_EXT, pAllocator);
         }
     }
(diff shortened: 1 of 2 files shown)


layers/object_tracker.cpp
```diff
@@ -3589,8 +3589,8 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
     VkResult result = get_dispatch_table(ot_device_table_map, device)
                           ->CreateGraphicsPipelines(device, pipelineCache, createInfoCount, pCreateInfos, pAllocator, pPipelines);
     lock.lock();
-    if (result == VK_SUCCESS) {
-        for (uint32_t idx2 = 0; idx2 < createInfoCount; ++idx2) {
+    for (uint32_t idx2 = 0; idx2 < createInfoCount; ++idx2) {
+        if (pPipelines[idx2] != nullptr) {
```
tobine: Compare should be against VK_NULL_HANDLE. Compare w/ nullptr breaks android build.

layers/object_tracker.cpp
```diff
@@ -3631,8 +3631,8 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateComputePipelines(VkDevice device, VkPipelin
     VkResult result = get_dispatch_table(ot_device_table_map, device)
                           ->CreateComputePipelines(device, pipelineCache, createInfoCount, pCreateInfos, pAllocator, pPipelines);
     lock.lock();
-    if (result == VK_SUCCESS) {
-        for (uint32_t idx1 = 0; idx1 < createInfoCount; ++idx1) {
+    for (uint32_t idx1 = 0; idx1 < createInfoCount; ++idx1) {
+        if (pPipelines[idx1] != nullptr) {
```
tobine: Compare should be against VK_NULL_HANDLE. Compare w/ nullptr breaks android build.

layers/unique_objects.cpp
```diff
@@ -428,13 +428,15 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateComputePipelines(VkDevice device, VkPipelin
...
     delete[] local_pCreateInfos;
-    if (VK_SUCCESS == result) {
+    {
         uint64_t unique_id = 0;
         std::lock_guard<std::mutex> lock(global_lock);
         for (uint32_t i = 0; i < createInfoCount; ++i) {
-            unique_id = global_unique_id++;
-            my_device_data->unique_id_mapping[unique_id] = reinterpret_cast<uint64_t &>(pPipelines[i]);
-            pPipelines[i] = reinterpret_cast<VkPipeline &>(unique_id);
+            if (pPipelines[i] != nullptr) {
```
tobine: Compare should be against VK_NULL_HANDLE. Compare w/ nullptr breaks android build.

layers/unique_objects.cpp
```diff
@@ -484,13 +486,15 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
...
```
tobine: Compare should be against VK_NULL_HANDLE. Compare w/ nullptr breaks android build.

yavn:
Thanks. I take it that changes make sense otherwise, right?

tobine:
Yep, changes make sense and look good now.

chrisforbes:
Landed in master a few days ago.

chrisforbes:
This is causing piles of breakage on Linux+Nvidia. We're ending up tracking nonsense objects -- if we skip calling down the chain, nobody bothers to null out the returned pipeline handles.

yavn:
Not sure if I understand. I don't see a place where the call into a driver could be skipped and that would result in using garbage handles. Or do you mean that the driver is returning non-NULLs? The latter case would be a driver bug I suppose.

chrisforbes:
When a layer returns VK_ERROR_VALIDATION_FAILED_EXT instead of calling down the chain, it's currently leaving the returned pipeline handles array as whatever junk the called passed in. If we bail out before calling down, we need to clear all the handles to VK_NULL_HANDLE.

chrisforbes:
You can observe this fairly easily by running the tests.

chrisforbes:
I'll do the followup work to make this correct tomorrow.

PR #1193, opened 2016-11-23
Tobin vu error cleanup

tobine:
A mixture of change here to get the validation error enum in source in line with what's claimed to be implemented in the database file. This makes the stats script happy regarding souce<->database cross-checking.

Split a couple of checks to be more fine grained as they covered multiple unique error enum cases.
Added error enums to a couple of checks and added a test for a case for one of those checks as well.
Removed a check from DB that's not actually in code yet.
Made an exception for a unique check associated with VK_AMD_negative_viewport_height that doesn't have an error with it, but allows for negative height input values.


layers/vk_validation_stats.py
```diff
@@ -129,6 +129,8 @@ class ValidationSource:
     def __init__(self, source_file_list):
         self.source_files = source_file_list
         self.enum_count_dict = {} # dict of enum values to the count of how much they're used
+        # 1790 is a special case that provides an exception when and extension is enabled. No specific error is flagged, but the exception is handled so add it here
```
mark-lunarg: typo

karl-lunarg:
+1

tobine:

PR #1194, opened 2016-11-24
Modernize PV layer data handling

chrisforbes:
This series splits the device data from the instance data, and moves the extension struct and dispatch tables into those layer data structures.

This is mostly reorganization, but there are also some real bugfixes in here, where the layer assumed physical devices had their own layer_data instance [they don't; the dispatch key is shared with the instance] and so fouled up if there were multiple GPUs in the system.


chrisforbes:
Last few commits are more gutting of the swapchain layer. Not very exciting...

chrisforbes:
Pushed to master.

mark-lunarg:
Looks like this commit "layers: move device dispatch table into layer_data in PV  …	bd52880" breaks the windows build.

chrisforbes:
@mark-lunarg should be fixed in master now. I'd let the tools reorder #includes...

PR #1196, opened 2016-11-25
Fixes, that validation layers do work for VK_KHR_display under Linux

FslNopper:
At the moment, if VK_KHR_display extension is used, the validation layers do crash.

The fixes in this pull request do solve the problem.
Tested on i.MX8 using Linux OS.

### DIFF
--- a/layers/unique_objects.cpp
+++ b/layers/unique_objects.cpp
@@ -676,6 +676,30 @@ VKAPI_ATTR VkResult VKAPI_CALL GetDisplayModePropertiesKHR(VkPhysicalDevice phys
     }
     return result;
 }
+
+VKAPI_ATTR VkResult VKAPI_CALL GetDisplayPlaneCapabilitiesKHR(VkPhysicalDevice physicalDevice, VkDisplayModeKHR mode,
+															  uint32_t planeIndex, VkDisplayPlaneCapabilitiesKHR* pCapabilities) {
+    layer_data *dev_data = get_my_data_ptr(get_dispatch_key(physicalDevice), layer_data_map);
+    {
+        std::lock_guard<std::mutex> lock(global_lock);
+        auto it = dev_data->unique_id_mapping.find(reinterpret_cast<uint64_t &>(mode));
+        if (it == dev_data->unique_id_mapping.end())
+        {
+            uint64_t unique_id = global_unique_id++;
+            dev_data->unique_id_mapping[unique_id] = reinterpret_cast<uint64_t &>(mode);
+
+            mode = reinterpret_cast<VkDisplayModeKHR &>(unique_id);
+        }
+        else
+        {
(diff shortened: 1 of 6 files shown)


chrisforbes:
Changes look good -- perhaps rearrange the patches to fix the scripts upfront and then not have WIP states?

FslNopper:
Okay, fixed the commit history.

chrisforbes:
👍 good to go, I think.

mark-lunarg:
LGTM too.

scripts/vulkan.py
```diff
@@ -1375,6 +1375,7 @@ def __init__(self, name, headers, objects, protos, ifdef = None):
 common_exts = [VK_VERSION_1_0,
                VK_KHR_surface,
                VK_KHR_swapchain,
+               VK_KHR_display,
```
mark-lunarg: @FslNopper, I am inclined to just add these two extensions into the common_exts list.  They should compile on Android even if nobody ever uses them, correct? Is there anything in the spec or xml indicating that these are platform-specific?

FslNopper: No, they are not platform specific. Also, agree, if not used, they should not harm and everything should be build-able.
I was keeping the list without the display extension, as I did not know the later impact in the Android build.

layers/unique_objects.cpp
```diff
@@ -676,6 +676,30 @@ VKAPI_ATTR VkResult VKAPI_CALL GetDisplayModePropertiesKHR(VkPhysicalDevice phys
...
     return result;
 }
+
+VKAPI_ATTR VkResult VKAPI_CALL GetDisplayPlaneCapabilitiesKHR(VkPhysicalDevice physicalDevice, VkDisplayModeKHR mode,
+															  uint32_t planeIndex, VkDisplayPlaneCapabilitiesKHR* pCapabilities) {
+    layer_data *dev_data = get_my_data_ptr(get_dispatch_key(physicalDevice), layer_data_map);
+    {
+        std::lock_guard<std::mutex> lock(global_lock);
+        auto it = dev_data->unique_id_mapping.find(reinterpret_cast<uint64_t &>(mode));
+        if (it == dev_data->unique_id_mapping.end())
```
FslNopper: Sorry, also forgot about this.

mark-lunarg:
Pushed changes to Master -- thanks Norbert!

commit 6f2ed666809272002a31b3b4f8adf6581cb41819
Author: Norbert Nopper <Norbert.Nopper@nxp.com>
Date:   Fri Nov 25 07:55:13 2016 +0100

    layers: Complete support for VK_KHR_Display ext

    Fixed using display extension only crash.
    Added vkGetDisplayPlaneCapabilitiesKHR.
    Fixes, that unique objects works for display extension.

    Change-Id: I66727a430b9d55bfa40fdddc884f40c91aa29f11

PR #1200, opened 2016-11-28
layers: Remove bogus CreateSharedSwapchainsKHR impl in swapchain

chrisforbes:
This wasn't even wired up. Parameter validation also now does these
checks.


chrisforbes:
Pushed to master.

PR #1202, opened 2016-11-28
Fix another batch of unexpected errors in tests

chrisforbes:
This is just assorted low-hanging fruit.


chrisforbes:
Pushed to master.

PR #1204, opened 2016-11-28
Core validation: add an errror message when wrong layout is used for clearing

scygan:


layers/core_validation.cpp
```diff
@@ -8301,6 +8301,52 @@ static bool VerifyDestImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node,
...
+                    "doesn't match the current layout %s.",
+                    string_VkImageLayout(destImageLayout), string_VkImageLayout(node.layout));
+            }
+        }
+    }
+
+    if (destImageLayout != VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL) {
+        if (destImageLayout == VK_IMAGE_LAYOUT_GENERAL) {
+            auto image_state = getImageState(dev_data, image);
+            if (image_state->createInfo.tiling != VK_IMAGE_TILING_LINEAR) {
```
stroyan: Why is VK_IMAGE_TILING_LINEAR a special case here?
I don't see any direct connection to it in the spec.

scygan: This is for PERF warning few lines below: typically, it is not OK to use GENERAL layout of TILING_OPTIMAL image, because it is slower than OPTIMAL layouts.

However, it does not matter for LINEAR images, as these have non-opaque storage that is not affected for layout. 

I have copied this from copyImage logic. I have nothing against removing that check, as LINEAR images are (and should be) rare.

layers/core_validation.cpp
```diff
@@ -8301,6 +8301,52 @@ static bool VerifyDestImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node,
...
+                skip_call |=
+                    log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT, 0,
+                    __LINE__, DRAWSTATE_INVALID_IMAGE_LAYOUT, "DS", "Cannot clear an image whose layout is %s and "
+                    "doesn't match the current layout %s.",
+                    string_VkImageLayout(destImageLayout), string_VkImageLayout(node.layout));
+            }
+        }
+    }
+
+    if (destImageLayout != VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL) {
```
tobine: I'd like to see this check hoisted before the check above as not being in one of the 2 defined acceptable layouts is a more fundamental issue than mis-matched layouts and will be easier to fix first.

layers/core_validation.cpp
```diff
@@ -8301,6 +8301,52 @@ static bool VerifyDestImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node,
...
+            VkImageSubresource sub = { range.aspectMask, level, layer };
+            IMAGE_CMD_BUF_LAYOUT_NODE node;
+            if (!FindLayout(cb_node, image, sub, node)) {
+                SetLayout(cb_node, image, sub, IMAGE_CMD_BUF_LAYOUT_NODE(destImageLayout, destImageLayout));
+                continue;
+            }
+            if (node.layout != destImageLayout) {
+                skip_call |=
+                    log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT, 0,
+                    __LINE__, DRAWSTATE_INVALID_IMAGE_LAYOUT, "DS", "Cannot clear an image whose layout is %s and "
```
tobine: For all errors we're now using the unique error enums defined in vk_validation_errors_messages.h. This is error 1085 for Color case and 1100 for DepthStencil case. Enum should be chosen based on calling function and corresponding error message should be appended to the end of the string here.

layers/core_validation.cpp
```diff
@@ -8301,6 +8301,52 @@ static bool VerifyDestImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node,
...
+            auto image_state = getImageState(dev_data, image);
+            if (image_state->createInfo.tiling != VK_IMAGE_TILING_LINEAR) {
+                // LAYOUT_GENERAL is allowed, but may not be performance optimal, flag as perf warning.
+                skip_call |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_PERFORMANCE_WARNING_BIT_EXT,
+                                     (VkDebugReportObjectTypeEXT)0, 0, __LINE__, DRAWSTATE_INVALID_IMAGE_LAYOUT, "DS",
+                                     "Layout for cleared image should be TRANSFER_DST_OPTIMAL instead of GENERAL.");
+            }
+        } else {
+            skip_call |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
+                                 DRAWSTATE_INVALID_IMAGE_LAYOUT, "DS", "Layout for cleared image is %s but can only be "
```
tobine: This is error 1086 for Color case and 1101 for DepthStencil case. Enum should be chosen based on calling function and corresponding error message should be appended to the end of the string here.

layers/core_validation.cpp
```diff
@@ -8301,6 +8301,52 @@ static bool VerifyDestImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node,
     return skip_call;
 }
 
+static bool VerifyClearImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node, VkImage image,
+                                  VkImageSubresourceRange range, VkImageLayout destImageLayout) {
```
tobine: Calling function name should be passed in as last parameter. See various "const char *func_name" examples in core_validation.

layers/core_validation.cpp
```diff
@@ -8858,6 +8904,9 @@ VKAPI_ATTR void VKAPI_CALL CmdClearColorImage(VkCommandBuffer commandBuffer, VkI
     } else {
         assert(0);
     }
+    for (uint32_t i = 0; i < rangeCount; ++i) {
+        skip_call |= VerifyClearImageLayout(dev_data, cb_node, image, pRanges[i], imageLayout);
```
tobine: Pass "vkCmdClearColorImage()" as last param so caller can be clarified in the error msg & so the correct unique error enum can be used.

layers/core_validation.cpp
```diff
@@ -8888,6 +8937,9 @@ CmdClearDepthStencilImage(VkCommandBuffer commandBuffer, VkImage image, VkImageL
...
```
tobine: Pass "vkCmdClearDepthStencilImage()" as last param so caller can be clarified in the error msg & so the correct unique error enum can be used.

tests/layer_validation_tests.cpp
```diff
@@ -10477,6 +10477,26 @@ TEST_F(VkLayerTest, InvalidImageLayout) {
...
+    VkImageSubresourceRange clearRange;
+    clearRange.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
+    clearRange.baseMipLevel = 0;
+    clearRange.baseArrayLayer = 0;
+    clearRange.layerCount = 1;
+    clearRange.levelCount = 1;
+
+    m_errorMonitor->SetDesiredFailureMsg(
+        VK_DEBUG_REPORT_ERROR_BIT_EXT,
+        "Layout for cleared image is VK_IMAGE_LAYOUT_UNDEFINED but can only be TRANSFER_DST_OPTIMAL or GENERAL.");
```
tobine: Once unique error enums are added to checks in validation, VALIDATION_ERROR_01086 can be used here instead of the string.

tests/layer_validation_tests.cpp
```diff
@@ -10477,6 +10477,26 @@ TEST_F(VkLayerTest, InvalidImageLayout) {
...
+    m_commandBuffer->ClearColorImage(dst_image, VK_IMAGE_LAYOUT_UNDEFINED, &clearValue, 1, &clearRange);
+    m_errorMonitor->VerifyFound();
+        m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "Cannot clear an image whose layout is "
```
tobine: Once unique error enums are added to checks in validation, VALIDATION_ERROR_01085 can be used here instead of the string.

scyganINTEL:
> There's a few updates I'd like to see made before this lands which I've noted in review comments.
> Are you ok if we just do the updates on top of your changes? 

Sure, Thanks You for that! (I am going to be mostly unavailable in Dec, so please commit at will).

tobine:
This was merged in #1239

PR #1207, opened 2016-11-29
toolchain: Switch back to tracked revisions on Android

cnorthrop:
This allows us to recreate the entire LVL package from any point in history.
We continue to pull known good revisions from AOSP.


cnorthrop:
cherry-picked to master

PR #1209, opened 2016-11-29
Shader checker: Validate RelaxedPrecision match across interfaces

### LINKED ISSUE #1077: RelaxedPrecision matching in cross-stage interfaces
CTS team found out that validation layers don't currently check that RelaxedPrecision decoration matches on cross-stage interfaces. For example following shader pair violates that rule when compiled using glslang:

Vertex:

```
#version 320 es
layout(location = 0) in highp vec4 a_position;
layout(location = 0) out highp vec4 v_color;
void main (void)
{
    gl_Position = a_position;
    v_color = vec4(1.0, 0.0, 0.0, 1.0);
}
```

Fragment:

```
#version 320 es
layout(location = 0) in mediump vec4 v_color;
layout(location = 0) out mediump vec4 o_color;
void main (void)
{
    o_color = v_color;
}
```

This is of high priority since currently drivers just happily accept this kind of binaries. Validation should be updated to flag mismatching interfaces as soon as possible to motivate apps to fix this.

See Khronos internal CTS issue 518 and spec issue 549.

chrisforbes:
Fixes #1077


chrisforbes:
Pushed to master.

PR #1211, opened 2016-11-30
loader: Fix cast warning in wsi.c

null77:
In 32-bits, this was downcasting a VkSurfaceKHR and comparing it
with NULL. Since VkSurfaceKHR is defined as a uint64_t, it's better
to cast the NULL to a VkSurfaceKHR to avoid a warning.

PTAL


MarkY-LunarG:
There are other instances of this.  I'll fix the rest.

PR #1212, opened 2016-11-30
Tobin gh1165 descriptor oo bindings

### LINKED ISSUE #1165: Descriptor consecutive binding updates are broken if layout has out-of-order bindings
Consider a descriptor set layout that has bindings out of order, such as 1, 2, 0.
All the descriptors are of the same type and meet the requirements for consecutive binding updates.
If you attempt to do a descriptor update starting at binding 0 that rolls over to bindings 1 & 2, you'll get a validation error.

I don't believe this should be the case based on spec language:

`If the dstBinding has fewer than descriptorCount array elements remaining starting from dstArrayElement, then the remainder will be used to update the subsequent binding - dstBinding+1 starting at array element zero. This behavior applies recursively, with the update affecting consecutive bindings as needed to update all descriptorCount descriptors. All consecutive bindings updated via a single VkWriteDescriptorSet structure must have identical descriptorType and stageFlags, and must all either use immutable samplers or must all not use immutable samplers.`

Need to fix validation to correctly handle out of order bindings.

tobine:
Fixes #1165 

Handle descriptor updates for consecutive updates with out-of-order bindings and update a positive test to prove that this works.


chrisforbes:
I'm not sure this is exercised thoroughly enough by that test. Picking just the test change, it still passes.

chrisforbes:
I see it does clean up all the unexpected errors in that test, but a hard failure would be even better.

layers/descriptor_sets.cpp
```diff
@@ -74,6 +77,18 @@ bool cvdescriptorset::DescriptorSetLayout::ValidateCreateInfo(debug_report_data
     return skip;
 }
 
+// Return the number of descriptors for the given binding and all successive bindings
+uint32_t cvdescriptorset::DescriptorSetLayout::GetConsecutiveDescriptorCountFromBinding(uint32_t binding) const {
+    // If binding is invalid we'll return 0
+    uint32_t binding_count = 0;
+    auto bi_itr = binding_to_index_map_.find(binding);
+    while (bi_itr != binding_to_index_map_.end()) {
```
chrisforbes: I was expecting to see this solved by sorting the bindings at descriptor set layout creation time. Help me understand why this approach is better?

tobine: This is simply less invasive to the existing code and behavior. The behavior of the class is well-documented in the header and this update is consistent with the defined behavior.
My initial thought when I filed the issue was to sort bindings, but upon reading the documentation the distinction between bindings and global layout indices is clearly delineated and this fix seems reasonable and consistent with that explanation.

tobine:
The issue with the test still passing even in the absence of the fix appears to be a bug with the errorMonitor. We're expecting success and getting errors and still saying that the test passes... needs further investigation.

layers/descriptor_sets.cpp
```diff
@@ -1169,14 +1195,15 @@ bool cvdescriptorset::DescriptorSet::ValidateWriteUpdate(const debug_report_data
         *error_msg = error_str.str();
         return false;
     }
-    if ((start_idx + update->descriptorCount) > p_layout_->GetTotalDescriptorCount()) {
+    if (update->descriptorCount >
+        (p_layout_->GetConsecutiveDescriptorCountFromBinding(update->dstBinding) - update->dstArrayElement)) {
```
chrisforbes: I'm not sure how this helps. `GetConsecutiveDescriptorCountFromBinding(x)` is equivalent to `GetTotalDescriptorCount() - GetGlobalStartIndexFromBinding(x)` except has to chase up and down a std::map to do it.

tobine: I don't believe this is true. Since bindings can be out-of-order GetGlobalStartIndexFromBinding(x) can currently return various values based on how the layout is defined.
Consider a layout with 10 bindings each of 1 descriptor. Binding 0 could land at global index 0 or 9 or anywhere in-between.

tobine:
Discussed this with Chris and we agreed that this is ok as short-term solution to handle out-of-order bindings. Filed #1227 and #1228 to record some issues that need to be addressed.

karl-lunarg:
FWIW, these commits pass run_all_tests (after a rebase).  And since there are follow-on issues created: :+1:

tobine:

PR #1213, opened 2016-11-30
Karl object tracker tests update2

karl-lunarg:
This pull request reconciles the status of validation layer **tests** involving object tracker.  The steps are:

_1) Fix coding errors in the tests - mostly fix double-deletion of objects, so that object_tracker doesn't emit errors for these._ EDIT: REMOVED
2) Update the database to set the Test column of object tracker tests from Unknown to None.  We know that there are very few tests explicitly written for object tracker, so this is a good start.
3) For the few object tracker tests that do exist, update the DB with the test name and change the test to use the error enum string.


tests/layer_validation_tests.cpp
```diff
@@ -4424,7 +4424,6 @@ TEST_F(VkLayerTest, FramebufferInUseDestroyedSignaled) {
     m_errorMonitor->VerifyFound();
     // Wait for queue to complete so we can safely destroy everything
     vkQueueWaitIdle(m_device->m_queue);
-    vkDestroyFramebuffer(m_device->device(), fb, nullptr);
```
chrisforbes: I'm not sure about this. The DestroyFramebuffer call above /fails/.

karl-lunarg: It is possible that the command buffer execution completes before the attempt to destroy it in flight.

And, object_tracker calls DestroyObject to change the state of the object to destroyed in the tracking data whether or not the device driver succeeds in destroying the FB or not.

karl-lunarg: In the second DestroyFramebuffer call, object_tracker determined that the framebuffer object was already deleted.

So, if CV or some other layer is supposed to detect that the CB is in flight which means the FB cannot be deleted, then something is marking the FB object as deleted anyway.

karl-lunarg: Yes, CV does issue an error because it does detect that the FB is in flight.  CV also skips the call to call down the chain, leaving the FB in place.  All good.

The trouble is that object_tracker, on its own, marks the object as destroyed on the first call, making it think that it is destroyed on the second call.

If we're going to merge object_tracker into CV, then I suppose this could get resolved then, at least more easily.

In that case, we'll have to put this Destroy back (as well as the others I suppose).  If we do that, we'll have to put up with some unexpected errors until the merge.  If we leave the destroys out, then we'll probably see new "need to destroy the xxx object before destroying the device" sorts of messages after the merge.

I'm thinking it is probably better to put back the destroys (essentially revert the 7474065 commit)

Thoughts?

karl-lunarg:
I reverted the "fix double destroy" commit.  (that made the comment string for that commit disappear)
Please review the rest.  Now the changes just consist of updating the test column in the DB and using the error enum strings in the existing tests.

karl-lunarg:
pushed to master

PR #1214, opened 2016-11-30
layers: update valid usage

jeremy-lunarg:
Change-Id: Id67419f6a1f45e791843bb6eee750f8df6c8c889


layers/core_validation.cpp
```diff
@@ -684,30 +684,41 @@ static void clear_cmd_buf_and_mem_references(layer_data *dev_data, const VkComma
...
-                           ", unable to find that object referenced by mem obj 0x%" PRIxLEAST64,
-                    object_type_to_string(type), handle, (uint64_t)mem);
+        if (msgCode == -1) {
+            // TODO: Fix callers with msgCode == -1 to use correct validation checks.
+            return log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, type, handle, __LINE__,
+                           MEMTRACK_INVALID_USAGE_FLAG, "MEM", "While trying to clear mem binding for %s obj 0x%" PRIxLEAST64
+                                                               ", unable to find that object referenced by mem obj 0x%" PRIxLEAST64,
+                           object_type_to_string(type), handle, (uint64_t)mem);
+        } else {
+            char const *valid_usage = (msgCode == -1) ? "" : validation_error_map[msgCode];
```
tobine: This line is not needed as anytime msgCode == -1 the "if" block above would be hit instead of this "else" block.

jeremy-lunarg: Whoops, this was a copy/paste. Thanks, will fix.

jeremy-lunarg: As for your questions. Our initial thrust is to update the "check" status in the database so my response might be short sighted. I'm okay with having multiple layers issue the same error (for now). Are you saying in this case that more than one valid usage is applicable to the condition?

jeremy-lunarg: Okay, I'm killing this PR after our discussion today. Thanks.

PR #1219, opened 2016-12-01
layers:Handle NULL DebugMarker function ptrs

### LINKED ISSUE #1074: Validation layers don't handle NULL func ptrs in dispatch table for extensions
The validation layers contain code like this:

``` cpp
VKAPI_ATTR void VKAPI_CALL CmdDebugMarkerBeginEXT(VkCommandBuffer commandBuffer, VkDebugMarkerMarkerInfoEXT *pMarkerInfo) {
    bool skip_call = false;
    layer_data *my_data = get_my_data_ptr(get_dispatch_key(commandBuffer), layer_data_map);
    assert(my_data != NULL);

    skip_call |= parameter_validation_vkCmdDebugMarkerBeginEXT(my_data->report_data, pMarkerInfo);

    if (!skip_call) {
        get_dispatch_table(pc_device_table_map, commandBuffer)->CmdDebugMarkerBeginEXT(commandBuffer, pMarkerInfo);
    }
}
```

I found it in `parameter_validation.cpp` and `object_tracker.cpp` at least.

Since these are extension functions there is no guarantee the ICD or loader will implement them, so the dispatch chain may end in NULL instead of a terminating function and the layer has to handle that and return without calling onwards.

tobine:
Fixes #1074

Object_tracker and parameter_checker intercept DebugMarker extension
functions which may not be supported by underlying device and
therefore could terminate with a NULL function call.
Updating these layers to handle this case by making sure that their
downstream function ptr is not null before calling down the chain.
If the downstream function ptr is null, considering that VK_SUCCESS
in the layers for functions with VkResult return code.

### DIFF
--- a/layers/object_tracker.cpp
+++ b/layers/object_tracker.cpp
@@ -3216,6 +3216,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDevice(VkPhysicalDevice physicalDevice, con
 
     layer_data *device_data = get_my_data_ptr(get_dispatch_key(*pDevice), layer_data_map);
     device_data->report_data = layer_debug_report_create_device(phy_dev_data->report_data, *pDevice);
+    layer_init_device_dispatch_table(*pDevice, &device_data->dispatch_table, fpGetDeviceProcAddr);
 
     // Add link back to physDev
     device_data->physical_device = physicalDevice;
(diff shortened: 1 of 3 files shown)


courtney-g:
It makes the most sense for extensions that may not be supported in the driver. That's why DebugMarker has issues as it's a validation layer thing, but not supported in drivers.

mark-lunarg:
Ah, @courtney-g, I see now.  Thanks for the explanation.

layers/object_tracker.cpp
```diff
@@ -3706,8 +3715,9 @@ VKAPI_ATTR void VKAPI_CALL CmdDebugMarkerBeginEXT(VkCommandBuffer commandBuffer,
     skip_call |=
         ValidateObject(commandBuffer, commandBuffer, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT, false, VALIDATION_ERROR_02014);
     lock.unlock();
-    if (!skip_call) {
-        get_dispatch_table(ot_device_table_map, commandBuffer)->CmdDebugMarkerBeginEXT(commandBuffer, pMarkerInfo);
+    layer_data *dev_data = get_my_data_ptr(get_dispatch_key(commandBuffer), layer_data_map);
+    if (dev_data->dispatch_table.CmdDebugMarkerBeginEXT) {
```
tobine: Shouldn't be. Will fix.

tobine:

PR #1221, opened 2016-12-01
Tobin kill print funcs

tobine:
I think all of this code can die but will leave PR for a couple days in case anyone still uses/cares about these info messages. I haven't cared about them for over a year.


mark-lunarg:
Hah!  Huge 👍 !

tobine:
Based on overwhelmingly positive response went ahead and just pushed this today.

PR #1223, opened 2016-12-01
Remove layer details doc and associated tests/scripts

mark-lunarg:


karl-lunarg:
Doesn't this doc have some value in tracking the completion state of non-valid-usage validation checks?

mark-lunarg:
At best it communicates that some checks exist for a class of errors -- i.e., "INDEX_OUT_OF_BOUNDS", which doesn't map to any particular check or have any degree of completeness for that class of check.  We can certainly leave it in the repo, though it has only seen a handful of updates in the last couple of months.

karl-lunarg:
OK.  I suppose that if we need to fish out any useful info later on, we can always find it in git.  How would you feel about adding a tag before this commit?
Also, the CI build is now failing because the doc is out of date with the code, so if you commit now, we won't have to fix that.

:+1:

mark-lunarg:
OK, if @tobine is OK with removing it outright, I'll add a tag just before the removal.

karl-lunarg:
Let me go ahead and fix it before you tag and delete it.  Might as well tag a working version.

tobine:
👎  I'd like to think on this one over the weekend. I'm fine getting rid of this stuff but it would be nice to have *some* level of automated checking with the stats script to take it's place. Initially I'm thinking if there are checks claimed in database file that aren't found in source or if there are tests claimed in database file that aren't found in source, then doc validation should fail.

Other thoughts?

tobine:
Please just sit on this PR. I just filed #1234 on myself that I'll work on now to replace this old framework with the stats script. I'll refer to this PR as I'm updating things. I suspect that I'll ultimately supersede this PR with my changes, but if there are any CLs I can pull directly I will.

mark-lunarg:
@tobine, sounds good.  Please lean towards batch files instead of powershell scripts for Windows.

mark-lunarg:
Replaced by MR #1242 .  Closing.

PR #1226, opened 2016-12-02
Fix macOS toolchain

cnorthrop:
Didn't realize this was broken.  It is required for vktrace.

### DIFF
--- a/update_external_sources.sh
+++ b/update_external_sources.sh
@@ -3,7 +3,13 @@
 
 set -e
 
-CURRENT_DIR="$(dirname "$(readlink -f ${BASH_SOURCE[0]})")"
+if [[ $(uname) == "Linux" ]]; then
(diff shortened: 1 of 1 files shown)


update_external_sources.sh
```diff
@@ -3,7 +3,17 @@
 
 set -e
 
-CURRENT_DIR="$(dirname "$(readlink -f ${BASH_SOURCE[0]})")"
+CORE_COUNT=0
```
stroyan: 1 seems like a safe initial value.  I don't see a defined behavior for 'make -j 0'

cnorthrop: Just trying to make it really obvious that something is wrong.  Probably leaving it undefined would work better, although that really buries some systems.

cnorthrop:
Updated to have sensible default CORE_COUNT.

cnorthrop:
cherry-picked to master

PR #1229, opened 2016-12-02
Smoke Android layers

cnorthrop:
This series allows us to package validation layers with Smoke and enable them on Android.

Since Smoke only exists in this repo for validating that layers are working correctly, the layers are packaged into the APK unconditionally.

This change is primarily driven by downstream needs for using Smoke to verify that persistently mapped buffer support is working with vktrace.

The biggest change is a direct port of arg parsing from Hologram that didn't make the original cutoff.

There should be no change in desktop functionality.

### DIFF
--- a/demos/smoke/Main.cpp
+++ b/demos/smoke/Main.cpp
@@ -21,6 +21,11 @@
 
 namespace {
 
+Game *create_game(const std::vector<std::string> &args)
+{
+    return new Smoke(args);
+}
+
 Game *create_game(int argc, char **argv)
 {
     std::vector<std::string> args(argv, argv + argc);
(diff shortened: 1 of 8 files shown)


cnorthrop:
I don't know of any documentation Courtney.  Here's a link to the parsing code though.

https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/demos/smoke/Game.h#L113

I've only tested --validate, but the other arguments should get parsed in the same way.

I'll add a note to BUILD.md showing how to invoke Smoke with --validate (just like Cube).

cnorthrop:
cherry-picked to master

PR #1237, opened 2016-12-06
Tobin pre post refactor

tobine:
Started to refactor a few functions for Pre/Post pattern and ran into RetireWorkOnQueue which still had a bit of validation mingled with state update so I pulled out the validation and split up related functions to do validation prior to calling down chain and only update state after the call down the chain.

Also added flags to the refactored API functions.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5223,6 +5223,8 @@ static void RetireFence(layer_data *dev_data, VkFence fence) {
 }
 
 static bool PreCallValidateWaitForFences(layer_data *dev_data, uint32_t fence_count, const VkFence *fences) {
+    if (dev_data->instance_data->disabled.wait_for_fences)
+        return false;
     bool skip = false;
     for (uint32_t i = 0; i < fence_count; i++) {
         skip |= verifyWaitFenceState(dev_data, fences[i], "vkWaitForFences");
@@ -5264,6 +5266,8 @@ WaitForFences(VkDevice device, uint32_t fenceCount, const VkFence *pFences, VkBo
 }
 
 static bool PreCallValidateGetFenceStatus(layer_data *dev_data, VkFence fence) {
+    if (dev_data->instance_data->disabled.get_fence_state)
+        return false;
     return verifyWaitFenceState(dev_data, fence, "vkGetFenceStatus");
 }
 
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -5280,30 +5306,62 @@ VKAPI_ATTR void VKAPI_CALL GetDeviceQueue(VkDevice device, uint32_t queueFamilyI
...
-    bool skip_call = false;
+    QUEUE_NODE *queue_state = nullptr;
     std::unique_lock<std::mutex> lock(global_lock);
-    auto pQueue = getQueueNode(dev_data, queue);
-    skip_call |= RetireWorkOnQueue(dev_data, pQueue, pQueue->seq + pQueue->submissions.size());
+    bool skip = PreCallValidateQueueWaitIdle(dev_data, queue, &queue_state);
     lock.unlock();
-    if (skip_call)
-        return VK_ERROR_VALIDATION_FAILED_EXT;
+    if (skip) return VK_ERROR_VALIDATION_FAILED_EXT;
```
mark-lunarg: did this get clang-formatted?

layers/core_validation.cpp
```diff
@@ -5302,17 +5302,33 @@ VKAPI_ATTR void VKAPI_CALL GetDeviceQueue(VkDevice device, uint32_t queueFamilyI
     }
 }
 
+static bool PreCallValidateQueueWaitIdle(layer_data *dev_data, VkQueue queue,
+                                         QUEUE_NODE **queue_state) {
+  *queue_state = getQueueNode(dev_data, queue);
```
chrisforbes: This whole commit looks like it got clang-formatted with the default style rather than the LVL style.

EDIT: Oh, I see the next commit fixes it.

layers/core_validation.cpp
```diff
@@ -4582,26 +4582,46 @@ static bool validateAndIncrementResources(layer_data *dev_data, GLOBAL_CB_NODE *
...
-                    skip_call |= log_msg(my_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT,
-                                         VK_DEBUG_REPORT_OBJECT_TYPE_QUERY_POOL_EXT, 0, 0, DRAWSTATE_INVALID_QUERY, "DS",
-                                         "Cannot get query results on queryPool 0x%" PRIx64
-                                         " with index %d which was guarded by unsignaled event 0x%" PRIx64 ".",
-                                         (uint64_t)(queryEventsPair.first.pool), queryEventsPair.first.index, (uint64_t)(event));
+// Note: This function assumes that the global lock is held by the calling thread.
+// For the given queue, verify the queue state up to the given seq number.
+// Currently the only check is to make sure that if there are events to be waited on prior to
+//  a QueryReset, make sure that all such events have been signalled.
+static bool VerifyQueueStateToSeq(layer_data *my_data, QUEUE_NODE *queue, uint64_t seq) {
```
chrisforbes: This no longer validates across semaphore operation pairs -- see tracking of, and then loop over `otherQueueSeqs` at the end of RetireWorkOnQueue. Have you proven that this is never interesting?

layers/core_validation.cpp
tobine: It did but one round of clang-format seemed to use default settings. I updated most of it. I'll update this and see which way correct clang-format settings want it.

layers/core_validation.cpp
```diff
@@ -4582,26 +4582,46 @@ static bool validateAndIncrementResources(layer_data *dev_data, GLOBAL_CB_NODE *
...
+    bool skip = false;
+    auto queue_seq = queue->seq;
+    // local copy of submissions as we're only simulating queue retirement
+    auto submissions = queue->submissions;
```
tobine: Yes, iterating is better.

chrisforbes:
That looks much better, 👍

tobine:
Pushed another CL to address review feedback.

tobine:

PR #1239, opened 2016-12-07
Cort clear dst layout (fixes for #1204)

cdwfs:
Incorporates feedback from @scyganINTEL's original PR #1204, and improves test coverage a bit.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -8178,11 +8178,36 @@ static bool VerifyDestImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node,
 }
 
 static bool VerifyClearImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node, VkImage image, VkImageSubresourceRange range,
-                                   VkImageLayout destImageLayout) {
+                                   VkImageLayout destImageLayout, const char *func_name) {
     bool skip_call = false;
 
     VkImageSubresourceRange resolvedRange = range;
     ResolveRemainingLevelsLayers(dev_data, &resolvedRange, image);
+
+    if (destImageLayout != VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL) {
+        if (destImageLayout == VK_IMAGE_LAYOUT_GENERAL) {
+            auto image_state = getImageState(dev_data, image);
+            if (image_state->createInfo.tiling != VK_IMAGE_TILING_LINEAR) {
+                // LAYOUT_GENERAL is allowed, but may not be performance optimal, flag as perf warning.
+                skip_call |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_PERFORMANCE_WARNING_BIT_EXT,
+                                     (VkDebugReportObjectTypeEXT)0, 0, __LINE__, DRAWSTATE_INVALID_IMAGE_LAYOUT, "DS",
+                                     "Layout for cleared image should be TRANSFER_DST_OPTIMAL instead of GENERAL.");
+            }
+        } else {
+            UNIQUE_VALIDATION_ERROR_CODE error_code = VALIDATION_ERROR_00000;
+            if (strcmp(func_name, "vkCmdClearColorImage()") == 0) {
+                error_code = VALIDATION_ERROR_01086;
+            } else if (strcmp(func_name, "vkCmdClearDepthStencilImage()") == 0) {
+                error_code = VALIDATION_ERROR_01101;
+            } else {
+                assert(0);
+            }
+            skip_call |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
+                                 error_code, "DS", "Layout for cleared image is %s but can only be "
+                                                   "TRANSFER_DST_OPTIMAL or GENERAL. %s",
+                                 string_VkImageLayout(destImageLayout), validation_error_map[error_code]);
+        }
+    }
 
     for (uint32_t levelIdx = 0; levelIdx < range.levelCount; ++levelIdx) {
         uint32_t level = levelIdx + range.baseMipLevel;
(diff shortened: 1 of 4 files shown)


layers/core_validation.cpp
```diff
@@ -8177,6 +8177,69 @@ static bool VerifyDestImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node,
...
+        for (uint32_t layerIdx = 0; layerIdx < range.layerCount; ++layerIdx) {
+            uint32_t layer = layerIdx + range.baseArrayLayer;
+            VkImageSubresource sub = {range.aspectMask, level, layer};
+            IMAGE_CMD_BUF_LAYOUT_NODE node;
+            if (!FindLayout(cb_node, image, sub, node)) {
+                SetLayout(cb_node, image, sub, IMAGE_CMD_BUF_LAYOUT_NODE(destImageLayout, destImageLayout));
+                continue;
+            }
+            if (node.layout != destImageLayout) {
+                UNIQUE_VALIDATION_ERROR_CODE error_code = VALIDATION_ERROR_00000;
```
tobine: Same comment as above here. Default to 01085 & have "else" assert that func_name is "vkCmdClearColorImage()"

layers/core_validation.cpp
```diff
@@ -8177,6 +8177,69 @@ static bool VerifyDestImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node,
...
+                if (strcmp(func_name, "vkCmdClearColorImage()") == 0) {
+                    error_code = VALIDATION_ERROR_01085;
+                } else if (strcmp(func_name, "vkCmdClearDepthStencilImage()") == 0) {
+                    error_code = VALIDATION_ERROR_01100;
+                } else {
+                    assert(0);
+                }
+                skip_call |= log_msg(
+                    dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT, 0,
+                    __LINE__, error_code, "DS", "Cannot clear an image whose layout is %s and "
```
tobine: Also prepend func_name here

layers/core_validation.cpp
```diff
@@ -8177,6 +8177,69 @@ static bool VerifyDestImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node,
     return skip_call;
 }
 
+static bool VerifyClearImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node, VkImage image, VkImageSubresourceRange range,
+                                   VkImageLayout destImageLayout, const char *func_name) {
+    bool skip_call = false;
```
tobine: Can rename this to "skip" to save chars.

layers/core_validation.cpp
```diff
@@ -8177,6 +8177,69 @@ static bool VerifyDestImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node,
     return skip_call;
 }
 
+static bool VerifyClearImageLayout(layer_data *dev_data, GLOBAL_CB_NODE *cb_node, VkImage image, VkImageSubresourceRange range,
+                                   VkImageLayout destImageLayout, const char *func_name) {
```
tobine: For variable names in internal functions (vs. API intercept functions) prefer no camelCase so destImageLayout->dest_image_layout.

tests/layer_validation_tests.cpp
```diff
@@ -10417,12 +10418,55 @@ TEST_F(VkLayerTest, InvalidImageLayout) {
...
     err = vkCreateImage(m_device->device(), &image_create_info, NULL, &dst_image);
     ASSERT_VK_SUCCESS(err);
+    image_create_info.format = VK_FORMAT_D32_SFLOAT;
+    image_create_info.usage |= VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT;
+    err = vkCreateImage(m_device->device(), &image_create_info, NULL, &depth_image);
+    ASSERT_VK_SUCCESS(err);
+
+    // Allocate memory
+    VkMemoryRequirements img_mem_reqs = {};
+    VkMemoryAllocateInfo memAlloc = {};
```
tobine: favor no camelCase for var names

tests/layer_validation_tests.cpp
```diff
@@ -10473,12 +10517,75 @@ TEST_F(VkLayerTest, InvalidImageLayout) {
...
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_01086);
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_01085);
+    m_commandBuffer->ClearColorImage(dst_image, VK_IMAGE_LAYOUT_UNDEFINED, &clearValue, 1, &clearRange);
+    m_errorMonitor->VerifyFound();
+    // Fail due to provided layout not matching actual current layout for color clear.
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_01085);
+    m_commandBuffer->ClearColorImage(dst_image, VK_IMAGE_LAYOUT_GENERAL, &clearValue, 1, &clearRange);
+    m_errorMonitor->VerifyFound();
+    // This one should work correctly
+    m_errorMonitor->ExpectSuccess();
```
tobine: In general we try to isolate positive tests and negative tests completely separate.
Based on that I'd say this last check is not needed. I understand that it is useful, though, as we already have the state setup so it's simple to add this final call and verify that it works. Based on that I'm fine keeping it but will defer to @mark-lunarg or @chrisforbes if they strenuously object.

tests/layer_validation_tests.cpp
```diff
@@ -10473,12 +10517,75 @@ TEST_F(VkLayerTest, InvalidImageLayout) {
...
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_01101);
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_01100);
+    m_commandBuffer->ClearDepthStencilImage(depth_image, VK_IMAGE_LAYOUT_UNDEFINED, &depthClearValue, 1, &clearRange);
+    m_errorMonitor->VerifyFound();
+    // Fail due to provided layout not matching actual current layout for depth clear.
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_01100);
+    m_commandBuffer->ClearDepthStencilImage(depth_image, VK_IMAGE_LAYOUT_GENERAL, &depthClearValue, 1, &clearRange);
+    m_errorMonitor->VerifyFound();
+    // This one should work correctly
+    m_errorMonitor->ExpectSuccess();
```
tobine: Same comment as above. I'm fine with or without this last call.

tests/layer_validation_tests.cpp
mark-lunarg: I don't think it's worth all this extra code being moved to info VkPositiveLayerTest, so I agree with Tobine that it isn't needed (here and below).

chrisforbes: I think this is OK for now.

PR #1241, opened 2016-12-08
Reorganize Linux WSI subsystem selection

Tony-LunarG:
Allow loader and layers to build with all available WSI selections while not forcing support on demos


BUILD.md
```diff
@@ -28,7 +28,7 @@ It should be straightforward to use it on other Linux distros.
```
karl-lunarg: While we are here, might as well update line 26 to just say "on the two most recent Ubuntu LTS versions".

BUILD.md
```diff
@@ -28,7 +28,7 @@ It should be straightforward to use it on other Linux distros.
 
 These packages are needed to build this repository:
 ```
-sudo apt-get install git cmake build-essential bison libx11-dev libxcb1-dev libxkbcommon-dev
+sudo apt-get install git cmake build-essential bison libx11-dev libxcb1-dev libxkbcommon-dev libmirclient-dev libwayland-dev
 ```
```
karl-lunarg: We really should have a new paragraph here describing how to select WSI build options.  Something like:

### WSI Support Build Options
By default, the Vulkan Loader and Validation Layers are built with support for all 4 Vulkan-defined WSI display systems, Xcb, Xlib, Wayland, and Mir.  It is recommended to build these modules with support for these
display systems to maximize their usability across Linux platforms.
If it is necessary to build these modules without support for one of the display systems, the appropriate CMake option of the form BUILD_WSI_xxx_SUPPORT can be set to OFF.   See the top-level CMakeLists.txt file for more info.

BUILD.md
```diff
@@ -169,6 +169,12 @@ Some demos that can be found in the dbuild/demos directory are:
 - cube: a textured spinning cube
 - smoke/smoke: A "smoke" test using a more complex Vulkan demo
 
+You can select which WSI subsystem is used to build the demos using a cmake option called DEMOS_WSI_SELECTION.
+Supported options are XCB, XLIB, WAYLAND, and MIR.  Note that you must build using the corresponding BUILD_WSI_*_SUPPORT enabled at the base repo level.
+For instance, creating a build that will use Xlib to build the demos, your cmake command line might look like:
```
karl-lunarg: Should mention that the default is XCB.

CMakeLists.txt
```diff
@@ -19,6 +19,8 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
     option(BUILD_WSI_XLIB_SUPPORT "Build Xlib WSI support" ON)
     option(BUILD_WSI_WAYLAND_SUPPORT "Build Wayland WSI support" ON)
     option(BUILD_WSI_MIR_SUPPORT "Build Mir WSI support" ON)
+    # Supported values for DEMOS_WSI_SELECTION are XCB, XLIB or WAYLAND.
```
karl-lunarg: If you put this here, then if someone adds Mir support to the demos, someone would have to remember to come back here to fix this.

It would be better to delete this line and add a #error preprocessor directive in a Mir-only code segment that tells the user Mir is not supported by the client.

CMakeLists.txt
```diff
@@ -12,60 +12,40 @@ set(API_NAME "Vulkan" CACHE STRING "API name to use when building")
...
-    #         MIR is stubbed and untested
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    include(FindPkgConfig)
     option(BUILD_WSI_XCB_SUPPORT "Build XCB WSI support" ON)
     option(BUILD_WSI_XLIB_SUPPORT "Build Xlib WSI support" ON)
-    option(BUILD_WSI_WAYLAND_SUPPORT "Build Wayland WSI support" OFF)
-    option(BUILD_WSI_MIR_SUPPORT "Build Mir WSI support" OFF)
+    option(BUILD_WSI_WAYLAND_SUPPORT "Build Wayland WSI support" ON)
+    option(BUILD_WSI_MIR_SUPPORT "Build Mir WSI support" ON)
+    # Supported values for DEMOS_WSI_SELECTION are XCB, XLIB or WAYLAND.
```
karl-lunarg: Same comment as in the demos/CMakeLists.txt file.   Someone will have to remember to come here and change this.  It would be better to indicate non-support with #error at compile time.

layers/CMakeLists.txt
```diff
@@ -1,5 +1,31 @@
...
+    if (BUILD_WSI_WAYLAND_SUPPORT)
+       add_definitions(-DVK_USE_PLATFORM_WAYLAND_KHR)
+    endif()
+
+    if (BUILD_WSI_MIR_SUPPORT)
+        add_definitions(-DVK_USE_PLATFORM_MIR_KHR)
+        include_directories(${MIR_INCLUDE_DIR})
+    endif()
+else()
+    message( FATAL_ERROR "Unrecognized value for DEMOS_WSI_SELECTION: ${DEMOS_WSI_SELECTION}" )
```
karl-lunarg: Doesn't seem like the right message.

BUILD.md
MarkY-LunarG: On Fedora core, the wayland is "wayland-devel".

BUILD.md
```diff
@@ -169,6 +169,12 @@ Some demos that can be found in the dbuild/demos directory are:
...
+
+cmake -H. -Bbuild -DCMAKE_BUILD_TYPE=Debug -DDEMOS_WSI_SELECTION=XLIB
+
```
MarkY-LunarG: Should we add information on how to disable the options?  Especially MIR?

Tony-LunarG:
Committed to master

PR #1242, opened 2016-12-08
Do doc validation based on database file

tobine:
Replaces #1223 

This kills the old doc generation script and details md file and updates the doc validation flow to use the stats script and database file instead.

Currently there are two error cases:

1. If a check is claimed to be implemented in the database but the ID is not found in the source files
2. If a test is claimed in the database but the testname is not found in the source files

Note that I just kept the existing scripts for doc validation as that's the path of least resistance to get these changes in.


mark-lunarg:
This all looks great.  I am unable to run the test or vk_validation_stats.py on Windows -- anybody else having this issue?

tobine:
I did test on Windows but I was using python2.7 on my windows system. I know that with python 3 you'll see an error.

mark-lunarg:
Ah, 'python' = python 3.5.1, 'py' = ptyhon 2.7.11 (cygwin) do not work, but 'python2.7' = 2.7.10 works fine.  Thanks.

tobine:
Replaced by #1248

PR #1245, opened 2016-12-09
MikeS object tracker helper error enums

mikes-lunarg:
Add more unique valid usage enums for object tracker layer.


layers/object_tracker.cpp
```diff
@@ -277,7 +277,9 @@ static void CreateObject(T1 dispatchable_object, T2 object, VkDebugReportObjectT
 }
 
 template <typename T1, typename T2>
-static void DestroyObject(T1 dispatchable_object, T2 object, VkDebugReportObjectTypeEXT object_type, const VkAllocationCallbacks *pAllocator) {
+static void DestroyObject(T1 dispatchable_object, T2 object, VkDebugReportObjectTypeEXT object_type,
```
karl-lunarg: (useless comment:) Right after I said "I thought we were done with Object Tracker", I remembered this DestroyObject section. :-)

mikes-lunarg:
Pushed to master

PR #1246, opened 2016-12-09
layers: GH1244, Fix Compressed BufferImageCopy False Positive Validation Errors

### LINKED ISSUE #1244: image transfer granularity is NOT validated collectly at vkCmdCopyBufferToImage()
Hello guys.

I got a error message from my application and it is below
[DS, 85]: vkCmdCopyBufferToImage(): pRegion[1].bufferOffset (7744) must be an even integer multiple of this command buffer's queue family image transfer granularity width (6).

As I guess, this validation layer only consider about non-compressed image.
In my application case, the texture format is VK_FORMAT_ASTC_6x6_UNORM_BLOCK
Its block size is 6 x 6 and the block size in byte is 16 thus its granularity should be multiple of 16 not 6.

Beneath is spec
VK_FORMAT_ASTC_6x6_UNORM_BLOCK
A four-component, ASTC compressed format where each 128-bit compressed texel block encodes a 6x6 rectangle of unsigned normalized RGBA texel data.

VkBufferImageCopy
If the calling command’s VkImage parameter is a compressed format image:
bufferOffset must be a multiple of the compressed texel block size in bytes

Thanks

mark-lunarg:
CmdCopyBufferToImage and CmdCopyImageToBuffer were validating compressed image attributes against the valid usage conditions for uncompressed images. Added check for compressed images and skipped this validation, and added TODO for future compressed VU addition.  Fixes #1244.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -8472,14 +8472,29 @@ static inline bool ValidateCopyBufferImageTransferGranularityRequirements(layer_
                                                                           const IMAGE_STATE *img, const VkBufferImageCopy *region,
                                                                           const uint32_t i, const char *function) {
     bool skip = false;
-    VkExtent3D granularity = GetScaledItg(dev_data, cb_node, img);
-    skip |= CheckItgSize(dev_data, cb_node, region->bufferOffset, granularity.width, i, function, "bufferOffset");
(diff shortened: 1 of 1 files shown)


mark-lunarg:
Pushed to Master.

PR #1247, opened 2016-12-09
scripts:Make stats script work with python3

tobine:
Just needed parens around the single print so that the
vk_validation_stats.py script will run with python3.

@mark-lunarg this should allow #1242 to run on Windows systems w/ python3 default.


mark-lunarg:
👍! Sweet!  Thanks Tobin.

tobine:
replaced by #1248

PR #1248, opened 2016-12-09
Tobin script stats database updates

tobine:
Replaces #1242 and #1247 

All of these CLs except the last two have already been approved. Wanted to put these all in together, though, so that stats script it correctly functional when it lands on master.

The two new CLs update the database file for new checks that landed but missed the database update and also updates the error enum parsing in the stats script to hit a couple missing cases.

@cdwfs I failed to catch the missing database updates for the checks you added. Note that all unique error enums that get added to the code are reflected in the database file, and the stats script enforces that. Once this PR lands, the enforcement will be automated in the run_all_tests script.


tobine:

PR #1255, opened 2016-12-12
tests: Fix WriteDescriptorSetIntegrityCheck

Tony-LunarG:
Addresses gh1201 Fix WriteDescriptorSetIntegrityCheck test

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -3882,9 +3882,9 @@ TEST_F(VkLayerTest, WriteDescriptorSetIntegrityCheck) {
     VkDescriptorPoolSize ds_type_count[4] = {};
     ds_type_count[0].type = VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER;
     ds_type_count[0].descriptorCount = 1;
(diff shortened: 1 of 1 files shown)


Tony-LunarG:
@chrisforbes I think this addesses gh1201, can you confirm?  It does get rid of the two unexpected validation errors.

tests/layer_validation_tests.cpp
```diff
@@ -3991,12 +3991,14 @@ TEST_F(VkLayerTest, WriteDescriptorSetIntegrityCheck) {
     VkBuffer dyub;
     err = vkCreateBuffer(m_device->device(), &buffCI, NULL, &dyub);
     ASSERT_VK_SUCCESS(err);
-    VkDescriptorBufferInfo buffInfo = {};
-    buffInfo.buffer = dyub;
-    buffInfo.offset = 0;
-    buffInfo.range = 1024;
-
-    descriptor_write.pBufferInfo = &buffInfo;
+    VkDescriptorBufferInfo buffInfo[2] = {};
```
chrisforbes: While you're in here, bind some memory to this buffer? We should be producing a validation error for what you're doing here, but we're not.

Tony-LunarG:
Added memory binding and committed to master

PR #1262, opened 2016-12-13
Disable header-reordering for clang-format

mark-lunarg:
By default, clang-format will sort header files when running the tool, which often breaks compilation. Added the option specified below to the main, tests, and layers version of these files to disable header reordering.

`SortIncludes: false`


mark-lunarg:
Pushed to Master.

renelindsay:
Excellent!  I've been stung by this too. :)

PR #1263, opened 2016-12-13
layers: Update Valid Usage enums in log_msg

mikew-lunarg:
Update Valid Usage enums in core_validation and update
their status in the VU database.
This completes Jira task VL-65.

Change-Id: I51ed327ad65f3a5d1f64bba01ad576c6656f88df


layers/core_validation.cpp
```diff
@@ -12317,9 +12325,10 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDevices(VkInstance instance, uin
         }
         return result;
     } else {
+        // This seems redundant with object_tracker
         log_msg(instance_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_INSTANCE_EXT, 0, __LINE__,
-                DEVLIMITS_INVALID_INSTANCE, "DL", "Invalid instance (0x%" PRIxLEAST64 ") passed into vkEnumeratePhysicalDevices().",
-                (uint64_t)instance);
+                VALIDATION_ERROR_00023, "DL", "Invalid instance (0x%" PRIxLEAST64 ") passed into vkEnumeratePhysicalDevices(). %s",
```
tobine: I don't see a database update for this id, but I see that it's already marked as 'Y' which I guess comes from object_tracker. Since this is such an early and critical thing to check, I'm fine with the redundancy here for now and we can remove it when object_tracker is merged to core_validation.

mikew-lunarg:
Note: deleting a branch too quickly after pushing to master can cause a race with Travis-CI.
Travis-CI will begin its integration builds, but that can fail if the branch is deleted too quickly.

Issue #1228, opened 2016-12-02
errorMonitor not correctly flagging errors on positive tests


tobine:
Noticed this in updating DynamicOffsetWithInactiveBinding for #1212 

On branch tobin_gh1212_positive_test_unexpected_errors I've pushed the updated version of that test without my fix. This causes unexpected errors in the test but even though it's ExpectSuccess() the test does not fail.

Tony-LunarG:
Fixed with 689ea76741

PR #1265, opened 2016-12-13
tests: Fix issue with validation of positive tests

Tony-LunarG:
Fix for GH #1228  m_otherMsgs were not being considered when looking for errors in positive tests


Tony-LunarG:
Committed to master

PR #1266, opened 2016-12-13
layers: Make generators Python 2.7-compatible.

null77:
ANGLE and Chromium are fixed with Python 2.7, so we need to update
the generator to something that is compatible with 2.7 and 3.2.
The unicode handling is different between the two, so:

* use io.open to allow specifying encoding with open
* use unicode literals in a few places (u'txt' instead of 'txt')

Change-Id: I1570a45e302a09bb1f99eabc84f50b2ec599944d

PTAL.. necessary for ANGLE to use the Vulkan layers at all.


mark-lunarg:
@null77, generator.py is pulled in directly from the Khronos github:Vulkan-Docs repository upon each header update so any changes for that file will need originate in Khronos gitlab:vulkan/vulkan or Khronos github:Vulkan-Docs.  We'll take a look at the changes for the other files and add them if they don't cause any issues.

null77:
Doh.. OK thanks. I guess I'll have to figure out how to contribute to the gitlab.

null77:
OK my changes have landed against the internal repo. How often do you roll in the changes into this repo?

mark-lunarg:
Looks like the changes made the 1.0.38 header update, so they should be pulled in probably tomorrow sometime.

null77:
Excellent, thank you!

mark-lunarg:
This PR is superseded by #1287.  Closing.

null77:
Mark, did the changes to the non-official builder python files make it in? Namely, the ones for parameter validation, threading, and unique objects?

mark-lunarg:
@null77, no, forgot all about them!  thanks for the reminder.  I'll pull the in as appropriate.  Sorry!

null77:
Thanks, looks like this made it in. Appreciate it.

mark-lunarg:
Pushed to Master.

PR #1271, opened 2016-12-14
layers: Fix object tracker memory leak

mikes-lunarg:
Upon repeated calls, GetDisplayPlaneSupportedDisplaysKHR will call
CreateObject multiple times with the same VkDisplayKHR handle.  Add a
check in CreateObject for an existing tracking object before allocating
a new one.

Change-Id: I50f634fbe8b7e69edbfeb4e6f0c050b7fccaa6e4


layers/object_tracker.cpp
```diff
@@ -263,17 +263,20 @@ static void CreateObject(T1 dispatchable_object, T2 object, VkDebugReportObjectT
     auto object_handle = handle_value(object);
     bool custom_allocator = pAllocator != nullptr;
 
-    log_msg(instance_data->report_data, VK_DEBUG_REPORT_INFORMATION_BIT_EXT, object_type, object_handle,
-            __LINE__, OBJTRACK_NONE, LayerName, "OBJ[0x%" PRIxLEAST64 "] : CREATE %s object 0x%" PRIxLEAST64, object_track_index++,
-            object_name[object_type], object_handle);
+    if (instance_data->object_map[object_type].find(object_handle) == instance_data->object_map[object_type].end()) {
```
tobine: could use `if (!instance_data->object_map[object_type].count()) {` instead

mikes-lunarg: I like it, easier to read and you don't have to construct an iterator

PR #1272, opened 2016-12-14
scripts: Add quiet option option to vk_helper.py.

null77:
Chromium typically does not allow build steps to output anything
during a build. This option allows for ANGLE to suppress normal
script output during execution.

PTAL.. a nice feature to have for Chromium + ANGLE integration.


null77:
Thanks!

PR #1273, opened 2016-12-14
scripts: Add quiet run option to lvl_genvk.py.

null77:
This helps integration with Chromium, similar to the changes to
vk-helper.py.

PTAL. Would also be nice as for the last script for ANGLE to use with Chromium builds.


mark-lunarg:
@null77, as with generate.py, this script comes as-is from the Github Khronos/Vulkan-Docs repo.  There is a merge step to remove their generators and add our own, but everything else gets clobbered automatically.  Fixing it upstream will take care of these issues.

null77:
Bleh.. I had checked in one Vulkan repo for this script, I guess it comes from a different repo? Oh well, I'll try to merge it upstream, thanks.

mark-lunarg:
Yeah, it's in Vulkan-Docs, but the script is called 'genvk.py' there, in the src/spec directory.  That's where our copy is sourced from.  Thanks man!

PR #1274, opened 2016-12-14
loader: Fix Clang warnings.

null77:
error: equality comparison with extraneous parentheses

error: comparison of array 'layer_prop->functions.str_gipa' equal to a null pointer is always false

PTAL.. some warnings I turned up when building the loader with Clang (on Windows).


loader/vk_loader_platform.h
```diff
@@ -234,7 +234,6 @@ loader_platform_thread_cond_broadcast(loader_platform_thread_cond *pCond) {
 #ifdef __cplusplus
 #include <iostream>
 #include <string>
-using namespace std;
```
MarkY-LunarG: Why are you removing this?  I noticed some of the layers that include this don't re-define "using namespace std;".  Are you only seeing this warning in a certain layer(s)?

mark-lunarg: @MarkY-LunarG, it is standard in LVL to avoid 'using' directives and to explicitly supply namespaces.

mikew-lunarg:
If used at all, "using namespace" should only be in .cpp files, not headers.

On Thu, Dec 15, 2016 at 11:32 AM, Mark Lobodzinski <notifications@github.com

loader/vk_loader_platform.h
null77: @MarkY-LunarG confirming what Mike and Mark said, it is a compiler warning (Clang in this case) to import namespaces in headers. I was seeing this when building layers, but @mark-lunarg asked me to move it into a loader-specific CL.

PR #1276, opened 2016-12-15
layers: Fix Clang warning.

null77:
error: using namespace directive in global context in header
error: comparison of integers of different signs: 'uint32_t' (aka 'unsigned int') and 'int'

PTAL.. some warnings I fixed as I was integrating the Clang build with ANGLE/Chromium.


layers/core_validation.cpp
```diff
@@ -11778,10 +11778,11 @@ static bool PreCallValidateCreateSwapchainKHR(layer_data *dev_data, VkSwapchainC
 
         // Validate pCreateInfo->imageExtent against
         // VkSurfaceCapabilitiesKHR::{current|min|max}ImageExtent:
-        if ((capabilities.currentExtent.width == -1) && ((pCreateInfo->imageExtent.width < capabilities.minImageExtent.width) ||
-                                                         (pCreateInfo->imageExtent.width > capabilities.maxImageExtent.width) ||
-                                                         (pCreateInfo->imageExtent.height < capabilities.minImageExtent.height) ||
-                                                         (pCreateInfo->imageExtent.height > capabilities.maxImageExtent.height))) {
+        if ((capabilities.currentExtent.width == std::numeric_limits<uint32_t>::max()) &&
```
mark-lunarg: This gives me the shivers.  Using -1 is just as bad, as the spec defines the value specifically as '0xFFFFFFFF'.  Can we use VK_WHOLE_SIZE or one of the other header defines instead?  If not, maybe make this a descriptive static const someplace and use it here and below?

loader/vk_loader_platform.h
```diff
@@ -234,7 +234,6 @@ loader_platform_thread_cond_broadcast(loader_platform_thread_cond *pCond) {
 #ifdef __cplusplus
 #include <iostream>
 #include <string>
-using namespace std;
 #endif // __cplusplus
```
mark-lunarg: Break this out into a separate 'loader' commit and change the other one to be just 'layers'.

null77: Done.. I put this in #1274.

null77:
Please take another look.

mark-lunarg:
Pushed to Master.

null77:
Thanks!

PR #1279, opened 2016-12-15
Remove redundant image layer checks, remove enum/struct_validator_helpers

mark-lunarg:
struct_validate_helper was unused, enum_validate_helper was only used for redundant checks that have been removed/replaced.  Part of switching vk-helper stuff over to being vk.xml-based.


mark-lunarg:
@tobine, any issues on the farm that need to be taken care of?

PR #1284, opened 2016-12-19
scripts:Update unique error enums as of 12/19/16

tobine:
Six updates and some new valid usages.

Cmd to do the update was:

python spec.py -update -remap 127-132:128-141:1465-1263:2168-1646,2:2373-1965


tobine:
@mark-lunarg good idea, opened up #1285 to track this.

tobine:

PR #1295, opened 2016-12-21
Add missing <string> dependency

mikejurka:


mark-lunarg:
Pushed to Master, thanks!

mikejurka:
Thanks. FYI I just signed the CLA, but just for some reason the CLAssistant is not updating the field above.

PR #1296, opened 2016-12-21
Use Python instead of greadlink to get full path on macOS

mikejurka:
Python is built-in on Macs; this avoids an unnecessary dependency on greadlink


tobine:
LGTM :+1:
Please sign CLA

mikejurka:
CLA signed, thanks!

Issue #1282, opened 2016-12-17
Layers: performance warning about unused vertex buffers, but no way to unbind them?


cdwfs:
Context: I have a VkRenderPass with two subpasses; the first renders some geometry, and the second does some simple post-processing. There's only one VkPipeline used in the second pass, and it has no vertex attribute bindings defined; the only draw call is a full-screen triangle which generates its VS output procedurally.

Today, I enabled PERFORMANCE_WARNING messages in the validation layers, and started receiving the following warning inside the second subpass's vkCmdDraw():
`PERFORMANCE WARNING: [DS] Code 24 : Vertex buffers are bound to command buffer (
0x5b00700) but no vertex buffers are attached to this Pipeline State Object (0x3
0).`
It looks like it's complaining that the vertex buffer bindings from a VkPipeline of the previous subpass are still active, even though the post-processing subpass isn't using them. Assuming that's accurate, what's the right way to unbind the buffers to avoid this warning? Calling vkCmdBindVertexBuffers with pBindings=NULL is an error; so is setting pBindings[0]=VK_NULL_HANDLE, or bindingCount=0 (though I'd expect that to be a no-op, if anything). Am I missing something, or is this a spurious warning?

chrisforbes:
Yeah, this warning logic falls apart for more complex recordings. @tobine, thoughts?

tobine:
Gave this some thought. Tried to think of a good way to update the message to allow for custom cases like this but don't have a great idea of wording. If anyone has an idea we can certainly tweak the message.
For now, I'm just thinking ignore it as you know what you're doing. In general for warnings and perf warnings, if you know what you're doing, you can ignore them. They're primarily intended for novice users. Here's the definition from spec for perf warning which allows for some leeway in what's reported:

`VK_DEBUG_REPORT_PERFORMANCE_WARNING_BIT_EXT indicates a potentially non-optimal use of Vulkan.`

cdwfs:
I don't mind performance warnings being ignorable, though if I'm regularly ignoring 90% of the warnings, I'm likely to start ignoring 100% of them. I don't plan to leave PERFORMANCE_WARNING enabled all the time (as I currently do with ERROR and WARNING during development); I'll save it for occasional spot checks.

[...]

PR #1299, opened 2016-12-21
Tobin draw dispatch pre post refactor

tobine:
Primary change here are to break out all of the CmdDraw* and CmdDispatch* functions to use the Pre/Post pattern.
There's some ugliness here with the active set & binding data structures now being declared at the top level of each function, set in the PreCallValidate function and then used in the PostCallRecord function. I plan to look at this a bit further and hopefully clean it up in a future commit. May make sense to store the data structs off of cmd buffer state struct so we're not passing them around.

Anyway, functionally this change seems to be working.

Also did some renaming and threw out some print code.

Finally snuck in one CL with a tweak for a perf warning related to #1282 as the change was touching code that was refactored in this PR as well.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -2853,7 +2853,7 @@ cvdescriptorset::DescriptorSet *getSetNode(const layer_data *my_data, VkDescript
 //     descriptor update must not overflow the size of its buffer being updated
 //  2. Grow updateImages for given pCB to include any bound STORAGE_IMAGE descriptor images
 //  3. Grow updateBuffers for pCB to include buffers from STORAGE*_BUFFER descriptor buffers
-static bool ValidateAndUpdateDrawtimeDescriptorState(
+static bool ValidateDrawtimeDescriptorState(
     layer_data *dev_data, GLOBAL_CB_NODE *pCB,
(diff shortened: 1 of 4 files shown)


layers/core_validation.cpp
```diff
@@ -3153,6 +3153,9 @@ static void UpdateDrawState(
         // For given active slots record updated images & buffers
         UpdateDrawtimeDescriptorState(my_data, cb_state, *active_set_binding_pairs);
     }
+    if (pPipe->vertexBindingDescriptions.size() > 0) {
+        cb_state->vtx_buffer_used = true;
```
chrisforbes: s/vtx/vertex/. We'd just got rid of all the instances of this.

layers/core_validation.cpp
```diff
@@ -8032,6 +8036,21 @@ static void MarkStoreImagesAndBuffersAsWritten(layer_data *dev_data, GLOBAL_CB_N
     }
 }
 
+static bool PreCallValidateCmdDraw(
+    layer_data *dev_data, VkCommandBuffer cmd_buffer, bool indexed, VkPipelineBindPoint bind_point, GLOBAL_CB_NODE **cb_state,
+    vector<std::tuple<cvdescriptorset::DescriptorSet *, std::map<uint32_t, descriptor_req>, std::vector<uint32_t> const *>>
```
chrisforbes: This type is absurd :(

tobine:

PR #1301, opened 2016-12-22
loader: guard gcc feature appropriately

fjhenigman:
Macro __GNUC__ should be used to guard gcc-specific code.
Also exclude clang because though it defines __GNUC__ it doesn't accept
this particular gcc feature.


MarkY-LunarG:
Tweaked commit message to fit within length.  But commited with 51b57689ecd37660f8bf875d8e8a192a12080f2f.

PR #1305, opened 2016-12-22
Tobin unique id stat updates

tobine:
Various unique error id updates:

- Added ids for new memory bound valid usage cases.
- Updated stats script to print info on duplicate location and line numbers
- Added exception mechanism to allow for some duplicates
- Fixed up a few other unique error ids


layers/descriptor_sets.cpp
```diff
@@ -1286,16 +1286,10 @@ bool cvdescriptorset::DescriptorSet::ValidateBufferUsage(BUFFER_STATE const *buf
...
     // First make sure that buffer is valid
     auto buffer_node = getBufferState(device_data_, buffer_info->buffer);
-    if (!buffer_node) {
-        std::stringstream error_str;
-        error_str << "Invalid VkBuffer: " << buffer_info->buffer;
-        *error_msg = error_str.str();
-        return false;
-    }
+    // Any invalid buffer should already be caught by object_tracker
+    assert(buffer_node);
```
mikew-lunarg: As policy, should layers use assert() and abnormally terminate?

tobine:

PR #1308, opened 2016-12-23
loader: make one variable plural and clean up a comment’s wrapping

mikejurka:


MarkY-LunarG:
Commit message first line should be less than 50 characters.  I think the next lines can be 70.  I'll merge it once you clean up the commit, perhaps something like:
```
   loader: clean up comment and variable

   Make one variable's name plural to clarify that it contains more than one
   layer.  Also, clean up a comment that wrapped prematurely to the next line.
```

I don't know if I can change the merge request titles from inside GitHub which is why I'm requesting you do it.  If it's too much, I can change it next week.

mikejurka:
Pushed a new commit with updated commit message.

MarkY-LunarG:
Thanks for cleaning it up!

PR #1320, opened 2016-12-28
tests:Clean up memory on early exit

tobine:
TestAliasedMemoryTracking has an early exit case that was failing to
free allocated memory. This was leading to a fail on some platforms.


tobine:
This was still failing on some platforms so force pushed CL with a couple more tweaks that now has this passing on our farm.

PR #1326, opened 2016-12-29
Move struct_size_helper files to use vk.xml-based codegen

mark-lunarg:
Added option-blocks for the two files, added codegen source to the existing helper-file-generator, removed references to the old codegen/source.

Safe_struct is next (and last!).


mark-lunarg:
@cnorthrop, these files are only really used in the VulkanTools repo, and it looks like there is a post-build step to copy them (though they were built as a side-effect) -- can you double-check the android-build changes with that in mind?

tobine:
I ran it on our LVL farm and it looks good. Didn't check VulkanTools android.

mark-lunarg:
Pushed to Master.

Issue #950, opened 2016-09-18
tests: Consider removing implicit BeginRenderPass on BeginCommandBuffer


chrisforbes:
This happens if the right set of initialization has happened, and doesn't otherwise. Some tests explicitly call EndRenderPass immediately after, since they need the state but don't want the render pass instance.

It would be clearer to just explicitly begin the renderpass.

lentinem:
Can you elaborate on this? End cannot be called if Begin is not called.

tobine:
@lentinem this is a suggestion for improving the validation test framework. Currently in the test framework if you call BeginCommandBuffer() it creates an implicit renderpass. @chrisforbes is suggesting that the framework stop doing that.

Tony-LunarG:
Fixed with 7dc45c3

PR #1329, opened 2017-01-02
Fix GH #950 Make explicit calls to vkCmdBeginRenderPass

Tony-LunarG:
Get rid of the BeginCommandBuffer function whose side effect
can be to call vkCmdBeginRenderPass and make the calls directly
from the tests.  Make the analgous changes for ending command
buffers and renderpasses.

Also gets rid of a couple of unexpected errors.


tests/layer_validation_tests.cpp
```diff
@@ -502,7 +469,10 @@ void VkLayerTest::VKTriangleTest(const char *vertShaderText, const char *fragSha
     descriptorSet.AppendBuffer(VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, constantBuffer);
 
     ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
-    ASSERT_VK_SUCCESS(BeginCommandBuffer());
+    m_commandBuffer->BeginCommandBuffer();
+    if (m_renderPass) {
```
chrisforbes: VKTriangleTest is /always/ going to have a renderpass, isn't it?

Tony-LunarG:
Committed to master

PR #1332, opened 2017-01-03
Update Android travis

cnorthrop:
Changes to improve build testing, mostly targeting Android:
- Move to earliest Android version that supports Vulkan (23)
- Switch to latest NDK
- Cache external build directories.  This improves build time from ~20 minutes to ~5 minutes.
- Return early is any test fails


cnorthrop:
cherry-picked to master - Thanks!

PR #1333, opened 2017-01-03
Remove safe_struct extension special cases

mark-lunarg:
During addition of the `Display `and `DisplayMode `extensions, the unique_objects code was incorrectly cut-pasted resulting in missing safe_struct helper errors.  These in turn were hacked around by adding special cases to the vk_helper safe_struct codegen.  Removed both.


mark-lunarg:
Cherry-picked to Master.

PR #1341, opened 2017-01-05
tests: Fix ImageLayerViewTests

renelindsay:
Change-Id: I64cf95cd9f7357879cfe72e9f856e0e5baaefb2d

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -14703,7 +14703,7 @@ TEST_F(VkLayerTest, ImageLayerViewTests) {
     // VIEW_CREATE_ERROR
     VkImageCreateInfo mutImgInfo = image.create_info();
     VkImage mutImage;
-    mutImgInfo.format = VK_FORMAT_R8_UINT;
+    mutImgInfo.format = VK_FORMAT_R8G8B8A8_UINT;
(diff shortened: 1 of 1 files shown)


renelindsay:
Hi Chris,

Are you saying that changing the color format from 'VK_FORMAT_R8_UINT' to
'VK_FORMAT_R8G8B8A8_UINT' is causing a memory-not-bound error?
What hardware are you running on?

Rene

chrisforbes:
I'm saying that the subtest you're modifying doesn't do what it claims to do. It's checking for memory-not-bound, but was set up to check for something else. If changing the format here ends up being the right thing, so be it, but make everything from :14697 onward in that test /consistent/ about what it's trying to do.

If you /do/ decide to change this format, also change the format in the nearby assert.

jeremy-lunarg:
Judging from Tobin's TODO above the suspicious SetDesiredFailureMsg, it appears to be a work around for a framework deficiency. Indeed, if you remove that line, the test generates two errors. One of them is regarding usage without bound memory.

chrisforbes:
Can we resolve that TODO (or just manually bind some memory to the image, but I know that's tedious noise) so this /just/ hits the intended case? We /do/ need to exercise the format class compatiblity check.

jeremy-lunarg:
I've recommended that Rene revert Tobin's change for now. It will still trigger an unexpected error, but at least it will exercise the validation layers as originally intended.

renelindsay:
Well spotted, thanks Chris.

PR #1353, opened 2017-01-10
layers: Reset Valid Usage database check_implemented status

mikew-lunarg:
Following the audit of existing Valid Usage enums in the LVL codebase,
update the VU database so that any remaining check_implemented == "U"
(Unknown) status codes now become "N" (Not implemented).

"U" is no longer permitted as a value for the check_implemented status.

Tweak database stats script to emit an error message if the
check_implemented value is anything other than "Y" or "N".

Change-Id: I5d214ce4d35a0a8d6d9b16656ef353b2dd134a2c


layers/vk_validation_stats.py
```diff
@@ -106,6 +106,8 @@ def read(self):
                 api = db_line[3]
                 error_str = db_line[4]
                 note = db_line[5]
+                if implemented not in ['Y', 'N']:
+                    print("ERROR: %s has invalid 'implemented' status of '%s'; must be 'Y' or 'N'" % (error_enum, implemented))
```
tobine: This should be printed out in red and cause the script to exit with a non-zero value.

mikew-lunarg:
Will do.  In the meantime, I'll create a separate PR with just the database updates.  The script changes will come separately.

PR #1354, opened 2017-01-10
layers: Transition each aspect individually in TransitionImageLayouts

aejsmith:
This fixes the possibility of spurious validation errors like the following, when attempting to transition multiple aspects of an image after only a subset of those aspects have previously been used in the command buffer:

```
Cannot query for VkImage 0x599 layout when combined aspect mask 6 has
multiple initial layout types: VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL and
VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL
```

The specific case where this bug was encountered was where both the depth and stencil aspects of an image were initially in the TRANSFER_DST_OPTIMAL layout. The first command in the command buffer to reference the image was a vkCmdClearDepthStencilImage on only the depth aspect of the image, followed later by a vkCmdPipelineBarrier to transition both aspects to the DEPTH_STENCIL_ATTACHMENT_OPTIMAL layout.

Since TransitionImageLayouts tries to look up the initial state for all aspects at the same time, it was picking up the state based on the depth aspect because of the previous use of that (which had both initialLayout and layout set to TRANSFER_DST_OPTIMAL), and then calling SetLayout for all aspects with only the new layout. Since the stencil aspect didn't have any currently recorded state, it was being added with both initialLayout and layout set to the new layout, causing a mismatch between the initialLayout for the two aspects and therefore the spurious error above.

Fix by updating the state for each aspect individually.


aejsmith:
Just fixed a mistake in TransitionImageAspectLayout - the first 2 returns should have returned false (*don't* skip the call...)

tobine:
@aejsmith yep, was just about to comment on that when I saw you had just changed it. Testing on our farm now. This looks like a great fix, though, thanks!

PR #1357, opened 2017-01-10
layers: Reset Valid Usage database check status

mikew-lunarg:
Following the audit of existing Valid Usage enums in the LVL codebase,
we can claim that there are no Valid Usage checks in an unknown
state; the existing VU checks are either implemented or not.

This commit resets the check_implemented flags in the VU database,
such that any remaining "U" (Unknown) entries now become "N" (Not
implemented).

Going forward, the check_implemented status flag must be either "Y"
or "N"; "U" is no longer permitted.

Change-Id: Iaf1807cadf02d57e3fa5676108e290d3bab9b271


tobine:
@mikew-lunarg Just to clarify, this initially sets a ton of checks that are actually handled by parameter_validation to "N." So "N" doesn't mean that the check is not implemented, it means that either the check is not implemented, or it's implemented but the unique error id hasn't been added to the source yet.

Thinking out loud of if there's any benefit to trying to distinguish such checks. Once error ids are integrated into the spec, the parameter_validation checks will get added to the generated code and the database can be updated for all of them.

Either way the next step is going to involve going through "N" checks and doing the work to make them "Y". Checks that parameter_validation handles can be skipped until spec integration is complete. Not sure there's a big issue with having checks marked "N" that may be actually implemented, just want to make sure all of the key people involved understand the distinction.

karl-lunarg:
@tobine  Yes, I think all your clarifications are correct and were well-understood by those involved.

The recent work up to now to add unique error ids to all existing checks was meant to minimize the number of VU tests marked as N that were already implemented.  That is, all implemented VU checks would have unique ids and would be true Y's, and the N's would be ones where the checks are not implemented.

The big exception are the PV codegened checks, and we agreed to accept and understand the "drop" in the apparent implementation %age, which would be repaired all at once when the spec integration for the PV codegen is complete.

So outside of the PV codegen checks, anything marked N are now clear targets for what needs to be implemented.

There may be some value in introducing a "P" state for the PV codegened checks. But if the spec integration is imminent, we could just live with the short term disconnect.

Do you think it would be better to "distinguish such checks" by marking them differently at this point?

tobine:
@karl-lunarg I think it's best to just live with all of the PV checks as "N" for a short while. The only people who would really benefit from marking them "P" for a short while should be the same people who understand that the checks are already implemented in PV. Not really worth the effort. Just wanted to make sure we're all on the same page.

PR #1358, opened 2017-01-10
build: Update travis config to catch build errors on intermediate steps

cnorthrop:
This breaks up the script blocks used to build.  Travis is only reporting the final result from the "- if" script blocks, masking failures downstream.  Another approach would be && command sequences together, but this is more readable.


cnorthrop:
cherry-picked to master

PR #1360, opened 2017-01-10
Add vk_safe_struct.* to helper_file_generator, kill vulkan.py, vk_helper.py

mark-lunarg:
Added generation of safe_struct files to the helper file generator and removed header-parsing scripts from the repo -- vulkan.py will move into the VulkanTools repo.


mark-lunarg:
Pushed to Master.

Issue #1355, opened 2017-01-10
Layers: vkAcquireNextImageKHR throws Access Violation Exception


ThomasL81:
Hello,
I was troubleshooting my application and came across an Access Violation Exception reading address 0x0 on the execution of vkAcquireNextImageKHR.

After building/debugging this repo's source and testing with the provided layers, I came to the conclusion I forgot to call GetSwapchainImagesKHR after creating the Swapchain object. I have the impression these images are cached by the core_validation layer. In the code it seems to be assumed that the images are there.

Wouldn't it be a better solution to post ERROR feedback when images where not found?

The statement throwing the assertion is the following (the vector assertion going out of bounds):
auto image = swapchain_data->images[*pImageIndex];

Contextual:
```
        // Mark the image as acquired.
        auto image = swapchain_data->images[*pImageIndex];
        auto image_state = getImageState(dev_data, image);
        image_state->acquired = true;
    }
    lock.unlock();
```
The debugger asserts on this piece of code in core_validation.cpp

My partial stacktrace on Visual Studio 2015 is:

` 	msvcp140d.dll!00007ffd1b670806()	Unknown
>	VkLayer_core_validation.dll!std::vector<VkImage_T * __ptr64,std::allocator<VkImage_T * __ptr64> >::operator[](unsigned __int64 _Pos) Line 1235	C++
 	VkLayer_core_validation.dll!core_validation::AcquireNextImageKHR(VkDevice_T * device, VkSwapchainKHR_T * swapchain, unsigned __int64 timeout, VkSemaphore_T * semaphore, VkFence_T * fence, unsigned int * pImageIndex) Line 12374	C++
 	VkLayer_parameter_validation.dll!parameter_validation::AcquireNextImageKHR(VkDevice_T * device, VkSwapchainKHR_T * swapchain, unsigned __int64 timeout, VkSemaphore_T * semaphore, VkFence_T * fence, unsigned int * pImageIndex) Line 4888	C++
 	vulkan-1.dll!vkAcquireNextImageKHR(VkDevice_T * device, VkSwapchainKHR_T * swapchain, unsigned __int64 timeout, VkSemaphore_T * semaphore, VkFence_T * fence, unsigned int * pImageIndex) Line 503	C
[...]

PR #1361, opened 2017-01-10
Github issue#1355 - Layers: vkAcquireNextImageKHR throws Access Viola…

ThomasL81:
…tion Exception. Added a new error enum value and created an error check in vkAcquireNextImageKHR to avoid index out of bounds exception (thrown in core_validation) within the application.

Pull Request is linked to issue #1355


mark-lunarg:
Updated commit message and source for style, changed to a Warning, and pushed to Master.

PR #1366, opened 2017-01-12
Infrastructure for Image layer merge, code segregation

mark-lunarg:
Add image_validation.cpp and image_validation.h to the repo -- these'll hold the image-related validation source.  Also, some reshuffling of some stuff formerly included in core_validation.cpp that needed to move into the header to make things visible to other modules, mostly related to the layer_data structures.  Also, moved a small portion of the CreateImage validation code into the new module as a proof-of-concept, for feedback and improvement.

The idea is that much of the current CV image-related validation can go into this new module, along with everything else that's in the image layer currently.  The long-term goal is to have CV contain only the top-level intercept functions, with all the other stuff residing elsewhere.

### DIFF
--- a/layers/image_validation.h
+++ b/layers/image_validation.h
@@ -1,0 +1,49 @@
+/* Copyright (c) 2015-2016 The Khronos Group Inc.
+ * Copyright (c) 2015-2016 Valve Corporation
+ * Copyright (c) 2015-2016 LunarG, Inc.
+ * Copyright (C) 2015-2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Mark Lobodzinski <mark@lunarg.com>
+ */
+#ifndef CORE_VALIDATION_IMAGE_VALIDATION_H_
+#define CORE_VALIDATION_IMAGE_VALIDATION_H_
+
+#include "core_validation_error_enums.h"
+#include "vk_validation_error_messages.h"
+#include "core_validation_types.h"
+#include "vk_layer_logging.h"
+#include "vk_layer_utils.h"
+#include "vk_safe_struct.h"
+#include "vulkan/vk_layer.h"
+#include <map>
+#include <memory>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+#endif // CORE_VALIDATION_IMAGE_VALIDATION_H_
--- a/layers/image_validation.h
+++ b/layers/image_validation.h
@@ -33,17 +33,9 @@
 #include <unordered_set>
 #include <vector>
 
-
-
-
-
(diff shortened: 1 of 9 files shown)


mark-lunarg:
We could rename the new module and have it contain image and buffer related stuff, or make it more general still.  Could eventually contain the create/destroy image/buffer, imageview/bufferview, imageusage/bufferusage, ImageLayouts, image/buffer memorybinding, and maybe even the CmdCopyImage, CmdBlitImage, CmdCopyImageToBuffer, CmdCopyBufferToImage, CmdUpdateBuffer, CmdFillBuffer, CmdClear*, CmdResolveImage, SparseImage* and SparseBuffer* related validation code.  Comments?

mark-lunarg:
Refactored the change to match the existing layer data access methods.

layers/image_validation.h
```diff
@@ -0,0 +1,41 @@
...
+#include <map>
+#include <memory>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+VK_LAYER_EXPORT void PostCallRecordCreateImage(std::unordered_map<VkImage, std::unique_ptr<IMAGE_STATE>> *imageMap,
+    std::unordered_map<VkImage, std::vector<ImageSubresourcePair>> *imageSubresourceMap,
+    std::unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE> *imageLayoutMap,
+    const VkImageCreateInfo *pCreateInfo, VkImage *pImage);
```
jeremy-lunarg: Can we pass by reference? Might be cleaner to consume?

layers/image_validation.h
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Mark Lobodzinski <mark@lunarg.com>
+ */
+#ifndef CORE_VALIDATION_IMAGE_VALIDATION_H_
+#define CORE_VALIDATION_IMAGE_VALIDATION_H_
+
+#include "core_validation_error_enums.h"
```
mikew-lunarg: can this list of #includes be culled, to those needed to support the single declaration/

layers/image_validation.h
tobine: Can only pass by reference if 'const' as defined in [style guide](https://google.github.io/styleguide/cppguide.html#Reference_Arguments)

mark-lunarg:
Implemented review feedback, changed names from image_validation* to buffer_validation.*, tested, and pushed to Master.

tobine:
This is ok, esp for initial framework example.

I don't like passing 3 separate members of dev_data as separate params.
One alternative to consider:
Only pass dev_data as the initial parameter. Define helper functions to retrieve ptrs to the maps of interest off of dev_data. Those helpers can initially live in core_validation.cpp. Prototypes for the helpers can be defined in image_validation.cpp. Long-term the helper definitions can migrate to a separate file that defines all such helpers and handles all the calls require knowledge of the internals of dev_data.

layers/image_validation.h
jeremy-lunarg: I think we should be careful here. We just chose to adopt this style (out of logistics; not because it was what we actually chose to begin with). There will be other situations where the code won't abide by the style guide (e.g. exceptions) and I'm okay with that. I think the main thing we should take from the style guide is naming conventions.

chrisforbes: I strongly prefer passing by ptr here rather than by ref-to-nonconst.

chrisforbes: Why is this VK_LAYER_EXPORT? You're not wanting to expose it across a shared object boundary.

jeremy-lunarg: That's fine. And I disagree. That's fine too. I just think we should be flexible here.

PR #1369, opened 2017-01-12
Tobin spec hack implicit count

tobine:
This updates the spec script to add "implicit" note to database for all implicit checks.
It also updates the stats script to count the non-implemented implicit checks, report that number, and report a potential total coverage number if all non-implemented implicit checks are handled in parameter_validation.

Currently >300 implicit checks are already implemented (probably all in object_tracker) so the best potential combined number of checks covered right now is 1699/2480 or ~69%.


karl-lunarg:
> Currently >300 implicit checks are already implemented (probably all in object_tracker) so the best potential combined number of checks covered right now is 1699/2480 or ~69%.

I thought we were kicking around a number over 1000 and they were in parameter checking instead of object tracker.

layers/vk_validation_stats.py
```diff
@@ -116,6 +117,8 @@ def read(self):
                 # Now build custom data structs
                 if 'Y' == implemented:
                     self.db_implemented_enums.append(error_enum)
+                elif 'implicit' in note: # only make not of non-implemented implicit checks
```
mikew-lunarg: s/not/note/

tobine:
@karl-lunarg There were 1347 total implicit checks. 1007 are not marked as implemented but most of those should be implemented in parameter_validation. The other 340 are already implemented, probably most (all?) in object_tracker.

Issue #1330, opened 2017-01-03
Cube example uniform buffer update synchronization


tksuoran:
It is not clear to me how the cube example synchronizes uniform buffer updates from CPU and reads from GPU. Shouldn't demo_update_data_buffer() wait for last previous command that reads from uniforms to be completed before it can go modify uniform memory? And once there is such synchronization, the same memory probably should not be used in subsequent draws, otherwise CPU and GPU execution gets effectively serialized.

Tony-LunarG:
Fixed with c44ffcd5

PR #1370, opened 2017-01-13
Demos: Cube use uniform buffer per swapchain image

Tony-LunarG:
GH #1330 points out that there is no synchronization around updating the uniform buffer in cube.  This fixes that by adding a uniform buffer per swapchain image and synchronizes mapped access using fences.  It also adds triple buffering and removes unneeded code to ignore missing fence errors.


Tony-LunarG:
Commited to master

PR #1381, opened 2017-01-17
layers: Fix vkCmdCopyImage intersection check for array layer/mip level

aejsmith:
RegionIntersects would return true if the source and destination array layers and mip level did *not* overlap, resulting in a spurious validation error when trying to copy between different array layers or mip levels of the same image.


mark-lunarg:
Edited commit message for length, tested, and pushed.  Thanks a bunch!

PR #1384, opened 2017-01-19
tests: Enhance Error Monitor functionality

daveh-lunarg:
Enhanced ErrorMonitor to allow any combination of error strings and enums to be set simultaneously.
Modified VerifyFound() to require *all* set strings/enums to be found, from previous behavior that was any.
VerifyNotFound() retains the previous behavior of failing on any match.
Added a Reset() function, which is called at the end of both Verify...() flavors.
Flag handling during SetDesired...() is changed to a logical OR of all flag parameters provided, but remains an explicit overwrite in ExpectSuccess() (allowing flags=0 to disable all messages).

2nd patch fixes 7 layer tests that were using multiple target strings/enums, and probably not getting the expected coverage previously.


tests/layer_validation_tests.cpp
```diff
@@ -5382,17 +5390,27 @@ TEST_F(VkLayerTest, InvalidCmdBufferDescriptorSetImageSamplerDestroyed) {
...
     m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "Cannot call vkFreeDescriptorSets() on descriptor set 0x");
     vkFreeDescriptorSets(m_device->device(), ds_pool, 1, &descriptorSet);
     m_errorMonitor->VerifyFound();
-    // Attempt to submit cmd buffer
+
+    // UGLY - The SetDesired/VerifyFound pair immediately above intercepts the vkFreeDescriptorSets() call before it reaches
+    //        the driver. But, we actually want the free to happen in order to provoke the check within QueueSubmit() below. We
+    //        repeat the call now without any failure messages queued, which allows the call to propagate all the way to the
+    //        driver.
+    vkFreeDescriptorSets(m_device->device(), ds_pool, 1, &descriptorSet);
```
mark-lunarg: Whaaaaaaaat?

tests/layer_validation_tests.cpp
```diff
@@ -122,38 +122,40 @@ class ErrorMonitor {
...
+    }
+
+    // ErrorMonitor will look for an error message containing the specified string(s)
+    void SetDesiredFailureMsg(VkFlags msgFlags, const char *msgString) {
+        test_platform_thread_lock_mutex(&m_mutex);
         m_desired_message_strings.insert(msgString);
-        m_msgFound = VK_FALSE;
-        m_msgFlags = msgFlags;
+        m_msgFlags |= msgFlags;
+        m_msgOutstandingCount++;
```
mark-lunarg: Make new vars lower_camel_case.  Extra points if there's a commit to fix the other vars!

tests/layer_validation_tests.cpp
```diff
@@ -10430,15 +10471,15 @@ TEST_F(VkLayerTest, MismatchCountQueueCreateRequestedFeature) {
                      "Use invalid Queue Family Index in vkCreateDevice");
     ASSERT_NO_FATAL_FAILURE(InitState());
 
-    const char *mismatch_count_message = "Call to vkEnumeratePhysicalDevices() "
-                                         "w/ pPhysicalDeviceCount value ";
+    // const char *mismatch_count_message = "Call to vkEnumeratePhysicalDevices() "
```
tobine: Can we kill these lines instead of commenting?

tests/layer_validation_tests.cpp
```diff
@@ -10430,15 +10471,15 @@ TEST_F(VkLayerTest, MismatchCountQueueCreateRequestedFeature) {
...
+    //                                      "w/ pPhysicalDeviceCount value ";
 
     const char *invalid_queueFamilyIndex_message = "Invalid queue create request in vkCreateDevice(). Invalid "
                                                    "queueFamilyIndex ";
 
     const char *unavailable_feature_message = "While calling vkCreateDevice(), requesting feature #";
 
-    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_WARNING_BIT_EXT, mismatch_count_message);
+    // m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_WARNING_BIT_EXT, mismatch_count_message);
```
tobine: Can we kill this line instead of commenting?

daveh-lunarg:
Added the QueueWaitIdle, updated to ErrorMonitor member names to google_style_, removed the commented lines.  Pushing.

PR #1385, opened 2017-01-19
Github issue #398: CmdClearAttachments Validation

### LINKED ISSUE #398: Validation: vkCmdClearAttachments validation not checking all parameters
In the spec for vkCmdClearAttachments (https://www.khronos.org/members/login/sandbox/vulkan/wsi/xhtml/vkspec.html#vkCmdClearAttachments) it lists several rules including:
- The rectangular region specified by a given element of pRects must be contained within the render area of the current render pass instance

Looking at the validation code for vkCmdClearAttachments in https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/parameter_validation.cpp#L3429 it doesn't appear this rule and many of the others from the spec are being checked.

mark-lunarg:
Bug was against pRects not getting validated against RenderArea, also added similar checks for layercount and basearraylayers and removed a bad existing check. Fixes #398.


layers/core_validation.cpp
```diff
@@ -8837,6 +8837,16 @@ CmdFillBuffer(VkCommandBuffer commandBuffer, VkBuffer dstBuffer, VkDeviceSize ds
         dev_data->dispatch_table.CmdFillBuffer(commandBuffer, dstBuffer, dstOffset, size, data);
 }
 
+// Returns true if sub_area is entirely contained within area
+static inline bool ContainsRect(VkRect2D area, VkRect2D sub_area) {
```
tobine: Definite nit-pick but "area" is a measurement of dim^2 that's independent of position and this check is specifically related to position so I'd prefer "rect" and "sub_rect" as param names.

layers/core_validation.cpp
```diff
@@ -8913,12 +8922,36 @@ VKAPI_ATTR void VKAPI_CALL CmdClearAttachments(VkCommandBuffer commandBuffer, ui
...
+                        log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_VIEW_EXT,
+                                reinterpret_cast<uint64_t &>(image_view), __LINE__, VALIDATION_ERROR_01125, "DS",
+                                "vkCmdClearAttachments() with aspects not present in image view: %s. %s",
+                                string_VkImageAspectFlagBits((VkImageAspectFlagBits)extra_aspects),
+                                validation_error_map[VALIDATION_ERROR_01125]);
+                }
+                for (uint32_t j = 0; j < rectCount; j++) {
+                    // The rectangular region specified by a given element of pRects must be contained within the render area of the
+                    // current render pass instance
+                    if (false == ContainsRect(pCB->activeRenderPassBeginInfo.renderArea, pRects[j].rect)) {
```
tobine: Perhaps simpler and cleaner as "if (!Contains...."

layers/core_validation.cpp
mark-lunarg: Will change.

mark-lunarg:
Will squash the DB commit.  Thanks.

mark-lunarg:
Updated for review feedback and pushed to Master.

PR #1386, opened 2017-01-19
Vulkaninfo: Vulkaninfo crashes when DISPLAY is invalid

ardaLunarG:
On Linux the following command crashes:
DISPLAY=foobar ./vulkaninfo

This patch fixes this issue.
https://github.com/LunarG/VulkanTools/issues/125

Change-Id: Ibb8d7b8b6cb4dc730aab9f5b4311e744c903029b

### DIFF
--- a/demos/vulkaninfo.c
+++ b/demos/vulkaninfo.c
@@ -884,6 +884,13 @@ static void AppCreateXcbWindow(struct AppInstance *inst) {
         exit(1);
     }
 
+    int conn_error = xcb_connection_has_error(inst->xcb_connection);
+    if (conn_error) {
+        printf("XCB failed to connect to the X server due to error:%d.\nExiting ...\n", conn_error);
+        fflush(stdout);
+        exit(1);
+    }
+
     setup = xcb_get_setup(inst->xcb_connection);
     iter = xcb_setup_roots_iterator(setup);
     while (scr-- > 0) {


demos/vulkaninfo.c
```diff
@@ -885,6 +885,12 @@ static void AppCreateXcbWindow(struct AppInstance *inst) {
     }
 
     setup = xcb_get_setup(inst->xcb_connection);
+    if (setup == NULL) {
```
karl-lunarg: Where is it documented that this is the right way to do this?
Wouldn't using xcb_connection_has_error() make more sense?

karl-lunarg:
```
 if (xcb_connection_has_error(inst->xcb_connection)) {
        printf("XCB connection has error.\nExiting ...\n");
        fflush(stdout);
        exit(1);
    }
```

works.

Also, you shouldn't repeat two error messages verbatim; else you can't tell which check failed.

It might also be good to get the DISPLAY env var and include it in the error message

ardaLunarG:
Yes you are right. Just checked the code xcb_get_setup returns 0 in error case not NULL
const xcb_setup_t *xcb_get_setup(xcb_connection_t *c)
{
    if(c->has_error)
        return 0;

I have updated the patch accordingly. If setup is 0 we enquiry XCB error number.
Print and exit.

karl-lunarg:
OK, but I STILL don't see anywhere in the XCB docs that xcb_get_setup() can return NULL.  (Yes, I see that it does return NULL for a bad DISPLAY variable on our implementation, but the specs say that it ALWAYS returns a pointer to an xcb_setup_t structure, unless I am missing something)

Why not just check the connection first, before calling get_setup?:

```
    inst->xcb_connection = xcb_connect(NULL, &scr);
        printf("XCB failed to connect to the X server.\nExiting ...\n");
        fflush(stdout);
        exit(1);
    }

     int conn_error = xcb_connection_has_error(inst->xcb_connection);
     if (conn_error)
         printf("XCB failed to connect to the X server due to error:%d.\nExiting ...\n", conn_error);
         fflush(stdout);
         exit(1);
    }

    setup = xcb_get_setup(inst->xcb_connection);

```

You can see that exactly this issue is discussed in:

http://xcb.pdx.freedesktop.narkive.com/44RRACqt/patch-2-2-xcb-get-setup-never-return-null

This implies that some versions of XCB might let xcb_get_setup() return NULL on a failure, while others return a pointer to a struct filled with zero.

ardaLunarG:
OK, trusting implicit "error check" is not right, as it can change in different XCB implementations.
Thus agree again to run our own check.

I have added xcb error check and we exit if there is an error.

karl-lunarg:
Thanks.  It seems the the XCB implementation we're using has a number of issues.  Here are the specs for xcb_connect:

> Connects to the X server specified by displayname. If displayname is NULL, uses the value of the DISPLAY environment variable. If a particular screen on that server is preferred, the int pointed to by screenp (if not NULL) will be set to that screen; otherwise the screen will be set to 0.
> 
> Always returns a non-NULL pointer to a xcb_connection_t, even on failure. Callers need to use xcb_connection_has_error() to check for failure. When finished, use xcb_disconnect() to close the connection and free the structure.

We're seeing xcb_connect() return NULL if there is no DISPLAY variable defined, which violates the first sentence in the second paragraph above.

Also note that the second sentence in the same paragraph says that we have to call xcb_connection_has_error() to check for failure.  That suggests that the latest version of this patch is doing the right thing.

ardaLunarG:
Merged [c18eb54c](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/commit/c18eb54c92b625ad2f5d2970eadc53c1d6f55aa5)

PR #1392, opened 2017-01-20
layers: Workaround QueueSubmit validation issue

Tony-LunarG:
When validation errors are found in QueueSubmit, the layer code aborts
without actually calling QueueSubmit or adding the command buffer to
any queue's work submission, causing it to be forever in flight. Fix
this by removing the command buffer from the in flight set before
aborting

Gets rid of 14 or so unexpected errors from vk_layer_validation_tests


tobine:
Is this to fix a crash/hang issue of some kind or just remove some unexpected errors?
The real solution here is to split up validation/statue update into pre/post functions and only update the state (such as cmd buffers in-flight) after validation passes and the call down the chain succeeds.
This takes a bit of work as QueueSubmit is a bit tangled, but I suspect it could be done in a couple of days. If this current patch is just to get around some unexpected errors I'd prefer to not put it in and prioritize the correct solution instead.

Tony-LunarG:
My thought was to improve the already existing workaround.  I'll take a look at something more ambitious next week.

tobine:
Thanks, @Tony-LunarG, sounds good. I'm happy to consult as needed and you can look at a previous [refactor](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/1288/commits/6f349ece74365ae490b3c901d456163fbe1bd509) I did of GetQueryPoolResults() as reference. Not as extensive as QueueSubmit() but gathers some local state during validation time that's then use to update persistent state Post-call.

PR #1402, opened 2017-01-24
Fix DSUsageBitsErrors test's unexpected errors

Tony-LunarG:
This test failed in SDK 1.0.39 testing due to call to CreateBufferView with a buffer created with invalid usage, and the driver returned VALIDATION_FAILED, causing an asset and the test to fail.  Fix both that call and the call to CreateImageView that was getting an error for similiar reasons.


Tony-LunarG:
Committed to master

PR #1413, opened 2017-01-26
build: Turn on "treat warnings as errors" on Win

karl-lunarg:
Address compiler warnings:

- Suppress warnings for inconsistently redefined snprintf
  on Windows with Visual Studio < 2015.  Warnings for
  preprocessor redefinitions are still enabled on VS 2015
  because snprintf is not redefined on VS 2015.

- Turn off "changed behavior" warning with Visual Studio < 2015
  for using initializers in constructors.  MSFT "fixed" some
  forms of these initializers so that they clear members instead
  of just leaving them alone.  Since clearing is usually desired,
  this isn't a problem in most cases.  But if an object is
  reinstantiated on top of an existing object, via placement new
  for example, the clearing may not be desired if the first object's
  data was meant to be preserved.  We don't count on this behavior.

- Suppress linker warning for linking the loader in debug build on
  Windows.  This warning is for mismatched runtime libraries and
  causes extra code to be linked in only for debug builds.  Ignoring
  the warning is generally regarded as better than the alternatives.

Change-Id: I8b0e311d3c80a8dd9ac0cc95f764ed1123efa86f

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -1,5 +1,4 @@
 cmake_minimum_required (VERSION 2.8.11)
-
 if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
     add_definitions(-DVK_USE_PLATFORM_WIN32_KHR -DWIN32_LEAN_AND_MEAN)
     set(DisplayServer Win32)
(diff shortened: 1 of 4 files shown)


layers/CMakeLists.txt
```diff
@@ -122,6 +121,7 @@ if (WIN32)
     set (CMAKE_C_FLAGS_RELEASE   "${CMAKE_C_FLAGS_RELEASE} -D_CRT_SECURE_NO_WARNINGS")
     set (CMAKE_CXX_FLAGS_DEBUG   "${CMAKE_CXX_FLAGS_DEBUG} -D_CRT_SECURE_NO_WARNINGS /bigobj")
     set (CMAKE_C_FLAGS_DEBUG     "${CMAKE_C_FLAGS_DEBUG} -D_CRT_SECURE_NO_WARNINGS /bigobj")
+    add_compile_options("$<$<AND:$<CXX_COMPILER_ID:MSVC>,$<VERSION_LESS:$<CXX_COMPILER_VERSION>,19>>:/wd4351>")
```
cnorthrop: Can you add a comment for this one too?  That's not trivial to parse.

karl-lunarg: Yes, sorry,  I missed that one.  Thanks.

karl-lunarg:
Pushed to master

PR #1414, opened 2017-01-26
Clang-Format LoaderAndValidationLayers Source Code

mark-lunarg:
1) Standardized the repo .clang-format files
2) Updated the repo using the single standard with the existing LLVM style
3) Changed to Google style and updated the repo


.clang-format
```diff
@@ -1,6 +1,6 @@
 ---
 # Use defaults from LLVM style with the following exceptions:
```
jeremy-lunarg: Comment should say google; not LLVM.

mark-lunarg:
Also worked around a problem in core_validation that resulted in vk_validation_stats.py failing.  Added a TODO for a proper script fix.

Issue #945, opened 2016-09-16
Layers: QueueSubmit checks & side effects not properly separated


chrisforbes:
CV gets confused about which command buffers are in flight when a queue submission is skipped due to a validation error.

chrisforbes:
This is responsible for a good chunk of the unexpected errors in the tests.

tobine:
@Tony-LunarG Here's an old bug for tracking the QueueSubmit refactor. Assigning to you as you expressed interest in untangling this.

mark-lunarg:
@Tony-LunarG, can we close this guy now?

PR #1415, opened 2017-01-26
GH #945 - Separate validation and state record in QueueSubmit

Tony-LunarG:
Refactor QueueSubmit into pre-call validation and post-call state update.  Eliminates several unexpected errors in LVTests.  Also fixes a test that incorrectly relied on previous behavior.


layers/core_validation.cpp
```diff
@@ -4833,19 +4897,14 @@ QueueSubmit(VkQueue queue, uint32_t submitCount, const VkSubmitInfo *pSubmits, V
...
-                    pSemaphore->signaled = false;
-                } else {
+                if (!pSemaphore->signaled) {
                     skip_call |=
                         log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_SEMAPHORE_EXT,
                                 reinterpret_cast<const uint64_t &>(semaphore), __LINE__, DRAWSTATE_QUEUE_FORWARD_PROGRESS, "DS",
                                 "Queue 0x%p is waiting on semaphore 0x%" PRIx64 " that has no way to be signaled.", queue,
                                 reinterpret_cast<const uint64_t &>(semaphore));
+                } else {
+                    pSemaphore->signaled = false;
```
tobine: Not sure about this. This is down PreCallValidate path, right? If so, semaphore state shouldn't be updated down this path.

layers/core_validation.cpp
```diff
@@ -4862,32 +4921,27 @@ QueueSubmit(VkQueue queue, uint32_t submitCount, const VkSubmitInfo *pSubmits, V
                                 queue, reinterpret_cast<const uint64_t &>(semaphore),
                                 reinterpret_cast<uint64_t &>(pSemaphore->signaler.first));
                 } else {
-                    pSemaphore->signaler.first = queue;
-                    pSemaphore->signaler.second = pQueue->seq + pQueue->submissions.size() + 1;
                     pSemaphore->signaled = true;
```
tobine: Same comment as above, updating semaphore state down validation path.

layers/core_validation.cpp
```diff
@@ -4483,37 +4483,32 @@ static bool ValidateAndIncrementBoundObjects(layer_data *dev_data, GLOBAL_CB_NOD
             // TODO : Merge handling of other objects types into this code
             break;
         }
-        if (!base_obj) {
+        if (base_obj && increment) {
+            base_obj->in_use.fetch_add(1);
+        } else if (!base_obj) {
```
tobine: We shouldn't flag errors down the state update path so this case should include "&& !increment" or perhaps assert(!increment) in the else if block.

layers/core_validation.cpp
```diff
@@ -4862,32 +4921,27 @@ QueueSubmit(VkQueue queue, uint32_t submitCount, const VkSubmitInfo *pSubmits, V
...
-                    semaphore_signals.push_back(semaphore);
                 }
             }
         }
-
-        std::vector<VkCommandBuffer> cbs;
-
         for (uint32_t i = 0; i < submit->commandBufferCount; i++) {
             auto cb_node = getCBNode(dev_data, submit->pCommandBuffers[i]);
             skip_call |= ValidateCmdBufImageLayouts(dev_data, cb_node);
```
chrisforbes: There's a SetLayout inside ValidateCmdBufImageLayouts you'll still need to untangle.

layers/core_validation.cpp
```diff
@@ -4483,37 +4483,32 @@ static bool ValidateAndIncrementBoundObjects(layer_data *dev_data, GLOBAL_CB_NOD
             // TODO : Merge handling of other objects types into this code
```
chrisforbes: Isn't the bulk of this `GetStateStructPtrFromObject`? Possibly something for a followup.

layers/core_validation.cpp
```diff
@@ -4401,9 +4401,9 @@ static bool ValidateCmdBufImageLayouts(layer_data *dev_data, GLOBAL_CB_NODE *pCB
     return skip_call;
 }
 
-// Loop through bound objects and increment their in_use counts
-//  For any unknown objects, flag an error
-static bool ValidateAndIncrementBoundObjects(layer_data *dev_data, GLOBAL_CB_NODE const *cb_node) {
+// Loop through bound objects and either increment their in_use counts
```
chrisforbes: Ick. Separating the traversal from the choice of side effect is a bigger change though.

layers/core_validation.cpp
```diff
@@ -4691,7 +4685,7 @@ static bool validateCommandBufferState(layer_data *dev_data, GLOBAL_CB_NODE *pCB
     if (dev_data->instance_data->disabled.command_buffer_state)
         return skip;
     // Validate ONE_TIME_SUBMIT_BIT CB is not being submitted more than once
-    if ((pCB->beginInfo.flags & VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT) && (pCB->submitCount > 1)) {
+    if ((pCB->beginInfo.flags & VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT) && (pCB->submitCount > 0)) {
```
chrisforbes: This doesn't work if the command buffer is submitted multiple times in one batch.

chrisforbes: [You might want a tentative submit count, or some other tracking structure instead]

layers/core_validation.cpp
chrisforbes: This is tricky. Submits within one batch may wait/signal the same semaphore multiple times. Going to need to track tentative signaled state to make the validation correct, just like the submit count case above (or similar)

layers/core_validation.cpp
chrisforbes: Not that simple, see previous comment

layers/core_validation.cpp
```diff
@@ -4800,28 +4815,77 @@ ValidateFenceForSubmit(layer_data *dev_data, FENCE_NODE *pFence)
...
+                incrementResources(dev_data, cb_node);
+                if (!cb_node->secondaryCommandBuffers.empty()) {
+                    for (auto secondaryCmdBuffer : cb_node->secondaryCommandBuffers) {
+                        GLOBAL_CB_NODE *pSubCB = getCBNode(dev_data, secondaryCmdBuffer);
+                        incrementResources(dev_data, pSubCB);
+                    }
+                }
+            }
+        }
+        pQueue->submissions.emplace_back(*cbs, semaphore_waits, semaphore_signals,
```
chrisforbes: No, the contents of `cbs` here is wrong. It should be different for each submit, not the same across the whole batch! Can just collect them here in the loop.

layers/core_validation.cpp
chrisforbes: Note this has the same messy interactions with the validation that the other submit side effects do

Tony-LunarG:
Ok, second try.  Probably still needs work but it should be closer to right.  @chrisforbes @tobine     suggestions welcome

Tony-LunarG:
Sounds good - no rush on this. Thanks

layers/core_validation.cpp
```diff
@@ -3486,6 +3516,11 @@ bool FindLayouts(const layer_data *my_data, VkImage image, std::vector<VkImageLa
     return true;
 }
 
+// Set the layout in supplied map
+void SetLayout(unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE> imageLayoutMap, ImageSubresourcePair imgpair, const VkImageLayout &layout) {
```
chrisforbes: No reason to pass layout by `const &`. It's an enum, by value is fine.

layers/core_validation.cpp
```diff
@@ -4349,12 +4384,20 @@ static bool ValidateCmdBufImageLayouts(layer_data *dev_data, GLOBAL_CB_NODE *pCB
...
+            SetLayout(imageLayoutMap, cb_image_data.first, cb_image_data.second.layout);
         }
     }
     return skip_call;
 }
 
+static void UpdateCmdBufImageLayouts(layer_data *dev_data, GLOBAL_CB_NODE *pCB) {
+    for (auto cb_image_data : pCB->imageLayoutMap) {
+        VkImageLayout imageLayout;
+        FindLayout(dev_data, cb_image_data.first, imageLayout);
```
chrisforbes: Why call `FindLayout` here?

layers/core_validation.cpp
```diff
@@ -4315,11 +4350,11 @@ static bool ValidateStageMaskGsTsEnables(layer_data *dev_data, VkPipelineStageFl
 // This validates that the initial layout specified in the command buffer for
 // the IMAGE is the same
 // as the global IMAGE layout
-static bool ValidateCmdBufImageLayouts(layer_data *dev_data, GLOBAL_CB_NODE *pCB) {
+static bool ValidateCmdBufImageLayouts(layer_data *dev_data, GLOBAL_CB_NODE *pCB, unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE> imageLayoutMap) {
```
chrisforbes: Ditto re copying the map.

layers/core_validation.cpp
```diff
@@ -4882,7 +4926,8 @@ static bool PreCallValidateQueueSubmit(layer_data *dev_data, VkQueue queue, uint
         }
         for (uint32_t i = 0; i < submit->commandBufferCount; i++) {
             auto cb_node = getCBNode(dev_data, submit->pCommandBuffers[i]);
-            skip_call |= ValidateCmdBufImageLayouts(dev_data, cb_node);
+            unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE> localImageLayoutMap = dev_data->imageLayoutMap;
```
chrisforbes: Scoped too tightly. This needs to be maintained across all submissions in the batch. Suppose a batch consisting of two command buffers which use the same image (either in one submission, or two).

layers/core_validation.cpp
```diff
@@ -4809,7 +4809,8 @@ static bool ValidateFenceForSubmit(layer_data *dev_data, FENCE_NODE *pFence) {
 }
 
 static void PostCallRecordQueueSubmit(layer_data *dev_data, VkQueue queue, uint32_t submitCount, const VkSubmitInfo *pSubmits,
-                                      VkFence fence, std::vector<VkCommandBuffer> *cbs) {
+                                      VkFence fence) {
+    std::vector<VkCommandBuffer> cbs;
```
chrisforbes: Still too far out; needs to be scoped to the submission. Suppose a batch containing two submissions. You don't want to be adding submission 0's CBs to pQueue->submissions on :4867 the second time around!

tests/layer_validation_tests.cpp
```diff
@@ -5576,6 +5576,7 @@ TEST_F(VkLayerTest, InvalidCmdBufferDescriptorSetImageSamplerDestroyed) {
     Draw(1, 0, 0, 0);
     m_commandBuffer->EndRenderPass();
     m_commandBuffer->EndCommandBuffer();
+    vkQueueSubmit(m_device->m_queue, 1, &submit_info, VK_NULL_HANDLE);
```
chrisforbes: :+1: for the test change.

tests/layer_validation_tests.cpp
```diff
@@ -6986,6 +6987,101 @@ TEST_F(VkLayerTest, NoBeginCommandBuffer) {
...
+    submit_info[2].waitSemaphoreCount = 1;
+    submit_info[2].pWaitSemaphores = &semaphore2;
+    submit_info[2].signalSemaphoreCount = 0;
+    submit_info[2].pSignalSemaphores = nullptr;
+    submit_info[2].pWaitDstStageMask = flags;
+    vkQueueSubmit(m_device->m_queue, 3, submit_info, VK_NULL_HANDLE);
+    vkQueueWaitIdle(m_device->m_queue);
+    vkDestroySemaphore(m_device->device(), semaphore1, NULL);
+    vkDestroySemaphore(m_device->device(), semaphore2, NULL);
+}
```
chrisforbes: Need additional tests to verify the logic for `COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT` and (lack of) `COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT` where the same command buffer appears multiple times in the same batch.

tests/layer_validation_tests.cpp
```diff
@@ -6986,6 +6987,101 @@ TEST_F(VkLayerTest, NoBeginCommandBuffer) {
...
+    submit_info[0].pCommandBuffers = &cmd_bufs[0];
+    submit_info[0].signalSemaphoreCount = 1;
+    submit_info[0].pSignalSemaphores = &semaphore1;
+    submit_info[0].waitSemaphoreCount = 0;
+    submit_info[0].pWaitDstStageMask = nullptr;
+    submit_info[0].pWaitDstStageMask = flags;
+    submit_info[1].sType = VK_STRUCTURE_TYPE_SUBMIT_INFO;
+    submit_info[1].commandBufferCount = 1;
+    submit_info[1].pCommandBuffers = &cmd_bufs[1];
+    submit_info[1].waitSemaphoreCount = 1;
```
chrisforbes: Consider splitting the exercising of semaphore tracking out into its own test?

layers/core_validation.cpp
```diff
@@ -4652,7 +4652,7 @@ static bool PreCallValidateQueueSubmit(layer_data *dev_data, VkQueue queue, uint
 
     unordered_set<VkSemaphore> signaled_semaphores;
     unordered_set<VkSemaphore> unsignaled_semaphores;
-    unordered_set<VkCommandBuffer> current_cmds;
+    vector<VkCommandBuffer> current_cmds;
```
chrisforbes: Might have considered `unordered_map<VkCommandBuffer, int>` which allows :4702 to be `int cb_submit_count = ++current_cmds[submit->pCommandBuffers[i]];` and then :4704 to just use `cb_submit_count`, but what you've done works and I think is fine.

tests/layer_validation_tests.cpp
```diff
@@ -6987,6 +6987,127 @@ TEST_F(VkLayerTest, NoBeginCommandBuffer) {
...
+    submit_info[2].pNext = nullptr;
+    submit_info[2].commandBufferCount = 2;
+    submit_info[2].pCommandBuffers = &cmd_bufs[2];
+    submit_info[2].waitSemaphoreCount = 1;
+    submit_info[2].pWaitSemaphores = &semaphore2;
+    submit_info[2].signalSemaphoreCount = 0;
+    submit_info[2].pSignalSemaphores = nullptr;
+    submit_info[2].pWaitDstStageMask = flags;
+    vkQueueSubmit(m_device->m_queue, 3, submit_info, VK_NULL_HANDLE);
+    vkQueueWaitIdle(m_device->m_queue);
```
chrisforbes: You still don't assert anything in this test.

tests/layer_validation_tests.cpp
```diff
@@ -6987,6 +6987,127 @@ TEST_F(VkLayerTest, NoBeginCommandBuffer) {
...
+    submit_info[0].commandBufferCount = 2;
+    submit_info[0].pCommandBuffers = duplicates;
+    cmd_bufs[1] = cmd_bufs[0];
+    submit_info[0].signalSemaphoreCount = 0;
+    submit_info[0].pSignalSemaphores = nullptr;
+    vkQueueSubmit(m_device->m_queue, 1, submit_info, VK_NULL_HANDLE);
+    vkQueueWaitIdle(m_device->m_queue);
+
+    cb_binfo.flags = VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT |
+            VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT;
```
chrisforbes: This combination of flags doesn't make sense? What is this trying to do?

tests/layer_validation_tests.cpp
```diff
@@ -6987,6 +6987,127 @@ TEST_F(VkLayerTest, NoBeginCommandBuffer) {
...
+    vkCmdSetViewport(cmd_bufs[4], 0, 1, &viewport);
+    vkEndCommandBuffer(cmd_bufs[4]);
+    VkCommandBuffer duplicates[2] = {cmd_bufs[4], cmd_bufs[4]};
+
+    submit_info[0].commandBufferCount = 2;
+    submit_info[0].pCommandBuffers = duplicates;
+    cmd_bufs[1] = cmd_bufs[0];
+    submit_info[0].signalSemaphoreCount = 0;
+    submit_info[0].pSignalSemaphores = nullptr;
+    vkQueueSubmit(m_device->m_queue, 1, submit_info, VK_NULL_HANDLE);
```
chrisforbes: You expect this submission to fail. Assert that it does.

tests/layer_validation_tests.cpp
```diff
@@ -6987,6 +6987,127 @@ TEST_F(VkLayerTest, NoBeginCommandBuffer) {
...
+    vkQueueWaitIdle(m_device->m_queue);
+
+    cb_binfo.flags = VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT |
+            VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT;
+    vkBeginCommandBuffer(cmd_bufs[4], &cb_binfo);
+    vkCmdSetViewport(cmd_bufs[4], 0, 1, &viewport);
+    vkEndCommandBuffer(cmd_bufs[4]);
+    vkQueueSubmit(m_device->m_queue, 1, submit_info, VK_NULL_HANDLE);
```
chrisforbes: Same here -- this submission fails. Ensure that it did.

layers/core_validation.cpp
```diff
@@ -4608,31 +4689,25 @@ VKAPI_ATTR VkResult VKAPI_CALL QueueSubmit(VkQueue queue, uint32_t submitCount,
...
+
                 // Potential early exit here as bad object state may crash in delayed function calls
-                if (skip_call) return result;
+                if (skip_call) {
+                    return true;
+                }
+
                 // Call submit-time functions to validate/update state
                 for (auto &function : cb_node->validate_functions) {
                     skip_call |= function();
```
chrisforbes: Might be OK to say this is just going to be imperfect for now, and deal with `validate_functions` being a mess later?

tests/layer_validation_tests.cpp
Tony-LunarG: Yes, the test is a work in progress to try to get the validation right, and then I plan to split it into positive and negative tests with better names

tests/layer_validation_tests.cpp
```diff
@@ -6986,6 +6987,127 @@ TEST_F(VkLayerTest, NoBeginCommandBuffer) {
...
+    submit_info[2].pCommandBuffers = &cmd_bufs[2];
+    submit_info[2].waitSemaphoreCount = 1;
+    submit_info[2].pWaitSemaphores = &semaphore2;
+    submit_info[2].signalSemaphoreCount = 0;
+    submit_info[2].pSignalSemaphores = nullptr;
+    submit_info[2].pWaitDstStageMask = flags;
+    vkQueueSubmit(m_device->m_queue, 3, submit_info, VK_NULL_HANDLE);
+    vkQueueWaitIdle(m_device->m_queue);
+
+    vkBeginCommandBuffer(cmd_bufs[4], &cb_binfo);
```
chrisforbes: Split these negative cases into their own test?

tests/layer_validation_tests.cpp
chrisforbes: :+1: alright, ping me again when you're happy with it. I think this is getting pretty close.

Tony-LunarG:
@chrisforbes could you have another look.  I think its good or very close at this point.  Thanks!

mark-lunarg:
Tony, if Chris has additional input on the newest changes we can address them in a follow-on series.  If the testing all looks good go ahead and push it as it's holding up things elsewhere.  👍

layers/core_validation.cpp
```diff
@@ -4363,6 +4357,27 @@ static bool validateCommandBufferState(layer_data *dev_data, GLOBAL_CB_NODE *pCB
     return skip;
 }
 
+static bool validateResources(layer_data *dev_data, GLOBAL_CB_NODE *cb_node) {
+    bool skip_call = false;
+
+    // First Increment for all "generic" objects bound to cmd buffer, followed by special-case objects below
```
tobine: This comment is incorrect now. No increment here.

Tony-LunarG:
Merged

PR #1419, opened 2017-01-27
layers: Fix for VkBufferImageCopy::bufferOffset validation

MarcinKantochAMD:
According to the Vulkan spec:

"If the the calling command’s VkImage parameter’s format is not a depth/stencil
format, then VkBufferImageCopy::bufferOffset must be a multiple of the format’s
element size"

Currently the image validation layer is checking unconditionally if
VkBufferImageCopy::bufferOffset is a multiple of the format’s element size
and reporting error otherwise.

Added code that skips the alignment check if the VkImage parameter’s format is
combined depth/stencil.

### DIFF
--- a/layers/image.cpp
+++ b/layers/image.cpp
@@ -843,9 +843,10 @@ static bool ValidateBufferImageCopyData(layer_data *dev_data, uint32_t regionCou
                 }
             }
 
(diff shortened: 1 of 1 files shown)


mark-lunarg:
Fixed commit message for style, tested and pushed to Master.  Thanks @MarcinKantochAMD!

PR #1421, opened 2017-01-27
build: Enable declaration hiding warning on Windows

### LINKED ISSUE #1388: Add additional compiler warnings / errors to cmake build
Often I update ANGLE's build of the layers to latest and I find a common warning:

..\third_party\vulkan-validation-layers\src\loader\loader.c(5006): warning C4456: declaration of 'i' hides previous local declaration

Assuming this isn't a desired programming practice (making multiple nested "i" iteration variables) then it would be nice if there was a way to enable this warning in the default Vulkan build, potentially so it could be caught during the submission step by an automated builder. At least it would be useful to know how to control warnings from the cmake files.

Any interest in taking on this work from the maintainers?

karl-lunarg:
Fixes #1388
Turn on the Windows compiler option (4456) to report
hidden declarations.
Fix all places where this was occurring.

Change-Id: I3346d87da8b70d6299c206fcac68520a091ed1a6

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,9 +57,12 @@ if (CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
 endif()
 
 if(WIN32)
-    # Disable RTTI, Treat warnings as errors
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /WX")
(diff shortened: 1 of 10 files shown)


karl-lunarg:
We never agreed or discussed turning on this warning, so this PR also constitutes a proposal to turn it on.  If there's a lot of push back, we can reconsider.   But I think it helps prevent confusing code and discourages things like using a loop index 'i' over a really big loop.

karl-lunarg:
pushed to master

PR #1422, opened 2017-01-27
docs: Update build dependencies

renelindsay:
libx11-xcb-dev is used by WSI-Window.
libx11-dev and libxcb1-dev depend on libx11-xcb-dev.
So, installing libx11-xcb-dev will install all three.

Change-Id: Id3909919bbd48794df2fcad10b67ea3f8dbb203f


stroyan:
You wrote-
    libx11-dev and libxcb1-dev depend on libx11-xcb-dev.
I think you meant
    libx11-xcb-dev depends on libx11-dev and libxcb1-dev.

And that does mean that installing libx11-xcb-dev will install all three.

renelindsay:
Yes Mike, that's what I meant.
I'll correct the working, thanks.

PR #1423, opened 2017-01-27
Initial batch of image late migration

mark-lunarg:
Moved guts of CreateImage, DestroyImage, CreateRenderPass and ClearColorImage.  Also fixed up some VUs.


layers/parameter_validation.cpp
```diff
@@ -3754,6 +3754,16 @@ static bool PreCreateRenderPass(layer_data *dev_data, const VkRenderPassCreateIn
     bool skip = false;
     uint32_t max_color_attachments = dev_data->device_limits.maxColorAttachments;
 
+    for (uint32_t i = 0; i < pCreateInfo->attachmentCount; ++i) {
+        if (pCreateInfo->pAttachments[i].format == VK_FORMAT_UNDEFINED) {
```
mikes-lunarg: Should we be checking more that just VK_FORMAT_UNDEFINED here? Maybe also <= VK_FORMAT_END_RANGE

mark-lunarg: Yeah, that particular check is handled in parameter-validation, but it is also cool with FORMAT_UNDEFINED (hence, this check).

PR #1426, opened 2017-01-30
loader: Improve Linux manifest search

Ralith:
This builds on the fixes introduced in #1086 to extend support to more systems and improve flexibility on existing systems. Searching XDG_CONFIG_DIRS and XDG_DATA_DIRS instead of hard-coding specific paths improves portability, particularly on platforms which don't implement the Filesystem Hierarchy Standard. For example, this allows the loader to function correctly out of the box on NixOS.

Use of the `location` parameter to `loader_get_manifest_files` on non-windows was discontinued entirely in favor of the new search procedure, and variable/macro names used to identify specific search subdirectories were adjusted from `home_` to `relative_` to reflect their broadened use. This allowed substantial simplification of vk_loader_platform.h. The `DATADIR` macros were removed entirely because /usr/share (and /usr/local/share) are searched by default as per [the XDG Base Directory spec](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html), although most systems will get these from the environment variable instead. The `SYSCONFDIR` macros were retained because XDG defaults do not subsume them.

Future work could include extending the search to XDG_CONFIG_HOME, which defaults to $HOME/.config. This would improve consistency (because system config dirs are already being searched) but probably won't benefit existing applications.

This patch was based on work by @corngood to implement the same functionality on version 1.0.26.0 in https://github.com/NixOS/nixpkgs/pull/19936, and represents an effort to upstream the changes necessary for vulkan to work correctly on NixOS.


loader/loader.c
```diff
@@ -2638,16 +2638,89 @@ static VkResult loader_get_manifest_files(const struct loader_instance *inst, co
...
+        loc_read = &xdgconfdirs[0];
+        for (const char *x = loc_read;; ++x) {
+            if (*x == PATH_SEPARATOR || *x == '\0') {
+                const size_t s = x - loc_read;
+                memcpy(loc_write, loc_read, s);
+                loc_write += s;
+                memcpy(loc_write, relative_location, rel_size);
+                loc_write += rel_size;
+                *loc_write++ = PATH_SEPARATOR;
+                if (*x == 0)
```
MarkY-LunarG: What are you doing here?  Does this do anything more than line 2676 since you haven't changed x?

loader/loader.c
```diff
@@ -2638,16 +2638,89 @@ static VkResult loader_get_manifest_files(const struct loader_instance *inst, co
...
+                    break;
+                loc_read = ++x;
+            }
+        }
+        --loc_write;
+        *loc_write = '\0';
+#else
+        memcpy(loc_write, location, loc_size);
+        loc[loc_size-1] = '\0';
```
MarkY-LunarG: Nit: Since you use loc_write in the command above, I would use loc_write here (instead of loc) just for clarity.

loader/loader.c
Ralith: This breaks the loop, while the if statement (which must run for every path in the set) doesn't.

MarkY-LunarG: <facepalm> You're right.  Sorry, that should have been obvious.

MarkY-LunarG: Sorry.  You're correct.

Ralith:
I went ahead and made `loc_write` used more consistently. This had the side benefit of making it easier to assert against a buffer overrun.

PR #1427, opened 2017-01-30
vulkaninfo: Print human-readable heap sizes

hesiod:
Since the size of a memory heap can be rather substantial, it is helpful
to have the size additionally printed out in a human-readable format,
i.e. using prefixes.

**NB:** I haven't checked if Windows has any quirks regarding printf's `%f` format specifier and more specifically whether Windows supports the precision specifier I used here. I only tested this on Linux.


mark-lunarg:
Squashed the commits, changed a couple of var names for new style, pushed to Master.  Thanks Tobias!

Issue #283, opened 2016-04-06
Validate that sparse residency does not use linear


tobine:
From spec section 28.1:

```
A sparse image created using VK_IMAGE_CREATE_SPARSE_RESIDENCY_BIT supports all
non-compressed color formats with power-of-two texel size that non-sparse usage supports. Additional 
formats may also be supported and can be queried via vkGetPhysicalDeviceSparseImageFormatProperties.
VK_IMAGE_TILING_LINEAR tiling is not supported.
```

daveh-lunarg:
Added this check to parameter_validation, along with a half dozen related valid usage checks.  Submitted as PR #1429.

PR #1429, opened 2017-01-30
Resolve github issue #283, plus related VU checks.

daveh-lunarg:
Added a check requested in github issue #283 to catch a request for sparse residency images with linear tiling.  Also added a half dozen other checks of valid usages for sparse residency of images.  Added two tests to verify the new checks.  Added a GetPhysicalDeviceFeatures() function to the test framework, to make it easier to mask out specific device features, in order to facilitate these tests.


Tony-LunarG:
Fine by me

PR #1431, opened 2017-01-31
Github 1390: CmdBeginRenderpass issues

### LINKED ISSUE #1390: Incorrect error message from vkCmdBeginRenderPass related to LOAD_OP
There appears to be two issues with the message: e.g.
"In vkCmdBeginRenderPass() the VkRenderPassBeginInfo struct has a clearValueCount of 2 but only first 1 entries in pClearValues array are used. The highest index attachment in renderPass 0x5c6f14d230 that uses VK_ATTACHMENT_LOAD_OP_CLEAR is 1 - other pClearValues are ignored."

The first is the highest index attachment that uses LOAD_OP_CLEAR appears to be one higher than the actual value, I suspect because the printf is composed with clear_op_size, and a highest index of 0 necessarily means a size of 1.

The second is the renderPass handle (as above, 0x5c6f14d230) has no resemblance to the created handle as seen from the client code, in my case 0x6b7. I am unsure of the cause. Locally I can see from the callstack that it has passed through:

 	VkLayer_core_validation.dll!00007fff7a41eac0() 	
 	[Frames below may be incorrect and/or missing, no symbols loaded for VkLayer_core_validation.dll]	
 	VkLayer_core_validation.dll!00007fff7a4997ff() 	
 	VkLayer_core_validation.dll!00007fff7a3e0042() 	
 	VkLayer_object_tracker.dll!00007fff8308dffd() 	
 	VkLayer_parameter_validation.dll!00007fff7a214342() 	
 	VkLayer_threading.dll!00007fff9889bb92() 	

so perhaps another layer has messed with it, and/or my layer order is incorrect.

mark-lunarg:
Validation was printing incorrect handle values and one of the error messages was incorrect.  Fixes #1390.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -10838,24 +10838,25 @@ VKAPI_ATTR void VKAPI_CALL CmdBeginRenderPass(VkCommandBuffer commandBuffer, con
             if (clear_op_size > pRenderPassBegin->clearValueCount) {
                 skip_call |= log_msg(
                     dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_RENDER_PASS_EXT,
-                    reinterpret_cast<uint64_t &>(renderPass), __LINE__, VALIDATION_ERROR_00442, "DS",
+                    reinterpret_cast<uint64_t &>(renderPass->renderPass), __LINE__, VALIDATION_ERROR_00442, "DS",
                     "In vkCmdBeginRenderPass() the VkRenderPassBeginInfo struct has a clearValueCount of %u but there must "
                     "be at least %u entries in pClearValues array to account for the highest index attachment in renderPass "
                     "0x%" PRIx64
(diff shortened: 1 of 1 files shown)


layers/core_validation.cpp
```diff
@@ -10845,7 +10845,7 @@ VKAPI_ATTR void VKAPI_CALL CmdBeginRenderPass(VkCommandBuffer commandBuffer, con
                     " that uses VK_ATTACHMENT_LOAD_OP_CLEAR is %u. Note that the pClearValues array "
                     "is indexed by attachment number so even if some pClearValues entries between 0 and %u correspond to "
                     "attachments that aren't cleared they will be ignored. %s",
-                    pRenderPassBegin->clearValueCount, clear_op_size, reinterpret_cast<uint64_t &>(renderPass), clear_op_size,
+                    pRenderPassBegin->clearValueCount, clear_op_size, reinterpret_cast<uint64_t &>(renderPass->renderPass), clear_op_size,
```
chrisforbes: You've missed two. Object reported (not in the message) is also the address of our state struct rather than the handle, in both cases.

mark-lunarg:
Fix errors Chris found, took Dave's suggestion and pushed to Master.

PR #1436, opened 2017-02-01
layers:Move gfx pipe check to create time

### LINKED ISSUE #1424: Error message at PSO create time instead of runtime
I got the message:
Render pass subpass 0 mismatch with blending state defined and blend state attachment count 1 while subpass color attachment count 0 in Pipeline (0x7f81)!

Which took me a while to figure out because the mistmach is between the renderpass and the color attachment on the blendstate, which are both known at PSO creation time, but it looks like there is no matching here, instead it matches the bound renderpass against the color attachment count at runtime. It would be far easier to debug if a check happened at creation time.

tobine:
Fixes #1424 

The check to verify that the attachmentCount for blend state of a
Graphics Pipeline matches the attachmentCount for the given subpass of
the creation renderPass was occurring at draw time but can and should
be done at Gfx Pipeline creation time.

This change moves the check to creation time and updates the database
and test appropriately.


tests/layer_validation_tests.cpp
```diff
@@ -10414,23 +10413,6 @@ TEST_F(VkLayerTest, NumBlendAttachMismatch) {
...
-    VkViewport viewport = {0, 0, 16, 16, 0, 1};
-    VkRect2D scissor = {{0, 0}, {16, 16}};
-    vkCmdSetViewport(m_commandBuffer->handle(), 0, 1, &viewport);
-    vkCmdSetScissor(m_commandBuffer->handle(), 0, 1, &scissor);
-
-    // Render triangle (the error should trigger on the attempt to draw).
-    Draw(3, 1, 0, 0);
-
-    // Finalize recording of the command buffer
-    m_commandBuffer->EndRenderPass();
```
chrisforbes: Still plenty more junk can be deleted from this test. We don't need a descriptor set, etc since we're not drawing.

PR #1437, opened 2017-02-01
Move more stuff to the buffer_validation module

mark-lunarg:
1) Move some image-related helper routines from CV to buffer_validation:

- ResolveRemainingLevelsLayers
- VerifyClearImageLayout
- RecordClearImageLayout
- PreCallValidateCmdClearColorImage
- PreCallRecordCmdClearImage
- PreCallValidateCmdClearDepthStencilImage

2) Move CmdClearDepthStencilImage validation from image.cpp to buffer_validation

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -415,19 +415,36 @@ void PreCallRecordCmdClearImage(core_validation::layer_data *dev_data, VkCommand
     }
 }
 
-bool PreCallValidateCmdClearDepthStencilImage(core_validation::layer_data *dev_data, VkCommandBuffer commandBuffer, VkImage image,
-                                              VkImageLayout imageLayout, uint32_t rangeCount,
+bool PreCallValidateCmdClearDepthStencilImage(core_validation::layer_data *device_data, VkCommandBuffer commandBuffer,
+                                              VkImage image, VkImageLayout imageLayout, uint32_t rangeCount,
                                               const VkImageSubresourceRange *pRanges) {
     bool skip = false;
+    const debug_report_data *report_data = core_validation::GetReportData(device_data);
+
     // TODO : Verify memory is in VK_IMAGE_STATE_CLEAR state
-    auto cb_node = getCBNode(dev_data, commandBuffer);
-    auto image_state = getImageState(dev_data, image);
+    auto cb_node = getCBNode(device_data, commandBuffer);
+    auto image_state = getImageState(device_data, image);
     if (cb_node && image_state) {
-        skip |= ValidateMemoryIsBoundToImage(dev_data, image_state, "vkCmdClearDepthStencilImage()", VALIDATION_ERROR_02528);
-        skip |= ValidateCmd(dev_data, cb_node, CMD_CLEARDEPTHSTENCILIMAGE, "vkCmdClearDepthStencilImage()");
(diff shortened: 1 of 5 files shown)


layers/buffer_validation.cpp
```diff
@@ -263,3 +263,189 @@ bool ValidateImageAttributes(core_validation::layer_data *device_data, IMAGE_STA
...
+    }
+}
+
+bool VerifyClearImageLayout(core_validation::layer_data *device_data, GLOBAL_CB_NODE *cb_node, VkImage image,
+                            VkImageSubresourceRange range, VkImageLayout dest_image_layout, const char *func_name) {
+    bool skip = false;
+    const debug_report_data *report_data = core_validation::GetReportData(device_data);
+
+    VkImageSubresourceRange resolved_range = range;
+    ResolveRemainingLevelsLayers(device_data, &resolved_range, image);
```
tobine: Would be better to do a single image_state look-up before this call and pass in image_state ptr instead of image, then can skip the look-ups in the function and below as well.

layers/buffer_validation.cpp
```diff
@@ -263,3 +263,189 @@ bool ValidateImageAttributes(core_validation::layer_data *device_data, IMAGE_STA
...
+
+    // TODO : Verify memory is in VK_IMAGE_STATE_CLEAR state
+    auto cb_node = getCBNode(device_data, commandBuffer);
+    auto image_state = getImageState(device_data, image);
+    if (cb_node && image_state) {
+        skip |= ValidateMemoryIsBoundToImage(device_data, image_state, "vkCmdClearDepthStencilImage()", VALIDATION_ERROR_02528);
+        skip |= ValidateCmd(device_data, cb_node, CMD_CLEARDEPTHSTENCILIMAGE, "vkCmdClearDepthStencilImage()");
+        skip |= insideRenderPass(device_data, cb_node, "vkCmdClearDepthStencilImage()", VALIDATION_ERROR_01111);
+        for (uint32_t i = 0; i < rangeCount; ++i) {
+            skip |= VerifyClearImageLayout(device_data, cb_node, image, pRanges[i], imageLayout, "vkCmdClearDepthStencilImage()");
```
tobine: Actually, we've got the image_state here already so can pass it down at this point and kill 3 separate look-ups down the chain.

layers/buffer_validation.cpp
```diff
@@ -263,3 +263,189 @@ bool ValidateImageAttributes(core_validation::layer_data *device_data, IMAGE_STA
...
+    }
+}
+
+bool PreCallValidateCmdClearColorImage(core_validation::layer_data *dev_data, VkCommandBuffer commandBuffer, VkImage image,
+                                       VkImageLayout imageLayout, uint32_t rangeCount, const VkImageSubresourceRange *pRanges) {
+    bool skip = false;
+    // TODO : Verify memory is in VK_IMAGE_STATE_CLEAR state
+    auto cb_node = core_validation::getCBNode(dev_data, commandBuffer);
+    auto image_state = getImageState(dev_data, image);
+    if (cb_node && image_state) {
```
tobine: All of the if checks like this can theoretically be killed and replaced with asserts. Object_tracker will trigger before these look-ups should fail. Don't have to update this now but if not it should definitely be done when object_tracker is integrated.

mark-lunarg:
Added suggested changes and pushed to Master.

PR #1439, opened 2017-02-02
Tolerate null handles in destroy and free routines

Tony-LunarG:
Several CTS tests call destroy routines with null handles.  This
keeps validation layers from crashing or reporting false positives


Tony-LunarG:
I forgot to add a test for these - working on it now

Tony-LunarG:
Committed to master

PR #1440, opened 2017-02-02
tests: Fix error in InUseDestroyedSignaled

jeremy-lunarg:
Expect the unexpected.

Change-Id: I11e2b54b189beec6c709a09eac47802bf75fdee1


tests/layer_validation_tests.cpp
```diff
@@ -11504,7 +11503,10 @@ TEST_F(VkLayerTest, InUseDestroyedSignaled) {
     submit_info.sType = VK_STRUCTURE_TYPE_SUBMIT_INFO;
     submit_info.commandBufferCount = 1;
     submit_info.pCommandBuffers = &m_commandBuffer->handle();
-    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, submit_with_deleted_event_message);
+    const char *deleted_event_message[] = {"Cannot submit cmd buffer using deleted event 0x",
```
mark-lunarg: I still think these strings should just go into the SDFM calls instead of getting their own single-use array.

jeremy-lunarg: Agreed. Changed. Thanks.

Issue #1171, opened 2016-11-17
Complete VU checks for BindImageMemory


tobine:
Image version of #1170. BindImageMemory needs to be updated to check all the VUs from the spec, and should also be refactored to use the Pre/Post pattern.

mark-lunarg:
In progress.

tobine:
@cdwfs Mark has pushed mark-master-GH1171 to master which has some refactoring and re-naming but no new checks. You can use that as a starting point or start from scratch if you'd like.
This issue includes the fact that all [vkBindImageMemory](https://www.khronos.org/registry/vulkan/specs/1.0-extensions/xhtml/vkspec.html#vkBindImageMemory) valid usage requirements need to have checks and tests implemented for them. The key checks of interest for this issue are [VALIDATION_ERROR_00803-VALIDATION_ERROR_00806](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/vk_validation_error_database.txt#L795)
There are extension and implicit valid usages beyond that. Don't worry about either of those for this issue as extension only applies to particular devices and the implicit cases should all be handed already by parameter_checker and object_tracker.

For an initial PR, I recommend just doing the Pre/Post refactor and adding a single check and test. Then we can review and any feedback may apply to further checks and tests.
Some checks may already be implemented in which case make sure they're using the correct VALIDATION_ERROR_##### and that the database file is correctly updated to reflect that they're implemented.

No rush on this. Should provide some steady work to help you continue growing in the codebase. I'll also assign #1171 to you which is similar but for BindBufferMemory. You can do that as a follow-on once this is complete.

cdwfs:
Fixed by #1446

Issue #1170, opened 2016-11-17
Complete VU checks for BindBufferMemory


tobine:
This function's a bit of a mess in core_validation. It needs to be updated to check all the VUs from the spec, but it should also be refactored to use the Pre/Post pattern.

cdwfs:
Fixed by #1446

PR #1446, opened 2017-02-02
VU tests for vkBindImageMemory() / vkBindBufferMemory()

cdwfs:
Fixes for #1170 and #1171.


tests/layer_validation_tests.cpp
```diff
@@ -2685,6 +2685,54 @@ TEST_F(VkLayerTest, BindInvalidMemory) {
...
+        vkGetBufferMemoryRequirements(device(), buffer, &buffer_mem_reqs);
+        VkMemoryAllocateInfo image_alloc_info = {}, buffer_alloc_info = {};
+        image_alloc_info.sType = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO;
+        image_alloc_info.allocationSize = image_mem_reqs.size;
+        buffer_alloc_info.sType = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO;
+        buffer_alloc_info.allocationSize = buffer_mem_reqs.size;
+        // Create a mask of available memory types *not* supported by these resource.
+        VkPhysicalDeviceMemoryProperties memory_properties = {};
+        vkGetPhysicalDeviceMemoryProperties(m_device->phy().handle(), &memory_properties);
+        uint32_t image_unsupported_mem_type_bits = ((1 << memory_properties.memoryTypeCount) - 1) & ~image_mem_reqs.memoryTypeBits;
```
tobine: Is there a theoretic possibility that all memory type bits could be returned as supported for either buffer of image? If this is the case then should have any early out check in case all bits are returned as supported.

layers/core_validation.cpp
```diff
@@ -5760,8 +5773,9 @@ VKAPI_ATTR VkResult VKAPI_CALL BindBufferMemory(VkDevice device, VkBuffer buffer
         // Track and validate bound memory range information
         auto mem_info = getMemObjInfo(dev_data, mem);
         if (mem_info) {
-            skip_call |= InsertBufferMemoryRange(dev_data, buffer, mem_info, memoryOffset, buffer_state->requirements);
-            skip_call |= ValidateMemoryTypes(dev_data, mem_info, buffer_state->requirements.memoryTypeBits, "vkBindBufferMemory()",
+            skip_call |=
+                InsertBufferMemoryRange(dev_data, buffer, mem_info, memoryOffset, buffer_state->requirements, "vkBindBufferMemory");
+            skip_call |= ValidateMemoryTypes(dev_data, mem_info, buffer_state->requirements.memoryTypeBits, "vkBindBufferMemory",
```
tobine: Append "()" to function strings on these two lines

layers/core_validation.cpp
```diff
@@ -11417,7 +11431,7 @@ static bool PreCallValidateBindImageMemory(layer_data *dev_data, VkImage image,
         auto mem_info = getMemObjInfo(dev_data, mem);
         if (mem_info) {
             skip |= InsertImageMemoryRange(dev_data, image, mem_info, memoryOffset, image_state->requirements,
-                                           image_state->createInfo.tiling == VK_IMAGE_TILING_LINEAR);
+                                           image_state->createInfo.tiling == VK_IMAGE_TILING_LINEAR, "vkBindImageMemory");
```
tobine: "()" here and below as well.

cdwfs:
@mark-lunarg -- I didn't test them in isolation, but the individual commits should be fairly self-contained; one adds a new VU error report, another adds a test for that error. Would you rather I go back and fold the database updates into each individual commit, or just collapse everything into two commits (one layers, one tests)?

layers/core_validation.cpp
```diff
@@ -783,10 +783,22 @@ static bool SetMemBinding(layer_data *dev_data, VkDeviceMemory mem, uint64_t han
...
         BINDABLE *mem_binding = GetObjectMemBinding(dev_data, handle, type);
         assert(mem_binding);
-        // TODO : Add check here to make sure object isn't sparse
-        //  VALIDATION_ERROR_00792 for buffers
-        //  VALIDATION_ERROR_00804 for images
-        assert(!mem_binding->sparse);
+        if (mem_binding->sparse) {
+            UNIQUE_VALIDATION_ERROR_CODE error_code = VALIDATION_ERROR_00804;
+            const char *handle_type = "IMAGE";
+            if (strcmp(apiName, "vkBindBufferMemory") == 0) {
```
cdwfs: I meant to ask what the prevailing convention was for () at the end of function calls; thanks for reminding me. I'll add the parens presently.

tests/layer_validation_tests.cpp
cdwfs: Great catch; I'll add that check.

cdwfs:
All previous suggestions have been incorporated.

tobine:
Note that there's still a state update down the PreCallValidateBindImageMemory()->SetMemBinding() path.

No PreCallValidate* functions should update layer state. The state update should only occur after the call down the chain has succeeded, the state update should then occur down the PostCallRecord* chain.

I recommend creating a new function called "ValidateMemBinding()" that contains almost all of the code from the current SetMemBinding(). Then change SetMemBinding() to return void and the body of the function will just be the state updates that are performed in the final "else" block of the current function.
PreCallValidateBindImageMemory() would then call ValidateMemBinding() and PostCallRecordBindImageMemory() would call SetMemBinding().

cdwfs:
@tobine: The SetMemBinding() refactor is complete. Along the way, it was necessary to split BindBufferMemory() into Pre/Post calls, so that's done too now. Let me know if there's anything else you'd like to see!

mark-lunarg:
@cdwfs, please take a look at the `contributing.md` doc for guidelines on commit messages -- these look overly long.

cdwfs:
@mark-lunarg Shortened!

layers/core_validation.cpp
```diff
@@ -732,8 +732,31 @@ bool ValidateMemoryIsBoundToBuffer(const layer_data *dev_data, const BUFFER_STAT
 //  Otherwise, add reference from objectInfo to memoryInfo
 //  Add reference off of objInfo
 // TODO: We may need to refactor or pass in multiple valid usage statements to handle multiple valid usage conditions.
```
tobine: This comment now applies to ValidateSetMemBindings()

cdwfs: Done

PR #1447, opened 2017-02-03
Tobin gh880 descriptor limits

tobine:
Add 2 new checks for descriptor set buffer range limits.
Add a new test to hit those checks and 2 related limits that were untested.


tobine:

PR #1448, opened 2017-02-03
Migrate CmdClearAttachments and CmdCopyImage to buffer module

mark-lunarg:


tobine:
This looks good but running through our farm to make sure. Will approve in an hour or so if that passes.

mark-lunarg:
Pushed to Master.

PR #1452, opened 2017-02-04
GH706 -- Allow validation with NV_GLSL_SHADER extension

### LINKED ISSUE #706: Shader Validation and VK_NV_glsl_shader extension
When using the NVidia driver and enabling the VK_NV_glsl_shader extension, the shader validation appears to assume the shader must be SPIRV and throws this error:

`SC: SPIR-V module not valid: Invalid SPIR-V magic number. Exception thrown at 0x00007FF8F9DD51E4 (VkLayer_core_validation.dll) in Viewer.exe: 0xC0000005: Access violation reading location 0x000001A629FE1574.`

The extension allows GLSL shaders to be passed in instead of SPIR-V. The debug layer does not allow for non-SPIRV shaders when this extension is present.  If the exension is _enabled_ and the shader signature is not SPIRV, we should probably skip continuing with the validator.

mark-lunarg:
Fixes #706.  Apps can enable this extension and use straight GLSL, which needs to be ignored by the validation layers which assume SPIR-V.  Added a disable for shader validation, set an enable flag in the CV layer for the extension, and added conditional validation based on whether a shader-module is SPIR-V or otherwise. 

The test commit is only for context and will not be pushed.  It uses GLSL for one of the two shadermodules in cube.  This change _should_ allow a mix of validated SPIR-V shaders and ignored GLSL modules.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -3158,7 +3158,8 @@ static bool verifyPipelineCreateState(layer_data *my_data, std::vector<PIPELINE_
                              validation_error_map[VALIDATION_ERROR_02122]);
     }
 
-    if (!validate_and_capture_pipeline_shader_state(my_data->report_data, pPipeline, &my_data->enabled_features,
+    if (!GetDisables(my_data)->shader_validation &&
+        !validate_and_capture_pipeline_shader_state(my_data->report_data, pPipeline, &my_data->enabled_features,
                                                     my_data->shaderModuleMap)) {
         skip_call = true;
     }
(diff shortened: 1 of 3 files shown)


layers/core_validation.cpp
```diff
@@ -9907,19 +9913,23 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateShaderModule(VkDevice device, const VkShade
...
 
-        auto result = spvValidate(ctx, &binary, &diag);
-        if (result != SPV_SUCCESS) {
-            skip_call |= log_msg(dev_data->report_data,
-                                 result == SPV_WARNING ? VK_DEBUG_REPORT_WARNING_BIT_EXT : VK_DEBUG_REPORT_ERROR_BIT_EXT,
-                                 VkDebugReportObjectTypeEXT(0), 0, __LINE__, SHADER_CHECKER_INCONSISTENT_SPIRV, "SC",
-                                 "SPIR-V module not valid: %s", diag && diag->error ? diag->error : "(no error text)");
+        spv_valid = spvValidate(ctx, &binary, &diag);
+        if (spv_valid != SPV_SUCCESS) {
+            static const uint32_t kSpirvMagicNumber = 0x07230203;
```
karl-lunarg: spv::MagicNumber is available for your use here.

chrisforbes:
Looks like it does what it says on the tin.

This is going to have odd effects downstream when we try to track and validate resources affected by a draw-- we'll be assuming that GLSL shaders use no resources. I suppose that's slightly better than assuming everything bound is used.

Another potential direction would be to have the layer compile the GLSL to spirv itself in this case, and validate on the basis of that. That would require having the compiler available, which isn't ideal, and we may have slight mismatches between what glslang does with the shader, and what nvidia's GLSL frontend does...

layers/core_validation.cpp
```diff
@@ -259,13 +259,17 @@ struct shader_module {
     // A mapping of <id> to the first word of its def. this is useful because walking type
     // trees, constant expressions, etc requires jumping all over the instruction stream.
     unordered_map<unsigned, unsigned> def_index;
+    bool validate;
```
karl-lunarg: "validate" is a little vague and is a verb that is being used here as a state.  Something like has_valid_spirv or shader_valid might be better.  Then something like `if (shaders[vertex_stage] && shaders[vertex_stage]->has_valid_spirv) {` reads a little better, at least for me.

karl-lunarg:
@chrisforbes Yes, we talked about the option to compile the GLSL into SPIR-V and had the same concerns about mismatches and compiler availability.

mark-lunarg:
Renamed variable as Karl suggested, pushed to Master.

sergeyreznik:
@mark-lunarg How long do you think it would take to appear in new build of LunarG SDK?

mark-lunarg:
The next one is roughly scheduled for two-to-three weeks.

sergeyreznik:
@mark-lunarg is it included in 1.0.42.0?

PR #1456, opened 2017-02-07
Move/Consolidate CmdBlitImage/CmdResolveImage validation

mark-lunarg:
Split validation & state updates, refactored for pre/post updates, moved image related stuff to buffer module, removed (or changed to warnings) some bad validation checks, fixed tests and updated database when VUs were available for existing checks.

The only two items left in the image layer are the buffer/image copy routines, which have in-flight changes from DaveH.  After that, they'll get moved and this layer will be completely absorbed.


layers/buffer_validation.cpp
```diff
@@ -1269,21 +1269,170 @@ void PreCallRecordCmdResolveImage(core_validation::layer_data *device_data, GLOB
...
+
+                VkImageAspectFlags srcAspect = pRegions[i].srcSubresource.aspectMask;
+                VkImageAspectFlags dstAspect = pRegions[i].dstSubresource.aspectMask;
+
+                if (srcAspect != dstAspect) {
+                    std::stringstream ss;
+                    ss << "vkCmdBlitImage: Image aspects of depth/stencil images should match";
+                    // TODO: Verify against Valid Use section of spec, if this case yields undefined results, then it's an error
+                    skip |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
+                                    reinterpret_cast<uint64_t>(cb_node->commandBuffer), __LINE__, DRAWSTATE_INVALID_IMAGE_ASPECT,
```
tobine: This is VALIDATION_ERROR_01303

layers/core_validation.cpp
```diff
@@ -8429,34 +8429,6 @@ VKAPI_ATTR void VKAPI_CALL GetImageSubresourceLayout(VkDevice device, VkImage im
     VkSubresourceLayout *pLayout) {
     bool skipCall = false;
     layer_data *device_data = get_my_data_ptr(get_dispatch_key(device), layer_data_map);
-    VkFormat format;
```
tobine: I believe these checks are valid but it's just an issue of needing a VU for the spec language. See [this](https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VkImageAspectFlagBits) spec language:

`The mask must be only VK_IMAGE_ASPECT_COLOR_BIT, VK_IMAGE_ASPECT_DEPTH_BIT or VK_IMAGE_ASPECT_STENCIL_BIT if format is a color, depth-only or stencil-only format, respectively. If using a depth/stencil format with both depth and stencil components, aspectMask must include at least one of VK_IMAGE_ASPECT_DEPTH_BIT and VK_IMAGE_ASPECT_STENCIL_BIT, and can include both.`

mark-lunarg: I had convinced myself that this didn't apply in this case, since the API asks the driver to return the info on that particular aspect of an image, but since there's no return code and no way to report an error I am now in full agreement with you and will kill that change.

layers/buffer_validation.cpp
mark-lunarg: Done.  Check was incorrect anyhow (only validated d/s images).  Moved up and set ID.

Issue #777, opened 2016-07-19
Add validation for null pColorBlendState at pipeline creation (with rasterizer discard disabled)


SaschaWillems:
The current spec states that you must pass a valid pointer to pColorBlendState for the pipeline if you have subpasses using color attachments and rasterizer discard is not enabled:

> If the rasterizerDiscardEnable member of pRasterizationState is VK_FALSE, and subpass uses
> color attachments, pColorBlendState must be a pointer to a valid
> VkPipelineColorBlendStateCreateInfo structure

Having a renderpass with a color (and) depth attachment in a single subpass and setting pColorBlendState to NULL with rasterizerDiscardEnable = VK_FALSE for a pipeline that uses this renderpass does not trigger any validation errors or warnings.

There is a validation for the attachmentCount of the colorBlendState, but nothing if the colorBlendState is set to null.

mikew-lunarg:
This check corresponds to Valid Usage VALIDATION_ERROR_02116

mikew-lunarg:
Candidate check code is implemented on branch mikew_gh777.
Pull request is awaiting test code to exercise the check.

mikew-lunarg:
Pull request #1457

mikew-lunarg:
Merged to master at commit b502f14620de42454ef8906b6d76e9a176c52067

PR #1457, opened 2017-02-07
Implement VU 2116 check and test for GH-777

mikew-lunarg:


tests/layer_validation_tests.cpp
```diff
@@ -3294,6 +3295,53 @@ TEST_F(VkLayerTest, DisabledIndependentBlend) {
...
+
+namespace {
+
+// Perturb a good Pipeline's configuration to trigger VALIDATION_ERROR_02116
+class PipelineObj_VU2116 : public VkPipelineObj {
+   public:
+    PipelineObj_VU2116(VkDeviceObj *dev_obj) : VkPipelineObj(dev_obj) {}
+
+    VkResult init_try(const vk_testing::Device &dev, const VkGraphicsPipelineCreateInfo &info) {
+        const_cast<VkGraphicsPipelineCreateInfo &>(info).pColorBlendState = nullptr;
```
chrisforbes: Ick, having to cast away const to make this work...

I think it would be better to add this flexibility to `VkPipelineObj::CreateVKPipeline` rather than intercepting `init_try` to mangle the structure after the fact. There are precedents for this (`VkPipelineObj::Set*`), although not quite what you want if you need to be able to set the ptr to null.

layers/core_validation.cpp
```diff
@@ -3277,6 +3277,29 @@ static bool verifyPipelineCreateState(layer_data *my_data, std::vector<PIPELINE_
             }
         }
     }
+
+    // VALIDATION_ERROR_02116 : If the rasterizerization is enabled, and subpass uses color attachments, pColorBlendState must be
```
chrisforbes: Creative spelling of rasterization :)

layers/core_validation.cpp
```diff
@@ -3277,6 +3277,29 @@ static bool verifyPipelineCreateState(layer_data *my_data, std::vector<PIPELINE_
...
+    // valid pointer
+    if (pPipeline->graphicsPipelineCI.pRasterizationState &&
```
tobine: We now have 3 if statements in a row that all check "pPipeline->graphicsPipelineCI.pRasterization". See lines 3255 & 3265 above. Can you collapse them all under one check please?

layers/core_validation.cpp
```diff
@@ -3277,6 +3277,29 @@ static bool verifyPipelineCreateState(layer_data *my_data, std::vector<PIPELINE_
...
+        (pPipeline->graphicsPipelineCI.pRasterizationState->rasterizerDiscardEnable == VK_FALSE)) {
+        auto subpass_desc = renderPass ? &renderPass->createInfo.pSubpasses[pPipeline->graphicsPipelineCI.subpass] : nullptr;
+        if (subpass_desc) {
```
chrisforbes: To here, the condition is the same as the previous check. Slight preference toward combining the blocks.

layers/core_validation.cpp
```diff
@@ -3277,6 +3277,29 @@ static bool verifyPipelineCreateState(layer_data *my_data, std::vector<PIPELINE_
...
+            uint32_t color_attachment_count = 0;
+            for (uint32_t i = 0; i < subpass_desc->colorAttachmentCount; ++i) {
+                if (subpass_desc->pColorAttachments[i].attachment != VK_ATTACHMENT_UNUSED) {
+                    ++color_attachment_count;
+                }
```
chrisforbes: Meat of the check all looks correct :+1:

mikew-lunarg:
@tobine @chrisforbes Please review this latest PR.

Layer checking code has been refactored to remove duplicate checks.
Test framework has been refactored to separate initialization from execution during CreateVKPipeline()

PR #1462, opened 2017-02-08
tests: Fix UpdateBufferWithinRenderPass

renelindsay:
Change-Id: I25815aa9e09bd8f2a5f2e485ecbbcbed9dcb8275


tests/layer_validation_tests.cpp
```diff
@@ -8338,10 +8338,9 @@ TEST_F(VkLayerTest, UpdateBufferWithinRenderPass) {
     dstBuffer.init_as_dst(*m_device, (VkDeviceSize)1024, reqs);
 
     VkDeviceSize dstOffset = 0;
-    VkDeviceSize dataSize = 1024;
-    const void *pData = NULL;
-
-    vkCmdUpdateBuffer(m_commandBuffer->GetBufferHandle(), dstBuffer.handle(), dstOffset, dataSize, pData);
+    VkDeviceSize dataSize = 8;
```
tobine: Rather than statically declaring dataSize before creating the array, I'd prefer to dynamically set it after the array is created w/ something like `VkDeviceSize dataSize = (VkDeviceSize)(sizeof(Data)/sizeof(uint32_t));`
That way if array ever changes, size is automatically updated.

PR #1468, opened 2017-02-09
Function refactoring/move more image/buffer routines out of core_validation

mark-lunarg:
Separated CmdCopyImage/CmdCopyBuffer validation and state updates, refactored for pre/post call style, moved to buffer validation along with a scad of image/buffer support routines.


layers/buffer_validation.h
```diff
@@ -139,4 +139,18 @@ bool ValidateLayouts(core_validation::layer_data *dev_data, VkDevice device, con
 bool ValidateMapImageLayouts(core_validation::layer_data *dev_data, VkDevice device, DEVICE_MEM_INFO const *mem_info,
                              VkDeviceSize offset, VkDeviceSize end_offset);
 
+bool ValidateImageUsageFlags(layer_data *dev_data, IMAGE_STATE const *image_state, VkFlags desired, VkBool32 strict,
+    int32_t const msgCode, char const *func_name, char const *usage_string);
```
tobine: clang-format?

layers/buffer_validation.cpp
```diff
@@ -1886,3 +1886,157 @@ bool PreCallValidateCreateBufferView(layer_data *device_data, const VkBufferView
...
+        // Validate correct image aspect bits for desired formats and format consistency
+        skip |= ValidateImageAspectMask(device_data, image_state->image, image_format, aspect_mask, "vkCreateImageView()");
+    }
+    return skip;
+}
+
+void PostCallRecordCreateImageView(layer_data *device_data, const VkImageViewCreateInfo *create_info,
+                                                 VkImageView view) {
+    (*GetImageViewMap(device_data))[view] = std::unique_ptr<IMAGE_VIEW_STATE>(new IMAGE_VIEW_STATE(view, create_info));
+    ResolveRemainingLevelsLayers(device_data, &(*GetImageViewMap(device_data))[view].get()->create_info.subresourceRange,
```
tobine: These are a couple of difficult lines of code

layers/core_validation_types.h
```diff
@@ -742,6 +757,8 @@ SWAPCHAIN_NODE *GetSwapchainNode(const layer_data *, VkSwapchainKHR);
 GLOBAL_CB_NODE *GetCBNode(layer_data const *my_data, const VkCommandBuffer cb);
 RENDER_PASS_STATE *GetRenderPassState(layer_data const *my_data, VkRenderPass renderpass);
 FRAMEBUFFER_STATE *GetFramebufferState(const layer_data *my_data, VkFramebuffer framebuffer);
+COMMAND_POOL_NODE *GetCommandPoolNode(layer_data *dev_data, VkCommandPool pool);
+PHYS_DEV_PROPERTIES_NODE *GetPhysDevProperties(layer_data *device_data);
```
tobine: There's also functions below for GetPhysicalDeviceLimits(), GetPhysicalDevice() and GetPhysicalDeviceProperties()...
Can you take a look at those and see if consolidation makes sense? I suspect that this new function would have all of the data needed by all of the other functions, yes?

layers/buffer_validation.cpp
mark-lunarg: Yep, they are.  Unstacked them a bit.

layers/buffer_validation.h
mark-lunarg: Check.

layers/core_validation_types.h
mark-lunarg: Refactored the formatProperties and imageFormatProperties so that GetPhysicalDevice could be nuked.   Also took GetPhysicalDeviceLimits out of descriptor sets and buffer_validation and replaced those calls with GetPhysDevProperties.

mark-lunarg:
Added suggestions, pushed to Master.

PR #1470, opened 2017-02-10
Add KHR/KHX/NN extensions to threading layer

mark-lunarg:
These were skipped due to problems in codegen. Added support for new objects into the layer, fixed codegen.

### DIFF
--- a/scripts/threading_generator.py
+++ b/scripts/threading_generator.py
@@ -137,44 +137,26 @@ class ThreadOutputGenerator(OutputGenerator):
             if ((elem.tag is not 'type') and (elem.tail is not None)) and '*' in elem.tail:
                 ispointer = True
         return ispointer
+
+    # Check if an object is a non-dispatchable handle
+    def isHandleTypeNonDispatchable(self, handletype):
+        handle = self.registry.tree.find("types/type/[name='" + handletype + "'][@category='handle']")
+        if handle is not None and handle.find('type').text == 'VK_DEFINE_NON_DISPATCHABLE_HANDLE':
+            return True
+        else:
+            return False
+
+    # Check if an object is a dispatchable handle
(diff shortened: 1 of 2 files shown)


scripts/threading_generator.py
```diff
@@ -392,6 +398,7 @@ def genCmd(self, cmdinfo, name):
             'vkFreeCommandBuffers',
             'vkCreateDebugReportCallbackEXT',
             'vkDestroyDebugReportCallbackEXT',
+            'vkCreateSwapChainKHR',
```
MarkY-LunarG: We probably need a hook into vkCreateSharedSwapchainKHR as well.

stroyan: This adds a declaration for a function that would go into layers/threading.cpp.
But I don't see your changes to that file.

mark-lunarg: @MarkY-LunarG , CreateSharedSwapchains is covered by this layer.  @stroyan, I'll add the routine for CreateSwapChainKHR.

mark-lunarg:
@stroyan, removed CreateSwapchain from the special_functions list.  Also removed the hard-coded object lists and just checked the handles -- I noticed that PhysicalDevice was purposefully skipped and kept things that way, but also noticed that VkDescriptorSet was missing from the list of non-dispatchable objects and added that in (along with supporting changes).  Things seem fine this way, is there some non-obvious reason this object should be excluded?

stroyan:
The hard-coded lists of object types evolved as the list of externsync parameters expanded.
There is no need to use ReadObject counters for types that never appear in externsync parameters, (or for types that only appear as externsync parameters with WriteObject counters).  But it seems that all the handle types now are listed as externsync at least some of the time.  Dropping the hardcoded lists will make it a bit easier to maintain when new handle types appear.

PR #1471, opened 2017-02-10
layers: Fix CheckPreserved to allow input attachment

Tony-LunarG:
Attachments that are used (input / color / depth attachment)
in a subpass are not allowed in the preserved list, so if they
are used as a color, depth  or input attachment they should
be considered preserved.  Color and depth attachments were 
accounted for, this commit adds input attachment

Change-Id: Id87c3c1546092bc64b293bead1d997e0f61f7990


Tony-LunarG:
Merged

PR #1473, opened 2017-02-10
Change check for mem map for error to warning.

lentinem:
Currently the spec allows memory to be mapped regardless of format.
However, because the memory cannot be used by the device in a predicable
way unless the format is PREINITIALIZED or GENERAL we still need a
warning here.


mark-lunarg:
Pushed to Master.

PR #1477, opened 2017-02-13
tests: vk_layer_validation_tests depends on layer targets.

cdwfs:
After my fifth time of forgetting to rebuild the layers before running the validation tests, I decided to add an explicit build dependency in CMake. I figured it might help others as well, but feel free to reject if it doesn't seem useful.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -114,6 +114,14 @@ endif()
 if(WIN32)
    target_link_libraries(vk_layer_validation_tests ${LIBVK} gtest gtest_main VkLayer_utils ${GLSLANG_LIBRARIES})
 endif()
+add_dependencies(vk_layer_validation_tests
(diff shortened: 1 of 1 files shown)


tests/CMakeLists.txt
```diff
@@ -122,4 +131,4 @@ set_target_properties(vk_loader_validation_tests
 target_link_libraries(vk_loader_validation_tests ${LIBVK} gtest gtest_main VkLayer_utils ${GLSLANG_LIBRARIES})
 
 add_subdirectory(gtest-1.7.0)
-add_subdirectory(layers)
+add_subdirectory(layers)
```
stroyan: You should put the newline at end of file back in.

cdwfs:
Trailing newline restored. I also removed the dependency on the image layer (RIP).

Issue #1478, opened 2017-02-14
Confusing messages when command buffer is in CB_INVALID state


Novum:
In case a CB gets invalidated the validation layer will complain about 
    "You must call vkBeginCommandBuffer() before this call to %s"

This is really confusing. It should say "the command buffer was invalidated" or something similar. I spent quite some time trying to figure out why the CB was not started before realizing this.

mark-lunarg:
Fixed in PR #1479.

PR #1479, opened 2017-02-14
GH1478, Clarify Invalidated-CB Validation Error

mark-lunarg:
Added a bit of state to report_error_no_cb_begin so that if the state was marked as CB_INVALID it would output a different message than "Missing BeginCommandBuffer call".  Fixes GH #1478.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -7341,32 +7341,36 @@ VKAPI_ATTR void VKAPI_CALL CmdBindIndexBuffer(VkCommandBuffer commandBuffer, VkB
     auto buffer_state = GetBufferState(dev_data, buffer);
     auto cb_node = GetCBNode(dev_data, commandBuffer);
     if (cb_node && buffer_state) {
-        skip_call |= ValidateMemoryIsBoundToBuffer(dev_data, buffer_state, "vkCmdBindIndexBuffer()", VALIDATION_ERROR_02543);
-        std::function<bool()> function = [=]() {
-            return ValidateBufferMemoryIsValid(dev_data, buffer_state, "vkCmdBindIndexBuffer()");
-        };
-        cb_node->validate_functions.push_back(function);
-        skip_call |= ValidateCmd(dev_data, cb_node, CMD_BINDINDEXBUFFER, "vkCmdBindIndexBuffer()");
-        UpdateCmdBufferLastCmd(cb_node, CMD_BINDINDEXBUFFER);
-        VkDeviceSize offset_align = 0;
-        switch (indexType) {
-            case VK_INDEX_TYPE_UINT16:
-                offset_align = 2;
-                break;
-            case VK_INDEX_TYPE_UINT32:
-                offset_align = 4;
-                break;
-            default:
-                // ParamChecker should catch bad enum, we'll also throw alignment error below if offset_align stays 0
-                break;
+        if (cb_node->state == CB_RECORDING) {
+            skip_call |= ValidateMemoryIsBoundToBuffer(dev_data, buffer_state, "vkCmdBindIndexBuffer()", VALIDATION_ERROR_02543);
+            std::function<bool()> function = [=]() {
+                return ValidateBufferMemoryIsValid(dev_data, buffer_state, "vkCmdBindIndexBuffer()");
+            };
+            cb_node->validate_functions.push_back(function);
+            skip_call |= ValidateCmd(dev_data, cb_node, CMD_BINDINDEXBUFFER, "vkCmdBindIndexBuffer()");
+            UpdateCmdBufferLastCmd(cb_node, CMD_BINDINDEXBUFFER);
+            VkDeviceSize offset_align = 0;
+            switch (indexType) {
+                case VK_INDEX_TYPE_UINT16:
+                    offset_align = 2;
+                    break;
+                case VK_INDEX_TYPE_UINT32:
+                    offset_align = 4;
+                    break;
+                default:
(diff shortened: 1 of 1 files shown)


mark-lunarg:
Yep, will do.

mark-lunarg:
This seems much better.

layers/core_validation.cpp
```diff
@@ -3440,12 +3434,28 @@ bool ValidateCmd(layer_data *dev_data, GLOBAL_CB_NODE *pCB, const CMD_TYPE cmd,
                 break;
         }
     }
-    if (pCB->state != CB_RECORDING) {
-        skip_call |= report_error_no_cb_begin(dev_data, pCB->commandBuffer, caller_name);
+    if (cb_state->state != CB_RECORDING) {
+        if (cb_state->state == CB_INVALID) {
+            for (auto obj : cb_state->broken_bindings) {
+                const char *type_str = object_type_to_string(obj.type);
+                // Descriptor sets are a special case that can be either destroyed or updated to invalidated a CB
```
tobine: typo "invalidated"->"invalidate"

layers/core_validation.cpp
```diff
@@ -3440,12 +3434,28 @@ bool ValidateCmd(layer_data *dev_data, GLOBAL_CB_NODE *pCB, const CMD_TYPE cmd,
...
+                const char *cause_str =
+                    (obj.type == VK_DEBUG_REPORT_OBJECT_TYPE_DESCRIPTOR_SET_EXT) ? "destroyed or updated" : "destroyed";
+                skip |= log_msg(
+                    dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
+                    reinterpret_cast<uint64_t &>(cb_state->commandBuffer), __LINE__, DRAWSTATE_INVALID_COMMAND_BUFFER, "DS",
+                    "%s: You are submitting command buffer 0x%p that is invalid because bound %s 0x%" PRIxLEAST64 " was %s.",
```
tobine: Instead of "submitting" should be "add cmd <CMD_STRING> to"

layers/core_validation.cpp
```diff
@@ -8511,45 +8490,38 @@ VKAPI_ATTR void VKAPI_CALL CmdPushConstants(VkCommandBuffer commandBuffer, VkPip
...
-                                 "vkCmdPushConstants() Push constant range [%d, %d) "
-                                 "with stageFlags = 0x%" PRIx32
-                                 " "
-                                 "not within flag-matching ranges in pipeline layout 0x%" PRIx64 ". %s",
-                                 offset, offset + size, (uint32_t)stageFlags, (uint64_t)layout,
-                                 validation_error_map[VALIDATION_ERROR_00988]);
+            skip |= log_msg(
+                dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0, __LINE__,
+                VALIDATION_ERROR_00988, "DS", "vkCmdPushConstants() Push constant range [%d, %d) with stageFlags = 0x%" PRIx32
+                                              "not within flag-matching ranges in pipeline layout 0x%" PRIx64 ". %s",
```
tobine: Killed a space before "not" that is causing InvalidPushConstants test to fail

layers/core_validation.cpp
```diff
@@ -3440,12 +3434,28 @@ bool ValidateCmd(layer_data *dev_data, GLOBAL_CB_NODE *pCB, const CMD_TYPE cmd,
                 break;
         }
     }
-    if (pCB->state != CB_RECORDING) {
-        skip_call |= report_error_no_cb_begin(dev_data, pCB->commandBuffer, caller_name);
+    if (cb_state->state != CB_RECORDING) {
+        if (cb_state->state == CB_INVALID) {
+            for (auto obj : cb_state->broken_bindings) {
```
tobine: It may make sense to roll this block of code into its own function (ReportInvalidCommandBuffer()) and call it here and from validateCommandBufferState(). You'd have to add an extra string to slightly customize the error message based on if cmd buffer is being submitted or having a cmd added to it.

layers/core_validation.cpp
```diff
@@ -3440,12 +3434,28 @@ bool ValidateCmd(layer_data *dev_data, GLOBAL_CB_NODE *pCB, const CMD_TYPE cmd,
...
+                const char *type_str = object_type_to_string(obj.type);
+                // Descriptor sets are a special case that can be either destroyed or updated to invalidated a CB
+                const char *cause_str =
+                    (obj.type == VK_DEBUG_REPORT_OBJECT_TYPE_DESCRIPTOR_SET_EXT) ? "destroyed or updated" : "destroyed";
+                skip |= log_msg(
+                    dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
```
tobine: Just a general comment that it may be argued that adding a command to an invalid command buffer is, in itself, a perfectly fine thing to do and doesn't warrant an error, but instead a warning.
Since the only way to "fix" an invalid cmd buffer is to reset it, though, I think it's better to error on the side of caution and make it an error, at least initially.
We already have a warning at the point at which the cmd buffer becomes invalid so if a user didn't see that, then it makes sense to bump it up to error when they try to add commands to the invalid cmd buffer.

mark-lunarg:
Implemented excellent review suggestions and pushed to Master.

PR #1482, opened 2017-02-15
tests: Add some env var testing to loader tests.

karl-lunarg:
Tests recently added env var parsing code added to loader.

Change-Id: Ic58d92dfeaed725f02a8f4a79c60d9b4b363aa3f

### DIFF
--- a/tests/run_loader_tests.sh
+++ b/tests/run_loader_tests.sh
@@ -2,6 +2,62 @@
 
 pushd $(dirname "$0") > /dev/null
 
+RunEnvironmentVariablePathsTest()
+{
+    # Check for proper handling of paths specified via environment variables.
+
+    # Set up a layer path that includes default and user-specified locations,
+    # so that the test app can find them.  Include some badly specified elements as well.
+    vk_layer_path="$VK_LAYER_PATH"
+    vk_layer_path+=":/usr/local/etc/vulkan/implicit_layer.d:/usr/local/share/vulkan/implicit_layer.d"
+    vk_layer_path+=":/tmp/carol:::"
+    vk_layer_path+=":/etc/vulkan/implicit_layer.d:/usr/share/vulkan/implicit_layer.d:$HOME/.local/share/vulkan/implicit_layer.d"
+    vk_layer_path+=":::::/tandy:"
+
+    # Set vars to include some "challenging" paths and run the test.
+    output=$(VK_LOADER_DEBUG=all \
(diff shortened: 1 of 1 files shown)


tests/run_loader_tests.sh
```diff
@@ -2,6 +2,62 @@
...
+       XDG_DATA_DIRS="::::/tmp/goober4:::::/tmp/goober5:/tmp/goober6/with spaces::::/tmp/goober7:" \
+       VK_LAYER_PATH=${vk_layer_path} \
+       GTEST_FILTER=CreateInstance.LayerPresent \
+       ./vk_loader_validation_tests 2>&1)
+
+    # Here is a path we expect to find.  The loader constructs these from the XDG* env vars. 
+    right_path="/tmp/goober/vulkan/icd.d:/tmp/goober2//vulkan/icd.d:/tmp/goober3/with spaces/vulkan/icd.d"
+    # There are other paths that come from SYSCONFIG settings established at build time.
+    # So we can't really guess at what those are here.
+    right_path+=".*"
```
MarkY-LunarG: Actually, do you need a delimiter first?

tests/run_loader_tests.sh
```diff
@@ -2,6 +2,62 @@
...
+    # Also expect to find these, since we added them.
+    right_path+="/tmp/goober4/vulkan/icd.d:/tmp/goober5/vulkan/icd.d:/tmp/goober6/with spaces/vulkan/icd.d:/tmp/goober7/vulkan/icd.d"
```
MarkY-LunarG: Delimiter first?

karl-lunarg:
> Test looks fine, but any idea why it failed in clang? Or was that another change?

That was caused by the problem fixed with #1481.  This branch didn't have that commit.  I just rebased and pushed.  Travis should rebuild it cleanly.

tests/run_loader_tests.sh
karl-lunarg: Maybe not *need*.  But it might be better to specify
`":.*:"` so as to be explicit about the delimiters that would separate line 26 and line 31.
But the `".*"` will match those delimiters anyway, so it works as-is.   I'll think about this change.

tests/run_loader_tests.sh
karl-lunarg: Addressed above.

mark-lunarg:
Hey Karl, is this good to push?

karl-lunarg:
> Hey Karl, is this good to push?
I think so.  Go ahead and push if you'd like.

PR #1484, opened 2017-02-15
tests: Prevent new unexpected errors

jeremy-lunarg:
Add a filter to the error monitor and require explicit function for
unexpected errors. This helps developers easily detect when they
inadvertently add an unexpected error.

Change-Id: I652951fb135795770b06b6805155cd5e4a0e0c62

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -36,6 +36,8 @@
 #include "vk_layer_config.h"
 #include "vk_validation_error_messages.h"
 #include "vkrenderframework.h"
+
+#include <algorithm>
 #include <limits.h>
 #include <unordered_set>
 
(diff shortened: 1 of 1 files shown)


tobine:
This looks great and I'm about the approve.
Ran this on our farm and confirmed that Linux/nvidia run doesn't report any of the unexpected errors in the log. Note that on android there are still a number of unexpected errors, available to review in 
[android_unexpected_errors.txt](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/files/778351/android_unexpected_errors.txt)

If you're feeling saintly, please add these as well in a follow-on. Otherwise we can add the exception filters on our end at some point.

jeremy-lunarg:
Yowzers... I hadn't even thought about that. How would you feel about an OS specific entry point  (e.g. SetAndroidUnexpectedError)?. My ongoing battle is to remove as many as these as possible but if I fix it for win/lin, it might still persist on android.

PR #1486, opened 2017-02-16
Complete migration and removal of image layer

mark-lunarg:
Consolidated and refactored image-buffer copy validation including separating validation from state updates and pre/post refactoring; fixed up the layer lists in demos and loader; removed image layer refs from documents, scripts, cmake/build scripts, and deleted the files.


layers/buffer_validation.cpp
```diff
@@ -2602,27 +2599,233 @@ bool ValidateBufferImageCopyData(layer_data *device_data, uint32_t regionCount,
...
+        skip |= VerifySourceImageLayout(device_data, cb_node, src_image_state->image, pRegions[i].imageSubresource, srcImageLayout,
+                                        VALIDATION_ERROR_01251);
+        skip |= ValidateCopyBufferImageTransferGranularityRequirements(device_data, cb_node, src_image_state, &pRegions[i], i,
+                                                                       "CmdCopyImageToBuffer");
+    }
     return skip;
 }
 
-bool PreCallValidateCmdCopyBufferToImage(layer_data *dev_data, VkImage dstImage, uint32_t regionCount,
+bool PreCallValidateCmdCopyBufferToImage(layer_data *device_data, VkImageLayout dstImageLayout, GLOBAL_CB_NODE *cb_node,
```
mikew-lunarg: nit: since renaming anyway, rename dstImageLayout per style guide

layers/core_validation.cpp
```diff
@@ -7732,12 +7732,15 @@ VKAPI_ATTR void VKAPI_CALL CmdCopyBufferToImage(VkCommandBuffer commandBuffer, V
...
                                                         regionCount, pRegions, "vkCmdCopyBufferToImage()");
     } else {
+        lock.unlock();
         assert(0);
         // TODO: report VU01244 here, or put in object tracker?
     }
-    lock.unlock();
-    if (!skip_call)
+    if (!skip_call) {
+        lock.unlock();
```
mikew-lunarg: Is unlock() meant to go before or after PreCallRecord?
Here it is before, below it is after.

layers/buffer_validation.cpp
mark-lunarg: Mike -- we don't typically rename Vulkan API arguments, even as they get passed down into helper functions.  This one comes in this way from the API and is consistent with regionCount, pRegions, etc.

layers/core_validation.cpp
mark-lunarg: Great catch!

PR #1490, opened 2017-02-17
layers:Refactor image layout validation

tobine:
Streamline some of the image layout validation code. At the time of
CmdBeginRenderPass() remove the layout transition from validation code
and perform the layout transitions all at once.
This allows a bit of transition code to be killed so that the initial
transition code can share TransitionSubpassLayouts() function.


chrisforbes:
@tobine is there any reason not to land this?

tobine:
@chrisforbes Nope, guess I just spaced this one. There's a merge conflict now so created #1659 to replace this one

PR #1493, opened 2017-02-19
loader: Configurable fallback search paths

Ralith:
This makes it easier for non-FHS distributions to behave well when the loader is used by a SUID process or in an otherwise unusual environment.


MarkY-LunarG:
Thanks for changes.

Issue #373, opened 2016-04-15
Validate memory allocation size in device_limits


karl-lunarg:
Moved from gitlab 117

When binding memory to an image or buffer the size of the memory must be of at least size specified with getimage/buffer memory requirements. Additionally, the returned memory type bits must be adhered to. This can be validated by querying this information when the physical device is created within mem tracker and then checking it at bind time with the properties of the allocation stored at alloc time.

Norbert Nopper @FslNopper  2 months ago
Also, do bounds checking in the mem tracker when uploading data to the memory.

tobine:
Assigning to @cdwfs as a good follow-on to his recent work. In a quick check I don't believe that we've validating this yet. This should be done for both buffers and images. We do make sure that that <img/buff_state_ptr>->requirements has the requirements, and we check alignments, but not size or memoryTypeBits that I can see.

Also in a quick check of the code I see that Insert[Buffer|Image]MemoryRange() is being called from the PreCallValidate* functions and updating the state. This should be re-factored so that the state update only occurs if validation passes.

cdwfs:
Thanks, I'll get right on it while the Bind*Memory() code is fresh in my head. Could you point me towards the relevant VU codes that still need to be checked/reported/tested? If I'm not mistaken, it should just be the size-related ones; the memory type bits were already being checked (see the ValidateMemoryTypes() function). I added a test for this in my last PR (VU 797, 806).

tobine:
Errors VALIDATION_ERROR_00797 & VALIDATION_ERROR_00806 related to memoryTypeBits say they're already implemented with tests in BindInvalidMemory so shouldn't need to do anything there after all.
VALIDATION_ERROR_02175 is for size limit on vkBindBufferMemory() and still needs to be done.
VALIDATION_ERROR_02179 is the equivalent check for Images.

cdwfs:
Fixed with #1495

PR #1495, opened 2017-02-21
GH-373: Add size/alignment checks to BindBufferMemory and BindImageMemory

cdwfs:
In addition to the new VU checks/tests for #373, this PR refactors Insert*MemoryRange() into pre/post-call format, and fixes some copy-paste errors in existing BindBufferMemory log_msg() calls.


mark-lunarg:
Hey Cort, first character after the : in commit messages should be capitalized and the subject line should be right at 50 characters.

layers/core_validation.cpp
```diff
@@ -5478,26 +5513,38 @@ static bool PreCallValidateBindBufferMemory(layer_data *dev_data, VkBuffer buffe
...
+                            buffer_handle, __LINE__, VALIDATION_ERROR_02174, "DS",
                             "vkBindBufferMemory(): memoryOffset is 0x%" PRIxLEAST64
                             " but must be an integer multiple of the "
                             "VkMemoryRequirements::alignment value 0x%" PRIxLEAST64
                             ", returned from a call to vkGetBufferMemoryRequirements with buffer. %s",
                             memoryOffset, buffer_state->requirements.alignment, validation_error_map[VALIDATION_ERROR_02174]);
         }
 
+        // Validate memory requirements size
+        if (buffer_state->requirements.size > mem_info->alloc_info.allocationSize - memoryOffset) {
```
cdwfs: Done

layers/core_validation.cpp
```diff
@@ -5539,6 +5586,12 @@ static void PostCallRecordBindBufferMemory(layer_data *dev_data, VkBuffer buffer
     std::unique_lock<std::mutex> lock(global_lock);
     auto buffer_state = GetBufferState(dev_data, buffer);
```
cdwfs: Done (in BindImageMemory as well)

layers/core_validation.cpp
```diff
@@ -5339,7 +5366,7 @@ static bool InsertMemoryRange(layer_data const *dev_data, uint64_t handle, DEVIC
         auto check_range = &obj_range_pair.second;
         bool intersection_error = false;
         if (rangesIntersect(dev_data, &range, check_range, &intersection_error)) {
-            skip_call |= intersection_error;
+            assert(!intersection_error);  // should have been caught in ValidateInsertMemoryRange()
```
cdwfs: Yeah, I should give this another look before I push.

cdwfs: I like the idea of letting callers opt out of the validation check in rangesIntersect(). I'm _almost_ tempted to hoist the whole check out into the caller, but there are too many call sites (and thus too much code duplication) for that . Would you prefer that I added a separate parameter to skip the check, or would it be sufficient to add an "if skip_call == NULL" check inside rangesIntersect(), and have calls from the state-update path pass NULL?

PR #1496, opened 2017-02-22
Check all devices for valid object handles

mikes-lunarg:
Add additional check in ValidateObject to distinguish between an invalid object and an object that was created on another device.  Add new layer test for this check.


layers/object_tracker.cpp
```diff
@@ -335,21 +335,42 @@ static void DestroyObject(T1 dispatchable_object, T2 object, VkDebugReportObject
...
+                        other_device_data.second->object_map[object_type].end() ||
+                    (object_type == VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT &&
+                     other_device_data.second->swapchainImageMap.find(object_handle) !=
+                         other_device_data.second->swapchainImageMap.end())) {
+                    // Object found on other device, report an error if object has a device parent error code
+                    if (wrong_device_code != VALIDATION_ERROR_UNDEFINED) {
+                        return log_msg(device_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, object_type, object_handle,
+                                       __LINE__, wrong_device_code, LayerName,
+                                       "Object 0x%" PRIxLEAST64
+                                       " was not created, allocated or retreived from the correct device. %s",
```
tobine: spelling: retrieved

layers/object_tracker.cpp
```diff
@@ -335,21 +335,42 @@ static void DestroyObject(T1 dispatchable_object, T2 object, VkDebugReportObject
...
 
     layer_data *device_data = GetLayerDataPtr(get_dispatch_key(dispatchable_object), layer_data_map);
+    // Look for object in device object map
     if (device_data->object_map[object_type].find(object_handle) == device_data->object_map[object_type].end()) {
         // If object is an image, also look for it in the swapchain image map
         if ((object_type != VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT) ||
             (device_data->swapchainImageMap.find(object_handle) == device_data->swapchainImageMap.end())) {
-            const char *error_msg = (error_code == VALIDATION_ERROR_UNDEFINED) ? "" : validation_error_map[error_code];
+            // Object not found, look for it in other device object maps
+            for (auto other_device_data : layer_data_map) {
```
tobine: This will attempt a second look-up in the known-missing "device_data" layer data map. Not ideal, but not horrible since this code is only hit in the error case which is already a slow path. Probably worthwhile to skip the map that's already been checked, though.

tests/layer_validation_tests.cpp
```diff
@@ -2389,6 +2389,43 @@ TEST_F(VkLayerTest, CreateUnknownObject) {
...
+    device_create_info.ppEnabledLayerNames = NULL;
+    device_create_info.pEnabledFeatures = &features;
+
+    VkDevice secondDevice;
+    ASSERT_VK_SUCCESS(vkCreateDevice(gpu(), &device_create_info, NULL, &secondDevice));
+
+    // Try to destroy the renderpass from the first device using the second device
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_00399);
+    vkDestroyRenderPass(secondDevice, m_renderPass, NULL);
+    m_errorMonitor->VerifyFound();
```
tobine: Need to destroy secondDevice here?

tests/layer_validation_tests.cpp
```diff
@@ -2389,6 +2389,43 @@ TEST_F(VkLayerTest, CreateUnknownObject) {
...
+    auto features = m_device->phy().features();
+    device_create_info.sType = VK_STRUCTURE_TYPE_DEVICE_CREATE_INFO;
+    device_create_info.pNext = NULL;
+    device_create_info.queueCreateInfoCount = 1;
+    device_create_info.pQueueCreateInfos = &queue_info;
+    device_create_info.enabledLayerCount = 0;
+    device_create_info.ppEnabledLayerNames = NULL;
+    device_create_info.pEnabledFeatures = &features;
+
+    VkDevice secondDevice;
```
tobine: Avoid camelCase for var names

mikes-lunarg:
Folding in review changes and rebasing on master

PR #1498, opened 2017-02-22
GH284 - Fix buffer size check on d+s image copies, squash a stencil-only test framework bug.

daveh-lunarg:
Fix aspect bit for DS in framework : 
Fixed a bug in VkDepthStencilObj::Init() that allowed only depth or depth+stencil formats. Added support for stencil-only.

Fix a bug in depth/stencil packing : 
Fixed the buffer size calculation when packing the depth or stencil component of a D+S image in/out of a buffer. Added test code to layer_validation_tests ImageBufferCopyTests().


layers/buffer_validation.cpp
```diff
@@ -2685,6 +2685,27 @@ static inline bool ValidtateBufferBounds(const debug_report_data *report_data, I
         VkDeviceSize buffer_height = (0 == pRegions[i].bufferImageHeight ? copy_extent.height : pRegions[i].bufferImageHeight);
         VkDeviceSize unit_size = vk_format_get_size(image_state->createInfo.format);  // size (bytes) of texel or block
 
+        // Handle special buffer packing rules for specific depth/stencil formats
+        if (pRegions[i].imageSubresource.aspectMask & VK_IMAGE_ASPECT_STENCIL_BIT) {
```
tobine: If we're going to need these checks anywhere else should roll it into a utility function that takes the format and aspect and returns size.

daveh-lunarg: Agreed, but reviewing the spec this seems to be a one-off, only used for this specific case of copying one aspect of a depth+stencil image in/out of a buffer.

PR #1500, opened 2017-02-22
GH1489 -- Bad CmdClearAttachments validation for secondary command buffers

### LINKED ISSUE #1489: PreCallValidateCmdClearAttachments - validating clear rectangles against unset render area, if vkCmdClearAttachments was recorded to a secondary command buffer
The PreCallValidateCmdClearAttachments does the following check:

if(cb_node->activeRenderPass) 
{
   for (uint32_t j = 0; j < rectCount; j++)  
   {
      if (false == ContainsRect(cb_node->activeRenderPassBeginInfo.renderArea, pRects[j].rect)) { // code }
   }
}

The check is only valid if both activeRenderPass and activeRenderPassBeginInfo are set. If vkCmdClearAttachments is recorded to a secondary command buffer, vkCmdBeginRenderPass has not been called yet, therefore activeRenderPassBeginInfo is not set. On the other hand activeRenderPass is set as it is taken from VkCommandBufferBeginInfo::pInheritanceInfo::renderPass. It causes a situation in which the clear rectangles are compared against a zero render area, which triggers a validation error.

mark-lunarg:
Fixes #1489:  RenderPass might not have been begun when a secondary command buffer has CmdClearAttachments recorded into it.  This validation will need to be deferred until QueueSubmit time -- a new 'validation incomplete' issue will be filed for this check.

This fix skips the record-time validation for the offending check if the renderArea data is not yet valid.

Added a positive test for this situation (thanks Tony).

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -1392,8 +1392,8 @@ bool PreCallValidateCmdClearAttachments(layer_data *device_data, VkCommandBuffer
         if (!hasDrawCmd(cb_node) && (cb_node->activeRenderPassBeginInfo.renderArea.extent.width == pRects[0].rect.extent.width) &&
             (cb_node->activeRenderPassBeginInfo.renderArea.extent.height == pRects[0].rect.extent.height)) {
             // There are times where app needs to use ClearAttachments (generally when reusing a buffer inside of a render pass)
-            // Can we make this warning more specific? I'd like to avoid triggering this test if we can tell it's a use that must
-            // call CmdClearAttachments. Otherwise this seems more like a performance warning.
+            // This warning should be made more specific. It'd be best to avoid triggering this test if it's a use that must call
+            // CmdClearAttachments.
             skip |=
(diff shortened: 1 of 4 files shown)


tobine:
Alright, I see #1501 so I'm thinking that this change is just to address the bug. I'm ok with it in that case. For the real fix, when validation is implemented at QueueSubmit time, we should just kill the initial, speculative validation and do all validation of this case at QueueSubmit time so the code is all in one place.

chrisforbes:
Isn't the new flag redundant with the CB type? -- you're never going to have this state present at recording time within a secondary command buffer.

I don't see why anything needs to be deferred to `QueueSubmit`. For the troublesome case, you have the information you need at `CmdExecuteCommands`.

mark-lunarg:
@chrisforbes, we talked about doing the validation at CEC-time, but figured it was safer at QueueSubmit-time since folks get touchy about getting errors for doing bad things they don't submit.  I was thinking that if the primary CB was recorded before the secondary that the info in activeRenderPassBeginInfo might be available.  I'll do some tests in the morning and if it's not will remove the flag noise and just check for CB type.

mark-lunarg:
Unsurprisingly, @chrisforbes was correct -- simplified by checking CB level and pushed to Master.

Issue #168, opened 2016-03-23
validation: device_limits : BindDescriptorSet could assert dynamic offsets are multiple of minUniformBufferOffsetAlignment


mccaffrey-jonathan:
layers/device_limits.cpp contains checks that the offset  member of the VkDescriptorBufferInfo structure is a multiple of minUniformBufferOffsetAlignment .

Another requirement that could be checked is that dynamic offsets for dynamic uniform buffer passed in BindDescriptorSet are multiples of minUniformBufferOffsetAlignment .

dustin-lunarg:
It looks like the core_validation layer is already validating the vkCmdBindDescriptorSets pDynamicOffsets values against the minUniformBufferOffsetAlignment limit.  See [here](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/core_validation.cpp#L7416).

@mccaffrey-jonathan Did you encounter a case where a dynamic offset was invalid and no validation error was generated?

dustin-lunarg:
This case is handled by the core_validation layer.

PR #1502, opened 2017-02-24
Fix bad compressed texture size validation errors

mark-lunarg:
This is for gitlab issue #168.

The image<->buffer copy validation was checking against texel block sizes, but needed the block size in bytes. Added a new helper function to access these values and corrected the bad validation errors and renamed existing util function to match the spec.

### DIFF
--- a/layers/vk_layer_utils.cpp
+++ b/layers/vk_layer_utils.cpp
@@ -189,7 +189,7 @@ const std::map<VkFormat, VULKAN_FORMAT_INFO> vk_format_table = {
     {VK_FORMAT_ETC2_R8G8B8A1_UNORM_BLOCK,   {8, 4, VK_FORMAT_COMPATIBILITY_CLASS_ETC2_RGBA_BIT}},
     {VK_FORMAT_ETC2_R8G8B8A1_SRGB_BLOCK,    {8, 4, VK_FORMAT_COMPATIBILITY_CLASS_ETC2_RGBA_BIT}},
     {VK_FORMAT_ETC2_R8G8B8A8_UNORM_BLOCK,   {16, 4, VK_FORMAT_COMPATIBILITY_CLASS_ETC2_EAC_RGBA_BIT}},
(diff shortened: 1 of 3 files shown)


layers/vk_layer_utils.cpp
```diff
@@ -651,6 +651,74 @@ VK_LAYER_EXPORT VkExtent2D vk_format_compressed_block_size(VkFormat format) {
...
+    case VK_FORMAT_ASTC_10x10_UNORM_BLOCK:
+    case VK_FORMAT_ASTC_10x10_SRGB_BLOCK:
+    case VK_FORMAT_ASTC_12x10_UNORM_BLOCK:
+    case VK_FORMAT_ASTC_12x10_SRGB_BLOCK:
+    case VK_FORMAT_ASTC_12x12_UNORM_BLOCK:
+    case VK_FORMAT_ASTC_12x12_SRGB_BLOCK:
+        block_size_in_bytes = 16;
+        break;
+    default:
+        break;
```
tobine: I'm thinking an assert would be good here. Shouldn't call this function with non-compressed format. Agree?

layers/vk_layer_utils.cpp
```diff
@@ -189,7 +189,7 @@ const std::map<VkFormat, VULKAN_FORMAT_INFO> vk_format_table = {
     {VK_FORMAT_ETC2_R8G8B8A1_UNORM_BLOCK,   {8, 4, VK_FORMAT_COMPATIBILITY_CLASS_ETC2_RGBA_BIT}},
     {VK_FORMAT_ETC2_R8G8B8A1_SRGB_BLOCK,    {8, 4, VK_FORMAT_COMPATIBILITY_CLASS_ETC2_RGBA_BIT}},
```
tobine: is this one correct?

mark-lunarg: These all match the spec (search for "64-bit compressed texel block"), but I am suspicious of VK_FORMAT_ETC2_R8G8B8A8_SRGB_BLOCK.

mark-lunarg:
Fix pushed to Master, cherry-picked into GL 1.0 branch for gitlab issue #168.

PR #1503, opened 2017-02-25
layers: Added more format checks to utils

ardaLunarG:


layers/vk_layer_utils.cpp
```diff
@@ -351,6 +351,95 @@ VK_LAYER_EXPORT bool vk_format_is_norm(VkFormat format) {
     return is_norm;
 };
 
+// Return true if format is of time UNORM
```
mark-lunarg: Hey Arda, I think all of these comments where they say 'time' they must mean 'type'.

ardaLunarG:
Corrected commands

ardaLunarG:
merged

PR #1504, opened 2017-02-25
layers: Fix invalid push constant checks (GH953)

### LINKED ISSUE #953: Layers: Update Push Constants Range Checks to Disallow Multiple Ranges in a Single Block
https://github.com/KhronosGroup/Vulkan-Docs/issues/340 is being approved within Khronos, so need to update checks and tests.

karl-lunarg:
Rework of invalid push constant checks, motivated mostly
by some spec updates and previous bad assumptions.

- Add check/test for a shader stage flag being set more
  than once in CreatePipelineLayout.
- Remove checks and tests having to do with warning about
  overlapping ranges in CreatePipelineLayout.  Ranges can
  certainly overlap when used in different stages.  And
  the prohibition of more than one range with the same
  stage flag removes the possibility of overlap in the same
  stage.  These checks were warnings and so probably not
  noticed.

Fixes #953
Finishes VU 00871

Change-Id: Icc9d3d84fa3132f266075f2f1d45c243e9e1a65f


layers/core_validation.cpp
```diff
@@ -6522,23 +6522,20 @@ VKAPI_ATTR VkResult VKAPI_CALL CreatePipelineLayout(VkDevice device, const VkPip
     }
     if (skip_call) return VK_ERROR_VALIDATION_FAILED_EXT;
 
-    // Each range has been validated.  Now check for overlap between ranges (if they are good).
-    // There's no explicit Valid Usage language against this, so issue a warning instead of an error.
+    // As of 1.0.28, there is a VU that states that a stage flag cannot appear more then once
```
mark-lunarg: then->than.  This might also fit on a single line.

layers/core_validation.cpp
```diff
@@ -6522,23 +6522,20 @@ VKAPI_ATTR VkResult VKAPI_CALL CreatePipelineLayout(VkDevice device, const VkPip
...
-                                     __LINE__, DRAWSTATE_PUSH_CONSTANTS_ERROR, "DS",
-                                     "vkCreatePipelineLayout() call has push constants with "
-                                     "overlapping ranges: %u:[%u, %u), %u:[%u, %u)",
-                                     i, minA, maxA, j, minB, maxB);
+            if (0 != (pCreateInfo->pPushConstantRanges[i].stageFlags & pCreateInfo->pPushConstantRanges[j].stageFlags)) {
+                skip_call |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, (VkDebugReportObjectTypeEXT)0, 0,
+                                     __LINE__, VALIDATION_ERROR_00871, "DS",
+                                     "vkCreatePipelineLayout() A stage flag appears more than once in push constant range list. %s",
+                                     validation_error_map[VALIDATION_ERROR_00871]);
+                goto pcr_loop_exit;
```
mark-lunarg: This _would be_ the only use of 'goto' in the layers.  Just sayin'.

tests/layer_validation_tests.cpp
```diff
@@ -6891,69 +6891,35 @@ TEST_F(VkLayerTest, InvalidPushConstants) {
...
+    // to the push constant range that has that shader's stage flag set.
+    // The shader's stage flag can only appear once in all the ranges, so the
+    // implementation can find the one and only range to map it to.
     const uint32_t ranges_per_test = 5;
-    struct OverlappingRangeTestCase {
+    struct DuplicateStageFlagsTestCase {
         VkPushConstantRange const ranges[ranges_per_test];
-        std::vector<char const *> const msg;
+    };
+    const std::array<DuplicateStageFlagsTestCase, 2> duplicate_stageFlags_tests = {
```
mikew-lunarg: // overlapping ranges are ok, but stage flag must appear only once.

layers/core_validation.cpp
tobine: Yes, just finishing the loops is what I would prefer. Since the app determines what to return in the callback, it's possible that the app could return false for an early case, but then true for a later case. An early return does not allow for that possibility.

layers/core_validation.cpp
karl-lunarg: fixed

layers/core_validation.cpp
karl-lunarg: OK, fixed as described.

karl-lunarg: Allowing multiple errors complicates things a bit for the layer test system, which doesn't handle multiple expected errors with the same error text.  Partly to combat this, but mostly to make a better error report, I added the indices of the two PC ranges that had the same stage flags set to the error message.  This is better than what's there now, which is largely redundant with the error text from the spec that is tacked on at the end.  I'll push when I finish testing.

tests/layer_validation_tests.cpp
karl-lunarg: added this comment

karl-lunarg:
I applied the changes mentioned in the reviews.  I also found that CTS tests using vkCmdPushConstants to  update part of a push constant range.  So I updated the check in this validation code to tolerate partial range updates.
Should be good to go now.

PR #1515, opened 2017-02-28
Buffer/Image checks for SHARING_MODE_CONCURRENT

mark-lunarg:
When using this sharing mode, buffers and images are created with a list of valid queue families.  Check that when a CB uses these objects that they're valid for the current queue.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -4571,8 +4571,8 @@ static bool PreCallValidateQueueSubmit(layer_data *dev_data, VkQueue queue, uint
         }
         for (uint32_t i = 0; i < submit->commandBufferCount; i++) {
             auto cb_node = GetCBNode(dev_data, submit->pCommandBuffers[i]);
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -4373,20 +4373,68 @@ static bool validateResources(layer_data *dev_data, GLOBAL_CB_NODE *cb_node) {
...
 }
 
+// Check that the queue family index of 'queue' matches one of the entries in pQueueFamilyIndices
+bool ValidImageBufferQueue(layer_data *dev_data, GLOBAL_CB_NODE *cb_node, const VK_OBJECT *object, VkQueue queue, uint32_t count,
+                           const uint32_t *indices) {
+    bool found = false;
+    bool skip = false;
+    auto queue_state = GetQueueState(dev_data, queue);
+    if (queue_state) {
+        for (uint32_t qfi = 0; qfi < count; qfi++) {
```
chrisforbes: Having something called `qfi` that's /not/ a queue family index but an index into an array here is confusing.

mark-lunarg: Sure, will do.

PR #1519, opened 2017-02-28
layers: Remove redundant validation

jeremy-lunarg:
Remove redundant validation from ValidateOrIncrementBoundObjects.

Change-Id: I98f9de491d83788cf0b19d4c606d239e1a915e58

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -4050,97 +4050,75 @@ static bool ValidateStageMaskGsTsEnables(layer_data *dev_data, VkPipelineStageFl
     return skip;
 }
 
-// Loop through bound objects and increment their in_use counts if increment parameter is true
-// or flag an error if unknown objects are found
-static bool ValidateOrIncrementBoundObjects(layer_data *dev_data, GLOBAL_CB_NODE const *cb_node, bool increment) {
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -4356,7 +4335,7 @@ static bool validateCommandBufferState(layer_data *dev_data, GLOBAL_CB_NODE *cb_
 static bool validateResources(layer_data *dev_data, GLOBAL_CB_NODE *cb_node) {
     bool skip_call = false;
 
-    skip_call |= ValidateOrIncrementBoundObjects(dev_data, cb_node, false);
+    IncrementBoundObjects(dev_data, cb_node, false);
```
Tony-LunarG: If validation is happening elsewhere, there's no reason to call IncrementBoundObjects here, nor does it need a bool parameter to tell it whether to validate or increment, it should always just increment

jeremy-lunarg:
Good call, Tony. I've made the change you suggested.

layers/core_validation.cpp
jeremy-lunarg: Fixed.

jeremy-lunarg:
Took Chris' PR instead.

PR #1525, opened 2017-03-01
loader: Fix handling of surface parameters

cubanismo:
The loader code generation scripts had a bug that
caused all the surface->ICD surface translation C
code to be placed inside a block that only ran if
the ICD didn't support the extension function,
making it effectively a no-op.  ICDs that
supported the extension were instead called with
the un-translated surface handle, which could
cause a crash if they were expecting their local
surface handle.

Also, speculatively fix up return handling in this
same class of functions by handling the void
return type.

### DIFF
--- a/scripts/loader_extension_generator.py
+++ b/scripts/loader_extension_generator.py
@@ -1090,13 +1090,14 @@ class LoaderExtensionOutputGenerator(OutputGenerator):
                     funcs += '                   "ICD associated with VkPhysicalDevice does not support '
                     funcs += base_name
                     funcs += '");\n'
+                    funcs += '    }\n'
 
                     if has_surface == 1:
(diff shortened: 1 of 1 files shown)


scripts/loader_extension_generator.py
```diff
@@ -1090,13 +1090,14 @@ def CreateTrampTermFuncs(self):
                     funcs += '                   "ICD associated with VkPhysicalDevice does not support '
                     funcs += base_name
                     funcs += '");\n'
+                    funcs += '    }\n'
```
cubanismo: Note this means the code will continue on and likely crash lacking ICD support.  I wasn't sure if that was the intent, but that matches the behavior before my change.  If not intended, I can return some error code in this case, but we'd have to decide which one is appropriate, maybe good old 'INITIALIZATION_FAILED'.

mark-lunarg:
LGTM.  @MarkY-LunarG, you'd better have a go.

Issue #1507, opened 2017-02-27
Core Validation: ValidtateBufferBounds may reports incorrect error for compressed image formats


TiemoJung:
Hi,

if the copy extend of one of the regions is smaller than the block size of the compressed image format, then the range check becomes invalid as `copy_extent.height - 1`
in https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/buffer_validation.cpp#L2724 underflows.

As it is a invalid copy command, it should be reported on an earlier check (ValidateImageBounds may be?)

daveh-lunarg:
There is a valid usage check that might have caught this (not yet implemented, #1270). _Might_, because the spec language is slightly vague in that it doesn't disallow an extent of 0 for w,h - which is why this check was deferred previously.  Since Tiemo has managed to hit it, I'll avoid the underflow for now and replace it with the official valid usage once it's clarified.

daveh-lunarg:
This issue should be resolved with PR1530.  Thanks for the report!

PR #1530, opened 2017-03-03
Bugfix and added VU checks to resolve github issue 1507

daveh-lunarg:
Github issue #1507 exposed a bug with computing buffer sizes for compressed formats when the extent is less than the compressed block size in any dimension. In 3 parts, first is the bug fix, second adds the VU checks that potentially _should_ have been tripped, and third is just a clang reformat that should not be confused with the code changes in the first two.


daveh-lunarg:
Good catch, @aejsmith. Fixed the bug and added a layer validation test for compressed mip level copies that would have caught it.  Will update this PR as soon as the patch clears our build farm.

Issue #1529, opened 2017-03-03
Validation incorrect for VK_KHR_descriptor_update_template extension


mark-lunarg:
This extension contains an API (`UpdateDescriptorSetWithTemplateKHR`) which passes Vulkan objects in a memory buffer.  This has implications for code generation in the unique_objects utility layer, and leads to incomplete validation by the object_tracker validation layer.  The core_validation layer will also need modifications as this API updates descriptor sets causing incorrect validation errors.

For unique_objects and object_tracker, he buffer will need to be shadowed and the objects replaced with wrapped handles.

PR #1535, opened 2017-03-07
Add ability to handle DescriptorUpdateTemplate extension to unique objects layer

mark-lunarg:
The UniqueObjects utility layer needed to unwrap vulkan object handles embedded in a buffer parameter of `vkUpdateDescriptorSetWithTemplateKHR()`.  This patch series parses that buffer, replacing the uniquified handles with the actual ones before passing the call to the hardware.  This takes care of part of issue #1529.

### DIFF
--- a/layers/unique_objects.cpp
+++ b/layers/unique_objects.cpp
@@ -613,24 +613,24 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDescriptorUpdateTemplateKHR(VkDevice device
                                                                  const VkAllocationCallbacks *pAllocator,
                                                                  VkDescriptorUpdateTemplateKHR *pDescriptorUpdateTemplate) {
     layer_data *dev_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
-    safe_VkDescriptorUpdateTemplateCreateInfoKHR *local_pCreateInfo = NULL;
+    safe_VkDescriptorUpdateTemplateCreateInfoKHR *local_create_info = NULL;
     {
         std::lock_guard<std::mutex> lock(global_lock);
         if (pCreateInfo) {
-            local_pCreateInfo = new safe_VkDescriptorUpdateTemplateCreateInfoKHR(pCreateInfo);
+            local_create_info = new safe_VkDescriptorUpdateTemplateCreateInfoKHR(pCreateInfo);
             if (pCreateInfo->descriptorSetLayout) {
-                local_pCreateInfo->descriptorSetLayout =
+                local_create_info->descriptorSetLayout =
                     (VkDescriptorSetLayout)
                         dev_data->unique_id_mapping[reinterpret_cast<const uint64_t &>(pCreateInfo->descriptorSetLayout)];
             }
             if (pCreateInfo->pipelineLayout) {
-                local_pCreateInfo->pipelineLayout =
+                local_create_info->pipelineLayout =
                     (VkPipelineLayout)dev_data->unique_id_mapping[reinterpret_cast<const uint64_t &>(pCreateInfo->pipelineLayout)];
             }
         }
     }
     VkResult result = dev_data->device_dispatch_table->CreateDescriptorUpdateTemplateKHR(
-        device, (const VkDescriptorUpdateTemplateCreateInfoKHR *)local_pCreateInfo, pAllocator, pDescriptorUpdateTemplate);
+        device, (const VkDescriptorUpdateTemplateCreateInfoKHR *)local_create_info, pAllocator, pDescriptorUpdateTemplate);
     if (VK_SUCCESS == result) {
         std::lock_guard<std::mutex> lock(global_lock);
         uint64_t unique_id = global_unique_id++;
@@ -638,7 +638,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDescriptorUpdateTemplateKHR(VkDevice device
         *pDescriptorUpdateTemplate = reinterpret_cast<VkDescriptorUpdateTemplateKHR &>(unique_id);
 
         // Shadow template createInfo for later updates
-        std::unique_ptr<TEMPLATE_STATE> template_state(new TEMPLATE_STATE(*pDescriptorUpdateTemplate, local_pCreateInfo));
+        std::unique_ptr<TEMPLATE_STATE> template_state(new TEMPLATE_STATE(*pDescriptorUpdateTemplate, local_create_info));
         dev_data->desc_template_map[unique_id] = std::move(template_state);
     }
     return result;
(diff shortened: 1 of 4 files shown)


layers/unique_objects.cpp
```diff
@@ -647,25 +649,93 @@ VKAPI_ATTR void VKAPI_CALL DestroyDescriptorUpdateTemplateKHR(VkDevice device,
...
+                auto image_entry = reinterpret_cast<VkDescriptorImageInfo *>(entry);
+                size = std::max(size, (char *)offset + sizeof(VkDescriptorImageInfo));
+
+                auto sampler_handle = dev_data->unique_id_mapping[reinterpret_cast<uint64_t &>(image_entry->sampler)];
+                patch_list.emplace_back(std::make_pair(offset, sampler_handle));
+                offset += sizeof(VkSampler);
+                auto image_view_handle = dev_data->unique_id_mapping[reinterpret_cast<uint64_t &>(image_entry->imageView)];
+                patch_list.emplace_back(std::make_pair(offset, image_view_handle));
+                offset += sizeof(VkImageView);
+                patch_list.emplace_back(std::make_pair(offset, image_entry->imageLayout));
```
chrisforbes: This emits a 64 bits fixup to a 32 bits value. You get away with it in a LE 64 bit build as you zero extend imageLayout and then scribble in the struct padding, but this is going to go badly for 32 bit builds, if this happens to be right at the end of the buffer.

[or on BE, if anyone decides to run the layers there-- but they probably also have other problems]

layers/unique_objects.cpp
```diff
@@ -647,25 +649,93 @@ VKAPI_ATTR void VKAPI_CALL DestroyDescriptorUpdateTemplateKHR(VkDevice device,
...
+    if (template_map_entry == dev_data->desc_template_map.end()) {
+        assert(0);
+    }
+    auto const &create_info = template_map_entry->second->create_info;
+    char *size = 0;
+    std::vector<std::pair<char *, uint64_t>> patch_list;
+
+    for (uint32_t i = 0; i < create_info->descriptorUpdateEntryCount; i++) {
+        for (uint32_t j = 0; j < create_info->pDescriptorUpdateEntries[i].descriptorCount; j++) {
+            char *offset =
```
chrisforbes: This is hard to follow -- you have offsets and sizes that are actually ptrs?

layers/unique_objects.h
```diff
@@ -33,6 +33,14 @@ namespace unique_objects {
 // All increments must be guarded by global_lock
 static uint64_t global_unique_id = 1;
 
+struct TEMPLATE_STATE {
+    VkDescriptorUpdateTemplateKHR desc_update_template;
+    safe_VkDescriptorUpdateTemplateCreateInfoKHR *create_info;
```
chrisforbes: Alternatively, just hold this by value -- the TEMPLATE_STATE is held by unique_ptr already, one layer will do...

layers/unique_objects.cpp
```diff
@@ -643,6 +647,7 @@ VKAPI_ATTR void VKAPI_CALL DestroyDescriptorUpdateTemplateKHR(VkDevice device,
                                                               const VkAllocationCallbacks *pAllocator) {
     layer_data *dev_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
     std::unique_lock<std::mutex> lock(global_lock);
+    dev_data->desc_template_map.erase(descriptorUpdateTemplate);
```
chrisforbes: [Either fix TEMPLATE_STATE so it correctly owns its `create_info` member, or you need to explicitly delete it here.]

scripts/helper_file_generator.py
```diff
@@ -525,9 +525,18 @@ def GenerateSafeStructHelperSource(self):
     # safe_struct source -- create bodies of safe struct helper functions
     def GenerateSafeStructSource(self):
         safe_struct_body = []
+        wsi_structs = ['VkXlibSurfaceCreateInfoKHR',
```
MarkY-LunarG: This almost seems like it should be a different change.

layers/unique_objects.h
```diff
@@ -55,7 +55,7 @@ struct layer_data {
     VkDebugReportCallbackCreateInfoEXT *tmp_dbg_create_infos;
     VkDebugReportCallbackEXT *tmp_callbacks;
 
-    std::unordered_map<VkDescriptorUpdateTemplateKHR, std::unique_ptr<TEMPLATE_STATE>> desc_template_map;
+    std::unordered_map<uint64_t, std::unique_ptr<TEMPLATE_STATE>> desc_template_map;
```
mark-lunarg: Done.

layers/unique_objects.h
mark-lunarg: Changed to by-value.

layers/unique_objects.cpp
```diff
@@ -647,25 +649,93 @@ VKAPI_ATTR void VKAPI_CALL DestroyDescriptorUpdateTemplateKHR(VkDevice device,
...
+            if ((create_info->pDescriptorUpdateEntries[i].descriptorType == VK_DESCRIPTOR_TYPE_SAMPLER) ||
+                (create_info->pDescriptorUpdateEntries[i].descriptorType == VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER) ||
+                (create_info->pDescriptorUpdateEntries[i].descriptorType == VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE) ||
+                (create_info->pDescriptorUpdateEntries[i].descriptorType == VK_DESCRIPTOR_TYPE_STORAGE_IMAGE) ||
+                (create_info->pDescriptorUpdateEntries[i].descriptorType == VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT)) {
+                auto image_entry = reinterpret_cast<VkDescriptorImageInfo *>(entry);
+                size = std::max(size, (char *)offset + sizeof(VkDescriptorImageInfo));
+
+                auto sampler_handle = dev_data->unique_id_mapping[reinterpret_cast<uint64_t &>(image_entry->sampler)];
+                patch_list.emplace_back(std::make_pair(offset, sampler_handle));
```
mark-lunarg: Much nicer!

layers/unique_objects.cpp
mark-lunarg: Reworked to let the compiler figure it out -- only copy structures now.

layers/unique_objects.cpp
mark-lunarg: Fixed as above

mark-lunarg:
Excellent review feedback incorporated, new version now in place.

scripts/helper_file_generator.py
mark-lunarg: It is -- these changes are in their own commit applied before any of the other changes.

mark-lunarg: This change is necessary because none of the KHR extensions were getting deep copies for their safe-struct members.  This is required for remapping handles in these extensions.

PR #1539, opened 2017-03-09
loader: Fix getenv probing

### LINKED ISSUE #1538: Fails to build with musl libc
`| ../loader/libvulkan.so.1.0.39: undefined reference to '__secure_getenv'`

musl libc does not provide secure_getenv() or __secure_getenv() so linking the loader fails.

Would you accept a patch that uses getenv() instead in this case?

jku:
__GNUC__ depends on the compiler not libc, but here we are actually
interested in libc: current code ends up using getenv() when compiler
is not gcc.

Also, it's better to check for features anyway instead of libc's and
their versions.

Fixes #1538.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,6 +303,10 @@ if(NOT WIN32)
 endif()
 
 if(UNIX)
+    include(CheckFunctionExists)
+    CHECK_FUNCTION_EXISTS(secure_getenv HAVE_SECURE_GETENV)
+    CHECK_FUNCTION_EXISTS(__secure_getenv HAVE___SECURE_GETENV)
+
     install(DIRECTORY "${PROJECT_SOURCE_DIR}/include/vulkan" DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 endif()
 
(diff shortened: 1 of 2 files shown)


jku:
Oh, a CLA. Lovely.

I'm sorry, I'm not taking this through legal for a 11 line patch. You will have to re-implement this or decide the contribution is too small to be under copyright in any case: for what it's worth I'd agree with that interpretation.

Sorry for the noise: I did not realize vulkan has a CLA.

MarkY-LunarG:
Ah, no worries.  You gave me enough info to continue.  It also was missing several steps.  According to [this page](http://www.vtk.org/Wiki/CMake_HowToDoPlatformChecks), you need to generate a config file which CMake will generate the #defines for.

I'm almost done with the actual fix.  Once that's done, I'll post the final commit here.

Thanks for the work!

BTW, the CLA is just an Apache license acceptance for any changes you submit.  But, I totally understand from an employee concept because other companies I've worked for have been just as stringent.

MarkY-LunarG:
Ok.  Fixed this with a984b99946b3c8a7d047861698972da26a032c61.

jku:
Thanks! I'm really far from a CMake expert so you may be right about the generated header but I had the impression it all just works without it as long as you remove CMakeCache.txt in between.
 
By the way, the Vulkan CLA is definitely not just license acceptance: It's also a patent grant (and those probably need to go through legal review in any large software company).

PR #1542, opened 2017-03-09
gh760 Checks and tests for VU 00303-00308

mikew-lunarg:
Implement Valid Usages checks that verify that a vkImageMemoryBarrier's oldLayout/newLayout are compatible with the VkImage's usage flags.

Add both negative and positive tests.  Fix a couple existing tests that the new checks revealed were incorrect.


mikew-lunarg:
bump

PR #1543, opened 2017-03-09
Tolerate use of vkUpdateDescriptorSetWithTemplateKHR in Core Validation

mark-lunarg:
Using this API updated descriptors without the knowledge of core_validation.  Added hooks and state updates/tracking for new UpdateTemplate objects and APIs into CV.  Validation, along with support for other APIs in this extension TBD.


layers/core_validation.cpp
```diff
@@ -11417,10 +11423,136 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDeviceGroupsKHX(
...
+VKAPI_ATTR VkResult VKAPI_CALL CreateDescriptorUpdateTemplateKHR(VkDevice device,
+                                                                 const VkDescriptorUpdateTemplateCreateInfoKHR *pCreateInfo,
+                                                                 const VkAllocationCallbacks *pAllocator,
+                                                                 VkDescriptorUpdateTemplateKHR *pDescriptorUpdateTemplate) {
+    layer_data *dev_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
+    safe_VkDescriptorUpdateTemplateCreateInfoKHR *local_create_info = NULL;
+    {
+        std::lock_guard<std::mutex> lock(global_lock);
+        if (pCreateInfo) {
+            local_create_info = new safe_VkDescriptorUpdateTemplateCreateInfoKHR(pCreateInfo);
```
tobine: Why create this here and pass it down the chain?
Is there some benefit I'm missing?
Seems like it would be preferable to only create it in the VK_SUCCESS case below and just pass pCreateInfo down the chain.

layers/core_validation.cpp
```diff
@@ -11417,10 +11423,136 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDeviceGroupsKHX(
...
+                                                              VkDescriptorUpdateTemplateKHR descriptorUpdateTemplate,
+                                                              const void *pData) {
+    layer_data *device_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
+    device_data->dispatch_table.UpdateDescriptorSetWithTemplateKHR(device, descriptorSet, descriptorUpdateTemplate, pData);
+
+    PostCallRecordUpdateDescriptorSetWithTemplateKHR(device_data, descriptorSet, descriptorUpdateTemplate, pData);
+
+}
+
+VKAPI_ATTR void VKAPI_CALL CmdPushDescriptorSetWithTemplateKHR(VkCommandBuffer commandBuffer,
```
tobine: Just making sure you intended to include this change & intercept code for this function below in this commit. Currently a no-op, but I know that it's piping that will be needed shortly.

layers/core_validation.cpp
```diff
@@ -11417,10 +11423,136 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDeviceGroupsKHX(
...
+    dev_data->desc_template_map.erase(descriptorUpdateTemplate);
+    lock.unlock();
+    dev_data->dispatch_table.DestroyDescriptorUpdateTemplateKHR(device, descriptorUpdateTemplate, pAllocator);
+}
+
+void PostCallRecordUpdateDescriptorSetWithTemplateKHR(layer_data *device_data, VkDescriptorSet descriptorSet,
+                                                      VkDescriptorUpdateTemplateKHR descriptorUpdateTemplate, const void *pData) {
+    auto const template_map_entry = device_data->desc_template_map.find(descriptorUpdateTemplate);
+    if (template_map_entry == device_data->desc_template_map.end()) {
+        assert(0);
```
tobine: object_tracker is not yet verifying that descriptorUpdateTemplate is valid. Would like to see that code in place so that the error is properly thrown before this assert.

layers/core_validation.cpp
```diff
@@ -11417,10 +11423,136 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDeviceGroupsKHX(
...
+        auto dst_array_element = create_info.pDescriptorUpdateEntries[i].dstArrayElement;
+
+        for (uint32_t j = 0; j < create_info.pDescriptorUpdateEntries[i].descriptorCount; j++) {
+            desc_writes.emplace_back();
+            auto &write_entry = desc_writes.back();
+
+            size_t offset = create_info.pDescriptorUpdateEntries[i].offset + j * create_info.pDescriptorUpdateEntries[i].stride;
+            char *update_entry = (char *)(pData) + offset;
+
+            if (dst_array_element >= binding_count) {
```
tobine: I'd like to see the code in this block added as function in DescriptorSetLayout class. Something like:
`binding_being_updated = layout_obj->GetNextValidBinding(binding_being_updated);`
Then could call
`layout_obj->GetDescriptorCountFromBinding(binding_being_updated);`

Also, I believe we should guard that update with a call to IsNextBindingConsistent() and flag error if it's not.

layers/core_validation.cpp
```diff
@@ -11417,10 +11423,136 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDeviceGroupsKHX(
...
+                    write_entry.pTexelBufferView = reinterpret_cast<VkBufferView *>(update_entry);
+                    break;
+                default:
+                    assert(0);
+                    break;
+            }
+            dst_array_element++;
+        }
+    }
+    cvdescriptorset::PerformUpdateDescriptorSets(device_data, static_cast<uint32_t>(desc_writes.size()), desc_writes.data(), 0,
```
tobine: This is good re-use.

layers/core_validation.cpp
mark-lunarg: Right, copy-paste error.

layers/core_validation.cpp
mark-lunarg: To come in a follow-on patch set.

layers/core_validation.cpp
mark-lunarg: Will do.

layers/core_validation.cpp
mark-lunarg: Yes, thanks.

mark-lunarg:
@tobine -- we do need to get this into the SDK point release, so I'll push this stuff and implement these changes as a follow-on patch set in a day or two.  Thanks for the feedback.

PR #1552, opened 2017-03-10
Fix handling of depth+stencil formats in shader validation

### LINKED ISSUE #1491: Reading D24_S8 as input attachment with stencil triggers incorrect validation.
If the stencil aspect of a D24_S8 image is read as UINT (with a VkImageView using only stencil aspect), incorrect validation occurs.

```
            } else if (get_format_type(rpci->pAttachments[index].format) != get_fundamental_type(module, use.second.type_id)) {
                if (log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VkDebugReportObjectTypeEXT(0), 0, __LINE__,
                            SHADER_CHECKER_INPUT_ATTACHMENT_TYPE_MISMATCH, "SC",
                            "Subpass input attachment %u format of %s does not match type used in shader `%s`", use.first,
                            string_VkFormat(rpci->pAttachments[index].format), describe_type(module, use.second.type_id).c_str())) {
                    pass = false;
                }
            }
```

get_format_type() thinks D24_S8 is FLOAT, but it could be both in a shader.

chrisforbes:
Fixes #1491.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -1665,6 +1665,10 @@ static unsigned get_format_type(VkFormat fmt) {
         case VK_FORMAT_A2B10G10R10_UINT_PACK32:
         case VK_FORMAT_A2R10G10B10_UINT_PACK32:
             return FORMAT_TYPE_UINT;
+        case VK_FORMAT_D16_UNORM_S8_UINT:
+        case VK_FORMAT_D24_UNORM_S8_UINT:
+        case VK_FORMAT_D32_SFLOAT_S8_UINT:
+            return FORMAT_TYPE_FLOAT | FORMAT_TYPE_UINT;
         default:
             return FORMAT_TYPE_FLOAT;
     }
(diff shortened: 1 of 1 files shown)


layers/core_validation.cpp
```diff
@@ -1772,7 +1775,7 @@ static bool validate_vi_against_vs_inputs(debug_report_data *report_data, VkPipe
             unsigned input_type = get_fundamental_type(vs, it_b->second.type_id);
 
             // Type checking
-            if (attrib_type != FORMAT_TYPE_UNDEFINED && input_type != FORMAT_TYPE_UNDEFINED && attrib_type != input_type) {
+            if (!(attrib_type & input_type)) {
```
tobine: Previously if either type was UNDEFINED then  we'd skip the error, but now if either (or both) is UNDEFINED we flag the error. Just want to make sure that's intended. Same comment below.

chrisforbes: I'm pretty sure that was just an accident of history.

mark-lunarg:
Hey @chrisforbes, is this one ready to go?

chrisforbes:
@mark-lunarg should be good to go.

PR #1560, opened 2017-03-10
Refactor UpdateTemplate changes for DescriptorSet module

mark-lunarg:
Changes to address feedback on CV UpdateTemplate changes.  @tobine, the only things I knowingly left out are the ObjectTracker updates for this new object and the validation check you mentioned for the getnextvalidbinding call.

There is a process issue getting new object types into the debug-report extension object type enums.  This'll be solved long term by getting a standalone object list enum into the vulkan header, or by us generating one.  This applies to all new extensions containing new objects.

I left out the validation check since this was the record path.  Will this get covered when validation is added?

### DIFF
--- a/layers/descriptor_sets.cpp
+++ b/layers/descriptor_sets.cpp
@@ -193,6 +193,14 @@ VkSampler const *cvdescriptorset::DescriptorSetLayout::GetImmutableSamplerPtrFro
     }
     return nullptr;
 }
+// Move to next valid binding having a non-zero binding count
+uint32_t cvdescriptorset::DescriptorSetLayout::GetNextValidBinding(const uint32_t binding) const {
+    uint32_t new_binding = binding;
+    do {
+        new_binding++;
+    } while (GetDescriptorCountFromBinding(new_binding) == 0);
+    return new_binding;
+}
 // For given index, return ptr to ImmutableSampler array
 VkSampler const *cvdescriptorset::DescriptorSetLayout::GetImmutableSamplerPtrFromIndex(const uint32_t index) const {
     assert(index < bindings_.size());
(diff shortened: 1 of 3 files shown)


layers/descriptor_sets.cpp
```diff
@@ -1136,6 +1144,73 @@ void cvdescriptorset::PerformUpdateDescriptorSets(const layer_data *dev_data, ui
...
+                    write_entry.pTexelBufferView = reinterpret_cast<VkBufferView *>(update_entry);
+                    break;
+                default:
+                    assert(0);
+                    break;
+            }
+            dst_array_element++;
+        }
+    }
+    cvdescriptorset::PerformUpdateDescriptorSets(device_data, static_cast<uint32_t>(desc_writes.size()), desc_writes.data(), 0,
```
tobine: Don't need namespace qualifier anymore

layers/descriptor_sets.h
```diff
@@ -137,6 +137,8 @@ class DescriptorSetLayout {
     //  These calls should be guarded by a call to "HasBinding(binding)" to verify that the given binding exists
     uint32_t GetGlobalStartIndexFromBinding(const uint32_t) const;
     uint32_t GetGlobalEndIndexFromBinding(const uint32_t) const;
+    // Helper function to get the next valid binding for a descriptor
+    uint32_t GetNextValidBinding(const uint32_t current_binding) const;
```
tobine: Don't need parameter names for header function definition. Prefer to exclude them to match existing code.

layers/descriptor_sets.h
```diff
@@ -285,6 +287,9 @@ bool ValidateUpdateDescriptorSets(const debug_report_data *, const core_validati
 // "Perform" does the update with the assumption that ValidateUpdateDescriptorSets() has passed for the given update
 void PerformUpdateDescriptorSets(const core_validation::layer_data *, uint32_t, const VkWriteDescriptorSet *, uint32_t,
                                  const VkCopyDescriptorSet *);
+// Similar to PerformUpdateDescriptorSets, this function will do the same for updating via templates
+void PerformUpdateDescriptorSetsWithTemplateKHR(layer_data *device_data, VkDescriptorSet descriptorSet,
```
tobine: Don't need parameter names for header function definition. Prefer to exclude them to match existing code.

PR #1561, opened 2017-03-10
Fixes from running vk_layer_validation_tests on AMD

Tony-LunarG:
These are changes to layer validation tests (and one layer change) to make them better behaved on my AMD system.


Tony-LunarG:
Committed to master

Issue #1551, opened 2017-03-09
Build and test failures with latest Android toolchain


cnorthrop:
The latest releases of the Android NDK (r14) and SDK Tools (25.3.1) are causing problems with our builds and tests.

If you've already upgraded, please roll back until we get them straightened out.  I'll update this issue when things are working again with latest tools, either through source changes or tool updates.

Links to working versions below:

Android SDK Tools:
Windows - https://dl.google.com/android/repository/tools_r25.2.3-windows.zip
Linux - https://dl.google.com/android/repository/tools_r25.2.5-linux.zip
OSX - https://dl.google.com/android/repository/tools_r25.2.3-macosx.zip

Android NDK (13b known to work):
http://developer.android.com/ndk/downloads/older_releases.html

cnorthrop:
PR #1562 works around the NDK problem, so Android NDK r14 should work now.

cnorthrop:
PR #1673 fixes the packaging problem for Linux and OSX.  Still need to update Windows scripts.

mark-lunarg:
@cnorthrop, is this still an issue?

cnorthrop:
Yes, still an issue for Windows.  Probably will remain an issue until we get everything migrated over to python scripts.

PR #1562, opened 2017-03-10
tests: Workaround Clang bug in Android NDK r14

cnorthrop:
When using Android NDK r14, incorrect arm64 assembly is generated unless we write directly to the last member of this struct.

This is a workaround for a bug that does not occur after clang 3688880, which should be included in next NDK release.

We will ensure the bug is fixed (and not just masked) separately.  This should unblock regression testing on r14.

This partially addresses issue #1551.


cnorthrop:
No, these were the only two tests that started failing.  Other tests have similar blocks of code, which makes me worry that it will be difficult to create a small reproducer.

cnorthrop:
cherry-picked to master

Issue #1395, opened 2017-01-23
Signed/unsigned mismatch warning with VC++ 2017


randomascii:
In the FindDependency function, currently at line 10220:

https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/core_validation.cpp#L10220

an int (dependent) is used to call std::find on a std::vector of unsigned int types, leading to "warning C4389: '==': signed/unsigned mismatch" with VC++ 2017. The warning is correct so a cast or some other workaround would seem appropriate.

static bool FindDependency(const int index, const int dependent, const std::vector<DAGNode> &subpass_to_node,
                           std::unordered_set<uint32_t> &processed_nodes) {
    // If we have already checked this node we have not found a dependency path so return false.
    if (processed_nodes.count(index))
        return false;
    processed_nodes.insert(index);
    const DAGNode &node = subpass_to_node[index];
    // Look for a dependency path. If one exists return true else recurse on the previous nodes.
    if (std::find(node.prev.begin(), node.prev.end(), dependent) == node.prev.end()) {

randomascii:
I can't upload a patch to my fork right now for some reason but this works:

    Suppress VC++ 2017 signed/unsigned warning

    When building Chrome with VC++ 2017 there is one warning, from passing
    an int to a call to std::vector<uint32_t>::find. This change adds a cast
    to avoid that warning, thus fixing issue #1395.

diff --git a/layers/core_validation.cpp b/layers/core_validation.cpp
index fd21f01..2d15031 100644
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -10217,7 +10217,7 @@ static bool FindDependency(const int index, const int dependent, const std::vect
     processed_nodes.insert(index);
     const DAGNode &node = subpass_to_node[index];
     // Look for a dependency path. If one exists return true else recurse on the previous nodes.
-    if (std::find(node.prev.begin(), node.prev.end(), dependent) == node.prev.end()) {
+    if (std::find(node.prev.begin(), node.prev.end(), static_cast<uint32_t>(dependent)) == node.prev.end()) {
         for (auto elem : node.prev) {
             if (FindDependency(elem, dependent, subpass_to_node, processed_nodes))
                 return true;

randomascii:
Pull request should be good to go.

mark-lunarg:
This issue has been fixed.

PR #1573, opened 2017-03-14
layers: Fix round tripping uint32_t through int

bungeman:
Commit 25afc3c "Suppress VC++ 2017 signed/unsigned warning" fixed
issue #1395. However, it did so by static_cast-ing a value back to a
type it already had been. Instead, just avoid round tripping the
'uint32_t' through 'int'.


mark-lunarg:
@bungeman, this looks fine.  Can you reformat the commit messages to match others in the repo?  In this case they'd have the format of "layers: blah blah blah".  After that it's good to go.

bungeman:
Updated commit message and squashed.

mark-lunarg:
Thanks!  Keep 'em coming.

PR #1575, opened 2017-03-14
Update Vulkan-LVL to 1.0.43 version of the header

mark-lunarg:
This update killed a couple of KHX extension APIs and added three new APIs from two new extensions.


mark-lunarg:
hpp comes from an external repo (grumble grumble).

PR #1577, opened 2017-03-14
Tobin val flags piping

tobine:
Finally getting around to piping this in. Only current option is to disable all of the flags that are present in validation which don't cover all of the validation checks, but is a nice POC to at least have something working. Tested with an update to cube that turns on validation and sets the VK_VALIDATION_CHECK_ALL_EXT enum of VK_EXT_validation_flags.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -3809,6 +3809,20 @@ static void checkInstanceRegisterExtensions(const VkInstanceCreateInfo *pCreateI
     }
 }
 
+// For the given ValidationCheck enum, set all relevant instance disabled flags to true
+void SetDisabledFlags(instance_layer_data *instance_data, VkValidationFlagsEXT *val_flags_struct) {
+    for (uint32_t i = 0; i < val_flags_struct->disabledValidationCheckCount; ++i) {
+        switch (val_flags_struct->pDisabledValidationChecks[i]) {
+            case VK_VALIDATION_CHECK_ALL_EXT:
+                // Set all disabled flags to true
+                instance_data->disabled.SetAll(true);
+                break;
+            default:
+                break;
+        }
+    }
+}
+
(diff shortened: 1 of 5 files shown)


layers/core_validation.cpp
```diff
@@ -6682,9 +6707,11 @@ VKAPI_ATTR VkResult VKAPI_CALL ResetDescriptorPool(VkDevice device, VkDescriptor
 // as well as DescriptorSetLayout ptrs used for later update.
 static bool PreCallValidateAllocateDescriptorSets(layer_data *dev_data, const VkDescriptorSetAllocateInfo *pAllocateInfo,
                                                   cvdescriptorset::AllocateDescriptorSetsData *common_data) {
+    // Always update common data
+    cvdescriptorset::UpdateAllocateDescriptorSetsData(dev_data, pAllocateInfo, common_data);
```
courtney-g: Does this belong here?
Seems unrelated to validation flags.

tobine: Enabling validation flags exposed a bug here so included in same CL as code would crash otherwise.
When flags were enabled we were failing to do this common state update.

courtney-g: Cool. Okay. LGTM

karl-lunarg:
Cool!  Is there any indication when we can define the rest of the VkValidationCheckEXT enum?   I'd love to be able to selectively turn off validation for GetQueryPoolResults().

tobine:
I own the extension so I think it should be easy for us to add any new enums for different cases that we'd like to disable.
The process for updates is basically add new enum(s) to VkValidationCheckEXT and then add the code to set the appropriate flags in the SetDisabledFlags() function.
Note that currently the flags only have an effect in core_validation. Ideally if we refactor everything to be under that layer so that things like object_tracker and parameter_validation checks are calls off of core_validation, then future flags can disable those classes of checks. Currently, if we wanted a flag to disable anything outside of core_validation it would just require piping the extension support into the separate layer as I've done here for core_validation.

PR #1581, opened 2017-03-14
build: Add pkgconfig file for the Vulkan loader

### LINKED ISSUE #757: Vulkan-LoaderAndValidationLayers does not provide pkg-config file ( ex. vulkan.pc)
current Vulkan-LoaderAndValidationLayers does not provide pkg-config file.
in linux environment, Makefile usually uses pkg-config tools for setting  name of library linking. 
at this point of time we should describe library linking information like -lvulkan. 
so for convenience of vulkan application, we need to provide vulkan.pc .

karl-lunarg:
Fixes #757
On Linux, the build configures a vulkan.pc pkgconfig
file, writing it to <build_dir>/loader.  The make 'install'
target then installs this file along with the others so that
it can be found with pkg-config.  Then, issuing:
$ pkg-config --libs --cflags vulkan
emits compilation flags for vulkan programs.

Since retrieving the header version from vulkan.h was needed in
order to put the Version in the pc file, this commit also uses
the header version to specify the SO version, so it no longer needs
to be changed for every release.

Change-Id: If83c17bfbb59f2ee05d5f86f5e957fc11871546b


karl-lunarg:
pushed to master.

PR #1582, opened 2017-03-15
Fix useless pkconfig file and install location

leigh123linux:
The current pkconfig file and install location isn't usable on any 64bit distro

### DIFF
--- a/loader/CMakeLists.txt
+++ b/loader/CMakeLists.txt
@@ -155,6 +155,6 @@ else()
         endforeach()
         configure_file("vulkan.pc.in" "vulkan.pc" @ONLY)
         install(FILES       "${CMAKE_CURRENT_BINARY_DIR}/vulkan.pc"
-                DESTINATION "${CMAKE_INSTALL_PREFIX}/lib/pkgconfig")
+                DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig")
     endif()
 endif()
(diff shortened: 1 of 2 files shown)


karl-lunarg:
When I try using this pull request with using default CMake options, the generated vulkan.pc file looks like:

```
prefix=/usr/local
exec_prefix=/usr/local
libdir=
includedir=/vulkan

Name: VULKAN
Description: Vulkan Loader
Version: 1.0.43
Libs: -L${libdir} -lvulkan  -lstdc++ -lm -lc
Cflags: -I${includedir}
```

Note that `libdir` is not getting set and so pkg-config won't set the `-L` option correctly.

Also `includedir` is not correct.

The revised `vulkan.pc.in` in this pull request looks like:

```
prefix=@CMAKE_INSTALL_PREFIX@
exec_prefix=@CMAKE_INSTALL_PREFIX@
libdir=@INSTALL_LIB_DIR@
includedir=@INCLUDE_INSTALL_DIR@/vulkan

Name: @CMAKE_PROJECT_NAME@
Description: Vulkan Loader
Version: @VK_API_VERSION@
Libs: -L${libdir} -lvulkan @PRIVATE_LIBS@
Cflags: -I${includedir}
```

Where are `INSTALL_LIB_DIR` and `INCLUDE_INSTALL_DIR` defined?

leigh123linux:
@karl-lunarg Thank you for the feedback, I have rebased to fix the issues raised.

> Where are INSTALL_LIB_DIR and INCLUDE_INSTALL_DIR defined?

My bad, they were defined by rpm packaging.

PR #1588, opened 2017-03-15
Fix potentially uninitialized VS2015 warning.

### LINKED ISSUE #1587: Potentially uninitialized variable use warning in VS2015
This warning is showing up when building the loader with the ANGLE/Chrome integration files, which use different sets of warnings.

Would the layers SDK devs be interested in landing a CL which fixes the loader warnings as well as the other few places in the demos/tests where this warning is triggered? I can also set it so it will be added to the list of always-on warnings to prevent future regressions.

The two warnings in question are 4701 (for variables) and 4703 (for pointers).

null77:
This fixes the a few occurences in demos, the loader and a test.
It also adds the warning to the always-on list so it will prevent
future regressions.

Fixes #1587.

Change-Id: I26f69e977b57749a3ab4ddb548ada95384131edc


mark-lunarg:
Improved commit message format and pushed to Master.  Thanks!

null77:
Thank you!

PR #1590, opened 2017-03-15
tests: Fail on unexpected errors

jeremy-lunarg:
Unexpected errors will now cause a test to fail. Explicitly calling
out unexpected errors in the test will still work, but should be used
reluctantly.

Change-Id: I5d121dc453c4acf497643bf10ad02459d0d96f15

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -229,7 +229,6 @@ class ErrorMonitor {
             }
 
             if (!found_expected) {
-                printf("Unexpected: %s\n", msgString);
                 other_messages_.push_back(errorString);
             }
         }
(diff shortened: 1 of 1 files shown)


mark-lunarg:
CI hates your patch -- might wanna do a quick check using clang.

mark-lunarg:
Travis CI failures unrelated to change.  Good to Go.

cnorthrop:
This is causing all kinds of test failures on our end.  We are short staffed this week due to Spring Break.  It will be next week before we can look at all of these.  Getting a green light from our larger test pool would have been good before moving forward with this.

jeremy-lunarg:
I brought this patch up yesterday in the F2F. Nobody said anything. What do you want me to do?

cnorthrop:
http://giphy.com/gifs/PFwKHjOcIoVUc

I guess leave it as is and we'll dig out next week.  I don't think we'll have anything to contribute until then anyway.  Thanks Jeremy.

jeremy-lunarg:
Reverting this for now. Sorry for the headache.

cnorthrop:
Thanks, Jeremy.  We'll add SetUnexpectedError to the new cases so you can re-enable this soon.

PR #1594, opened 2017-03-16
layers: Fix incorrect KHR/KHX ifdefs.

### LINKED ISSUE #1593: layers: Building with VK_USE_PLATFORM_WIN32_KHR and not VK_USE_PLATFORM_WIN32_KHX produces compile error.
I found this locally when testing with a new version of the repo with an ANGLE build. It seems a few ifdefs improperly check for the KHR define instead of the KHX. Will upload a CL to fix this.

null77:
In two places we were checking for the presence of the WIN32 KHR
define instead of the KHX define where the enums and methods are
defined.

Fixes #1593.

Change-Id: I0bf0c2e2ed734ed2c40a44c4cd7d468ba70ca9a4


mark-lunarg:
Thanks Jamie -- good catch!

null77:
No problem, thank you Mark.

PR #1599, opened 2017-03-18
core validation: deep copy IMAGE_STATE and BUFFER_STATE 

plohrmann:
Neither class was deep copying the createinfo.pQueueFamilyIndicies array, which could cause invalid warnings to be reported to the user.


plohrmann:
Hi Mark, That would be much appreciated, thank you!

On Tue, Mar 21, 2017 at 12:33 PM, Mark Lobodzinski <notifications@github.com

mark-lunarg:
Fixed commit msg, clang-formatted and pushed to Master.  Thanks Peter!

PR #1607, opened 2017-03-21
object_tracker: Fix uses of KHR ifdef to KHX.

null77:
In a few places, we were using VK_USE_PLATFORM_WIN32_KHR instead of
VK_USE_PLATFORM_WIN32_KHX.

Change-Id: I579ea848868396b9c2877e43ca24780476e55919


mark-lunarg:
Ammended message, added one more KHR->KHX change and pushed to Master.  Thanks Jamie.

null77:
Thanks Mark!

PR #1608, opened 2017-03-21
loader: Add guard around cmake header include.

null77:
This allows non-cmake platforms to compile the loader.

Change-Id: I6f6d8731fe1be60fd138e9fe1006c913fb029b47


MarkY-LunarG:
On these platforms, you're no longer using secure_getenv as a side-effect of this change.  Also, where are you documenting the #define used?  Is there any other places we have to do something when CMake is not enabled?  If so, we might want to create a standard (No Cmake in use) define.

null77:
Mark, are those APIs necessary or more secure alternatives to fallback functions? It seems on Windows they weren't available in any case.

I didn't see any other places where CMake was assumed (that affected my build), but the idea of having a standard define seems reasonable. I can rename it to something more general.

MarkY-LunarG:
They're only available on Linux.  On Windows we don't really use the header, although it's included because CMake generates it but without any defines since they're not available.

What I'm really just suggesting here is that you change the define from `VULKAN_NO_LOADER_CMAKE_CONFIG_H` to something more generic like `NON_CMAKE_BUILD` so others can use it if we run into problems elsewhere and in our other projects, because I doubt this is the last Cmake related item we'll run into.

null77:
Thanks for your advice, Mark. I've changed the define to VULKAN_NON_CMAKE_BUILD. PTAL.

PR #1617, opened 2017-03-23
Remove VK_KHR_maintenence1 false positives

mikes-lunarg:
Skip checks that are no longer valid if the VK_KHR_maintenence1 extension is enabled.  Additional validation checks required for VK_KHR_maintenence1 will be provided in a later patch set.

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -1295,13 +1295,16 @@ bool PreCallValidateCmdCopyImage(layer_data *device_data, GLOBAL_CB_NODE *cb_nod
                             ss.str().c_str());
         }
 
-        // For each region the layerCount member of srcSubresource and dstSubresource must match
-        if (regions[i].srcSubresource.layerCount != regions[i].dstSubresource.layerCount) {
-            std::stringstream ss;
-            ss << "vkCmdCopyImage: number of layers in source and destination subresources for pRegions[" << i << "] do not match";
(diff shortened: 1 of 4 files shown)


layers/buffer_validation.cpp
```diff
@@ -1406,22 +1411,24 @@ bool PreCallValidateCmdCopyImage(layer_data *device_data, GLOBAL_CB_NODE *cb_nod
...
 
-        // The source region specified by a given element of regions must be a region that is contained within srcImage
-        if (ExceedsBounds(&regions[i].srcOffset, &regions[i].extent, &(src_image_state->createInfo.extent))) {
-            std::stringstream ss;
-            ss << "vkCmdCopyImage: srcSubResource in pRegions[" << i << "] exceeds extents srcImage was created with";
-            skip |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
-                            reinterpret_cast<uint64_t &>(command_buffer), __LINE__, VALIDATION_ERROR_01175, "IMAGE", "%s. %s",
-                            ss.str().c_str(), validation_error_map[VALIDATION_ERROR_01175]);
-        }
+        if (src_image_state->createInfo.imageType == dst_image_state->createInfo.imageType) {
```
MarkY-LunarG: Why does this matter all of a sudden?  Might be nice to document why in a short comment.

mikes-lunarg: ExceedsBounds only checks pRegions.extent for bounds, which is incomplete when doing a 2d-array to 3d copy that maintenance1 allows.  When new validation is added for this special case it should be a little more clear.  I'll also add a comment to that effect.

PR #1619, opened 2017-03-24
layers: Removing redundant if in unique_objects

ardaLunarG:
Exact same if check performed multiple times.
Removing redundant one

Change-Id: Iff32a724e58a06b4a967ba2632e60cc5b4050238


ardaLunarG:
Removed terminating "

mark-lunarg:
Good to go -- push it!

PR #1622, opened 2017-03-27
Mark master shader cap fixes

mark-lunarg:
MikeW's update to the new glslang/spirv stuff enabled new checks in the spirv-validator.  One of these superceded a core_validation check causing failures in the layer validation tests.

Removed the CV check, updated and renamed the test and added Mike's changes to update the external tools commit IDs.

Note that after this patch series goes in, users will see LVT failures until update_external_sources is run.


tests/layer_validation_tests.cpp
```diff
@@ -13936,7 +13936,7 @@ TEST_F(VkLayerTest, CreatePipelineCheckShaderBadCapability) {
     ASSERT_NO_FATAL_FAILURE(Init());
     ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
 
-    const char *bad_capability_message = "Shader declares capability 53, not supported in Vulkan.";
+    // const char *bad_capability_message = "Shader declares capability 53, not supported in Vulkan.";
```
mikew-lunarg: squash this commit into the actual test enhancement?

mark-lunarg: Then testing would fail for your glslang update commit.

chrisforbes:
:+1: after the fact.

PR #1625, opened 2017-03-27
Layers: Add check for cyclic pNext chains, and duplicate pNext structures

### LINKED ISSUE #1621: cyclic pNext chain should be an error
When a pNext chain has a cycle the layers currently hang or spams validation messages.

Instead the validation layer should detect that and issue an error.

Test case:

```
struct fakeExtension{
VkStructureType sType;
void* pNext;
}
std::vector<fakeExtension> chain(5);
for(int i = 0; i< chain.size(); i++){
    chain[i].sType = 0xFFFFFFFF;//invalid ext type;
    if(i>0)
        chain[i].pNext = &chain[i-1];
}
chain[0].pNext = &chain[3];

VkInstanceCreateInfo createInfo = {};
//fill createInfo properly

createInfo.pNext = &chain.back();
VkInstance inst;
vkCreateInstance(&createInfo, null, &inst);
//should emit error.
```

mark-lunarg:
Fixes #1621.  Cyclic pNext chain errors would crash/lock up the layers.  Added check, error message and early exit.  Also plumbed in similar check for duplicates in the pNext chain, which is about to be added to the spec, along with a matching test.

### DIFF
--- a/layers/parameter_validation_utils.h
+++ b/layers/parameter_validation_utils.h
@@ -491,12 +491,17 @@ static bool validate_struct_pnext(debug_report_data *report_data, const char *ap
                                   const char *allowed_struct_names, const void *next, size_t allowed_type_count,
                                   const VkStructureType *allowed_types, uint32_t header_version) {
     bool skip_call = false;
+    std::unordered_set<const void *> cycle_check;
+    std::unordered_set<VkStructureType, std::hash<int>> unique_stype_check;
+
     const char disclaimer[] =
         "This warning is based on the Valid Usage documentation for version %d of the Vulkan header.  It "
         "is possible that you are using a struct from a private extension or an extension that was added "
         "to a later version of the Vulkan header, in which case your use of %s is perfectly valid but "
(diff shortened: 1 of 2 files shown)


layers/parameter_validation_utils.h
```diff
@@ -509,10 +514,31 @@ static bool validate_struct_pnext(debug_report_data *report_data, const char *ap
             const VkStructureType *end = allowed_types + allowed_type_count;
             const GenericHeader *current = reinterpret_cast<const GenericHeader *>(next);
 
+            cycle_check.insert(next);
+
+
             while (current != NULL) {
-                if (std::find(start, end, current->sType) == end) {
-                    std::string type_name = string_VkStructureType(current->sType);
+                if (cycle_check.find(current->pNext) != cycle_check.end()) {
```
karl-lunarg: Too bad you didn't use the "tortoise and hare" algorithm here which just needs 2 pointers and runs in O(n) time.  But since there are two other checks in the main loop and these chains should be short, this seems OK.

PR #1629, opened 2017-03-28
android: Update toolchain revisions and locations

cnorthrop:
Start pulling from github again for more recent bits.

Also update the script to handle changed url.

This is step one toward reunification with desktop toolchain versions.

Still need to update the windows script, but this gets Android tests back to passing for Linux and macOS builds.


cnorthrop:
cherry-picked to master

PR #1634, opened 2017-03-28
layers: refactor layer utils into 2 files

daveh-lunarg:
Split format-related helper fxns out of vk_layer_utils.cpp into a new
file vk_format_utils.cpp.  Remove the duplicated fxns in the render
framework and share the new format utils across both projects 
(vk_layer_utils has dependencies that prevents sharing it with 
the test framework). Rename the util fxns into CamelCase per
coding std.

This change looks huge but it's really just a global search/replace
of the format classification utility functions, which touch quite a 
few files. After this change the validation layers and the test 
framework share a common file of format helper fxns. 

Change-Id: I0f9a34bc5931dfca085dfdc8d4800664aec526cf


daveh-lunarg:
Stripped the leading 'Vk' off the format helper fxns, and rebased.

PR #1637, opened 2017-03-30
tests: Add assert for multiple initializations

daveh-lunarg:
Added an assert in render framework to flag an InitFramework() call
when the VkInstance is already initialized. Re-initializing without an
explicit shutdown was leaking Vk* resources and sowing potential ICD 
confusion. Fixed a few tests that tripped the assert.


Tony-LunarG:
Thanks for the changes - LGTM

daveh-lunarg:
Tony pointed out a much better fix for the multi-init tests that avoids using the ShutdownFramework() big hammer.

PR #1643, opened 2017-03-31
cube: Use VK_KHR_incremental_present extension

ianelliottus:
Show how to use the VK_KHR_incremental_present extension.  Other notes:

- There are a few diagnostic DbgMsg()'s, which can help show the usage.
- Added a "--incremental_present" command-line option to turn on use of VK_KHR_incremental_present
- Should compile and run on Windows, Linux, and Android, but the feature will only really be used on systems that support the extension.

PR #1644, opened 2017-04-03
Update to 1.0.46 Header PLUS Incremental_Present extension updates

mark-lunarg:
Moved to 1.0.46 version of the header, pulled in new validation checks for the `VK_KHR_incremental_present` extension as well as related demo changes.

This supersedes #1643.


MarkY-LunarG:
FYI, once this change goes in, I can remove the TODOs I have in place regarding the Update Descriptor set template tracking.

PR #1646, opened 2017-04-05
WIP -- specialization constant support

chrisforbes:
Patches 1-5 do some housekeeping in the test framework in order to make writing this test easier.
Patches 6-7 are adding new test cases (which currently fail).

Still to come: actual implementation to make the tests pass.

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -14669,6 +14669,54 @@ TEST_F(VkLayerTest, CreatePipelineMissingEntrypoint) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(VkLayerTest, CreatePipelineVsFsArraySizeDefaultSpecialization) {
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "Type mismatch");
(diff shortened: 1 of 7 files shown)


chrisforbes:
@tobine perhaps. It's easy to add one.

chrisforbes:
Note there's still some issues here:
  - We're not using the optimized module yet.
  - Need the SPIRV-Tools team to fix their opt pass (currently doing string parsing that we don't want)
  - cmake needs some more work

chrisforbes:
@tobine What do you think about landing 1-5 now, and the rest later?

tobine:
@chrisforbes Yes, landing 1-5 asap makes sense. Spin off into separate PR and run through our farm to be safe. Assuming clean it has my +1.

chrisforbes:
cmake for this is no good. I need to look at it a bit more.

mark-lunarg:
Can anything be done with the work here, or should we just archive this for now?

mark-lunarg:
@chrisforbes,  this repository will close for **write access** on **Sunday, 5/13/2018**.  If it is pushed before that time it will be present in the follow-on Vulkan-ValidationLayers repository on Monday, otherwise a new PR will be required in the new repo.

PR #1654, opened 2017-04-05
tests: Use logcat for output on Android

cnorthrop:
Use the following to see output when using the APK:
adb logcat | grep VulkanLayerValidationTests

Printfs still won't show up, but this should give a
better idea of how things are running on production
devices, including when a test hangs or crashes.


cnorthrop:
cherry-picked to master

PR #1657, opened 2017-04-06
build: Add notifications to travis config

cnorthrop:


cnorthrop:
Names are flexible.  I just went with Karl because he set up with travis config with me.  I think our current config only sends results on branch pushes and pull requests, not direct contributions to  master, which is a blind spot on this repo and VulkanTools.  I probably should have put that in the description.

cnorthrop:
cherry-picked to master, thanks!  Feel free to change the email addresses in there.  We can turn this off if it doesn't work out, but I think it gives another level of information.

PR #1663, opened 2017-04-11
Generically handle pNext extension chains in Unique Objects

mark-lunarg:
Prior to this patch, using any pNext-based structures containing a handle (except for AllocateMemory) would crash in the driver when using layers.

Added initial support to the unique-objects generator to unwrap Vulkan handles embedded in pNext extension chains.  Basically, if an API has any possible pNext structure that also contains a non-disatchable handle, a copy of the entire structure chain along with the unwrapped handles needs to be made and passed to the driver.

### DIFF
--- a/scripts/unique_objects_generator.py
+++ b/scripts/unique_objects_generator.py
@@ -132,7 +132,6 @@ class UniqueObjectsOutputGenerator(OutputGenerator):
             'vkDestroyInstance',
             'vkCreateDevice',
             'vkDestroyDevice',
-            'vkAllocateMemory',
             'vkCreateComputePipelines',
             'vkCreateGraphicsPipelines',
             'vkCreateSwapchainKHR',
--- a/scripts/unique_objects_generator.py
+++ b/scripts/unique_objects_generator.py
@@ -170,7 +170,8 @@ class UniqueObjectsOutputGenerator(OutputGenerator):
         self.structMembers = []        # List of StructMemberData records for all Vulkan structs
         self.extension_structs = []    # List of all structs or sister-structs containing handles
                                        # A sister-struct may contain no handles but shares <validextensionstructs> with one that does
-        self.structTypes = dict()    # Map of Vulkan struct typename to required VkStructureType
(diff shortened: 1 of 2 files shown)


scripts/unique_objects_generator.py
```diff
@@ -433,6 +450,78 @@ def getNdosInParameterList(self, item_list, create_func):
...
+        pnext_proc += '        switch (header->sType) {\n'
+        for item in self.extension_structs:
+            struct_info = self.struct_member_dict[item]
+            if struct_info[0].feature_protect is not None:
+                pnext_proc += '#ifdef %s \n' % struct_info[0].feature_protect
+            pnext_proc += '            case %s: {\n' % self.structTypes[item].value
+            pnext_proc += '                    safe_%s *safe_struct = reinterpret_cast<safe_%s *>(new safe_%s);\n' % (item, item, item)
+            pnext_proc += '                    safe_struct->initialize(reinterpret_cast<const %s *>(cur_pnext));\n' % item
+            # Generate code to unwrap the handles
+            tmp_decl = ''
```
tobine: Assuming uniquify_members() will always return values for these 3 then there's no need to init them to empty string

scripts/unique_objects_generator.py
```diff
@@ -433,6 +450,78 @@ def getNdosInParameterList(self, item_list, create_func):
...
+            pnext_proc += '                } break;\n'
+            if struct_info[0].feature_protect is not None:
+                pnext_proc += '#endif // %s \n' % struct_info[0].feature_protect
+            pnext_proc += '\n'
+        pnext_proc += '            default:\n'
+        pnext_proc += '                break;\n'
+        pnext_proc += '        }\n\n'
+        pnext_proc += '        // Save pointer to the first structure in the pNext chain\n'
+        pnext_proc += '        head_pnext = (head_pnext ? head_pnext : cur_ext_struct);\n\n'
+        pnext_proc += '        // For any extension structure but the first, link the last structs pNext to the current ext struct\n'
```
tobine: structs -> struct's

mark-lunarg:
Made requested changes and pushed to Master.

PR #1666, opened 2017-04-12
ObjectTracker Update: Type Enums and variable names

mark-lunarg:
Updated OT to use the layer object type enum instead of the debug-report object type enum.  Also fixed up some variable names, added coverage for new object types, and added a string helper function to the codegen.

This passes cleanly on our farm, but as there are new object types Tony as (graciously!) volunteered to run these changes through CTS tonight to check for regressions.

### DIFF
--- a/layers/object_tracker.cpp
+++ b/layers/object_tracker.cpp
@@ -141,7 +141,7 @@ static void AllocateCommandBuffer(VkDevice device, const VkCommandPool command_p
 
 static bool ValidateCommandBuffer(VkDevice device, VkCommandPool command_pool, VkCommandBuffer command_buffer) {
     layer_data *device_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
-    bool skip_call = false;
+    bool skip = false;
     uint64_t object_handle = reinterpret_cast<uint64_t>(command_buffer);
     if (device_data->object_map[VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT].find(object_handle) !=
         device_data->object_map[VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT].end()) {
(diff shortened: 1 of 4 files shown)


scripts/helper_file_generator.py
```diff
@@ -569,6 +578,9 @@ def GenerateObjectTypesHeader(self):
                     done = True
                     break
             if done == False:
+                if object_type == 'kVulkanObjectTypeDebugReportCallbackEXT':
```
chrisforbes: Makes one wonder why this isn't data driven the same way, but instead emits a giant switch...

mark-lunarg: Wonder no longer -- changed to a const array.  Sheesh.

layers/object_tracker.cpp
```diff
@@ -117,25 +117,26 @@ static void ValidateQueueFlags(VkQueue queue, const char *function) {
...
-                                  VkDebugReportObjectTypeEXT object_type, VkCommandBufferLevel level) {
+                                  VkCommandBufferLevel level) {
     layer_data *device_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
 
-    log_msg(device_data->report_data, VK_DEBUG_REPORT_INFORMATION_BIT_EXT, object_type,
+    log_msg(device_data->report_data, VK_DEBUG_REPORT_INFORMATION_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
             reinterpret_cast<const uint64_t>(command_buffer), __LINE__, OBJTRACK_NONE, LayerName,
             "OBJ[0x%" PRIxLEAST64 "] : CREATE %s object 0x%" PRIxLEAST64, object_track_index++,
-            string_VkDebugReportObjectTypeEXT(object_type), reinterpret_cast<const uint64_t>(command_buffer));
+            "VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT", reinterpret_cast<const uint64_t>(command_buffer));
```
mark-lunarg: Eh, not sure, but there's only two of these guys and they're just different enough to make it a wash, probably.  Still room for improvement, for sure.

PR #1667, opened 2017-04-12
Specialize prep

chrisforbes:
First 5 housekeeping patches from the specialization handling PR.


chrisforbes:
Added one additional change to the end to address some breakage in VkLayerTest.InUseDestroyedSignaled that was affecting all our android test runs.

PR #1668, opened 2017-04-13
Tobin gh1476 ds image layout val

### LINKED ISSUE #1476: Validate image layout for descriptor images
The spec section on "vkUpdateDescriptorSets()" has the following text regarding the imageLayout member of the VkDescriptorImageInfo struct:

`imageLayout is the layout that the image will be in at the time this descriptor is accessed. imageLayout is used in descriptor updates for types VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE, VK_DESCRIPTOR_TYPE_STORAGE_IMAGE, VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER, and VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT.`

Currently there is not validation to make sure that this layout matches the actual image layout at the time the descriptor is accessed. Add that validation.

This originates from #1454

tobine:
Fixes #1476 

Validate that image layout specified in image descriptor update matches actual layout at draw time.

Also add a test to verify this case.

There's spec language for this but not a specific VU so I'll make a spec MR for that separately and then update the check/test to include that VU when it's in.

### DIFF
--- a/layers/core_validation_types.h
+++ b/layers/core_validation_types.h
@@ -862,6 +862,7 @@ const CHECK_DISABLED *GetDisables(layer_data *);
 std::unordered_map<VkImage, std::unique_ptr<IMAGE_STATE>> *GetImageMap(core_validation::layer_data *);
 std::unordered_map<VkImage, std::vector<ImageSubresourcePair>> *GetImageSubresourceMap(layer_data *);
 std::unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE> *GetImageLayoutMap(layer_data *);
+std::unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE> const *GetImageLayoutMap(layer_data const *);
 std::unordered_map<VkBuffer, std::unique_ptr<BUFFER_STATE>> *GetBufferMap(layer_data *device_data);
 std::unordered_map<VkBufferView, std::unique_ptr<BUFFER_VIEW_STATE>> *GetBufferViewMap(layer_data *device_data);
 std::unordered_map<VkImageView, std::unique_ptr<IMAGE_VIEW_STATE>> *GetImageViewMap(layer_data *device_data);
--- a/layers/descriptor_sets.cpp
+++ b/layers/descriptor_sets.cpp
@@ -24,6 +24,7 @@
 #include "descriptor_sets.h"
 #include "vk_enum_string_helper.h"
 #include "vk_safe_struct.h"
+#include "buffer_validation.h"
 #include <sstream>
 #include <algorithm>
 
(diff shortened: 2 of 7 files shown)


layers/descriptor_sets.cpp
```diff
@@ -493,7 +501,30 @@ bool cvdescriptorset::DescriptorSet::ValidateDrawState(const std::map<uint32_t,
...
-
+                        // Verify Image Layout
+                        // TODO: VALIDATION_ERROR_02981 is the error physically closest to the spec language of interest, however
+                        //  there is no VUID for the actual spec language. Need to file a spec MR to add VU language for:
+                        // imageLayout is the layout that the image subresources accessible from imageView will be in at the time
+                        // this descriptor is accessed.
+                        // Copy first mip level into sub_layers and loop over each mip level to verify layout
+                        VkImageSubresourceLayers sub_layers;
+                        sub_layers.aspectMask = image_view_ci.subresourceRange.aspectMask;
+                        sub_layers.baseArrayLayer = image_view_ci.subresourceRange.baseArrayLayer;
```
chrisforbes: We don't have a version of VerifyImageLayout that takes a full VkImageSubresourceRange?

tests/layer_validation_tests.cpp
```diff
@@ -6071,6 +6071,237 @@ TEST_F(VkLayerTest, InvalidCmdBufferDescriptorSetImageSamplerDestroyed) {
...
+    VkImage image;
+    const VkFormat tex_format = VK_FORMAT_B8G8R8A8_UNORM;
+    const int32_t tex_width = 32;
+    const int32_t tex_height = 32;
+    VkImageCreateInfo image_create_info = {};
+    image_create_info.sType = VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO;
+    image_create_info.pNext = NULL;
+    image_create_info.imageType = VK_IMAGE_TYPE_2D;
+    image_create_info.format = tex_format;
+    image_create_info.extent.width = tex_width;
```
chrisforbes: VkImageObj?

tests/layer_validation_tests.cpp
```diff
@@ -6071,6 +6071,237 @@ TEST_F(VkLayerTest, InvalidCmdBufferDescriptorSetImageSamplerDestroyed) {
...
+    // Transition image layout to shader RO
+    VkImageMemoryBarrier img_barrier = {};
+    img_barrier.sType = VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER;
+    img_barrier.pNext = NULL;
+    img_barrier.srcAccessMask = VK_ACCESS_SHADER_READ_BIT;
+    img_barrier.dstAccessMask = VK_ACCESS_SHADER_READ_BIT;
+    img_barrier.oldLayout = VK_IMAGE_LAYOUT_UNDEFINED;
+    img_barrier.newLayout = VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL;
+    img_barrier.image = image;
+    img_barrier.srcQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED;
```
chrisforbes: If VkImageObj, this is just setLayout

layers/core_validation_types.h
```diff
@@ -862,6 +862,7 @@ const CHECK_DISABLED *GetDisables(layer_data *);
 std::unordered_map<VkImage, std::unique_ptr<IMAGE_STATE>> *GetImageMap(core_validation::layer_data *);
 std::unordered_map<VkImage, std::vector<ImageSubresourcePair>> *GetImageSubresourceMap(layer_data *);
 std::unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE> *GetImageLayoutMap(layer_data *);
+std::unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE> const *GetImageLayoutMap(layer_data const *);
```
tobine: still need non-const version

layers/descriptor_sets.cpp
tobine: No, I thought about adding that instead but this is the only place that we deal with verifying layouts on VkImageSubresourceRange currently so just did locally.

tobine:
I reworked test to be less verbose as @chrisforbes suggested & made a couple other changes.

PR #1671, opened 2017-04-16
layers: Special-case UO generation of QueuePresentKHR

### LINKED ISSUE #1670: vkQueuePresentKHR doesn't write to pResults with enabled VK_LAYER_GOOGLE_unique_objects layer
I've stumbled recently on a problem with `vkQueuePresentKHR()` function when supplying `pResults` pointer. Problem is that after calling this function with valid `pResults` pointer pointing to element of `VkResult` type (I use only one swapchain), the value in the object pointer by this pointer is not modified.

For example if I have a call:
```cpp
uint32_t imageIndex = /* index acquired before */;
VkResult presentResult = static_cast<VkResult>(some_invalid_value_like_9000); // happens also with valid values

VkPresentInfoKHR presentInfo;
presentInfo.sType = VK_STRUCTURE_TYPE_PRESENT_INFO_KHR;
presentInfo.pNext = nullptr;
presentInfo.waitSemaphoreCount = 1;
presentInfo.pWaitSemaphores = &waitSemaphore;
presentInfo.swapchainCount = 1;
presentInfo.pSwapchains = &swapchain;
presentInfo.pImageIndices = &imageIndex;
presentInfo.pResults = &presentResult;

VkResult presentAllResult = vkQueuePresentKHR(queue, &presentInfo);
```

then when I inspect values without that layer enabled I get:
```
presentResult == 0 (success)
presentAllResult == 0 (success)
```
but if I enable some layers including unique_objects one (it might be just it!) it is:
```
presentResults == 9000 (or whatever value was there before the call)
presentAllResult == 0 (succcess)
```

Because of that, I get a lot of warnings from other validation layers:
```
DS(ERROR): object: 0x2 type: 27 location: 11109 msgCode: 108: vkAcquireNextImageKHR: Application has already acquired the maximum number of images (0x2)
DS(ERROR): object: 0x2 type: 27 location: 11109 msgCode: 108: vkAcquireNextImageKHR: Application has already acquired the maximum number of images (0x3)
DS(ERROR): object: 0x2 type: 27 location: 11109 msgCode: 108: vkAcquireNextImageKHR: Application has already acquired the maximum number of images (0x3)
DS(ERROR): object: 0x2 type: 27 location: 11109 msgCode: 108: vkAcquireNextImageKHR: Application has already acquired the maximum number of images (0x3)
```

Of course everything works beside these warnings and these return values...

My SW/HW:
* Windows 7 x64
* Nvidia 660M GTX
* Checked on SDK 1.0.42.2 and 1.0.46.0, same on both
* Checked on drivers 375.* and newest 381.65

I've also checked code of `unique_objects.(h|cpp)` but I can't find anything that could be the reason without actually debugging whole thing.

chrisforbes:
VkPresentInfo::pResults is an /output/ array. The generated code makes a
copy on the way in (pointlessly, but other clients of safe_* might find
this reasonable), but neglects to copy it back on the way out.

Lift the current code generated for QueuePresentKHR back into the
special cases, and add the correct post-call copy.

Fixes #1670.


layers/unique_objects.cpp
```diff
@@ -585,6 +585,41 @@ VKAPI_ATTR VkResult VKAPI_CALL GetSwapchainImagesKHR(VkDevice device, VkSwapchai
...
+                }
+            }
+        }
+    }
+    VkResult result = dev_data->device_dispatch_table->QueuePresentKHR(queue, (const VkPresentInfoKHR *)local_pPresentInfo);
+
+    // pResults is an output array embedded in a structure. The code generator neglects to copy back from the safe_* version,
+    // so handle it as a special case here:
+    if (pPresentInfo && pPresentInfo->pResults) {
+        std::copy_n(local_pPresentInfo->pResults, pPresentInfo->swapchainCount, pPresentInfo->pResults);
```
tobine: Windows doesn't like this. Warning-as-error from appveyor:

> C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\xutility(2297): warning C4996: 'std::_Copy_n': Function call with parameters that may be unsafe - this call relies on the caller to check that the passed values are correct.

PR #1674, opened 2017-04-17
layers:Remove old INFO message

tobine:
INFO message in vkCreateDescriptorPool() is not needed and it was
returning a validation error if the callback returned "true" which is
just wrong. Killing the code altogether.


tobine:
Note that this fixes [GH13](https://github.com/googlesamples/vulkan-basic-samples/issues/13) from Google Samples repo.

PR #1677, opened 2017-04-18
cube: Throttle rendering rather than presentation

damienleone:
It is currently impossible to reliably throttle presentation per the
Vulkan spec.

The previous code was relying on fences returned by
vkAcquireNextImageKHR() to throttle.  The only information this fence
holds is whether it is possible to render to that image since the *last
time* it was presented, which could have happened several frames ago.

Instead, we can throttle the rendering by passing a fence to
vkQueueSubmit().

The previous code (only the cube.c version) was using a fence there to
synchronize a vkMapMemory() in demo_update_data_buffer(), which doesn't
seem necessary.

Before this commit, we were effectively throttling to the number of
frames in the swapchain rather than on FRAME_LAG.

In the FIFO present mode, this could schedule too much work in the
presentation channel (since we have to account for VBLANK events)
and thus causing undesired side effects, such as stutters when trying
to move the cube window on a desktop, which is I assume why the
throttle code was added in the first place.

### DIFF
--- a/demos/cube.c
+++ b/demos/cube.c
@@ -309,7 +309,6 @@ typedef struct {
     VkDeviceMemory uniform_memory;
     VkFramebuffer framebuffer;
     VkDescriptorSet descriptor_set;
-    VkFence fence;
 } SwapchainImageResources;
 
 struct demo {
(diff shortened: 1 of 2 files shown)


Tony-LunarG:
Looks fine to me but @ianelliottus could you take a look?

PR #1680, opened 2017-04-19
scripts:Updated Validation IDs for spec 1.0.48

tobine:
This was much nicer after that last update. Cmd line was:
python spec.py -update -remap 2937-3217:557-375:669-2819:739-536:
751-2106,4:1263-2956,5:2237-3104:2906-3203:2960-3222

Updated parser to BeautifulSoup so we can now use the HTML spec
directly from online again. You may have to install BeautifulSoup
python module separately to be able to run.

Had to manually import errors 911 and 912 which are only present in
the non-extension spec. This is an issue we're planning to fix when
unique ids are integrated into the spec which may acutally finally
happen within a month.


tobine:
Went ahead and landed this as the ID changes were relatively minor and don't want it sitting over the weekend and risking conflicts.

PR #1684, opened 2017-04-21
Core validation various tidy-ups

chrisforbes:


chrisforbes:
(trimmed off the last commit; it's not quite right)

PR #1686, opened 2017-04-23
vulkaninfo: Fix various entries in VkPhysicalDeviceFeatures output

hesiod:
The `VkPhysicalDeviceFeatures` dump missed the `multiViewport` and
`samplerAnisotropy` fields. The `alphaToOne` field was at the wrong position
(compared to the official definition in the Vulkan specification).

The fields `{vertex,fragment}StoresAndAtomics` as well as
`shaderTessellationAndGeometryPointSize` had the misleading and incorrect
names `{vertex,tessellation,geometry}SideEffects`, respectively.


mark-lunarg:
Thanks!

PR #1699, opened 2017-04-25
Embed VU codes in layer_validation_tests error messages

daveh-lunarg:
Append the VU code for easier identification of specific error messages in lvl tests.  Enabled in debug builds only.

2nd commit is a whole-file clang-format only, with no code changes.


mark-lunarg:
I like it!

PR #1710, opened 2017-04-27
vkCreateGraphicsPipeline 1.0.47 update

### LINKED ISSUE #1411: missing checks in VU error 01426
VU error 01426 complains if VK_PRIMITIVE_TOPOLOGY_PATCH_LIST is used with pTessellationState->patchControlPoints outside the valid range. However, the check doesn't seem to be checking the primitive topology or the active shader stages; thus, the error triggers even in (e.g.) VS/PS use cases with triangle list topology. I can avoid the error by passing pTessellationState=NULL when creating the pipeline, but I had assumed that struct would be ignored if tessellation is disabled; is that not the case?

mikes-lunarg:
1.0.47 added language to spec about when arguments to
vkCreateGraphicsPipelines are ignored.  Only perform validation on
pTessellationState when both tessellation control and evaluation shader
stages are present.  Only perform validation on pViewportState,
pMultisampleState, pDepthStencilState, pColorBlendState when
rasterization is enabled.

Fixes #1411


layers/parameter_validation.cpp
```diff
@@ -3054,409 +3054,403 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
...
-                    bool has_eval = false;
-
-                    for (uint32_t stage_index = 0; stage_index < pCreateInfos[i].stageCount; ++stage_index) {
-                        if (pCreateInfos[i].pStages[stage_index].stage == VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT) {
-                            has_control = true;
-                        } else if (pCreateInfos[i].pStages[stage_index].stage == VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT) {
-                            has_eval = true;
-                        }
+            if (pCreateInfos[i].pStages != nullptr) {
+                bool has_control = false;
```
chrisforbes: We have a slightly different pattern for this (with present stages / duplicated stages) in CV. I wonder if you want to move those checks here eventually.

layers/parameter_validation.cpp
```diff
@@ -3420,19 +3425,19 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
...
-                                               "VkColorComponentFlagBits", AllVkColorComponentFlagBits,
-                                               pCreateInfos[i].pColorBlendState->pAttachments[attachmentIndex].colorWriteMask, false);
+                            skip |= validate_ranged_enum(
+                                report_data, "vkCreateGraphicsPipelines",
+                                ParameterName("pCreateInfos[%i].pColorBlendState->pAttachments[%i].alphaBlendOp",
+                                              ParameterName::IndexVector{i, attachmentIndex}),
+                                "VkBlendOp", VK_BLEND_OP_BEGIN_RANGE, VK_BLEND_OP_END_RANGE,
+                                pCreateInfos[i].pColorBlendState->pAttachments[attachmentIndex].alphaBlendOp);
+
+                            skip |= validate_flags(
```
chrisforbes: Do you plan to squash this in with the previous?

layers/parameter_validation.cpp
```diff
@@ -3094,6 +3094,19 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
                                             "be VK_STRUCTURE_TYPE_PIPELINE_TESSELLATION_STATE_CREATE_INFO. %s",
                                             i, validation_error_map[VALIDATION_ERROR_00538]);
                         }
+
+                        if (pCreateInfos[i].pTessellationState->patchControlPoints == 0 ||
+                            pCreateInfos[i].pTessellationState->patchControlPoints >
+                                device_data->device_limits.maxTessellationPatchSize) {
+                            skip |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0,
```
chrisforbes: There's no live test case that covers this :(.

(There is a long-disabled one for the ==0 case).

Actual change looks fine.

cdwfs: I have some tests for various pTessellationState errors I'm preparing in a separate commit, unless somebody else wants to grab them sooner.

mikes-lunarg: That would be great Cort, thanks for the heads up!

layers/parameter_validation.cpp
mikes-lunarg: I'll squash the clang-format commit before pushing to master

layers/parameter_validation.cpp
mikes-lunarg: I think that makes sense.  Ideally we would have all of the related checks that don't require state from previous calls together here in PV.

layers/parameter_validation.cpp
```diff
@@ -3086,11 +3086,12 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
                             ParameterName("pCreateInfos[%i].pTessellationState->flags", ParameterName::IndexVector{i}),
                             pCreateInfos[i].pTessellationState->flags);
 
-                        if (pCreateInfos[i].pTessellationState->sType != VK_STRUCTURE_TYPE_PIPELINE_TESSELLATION_STATE_CREATE_INFO) {
+                        if (pCreateInfos[i].pTessellationState->sType !=
+                            VK_STRUCTURE_TYPE_PIPELINE_TESSELLATION_STATE_CREATE_INFO) {
                             skip |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0,
                                             __LINE__, VALIDATION_ERROR_00538, LayerName,
```
cdwfs: Wrong VUID; this should be 01427

mikes-lunarg: Fixed

Issue #1701, opened 2017-04-26
When creating a framebuffer, VKFramebufferCreateInfo.layers is not checked to be > 0


hrydgard:
https://www.khronos.org/registry/vulkan/specs/1.0/man/html/VkFramebufferCreateInfo.html says:

    layers must be greater than 0.

But if I try to create a framebuffer with validation enabled and layers = 0, no warning is issued.

Making a draw call with such a framebuffer bound crashes the driver on at least one major GPU family. I accidentally did this and had tons of "fun" debugging... Oh well :)

cdwfs:
width and height are also not currently validated as nonzero. I can take care of all three checks later today.

hrydgard:
Those are harder to accidentally leave out though :) But yeah, they should of course be checked too.

mikew-lunarg:
@hrydgard @cdwfs If resolved, could you close this issue?  Thanks

PR #1711, opened 2017-04-27
Check for framebuffer dimensions >0

cdwfs:
This fixes the specific issue raised in #1701, plus some related additions and cleanup in the vicinity.

One question for reviewers: should I prefer to add simple checks like these to parameter_validation rather than core_validation?


mark-lunarg:
We lean towards putting simple 'stateless' checks in parameter validation, but I know the idea is for them all to wind up in the same place someday.  But it's not worth redoing these guys.

layers/core_validation.cpp
```diff
@@ -8567,19 +8567,16 @@ static bool ValidateFramebufferCreateInfo(layer_data *dev_data, const VkFramebuf
...
-                                "width: %u, %u\n"
-                                "height: %u, %u\n"
-                                "layerCount: %u, %u\n",
-                                i, ivci.subresourceRange.baseMipLevel, i, mip_width, pCreateInfo->width, mip_height,
-                                pCreateInfo->height, ivci.subresourceRange.layerCount, pCreateInfo->layers);
+                    skip |= log_msg(
+                        dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0, __LINE__,
+                        VALIDATION_ERROR_00410, "DS",
+                        "vkCreateFramebuffer(): VkFramebufferCreateInfo attachment #%u mip level %u has dimensions smaller "
+                        "than the corresponding Here are the respective dimensions for attachment #%u, framebuffer:\n"
```
hrydgard: "than the corresponding Here are" 

Broken grammar? Seems a line got lost.

cdwfs: Fixed.

PR #1713, opened 2017-04-28
layers: Change size of X8_D24_UNORM_PACK32

Tony-LunarG:
Lots of validation errors in CTS due to this being 3 instead of 4.  Dave confirms 4 is correct.


mark-lunarg:
Push it!

tobine:
P-push it real good!

PR #1715, opened 2017-04-28
layers: Change size of D32_SFLOAT_S8_UINT

daveh-lunarg:
Change the size of the other format with 'optional' bytes, to be
consistent with the X8_D24_UNORM_PACK32 change. We'll now assume
the optional bytes when calculating buffer/image sizes for these
formats.

Complement to Tony's PR 1713.

Change-Id: I7c194b85611795cc680f76de04513c3b45a88c4b


chrisforbes:
Hang on a minute. The X8_D24 case was easy. In almost all cases we special-case this format anyway rather than using the value from FormatSize(), but:

- We'll be more conservative in validating total size of the image against `VkImageFormatProperties::maxResourceSize in` `vkCreateImage`. This seems "safe" but we'll get false positives on HW that does packed separate stencil. [The check vs maxResourceSize is pretty terrible in general-- at CreateImage time we really don't know what this number should be.]. One might say that if you're operating that close to maxResourceSize you're probably doing something else wrong though...

- What about D16_UNORM_S8_UINT? That's currently defined to be 3.

daveh-lunarg:
These two cases X8_D24 and D32_S8 are the only ones where the spec contains language about "optional" bits. The D16_S8 case is different because according to spec it is unambiguously 3 bytes.

We were previously assuming the optional bytes were not there (i.e. 3 / 5 bytes size) when computing buffer or image sizes, leading to the possibility of false positive validation failures on copies.  If we pick the larger "optional-bits-included" size we switch to the possibility of false negatives, where the memory area is smaller than we calculate and we miss an overflow. We kicked this around quite a while here until settling on the latter as the better option - with neither being especially good.

The whole idea of listing 'optional' format bits in the spec is a nightmare IMO.  I asked the WG for a clarification (https://gitlab.khronos.org/vulkan/vulkan/issues/768), but didn't get a clear answer back.

chrisforbes:
Ok, sounds good to me.

PR #1723, opened 2017-05-01
Add valid usage checks 1202..1207

daveh-lunarg:
Add the per-dimension VU checks to image copies.  Modify region extent helper fxn to verify a mip level's existence before calculating it's dimensions, and to treat array layers as depth for 2D images.  

Modified 2 existing copy tests to provoke each of the new VUs.


tests/layer_validation_tests.cpp
```diff
@@ -17125,76 +17125,186 @@ TEST_F(VkLayerTest, ImageFormatLimits) {
...
 TEST_F(VkLayerTest, CopyImageSrcSizeExceeded) {
     // Image copy with source region specified greater than src image size
-    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_01175);
-
     ASSERT_NO_FATAL_FAILURE(Init());
 
+    uint32_t queue_count = 1;
+    VkQueueFamilyProperties queue_props;
+    vkGetPhysicalDeviceQueueFamilyProperties(gpu(), &queue_count, &queue_props);
+    if (0 == queue_props.minImageTransferGranularity.width || 0 == queue_props.minImageTransferGranularity.height ||
```
chrisforbes: This seems confused?

1. You're guaranteed to be running on a graphics queue by the framework, so these values /must/ be  {1,1,1}.
2. You're /not/ guaranteed to be using queue family index 0.

tests/layer_validation_tests.cpp
```diff
@@ -17125,76 +17125,186 @@ TEST_F(VkLayerTest, ImageFormatLimits) {
...
     copy_region.srcSubresource.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
-    copy_region.srcSubresource.mipLevel = 0;
-    copy_region.srcSubresource.baseArrayLayer = 0;
-    copy_region.srcSubresource.layerCount = 0;
-    copy_region.srcOffset.x = 0;
-    copy_region.srcOffset.y = 0;
-    copy_region.srcOffset.z = 0;
     copy_region.dstSubresource.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
-    copy_region.dstSubresource.mipLevel = 0;
+    copy_region.srcSubresource.mipLevel = 0;  // invalid for source
```
chrisforbes: What exactly does this mean?

tests/layer_validation_tests.cpp
```diff
@@ -17125,76 +17125,186 @@ TEST_F(VkLayerTest, ImageFormatLimits) {
...
-    copy_region.dstOffset.z = 0;
-    copy_region.extent.width = 64;
-    copy_region.extent.height = 64;
-    copy_region.extent.depth = 1;
+    copy_region.srcSubresource.layerCount = 1;
+    copy_region.dstSubresource.layerCount = 1;
+    copy_region.srcOffset = {0, 0, 0};
+    copy_region.dstOffset = {0, 0, 0};
+
+    m_errorMonitor->ExpectSuccess();
```
chrisforbes: (And then you expect it to succeed here anyway. I think this test is slightly confused)

tests/layer_validation_tests.cpp
```diff
@@ -17125,76 +17125,186 @@ TEST_F(VkLayerTest, ImageFormatLimits) {
...
     m_commandBuffer->CopyImage(src_image.image(), VK_IMAGE_LAYOUT_GENERAL, dst_image.image(), VK_IMAGE_LAYOUT_GENERAL, 1,
                                &copy_region);
-    m_commandBuffer->EndCommandBuffer();
+    m_errorMonitor->VerifyNotFound();
```
chrisforbes: Why is the happy case in this test at all?

tests/layer_validation_tests.cpp
```diff
@@ -17125,76 +17125,186 @@ TEST_F(VkLayerTest, ImageFormatLimits) {
...
-    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_01176);
-
     ASSERT_NO_FATAL_FAILURE(Init());
 
-    VkImageObj src_image(m_device);
-    src_image.Init(64, 64, 1, VK_FORMAT_B8G8R8A8_UNORM, VK_IMAGE_USAGE_TRANSFER_SRC_BIT, VK_IMAGE_TILING_LINEAR, 0);
+    uint32_t queue_count = 1;
+    VkQueueFamilyProperties queue_props;
+    vkGetPhysicalDeviceQueueFamilyProperties(gpu(), &queue_count, &queue_props);
+    if (0 == queue_props.minImageTransferGranularity.width || 0 == queue_props.minImageTransferGranularity.height ||
```
chrisforbes: Similar comments to previous test.

tests/layer_validation_tests.cpp
daveh-lunarg: Code changed, comment didn't.  I'll remove it.

tests/layer_validation_tests.cpp
daveh-lunarg: I often preface a series of failing actions with a successful sanity check. The successful case doesn't feel like it merits a positive test on its own.  Is this frowned upon? Been doing it a while...

chrisforbes: I think it's /mostly/ harmless in this case. I'm not happy about some of the larger tests that hopelessly intertwine winning & losing cases.

tests/layer_validation_tests.cpp
daveh-lunarg: This was a workaround for an Android device in our farm (NexusPlayer) that advertises granularity (0,0,0) in the framework. It's a conformance failure for the device, so probably shouldn't have a workaround anyway.  I'll remove it.

PR #1727, opened 2017-05-01
layers: Migrate surface support check to core_validation

krOoze:
- fix memory leak of `pQueueFamilyIndexSupport`
- remove vestige of queue family support check in the swapchain layer

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -10238,6 +10238,29 @@ static bool PreCallValidateCreateSwapchainKHR(layer_data *dev_data, const char *
     auto most_recent_swapchain = surface_state->swapchain ? surface_state->swapchain : surface_state->old_swapchain;
 
     // TODO: revisit this. some of these rules are being relaxed.
+
+    // All physical devices and queue families are required to be able
+    // to present to any native window on Android; require the
+    // application to have established support on any other platform.
+    if (!dev_data->instance_data->androidSurfaceExtensionEnabled) {
+        auto support_predicate = [dev_data](decltype(surface_state->gpu_queue_support)::const_reference qs) -> bool {
+            // TODO: should restrict search only to queue families of VkDeviceQueueCreateInfos, not whole phys. device
+            return (qs.first.gpu == dev_data->physical_device) && qs.second;
+        };
+        const auto& support = surface_state->gpu_queue_support;
+        bool is_supported = std::any_of(support.begin(), support.end(), support_predicate);
+
+        if (!is_supported) {
+            if (log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_EXT,
+                        reinterpret_cast<uint64_t>(dev_data->device), __LINE__, VALIDATION_ERROR_01922, "DS",
+                        "%s: pCreateInfo->surface is not known at this time to be supported for presentation by this device. "
+                        "The vkGetPhysicalDeviceSurfaceSupportKHR() must be called beforehand, and it must return VK_TRUE support "
+                        "with this surface for at least one queue family of this device. %s",
+                         func_name, validation_error_map[VALIDATION_ERROR_01922]))
+                return true;
+        }
+    }
+
     if (most_recent_swapchain != old_swapchain_state || (surface_state->old_swapchain && surface_state->swapchain)) {
         if (log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_EXT,
                     reinterpret_cast<uint64_t>(dev_data->device), __LINE__, DRAWSTATE_SWAPCHAIN_ALREADY_EXISTS, "DS",
@@ -11188,7 +11211,7 @@ VKAPI_ATTR VkResult VKAPI_CALL GetPhysicalDeviceSurfaceSupportKHR(VkPhysicalDevi
         instance_data->dispatch_table.GetPhysicalDeviceSurfaceSupportKHR(physicalDevice, queueFamilyIndex, surface, pSupported);
 
     if (result == VK_SUCCESS) {
-        surface_state->gpu_queue_support[{physicalDevice, queueFamilyIndex}] = (*pSupported != 0);
+        surface_state->gpu_queue_support[{physicalDevice, queueFamilyIndex}] = (*pSupported == VK_TRUE);
     }
 
     return result;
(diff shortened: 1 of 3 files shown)


layers/core_validation.cpp
```diff
@@ -10238,6 +10238,28 @@ static bool PreCallValidateCreateSwapchainKHR(layer_data *dev_data, const char *
...
+    if (!dev_data->instance_data->androidSurfaceExtensionEnabled) {
+        auto support_it = surface_state->gpu_support.find(dev_data->physical_device);
+
+        if (support_it == surface_state->gpu_support.end()) {
+            if (log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_PHYSICAL_DEVICE_EXT,
+                        reinterpret_cast<uint64_t>(dev_data->physical_device), __LINE__, VALIDATION_ERROR_01922, "DS",
+                        "%s: Creating swapchain without calling vkGetPhysicalDeviceSurfaceSupportKHR for its surface. %s",
+                         func_name, validation_error_map[VALIDATION_ERROR_01922]))
+                return true;
+        } else if (!support_it->second) {
```
chrisforbes: Don't know which QF will be used at this point, but app is required to have determined that /some/ QF can. :+1:

layers/core_validation.cpp
```diff
@@ -11188,7 +11210,9 @@ VKAPI_ATTR VkResult VKAPI_CALL GetPhysicalDeviceSurfaceSupportKHR(VkPhysicalDevi
         instance_data->dispatch_table.GetPhysicalDeviceSurfaceSupportKHR(physicalDevice, queueFamilyIndex, surface, pSupported);
 
     if (result == VK_SUCCESS) {
-        surface_state->gpu_queue_support[{physicalDevice, queueFamilyIndex}] = (*pSupported != 0);
+        bool is_supported = (*pSupported == VK_TRUE);
+        surface_state->gpu_support[physicalDevice] = is_supported;
```
chrisforbes: This is no good, you need to OR in the new result, not replace it.

Suppose I query first a QF that can present, and then a QF that cannot (transfer queue or something). You'd like CreateSwapchainKHR to consider this GPU to be able to present to this surface in this case.

layers/core_validation.h
```diff
@@ -187,6 +187,7 @@ struct SURFACE_STATE {
     VkSurfaceKHR surface = VK_NULL_HANDLE;
     SWAPCHAIN_NODE *swapchain = nullptr;
     SWAPCHAIN_NODE *old_swapchain = nullptr;
+    std::unordered_map<VkPhysicalDevice, bool> gpu_support;
```
chrisforbes: It's unfortunate to have to add yet more tracking, but I see why. In `CreateSwapchainKHR`, you have no idea what QF will be used. Possibly consider just walking `gpu_queue_support` to get what you want instead-- it's going to be really small.

layers/swapchain.cpp
```diff
@@ -1005,34 +991,6 @@ VKAPI_ATTR VkResult VKAPI_CALL GetPhysicalDeviceSurfaceSupportKHR(VkPhysicalDevi
...
-        if ((result == VK_SUCCESS) && pSupported && pPhysicalDevice) {
-            // Record the result of this query:
-            SwpInstance *pInstance = pPhysicalDevice->pInstance;
-            SwpSurface *pSurface = (pInstance) ? pInstance->surfaces[surface] : NULL;
-            if (pSurface) {
-                pPhysicalDevice->supportedSurfaces[surface] = pSurface;
-                if (!pSurface->numQueueFamilyIndexSupport) {
-                    if (pPhysicalDevice->gotQueueFamilyPropertyCount) {
-                        pSurface->pQueueFamilyIndexSupport =
-                            (VkBool32 *)malloc(pPhysicalDevice->numOfQueueFamilies * sizeof(VkBool32));
```
chrisforbes: Good riddance.

krOoze: RIP C-ist memory management.

layers/core_validation.h
krOoze: I guess. I prefered pretier and simpler code here. I mean, what's the harm here...?
I could potentially do `unordered_map<VkPhysicalDevice, unordered_map<uint32_t,bool>>`, which would unify the two states AND avoid walking the queue families...

layers/core_validation.cpp
krOoze: Damn. Good catch.

layers/core_validation.cpp
krOoze: I just shamelessly copied code from presentKHR check here.

layers/core_validation.h
chrisforbes: `std::any_of` should be fine.

krOoze: It won't because it would check all the physical devices also

chrisforbes: If it's too ugly, let's go with the extra data.

krOoze: Nah, no way back now. 😋 It's not so bad. Just the predicate lambda won't be one-liner....

layers/core_validation.cpp
krOoze: done

layers/core_validation.cpp
```diff
@@ -10238,6 +10238,29 @@ static bool PreCallValidateCreateSwapchainKHR(layer_data *dev_data, const char *
     auto most_recent_swapchain = surface_state->swapchain ? surface_state->swapchain : surface_state->old_swapchain;
 
     // TODO: revisit this. some of these rules are being relaxed.
+
+    // All physical devices and queue families are required to be able
+    // to present to any native window on Android; require the
+    // application to have established support on any other platform.
+    if (!dev_data->instance_data->androidSurfaceExtensionEnabled) {
+        auto support_predicate = [dev_data](decltype(surface_state->gpu_queue_support)::const_reference qs) -> bool {
+            // TODO: should restrict search only to queue families of VkDeviceQueueCreateInfos, not whole phys. device
```
chrisforbes: Interesting subtle point. !/$ is pretty poor for this though.

krOoze: What's a "!/$" ??

chrisforbes: Sorry, "bang for buck"

krOoze: Eh, it is so "subtle" that people might be surprised by what they are required to do here (making such check worth it). Then again, I was too lazy to implement it at this moment....

chrisforbes:
This is running on our test farm now. I'll merge it once it's done.

PR #1729, opened 2017-05-02
layers: Convert capability handling to data

chrisforbes:
Replace the huge switch + duplicated calls with a map of ptr-to-member.

clang-format disabled on the table as it makes a mess.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -2163,162 +2163,77 @@ static bool validate_shader_capabilities(layer_data *dev_data, shader_module con
     auto report_data = dev_data->report_data;
     auto const & enabledFeatures = dev_data->enabled_features;
 
+    struct CapabilityInfo {
+        char const *name;
+        VkBool32 const VkPhysicalDeviceFeatures::*feature;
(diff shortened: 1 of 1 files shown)


layers/core_validation.cpp
```diff
@@ -2163,162 +2163,77 @@ static bool validate_shader_capabilities(layer_data *dev_data, shader_module con
...
+    };
+    // clang-format on
 
-                default:
-                    // Spirv-validator should catch these errors
-                    break;
+    for (auto insn : *src) {
+        if (insn.opcode() == spv::OpCapability) {
+            auto it = capabilities.find(insn.word(1));
+            if (it != capabilities.end()) {
```
tobine: Do we want an INFO msg in "else" case where we miss the map? Would be good to promote feedback there for seemingly un-handled capability.

chrisforbes: Well, spirv-tools should have already complained loudly in this case. Do we want to emit an INFO message, or should I just resurrect the comment explaining the spirv-tools interaction?

PR #1730, opened 2017-05-02
external: Update external reference ID for glslang

mikew-lunarg:


mikew-lunarg:
This PR was a bit premature: I'm investigating breakage on desktop right
now.

mikew-lunarg:
Upstream glslang seems to be modernizing their CMake, and recent changes are breaking desktop.  

I've moved back a couple commits, that seems to be working on my local machine and our CI farm.  
I added the update of the android pointer to this PR.  Please confirm it works on your  farm.

On our Android CI, I see many msgs of the form:
`warning: unable to get realpath for the library "/data/app/com.example.VulkanLayerValidationTests-1/lib/arm64/libVkLayer_threading.so". Will use given name.`
Dunno if that's benign?

I'll continue to investigate if there is breakage is upstream, or if those are intentional changes that we'll need to migrate to.  But at least this gives us something several weeks newer.

mikew-lunarg:
The modernized glslang CMakeFiles are using /usr/local as the default CMAKE_INSTALL_PREFIX.

I'm able to use the newer commits on Linux by specifying the previous local CMAKE_INSTALL_PREFIX value on the commandline inside update_external_sources.sh; see ec1145afc34f0f97ad8affd33d1e4e90e70440ac

The Windows version update_external_sources.bat was already doing that.

Please check if Android needs a similar change.

Issue #1712, opened 2017-04-28
Threading layer considers simultaneous use of VK_NULL_HANDLE to be an error.


AWoloszyn:
It looks like the threading layer treats VK_NULL_HANDLE as a valid object that cannot be used on multiple threads at the same time.

THREADING(ERROR): object: **0x0** type: 7 location: 0 msgCode: 1: THREADING ERROR : object of type VkFence is simultaneously used in thread 140010143512320 and thread 140010312374016

PR #1731, opened 2017-05-02
Handle VK_NULL_HANDLE in threading layer

stroyan:
This resolves GH #1712 by exempting VK_NULL_HANDLE from collision detection.
A new positive test confirms that the threading layer allows simultaneous use of VK_NULL_HANDLE.

### DIFF
--- a/layers/threading.h
+++ b/layers/threading.h
@@ -77,6 +77,9 @@ class counter {
     std::mutex counter_lock;
     std::condition_variable counter_condition;
     void startWrite(debug_report_data *report_data, T object) {
+        if (object == VK_NULL_HANDLE) {
+            return;
+        }
         bool skipCall = false;
         loader_platform_thread_id tid = loader_platform_get_thread_id();
         std::unique_lock<std::mutex> lock(counter_lock);
@@ -147,6 +150,9 @@ class counter {
     }
 
     void finishWrite(T object) {
+        if (object == VK_NULL_HANDLE) {
+            return;
+        }
         // Object is no longer in use
         std::unique_lock<std::mutex> lock(counter_lock);
         uses[object].writer_count -= 1;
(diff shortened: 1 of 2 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -23317,6 +23318,43 @@ TEST_F(VkPositiveLayerTest, LongSemaphoreChain)
     m_errorMonitor->VerifyNotFound();
 }
 
+extern "C" void *ReleaseNullFence(void *arg) {
+    struct thread_data_struct *data = (struct thread_data_struct *)arg;
+
+    for (int i = 0; i < 40000; i++) {
+        vkDestroyFence(data->device, VK_NULL_HANDLE, NULL);
+        if (data->bailout) {
```
chrisforbes: Data race here causes UB.

layers/threading.h
```diff
@@ -147,6 +150,9 @@ class counter {
     }
 
     void finishWrite(T object) {
+        if (object == VK_NULL_HANDLE) {
+            return;
```
chrisforbes: Impl looks obviously correct :+1:

tests/layer_validation_tests.cpp
stroyan: The race here on data->bailout seems benign.  It is only a Read-After-Write, which will pick up a missed new value on the next time around the loop.
Setting bailout in a thread layer callback is actually protected by the behavior of the threading layer.
The only writes to data->bailout are done in callbacks from simultaneous use, which can only happen
when both threads are simultaneously in vulkan calls.
Reworking SetBailout to a mutex protected mechanism could be done by changing to a call into a function similar to AnyDesiredMsgFound.
(AnyDesiredMsgFound, VerifyNotFound and other such functions actually lack test_platform_thread_lock_mutex(&mutex_); calls right now.)

PR #1733, opened 2017-05-02
build: set CMAKE_INSTALL_PREFIX explicitly

mikew-lunarg:
Recent changes to upstream CMakeFiles changed the default install
location.  Explicitly specify our desired install location.

Changed for Linux only; Windows was already doing this.

Change-Id: Id71945d426ac8f0a445dca3855bfa037870aaed9


update_external_sources.sh
```diff
@@ -81,7 +81,7 @@ function build_glslang () {
    cd "${BASEDIR}"/glslang
    mkdir -p build
    cd build
-   cmake -D CMAKE_BUILD_TYPE=Release ..
+   cmake -D CMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=install ..
```
cnorthrop: If I'm reading KhronosGroup/glslang@22afc38b5554925ca986b5dff5087864c586c6c3 correctly, this used to be the default we relied on, but now it has to be specified on.  LGTM, appears to generate the same result.

update_external_sources.sh
```diff
@@ -91,7 +91,7 @@ function build_spirv-tools () {
    cd "${BASEDIR}"/spirv-tools
    mkdir -p build
    cd build
-   cmake -D CMAKE_BUILD_TYPE=Release ..
+   cmake -D CMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=install ..
```
cnorthrop: We aren't calling "make install" on this, so I don't know that you need to specify it.  Probably doesn't hurt, but seems unrelated to the glslang changes?

mikew-lunarg: Yeah, was playing it safe.

update_external_sources.sh
mikew-lunarg: Yup, that's my understanding.

PR #1735, opened 2017-05-04
layers: Correct error db of VkDeviceQueueCreateInfo

krOoze:
- validation cmd is vkCreateDevice not Destroy
- non-implicit checks are implemented,
just didn't use the unique codes


mark-lunarg:
And THANKS by the way!

krOoze:
@mark-lunarg I assumed the file is edited manually (only way I can imagine to have some persistence of the error codes). Is there some script I am not aware of?

mark-lunarg:
@krOoze, yes, and it's complicated.  We update some of the fields manually and the rest are filled in when the database is periodically updated (as when there is a header update) via the layers/spec.py script.  I don't recall all the details, but obviously the 'implemented' field (Y or N) and the testcase field are hand-edited.   The other stuff gets auto-generated.

tobine:
Yes, "API" field of database is updated every time an updated spec is parsed.
This means that if the spec parsing is broken then any manual changes to API entries may be overwritten.
The spec.py script handles updating of VUIDs but big changes should be coming soon so discussing details at this point are mostly moot and if things are broken in a non-egregious manner then I'd prefer to just live with them a bit longer.

Key thing to understand is that VUIDs will soon be integrated into the spec generation process which will greatly simplify how we track them in this repo. However, it's going to require a 1-time transition from the current IDs to new IDs that are generated on the spec side. This should occur sometime this month (May 2017) and then the IDs will be fixed *forever* going forward.

PR #1738, opened 2017-05-04
layers: Improve queue family checks

### LINKED ISSUE #1720: Layers are too restrictive with gotQueueFamilyPropertyCount for vkGetPhysicalDeviceSurfaceSupportKHR
In validation layers there is this [condition](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/572ea79/layers/swapchain.cpp#L266) (simplified here):

`if( qfp_count && !qfp_is_null ) gotQueueFamilyPropertyCount = true;`

I think this is over-restrictive. E.g. this (IMO valid and reasonable) made up code would generate an validation error:

    bool fun( VkPhysicalDevice physicalDevice, VkSurfaceKHR surface ){
        uint32_t qfCount;
        vkGetPhysicalDeviceQueueFamilyProperties( physicalDevice, &qfCount, nullptr );

        if( qfCount > 0 ){
            VkBool32 isSupported;
            VkResult errorCode = vkGetPhysicalDeviceSurfaceSupportKHR(
                physicalDevice, queueFamily, surface, &isSupported
            );

            return isSupported;
        }
        
       return false;
    }

> ERROR: PhysicalDevice(0x17523E6A540): 24, Swapchain, vkGetPhysicalDeviceSurfaceSupportKHR() called before calling the vkGetPhysicalDeviceQueueFamilyProperties function.

<hr/>

I can PR this myself, given time.

<hr/>

Also the `phy_data->numOfQueueFamilies` seems to be rewritten every time on `vkGetPhysicalDeviceQueueFamilyProperties`.

And `phy_data->gotQueueFamilyPropertyCount` seems to be reset every `vkEnumeratePhysicalDevices`.

That gives an Error for many valid usages of Vulkan, and I suspect for some "reasonable" usages too,

krOoze:
WIP &mdash; needs some testing.
known problems:
 - says "max value" instead "count"
 - crashes on `vkCreateDevice` if requesting queue family without details
 - Use `CALL_STATE` better
 - forgot skip in createSwapchain
 - checking should be in `parameter_validation`

<hr />

- migrate checks from swapchain layer
- make queue family count checks more permissive
- fixes #1720


layers/swapchain.cpp
```diff
@@ -1401,19 +1132,21 @@ static PFN_vkVoidFunction intercept_khr_surface_command(const char *name, VkInst
 }
 
 static PFN_vkVoidFunction intercept_extension_instance_commands(const char *name) {
+    /*
```
mark-lunarg: Let's just delete this stuff rather than comment it out.

layers/core_validation.cpp
```diff
@@ -3463,42 +3463,77 @@ VKAPI_ATTR void VKAPI_CALL DestroyInstance(VkInstance instance, const VkAllocati
     layer_data_map.erase(key);
 }
 
-// Verify that queue family has been properly requested
-static bool ValidateRequestedQueueFamilyProperties(instance_layer_data *instance_data, VkPhysicalDevice gpu,
-                                                   const VkDeviceCreateInfo *create_info) {
+static bool validate_queue_family_indices(instance_layer_data *instance_data, VkPhysicalDevice gpu,
```
mark-lunarg: Function names must be UpperCamelCase().  Also, all changes must be run through clang-format.

layers/vk_validation_error_database.txt
```diff
@@ -56,12 +56,12 @@ VALIDATION_ERROR_00050~^~Y~^~Unknown~^~vkDestroyDevice~^~For more information re
...
-VALIDATION_ERROR_00058~^~N~^~Unknown~^~vkDestroyDevice~^~For more information refer to Vulkan Spec Section '4.3.2. Queue Creation' which states 'pQueuePriorities must be a pointer to an array of queueCount float values' (https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VkDeviceQueueCreateInfo)~^~implicit
-VALIDATION_ERROR_00059~^~N~^~Unknown~^~vkDestroyDevice~^~For more information refer to Vulkan Spec Section '4.3.2. Queue Creation' which states 'queueCount must be greater than 0' (https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VkDeviceQueueCreateInfo)~^~implicit
+VALIDATION_ERROR_00054~^~Y~^~Unknown~^~vkCreateDevice~^~For more information refer to Vulkan Spec Section '4.3.2. Queue Creation' which states 'queueFamilyIndex must be less than pQueueFamilyPropertyCount returned by vkGetPhysicalDeviceQueueFamilyProperties' (https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VkDeviceQueueCreateInfo)~^~
```
mark-lunarg: These changes were part of the last PR, right?  Might wanna rebase.

krOoze: Yes, already on it. I hoped GitHub is able to to do it seamlessly. Apparently not.

layers/core_validation.cpp
krOoze: Right sorry,
Weird convention... I read only one rule and assumed underscores everywhere. But apparently Google likes some mixed naming scheme...

layers/swapchain.cpp
krOoze: OK.

krOoze:
I got distracted by the implementation of pre-existing facilities dealing with vkGetPhysicalDeviceQueueFamilyProperties outputs. I will PR some updates to that before I continue with this.

PR #1740, opened 2017-05-05
Add VUs for vertex input attribute and binding descriptions

### LINKED ISSUE #1709: maxVertexInputAttributeOffset  is not checked in layers
There is a missing check in layers. Following spec wording mandates a restriction on VkVertexInputAttributeDescription.offset, that is not validated:

In the description of `VkVertexInputAttributeDescription`
>offset must be less than or equal to VkPhysicalDeviceLimits::maxVertexInputAttributeOffset

jeremy-lunarg:
Also fix #1709.


tests/layer_validation_tests.cpp
```diff
@@ -8726,6 +8726,493 @@ TEST_F(VkLayerTest, PSOLineWidthInvalid) {
     vkDestroyPipeline(m_device->device(), pipeline, NULL);
 }
 
+TEST_F(VkLayerTest, VALIDATION_ERROR_01407) {
```
mikew-lunarg: Is this a new convention for test function naming?  Should it be CamelCase?
I can see pros/cons both ways, just curious.

jeremy-lunarg: Just trolling the water to see if anybody responds. I named it this way for easier grepping and consistency with the usage throughout the code. I think it'd be nice to have unit tests for each VU when practical. Makes it easier to identify coverage and removes some of the subjective nature of test names.

Tony-LunarG: Can it still have a meaningful name with a comment identifying the VU ID?

mikew-lunarg: Understood, and I do see the pros.

One pattern this could break: ability to add a tiny test to an existing test function, to leverage the convenience bunch of similar setup.  That implies either a) dont use that pattern, or b) the function name must change to add a new VU?

FWIW I agree about grepping, but comments could suffice?

chrisforbes: @mikew-lunarg I'd strongly prefer if we stopped bolting things on to existing test cases. If setup hurts too much (and it clearly does!), let's address that.

jeremy-lunarg: IMO, we're not doing a very good job of **unit** testing when we combine multiple error states into the same test. We should not assume prior errors are free of side effects. Obviously, there are practical cases where it will probably be just fine. Agreed with @chrisforbes. We could be doing a better job of addressing the setup problem.

PR #1742, opened 2017-05-07
layers: Update DeviceQueueCreate checks

krOoze:
- remove some potential false-positives from QF check
- update error db with existing check
- use error db unique codes on existing checks
- move check that need state to core_validation
- deal with VK_QUEUE_FAMILY_IGNORED case
- improve error messages texts
- make messages return appropriate object
- move code that looks displaced to appropriate places
- add some locks


mark-lunarg:
This is looking much better -- thanks for going through all that trouble.  One thing I saw in a quick once-over was the commit message. It should match the other commits in the repo, something like: "layers: Update DeviceQueueCreate checks".  Thanks!

krOoze:
@mark-lunarg Right, easy fix. Have to re-commit anyway &mdash; forgot to include command name in some of the messages (and somehow missed that when testing).

layers/core_validation.cpp
```diff
@@ -10770,61 +10820,66 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDevices(VkInstance instance, uin
...
+        if (UNCALLED == pd_state->vkGetPhysicalDeviceQueueFamilyPropertiesState)
+            pd_state->vkGetPhysicalDeviceQueueFamilyPropertiesState = QUERY_COUNT;
+        pd_state->queue_family_count = count;
     } else {  // Save queue family properties
-        if (pd_state->queue_family_properties.size() < count) pd_state->queue_family_properties.resize(count);
-        for (uint32_t i = 0; i < count; i++) {
+        pd_state->vkGetPhysicalDeviceQueueFamilyPropertiesState = QUERY_DETAILS;
+        pd_state->queue_family_count = std::max(pd_state->queue_family_count, count);
+
+        pd_state->queue_family_properties.resize(std::max(static_cast<uint32_t>(pd_state->queue_family_properties.size()), count));
```
chrisforbes: This Enumerate* / Get* call tracking stuff is begging for a helper...

layers/parameter_validation.cpp
```diff
@@ -1497,55 +1498,106 @@ VKAPI_ATTR void VKAPI_CALL GetPhysicalDeviceMemoryProperties(VkPhysicalDevice ph
...
+        for (size_t i = 0; i < pCreateInfo->enabledExtensionCount; i++) {
+            skip |= validate_string(my_instance_data->report_data, "vkCreateDevice", "pCreateInfo->ppEnabledExtensionNames",
+                                    pCreateInfo->ppEnabledExtensionNames[i]);
+        }
+    }
+
+    if (pCreateInfo->pNext != NULL && pCreateInfo->pEnabledFeatures) {
+        // Check for get_physical_device_properties2 struct
+        struct std_header {
+            VkStructureType sType;
```
chrisforbes: Why are we declaring this everywhere?

layers/core_validation.cpp
```diff
@@ -3529,11 +3571,19 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDevice(VkPhysicalDevice gpu, const VkDevice
     instance_layer_data *instance_data = GetLayerDataPtr(get_dispatch_key(gpu), instance_layer_data_map);
     bool skip = false;
 
+    // TODO: object_tracker should perhaps do this instead
+    //       and it does not seem to currently work anyway -- the loader just crashes before this point
+    if (!GetPhysicalDeviceState(instance_data, gpu)) {
```
chrisforbes: The dispatch machinery can't work if the dispatchable object handle is bogus.

krOoze: This check was there before, I only moved it up.

So, should I remove it?

layers/core_validation.cpp
krOoze: Yup. In fact in my app I do treat all `Get*`/`Enunerate*` commands in a unified way. It is a rather ugly `template` (because of all the differing amout of parameters and return types).

Only some `template` class would suffice here, tracking the `callState`, `count` and the `array`...
Some distinction has to be made though between static and dynamic `Enumerate*` commands - so probably two tracking classes would be needed.

layers/parameter_validation.cpp
krOoze: Don't ask me. I did not touch this code.

krOoze:
@chrisforbes BTW I noticed that somewhere the checks are surrunded by a lock and somewhere not. Is this just ommision and should I add it everywhere?

chrisforbes:
Anywhere you're reaching into layer_data you need to be holding the lock.

layers/core_validation.cpp
chrisforbes: Probably removable. Should check what else is redundantly calling GetPhysicalDeviceState and see if you can pass this internal tracking structure in there rather than the VkPhysicalDevice.

All for a follow-up change.

layers/parameter_validation.cpp
chrisforbes: :+1: something to clean up later then.

layers/core_validation.cpp
krOoze: other similar commands do not bother at all or just use `assert`.
And AIS in the comment it is disfuctional anyway. Out with it then...

> Should check what else is redundantly calling GetPhysicalDeviceState and see if you can pass this internal tracking structure in there rather than the VkPhysicalDevice.

OK

krOoze:
Rebased. Seems fine.

krOoze:
Merged as part of #1746

PR #1743, opened 2017-05-08
layers: Remove unhelpful precheck in ValidateDrawState

chrisforbes:
This case is covered better by the error messages from `DescriptorSet::ValidateDrawState`.


chrisforbes:
Looks good on the farm.

PR #1745, opened 2017-05-08
layers: Don't skip validating image part of combined image+immut sampler

chrisforbes:


layers/descriptor_sets.cpp
```diff
@@ -410,139 +410,133 @@ bool cvdescriptorset::DescriptorSet::ValidateDrawState(const std::map<uint32_t,
...
-                                    error_str << "Descriptor in binding #" << binding << " at global descriptor index " << i
-                                              << " uses buffer " << buffer << " that references invalid memory " << mem_binding
-                                              << ".";
-                                    *error = error_str.str();
-                                    return false;
-                                }
+        auto end_idx = p_layout_->GetGlobalEndIndexFromBinding(binding);
+        auto array_idx = 0;  // Track array idx if we're dealing with array descriptors
+        for (uint32_t i = start_idx; i <= end_idx; ++i, ++array_idx) {
+            if (!descriptors_[i]->updated) {
```
tobine: For a sampler-only immutable descriptor this will now cause a false positive. I think the initial IsImmutableSampler() call above may have been added to avoid this case, then ImageSampler validation got added in "else" block later.

Ideally here we should flag an error for ImageSampler descriptor with that hasn't been updated (regardless of if it has immutable samplers or not) or for a Sampler descriptor that's NOT immutable and hasn't been updated.

chrisforbes: I don't think you get a false-positive here. Immutable sampler descriptors start off as updated.

chrisforbes: Second commit claws back another false-negative.

layers/descriptor_sets.cpp
```diff
@@ -964,7 +964,6 @@ cvdescriptorset::ImageSamplerDescriptor::ImageSamplerDescriptor(const VkSampler
     if (immut) {
         sampler_ = *immut;
         immutable_ = true;
-        updated = true;
```
tobine: yep

layers/descriptor_sets.cpp
tobine: Yep, this is good. I had looked only at the default constructor and was confused.

chrisforbes:
Farm is upset about this. Will fix.

chrisforbes:
Fixed on farm.

PR #1748, opened 2017-05-09
Check for the installed libvulkan.so.1

### LINKED ISSUE #1747: demos/smoke should check for libvulkan.so.1 instead of libvulkan.so
This would allow installing it along with the library, without the development symlink.

tjaalton:
and avoid depending on the development symlink

Fixes #1747


mark-lunarg:
Whenever this is handled, please see that the commit message is reformatted for length and style.

demos/smoke/ShellAndroid.cpp
```diff
@@ -151,7 +151,7 @@ void ShellAndroid::log(LogPriority priority, const char *msg) {
 }
 
 PFN_vkGetInstanceProcAddr ShellAndroid::load_vk() {
-    const char filename[] = "libvulkan.so";
+    const char filename[] = "libvulkan.so.1";
```
cnorthrop: Android doesn't have versioned shared objects, so this line shouldn't be changed.

tjaalton: fixed in the latest version, thanks!

PR #1750, opened 2017-05-09
layers: Add vkCreateRenderPass subpass attachment VUs

cdwfs:
This PR implements all the stateless VkSubpassDescription tests I could find (excluding the ones that involve parsing and analyzing SPIR-V code), including test coverage. Several existing tests were eliminated, where I found it easier to include them as part of the new CreateRenderPassAttachments test.

One question for reviewers: please double-check that I've intepreted VU 00349 correctly. The spec states:

> If the first use of an attachment in this render pass is as an input attachment, and the attachment is not also used as a color or depth/stencil attachment in the same subpass, then loadOp must not be VK_ATTACHMENT_LOAD_OP_CLEAR

Does "first use" include usage as a preserve attachment? My change assumes that it does since the spec doesn't explicitly exclude it, but I'd love to get a second opinion.


cdwfs:
Updates:

- The spec does indeed state that preserve attachments are not "used"; I updated the check for VU 00349 accordingly.
- To clarify, several subpass attachment VUs were already implemented; in those cases all I did was add test coverage for them. See the vk_validation_error_database.txt changes for details.

tobine:
Just an initial comment to check if you've looked at [ValidateLayouts()](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/buffer_validation.cpp#L2205) which is implemented in buffer_validation.cpp?

There's a [comment](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/core_validation.cpp#L8929) in about consolidating that function with ValidateRenderpassAttachmentUsage() that you're modifying here.

Not sure if that comment still makes sense offhand but if so would be nice to consolidate while you're making these changes or shortly thereafter.

On a quick glance I noticed that ValidateLayouts() also has some first-use tracking which prompted this comment. I'll continue review of your changes.

tests/layer_validation_tests.cpp
```diff
@@ -3685,196 +3685,185 @@ TEST_F(VkLayerTest, PipelineRenderpassCompatibility) {
...
+                                                 attachments.data(),
+                                                 (uint32_t)subpasses.size(),
+                                                 subpasses.data(),
+                                                 0,
+                                                 nullptr};
+        m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_00349);
+        err = vkCreateRenderPass(m_device->device(), &rpci_multipass, nullptr, &rp);
+        m_errorMonitor->VerifyFound();
+        if (err == VK_SUCCESS) vkDestroyRenderPass(m_device->device(), rp, nullptr);
+        attachments[input[0].attachment].loadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
```
tobine: Just friendly prep for potential future follow-on tests?

layers/core_validation.cpp
```diff
@@ -8834,6 +8835,48 @@ static bool ValidateRenderpassAttachmentUsage(layer_data *dev_data, const VkRend
...
+                    "attachment with a different layout. %s",
+                    i, j, attachment, layout, validation_error_map[VALIDATION_ERROR_00358]);
+            }
+
+            if (!skip && attachment != VK_ATTACHMENT_UNUSED && first_subpass_used[attachment] == -1) {
+                first_subpass_used[attachment] = i;
+                bool used_as_depth =
+                    (subpass.pDepthStencilAttachment != NULL && subpass.pDepthStencilAttachment->attachment == attachment);
+                bool used_as_color = false;
+                for (uint32_t k = 0; !used_as_color && k < subpass.colorAttachmentCount; ++k) {
```
cdwfs: Done.

layers/core_validation.cpp
```diff
@@ -8843,6 +8886,27 @@ static bool ValidateRenderpassAttachmentUsage(layer_data *dev_data, const VkRend
                                 validation_error_map[VALIDATION_ERROR_00356]);
             } else {
                 skip |= ValidateAttachmentIndex(dev_data, attachment, pCreateInfo->attachmentCount, "Preserve");
+
+                // According to the spec [Section 7, Render Passes], preserve attachments do not count as "used".
```
cdwfs: Done.

tests/layer_validation_tests.cpp
cdwfs: Just so!

cdwfs:
I moved the layout-related checks into ValidateLayouts() to join their brethren. Assuming AppVeyor succeeds this time, I'll go ahead and merge.

cdwfs:
Appveyor failure looks spurious; everything builds and runs correctly both locally and on the internal buildbot.

PR #1751, opened 2017-05-09
Suppress Errors/Warnings for GetPhysDevFormatProperties calls returning INVALID_FORMAT

mark-lunarg:
Received some reasonable feedback that the extra noise was unnecessary since this return code is perfectly valid, acceptable, and expected.  If anybody has requests to ignore other return codes, chime in here.

TLDR -- if `vkGetPhysicalDeviceImageFormatProperties`(), `vkGetPhysicalDeviceImageFormatProperties2KHR`() or `vkGetPhysicalDeviceExternalImageFormatPropertiesNV`() return "`VK_ERROR_FORMAT_NOT_SUPPORTED`", no log_msg call will be made.

### DIFF
--- a/layers/parameter_validation_utils.h
+++ b/layers/parameter_validation_utils.h
@@ -813,12 +813,16 @@ static std::string get_result_description(VkResult result) {
 *
 * @param report_data debug_report_data object for routing validation messages.
 * @param apiName Name of API call being validated.
+* @param ignore vector of VkResult return codes to be ignored
 * @param value VkResult value to validate.
 */
-static void validate_result(debug_report_data *report_data, const char *apiName, VkResult result) {
+static void validate_result(debug_report_data *report_data, const char *apiName, std::vector<VkResult> const &ignore,
+                            VkResult result) {
     if (result < 0 && result != VK_ERROR_VALIDATION_FAILED_EXT) {
+        if (std::find(ignore.begin(), ignore.end(), result) != ignore.end()) {
+            return;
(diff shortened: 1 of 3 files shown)


layers/parameter_validation_utils.h
```diff
@@ -813,12 +813,18 @@ static std::string get_result_description(VkResult result) {
 *
 * @param report_data debug_report_data object for routing validation messages.
 * @param apiName Name of API call being validated.
+* @param ignore vector of VkResult return codes to be ignored
 * @param value VkResult value to validate.
 */
-static void validate_result(debug_report_data *report_data, const char *apiName, VkResult result) {
+static void validate_result(debug_report_data *report_data, const char *apiName, const std::vector<VkResult> ignore,
```
jeremy-lunarg: pass ignore by ref?

mark-lunarg: Ah, sure.

layers/parameter_validation_utils.h
```diff
@@ -813,12 +813,18 @@ static std::string get_result_description(VkResult result) {
...
+                            VkResult result) {
     if (result < 0 && result != VK_ERROR_VALIDATION_FAILED_EXT) {
+        for (auto &results_to_ignore : ignore) {
```
jeremy-lunarg: I have a small preference for:

`if(std::find(ignore.begin(), ignore.end(), result) != ignore.end())`

PR #1758, opened 2017-05-10
layers: GH1752 Update ImageSubrange checks

### LINKED ISSUE #1752: Layers show false-positive when using VK_REMAINING_MIP_LEVELS
Layers are issuing false positive error when using `VK_REMAINING_MIP_LEVELS` on `vkCreateImageView`:

> ERROR: unknown(0x0): 768, IMAGE, vkCreateImageView called with baseMipLevel 0 for image 000000000000000B that only has 1 mip levels. For more information refer to Vulkan Spec Section '11.5. Image Views' which states 'If levelCount is not VK_REMAINING_MIP_LEVELS, levelCount must be non-zero and (baseMipLevel + levelCount) must be less than or equal to the mipLevels specified in VkImageCreateInfo when the image was created' (https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VkImageAspectFlagBits)

I am already working on a PR for this...

krOoze:
- fix #1752
- handle possible overflow of `level`+`count`
- show more informative msg including correct err code
- add some tests

I had to use the extended VU code + text. The unextended VU is missing in the database (though it has its own VUID in the spec).

Does not include 3D slicing tests (I don't have `VK_KHR_maintenance1` anyway).
The validation check for 3D slicing is included though. There's not much room for error (it is tied to the normal non 3D code). Someone should check it (and add the tests) just the same.


krOoze:
Ok, tests fixed here...

layers/buffer_validation.cpp
```diff
@@ -31,6 +33,19 @@
 
 #include "buffer_validation.h"
 
+// TODO: remove on NDK update (r15 will probably have proper STL impl)
+#ifdef __ANDROID__
+namespace std {
+
+template <typename T>
+std::string to_string(T var) {
```
chrisforbes: Not entirely happy with you defining stuff in std namespace. This is generally not acceptable, except for adding specializations for your own types.

layers/buffer_validation.cpp
```diff
@@ -2576,20 +2597,77 @@ bool ValidateImageAspectMask(layer_data *device_data, VkImage image, VkFormat fo
...
+
     if (subresourceRange.levelCount == 0) {
-        skip |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0, __LINE__,
-                        VALIDATION_ERROR_00768, "IMAGE", "%s called with 0 in subresourceRange.levelCount. %s", func_name,
-                        validation_error_map[VALIDATION_ERROR_00768]);
+        skip |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT,
+                        HandleToUint64(image_state->image), __LINE__, VALIDATION_ERROR_00768, "IMAGE", "%s: %s.levelCount is 0. %s",
+                        cmd_name, param_name, validation_error_map[VALIDATION_ERROR_00768]);
+    } else if (subresourceRange.levelCount == VK_REMAINING_MIP_LEVELS) {
+        // TODO: Not in the spec VUs. Probably missing -- KhronosGroup/Vulkan-Docs#416
```
chrisforbes: Are we waiting for this to be resolved before wanting to land this?

tests/layer_validation_tests.cpp
```diff
@@ -9375,6 +9375,182 @@ TEST_F(VkLayerTest, UpdateBufferWithinRenderPass) {
...
+    m_commandBuffer->BeginCommandBuffer();
+
+    VkClearColorValue clear_color;
+    memset(clear_color.uint32, 0, sizeof(uint32_t) * 4);
+    VkMemoryPropertyFlags reqs = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT;
+    const VkFormat tex_format = VK_FORMAT_B8G8R8A8_UNORM;
+    const int32_t tex_width = 32;
+    const int32_t tex_height = 32;
+    VkImageCreateInfo image_create_info = {};
+    image_create_info.sType = VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO;
```
chrisforbes: Should be able to use VkImageObj to get rid of most of this boilerplate

tests/layer_validation_tests.cpp
```diff
@@ -9375,6 +9375,182 @@ TEST_F(VkLayerTest, UpdateBufferWithinRenderPass) {
...
+    // Try baseLevel >= image.mipLevels with VK_REMAINING_ARRAY_LAYERS
+    {
+        m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                                             "vkCmdClearColorImage: pRanges[0].baseArrayLayer (= 1) is greater or equal to the "
+                                             "arrayLayers of the image when it was created (i.e. greater or equal to 1).");
+        VkImageSubresourceRange range = {VK_IMAGE_ASPECT_COLOR_BIT, 0, 1, 1, VK_REMAINING_ARRAY_LAYERS};
+        vkCmdClearColorImage(m_commandBuffer->GetBufferHandle(), dstImage.handle(), VK_IMAGE_LAYOUT_GENERAL, &clear_color, 1,
+                             &range);
+        m_errorMonitor->VerifyFound();
+    }
```
chrisforbes: Meat of this test looks good :+1:

layers/buffer_validation.cpp
cnorthrop: It does suck that we don't have it in the NDK yet.  Perhaps add a utility call that does the same, using to_string for all platforms except Android?

krOoze: It is a hack/workaround.
I **want** it to be unacceptable and to break the minute Android provides proper STL and this project migrates to that new NDK version.

Also, I already sneaked one of these in `core_validation.cpp` :yum:

Are you fine with it as explained, or should I try different approach?

chrisforbes: OK, let's go with this as you've written it.

layers/buffer_validation.cpp
krOoze: Doesn't feel like it should be my decision. IMO it is an obvious spec bug, and the resolution will not be surprising. Users can benefit in the meantime (it may be months, it may be days).

tests/layer_validation_tests.cpp
krOoze: Yea.
I just stole this stuff from the test below...

layers/buffer_validation.cpp
cnorthrop: Hah, okay.  Fine with me given precedent, although we'll have to decide if we want to force everyone to whatever NDK eventually supports it.  Looks like r16 or later now, per android-ndk/ndk#82

tests/layer_validation_tests.cpp
krOoze: Alright. Dealt with this + made some other code quality touches to the tests.

PR #1759, opened 2017-05-11
ImageMemBarriers within RenderPasses must have oldLayout == newLayout

mark-lunarg:
Fixes LX652.  Added this VU check, fixed broken tests, added new test for this VU and updated the database.

@Tony-LunarG, added you as this might cause issues in samples.  Just sayin'.


layers/core_validation.cpp
```diff
@@ -7666,56 +7666,56 @@ static bool ValidateBarriers(const char *funcName, VkCommandBuffer cmdBuffer, ui
...
-                                    VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT, reinterpret_cast<uint64_t>(cmdBuffer), __LINE__,
-                                    DRAWSTATE_INVALID_QUEUE_INDEX, "DS",
-                                    "%s: Image 0x%" PRIx64
-                                    " was created with sharingMode "
-                                    "of VK_SHARING_MODE_EXCLUSIVE, but srcQueueFamilyIndex %d"
-                                    " or dstQueueFamilyIndex %d is greater than " PRINTF_SIZE_T_SPECIFIER
-                                    "queueFamilies crated for this device.",
-                                    funcName, reinterpret_cast<const uint64_t &>(mem_barrier->image), src_q_f_index, dst_q_f_index,
-                                    dev_data->phys_dev_properties.queue_family_properties.size());
+        if (mem_barrier) {
```
chrisforbes: This null check doesn't actually work. If `imageMemBarrierCount` is erroneously >0 while `pImageMemBarriers` is `nullptr`, then `mem_barrier` is nonsense just above nullptr for the second and subsequent barriers. You'd be better to check for pImageMemBarriers being non-null outside the loop.

layers/core_validation.cpp
```diff
@@ -7667,6 +7667,19 @@ static bool ValidateBarriers(const char *funcName, VkCommandBuffer cmdBuffer, ui
     for (uint32_t i = 0; i < imageMemBarrierCount; ++i) {
         auto mem_barrier = &pImageMemBarriers[i];
         if (mem_barrier) {
+            if (pCB->activeRenderPass) {
+                if (mem_barrier->oldLayout != mem_barrier->newLayout) {
```
chrisforbes: Perhaps closer to the other code below that deals with layouts?

layers/core_validation.cpp
mark-lunarg: Yeah, and parameter_validation has already choked and died if this was so.  Will just remove the check.

layers/core_validation.cpp
mark-lunarg: Right, found the perfect spot!  Thanks.

PR #1760, opened 2017-05-11
layers: Validate pOffsets in vkCmdBindVertexBuffers

Tony-LunarG:
Addresses LunarExchange 659 and VU 01417


layers/core_validation.cpp
```diff
@@ -7068,6 +7068,12 @@ VKAPI_ATTR void VKAPI_CALL CmdBindVertexBuffers(VkCommandBuffer commandBuffer, u
                 return ValidateBufferMemoryIsValid(dev_data, buffer_state, "vkCmdBindVertexBuffers()");
             };
             cb_node->validate_functions.push_back(function);
+            if (pOffsets[i] > buffer_state->requirements.size) {
```
chrisforbes: Should compare to `buffer_state->createInfo.size`. A driver insisting on weird padding shouldn't cause larger offsets to be valid.

Probably also `>=`-- the VU says it must be less than.

Otherwise :+1:

tests/layer_validation_tests.cpp
```diff
@@ -12383,6 +12383,22 @@ TEST_F(VkLayerTest, VertexBufferInvalid) {
...
+    ASSERT_NO_FATAL_FAILURE(Init());
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+    static const float vbo_data[3] = {1.f, 0.f, 1.f};
+    VkConstantBufferObj vbo(m_device, sizeof(vbo_data), sizeof(float), (const void *)&vbo_data);
+    VkMemoryRequirements memory_reqs;
+    vkGetBufferMemoryRequirements(m_device->device(), vbo.handle(), &memory_reqs);
+    m_commandBuffer->BeginCommandBuffer();
+    m_commandBuffer->BeginRenderPass(m_renderPassBeginInfo);
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_01417);
+    BindVertexBuffer(&vbo, (VkDeviceSize)(memory_reqs.size + 1), 1);  // Offset past the end of the buffer
```
chrisforbes: late, but: Size here should match what you're testing.

Tony-LunarG: I think this is what you're asking for - pushed.

PR #1761, opened 2017-05-11
tests: Allow filtering loader tests on Windows

cnorthrop:
This mirrors what we can already do for layer tests.

Tested locally and on our farm with a test having problems.


cnorthrop:
cherry-picked to master - thanks!

PR #1763, opened 2017-05-13
layers: Add handleToUint64 log helper

krOoze:
Add unified helper that transforms any Vulkan handle into a `uint64_t`
for the purposes of the log function.

Q: Why in `vk_layer_logging.h`?
A: IMO any other uses should not exist in C++ code in the first place. This is here to conform with the C99 Vulkan API specific function which does require `uint64_t` (and for textual representation of the handle).

Q: Why not also Uint64ToHandle?
A: See above.


tobine:
This looks good. I've rebased and running on our internal farm now. Will push if there are no issues.

tobine:
Rebased and pushed to master. Thanks!

PR #1765, opened 2017-05-14
docs: updated LoaderAndLayerInterface.md

sjfricke:
According to the spec sheet, Vulkan layers **cannot** modify commands/features

https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#extended-functionality

### DIFF
--- a/loader/LoaderAndLayerInterface.md
+++ b/loader/LoaderAndLayerInterface.md
@@ -96,7 +96,7 @@ possible fashion.
 #### Layers
 
 Layers are optional components that augment the Vulkan system.  They can
-intercept, evaluate, and modify existing Vulkan functions on their way from the
+intercept and evaluate existing Vulkan functions on their way from the
 application down to the hardware.  Layers are implemented as libraries that can
 be enabled in different ways (including by application request) and are loaded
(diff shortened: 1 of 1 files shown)


jfroy:
The spec language feels vague. By modify, does it mean "modify the semantics of commands", i.e. "command Foo does A with arguments X, Y, Z" without the layer, and "command Foo does B with arguments X, Y, Z" with the layer, or does it (also) mean "modify the arguments of commands", i.e. "the ICD sees command Foo with arguments X, Y, Z" without the layer, and "the ICD sees command Foo with arguments W, Y, Z" with the layer.

mark-lunarg:
Perhaps modifying the specification makes more sense in this case.  This is an extreme limitation.

karl-lunarg:
Yes, this limitation is too extreme for layers.  For example, it should be allowed to implement a layer that presents a different (lower) set of PhysicalDevice limits than what is supported by the actual driver in order to "emulate" a device of lower capability.

We should reject his change to the LoaderAndLayerInterface spec doc and seek a loosening/clarification of "modify" in the Vulkan spec.

mark-lunarg:
As there is no further action here for this particular repository, closing this PR for now.  If anybody disagrees or comes up with some work here, please reopen.

PR #1766, opened 2017-05-14
layers: add support external_memory_win32

techlabxe:
Fixed not working properly when using extension 'external_memory_win32(NV, KHX)'
I was in trouble, so I added it easily.


krOoze:
Posibly wrong repo?
The original `vk.xml`: https://github.com/KhronosGroup/Vulkan-Docs/blob/1.0/src/spec/vk.xml

techlabxe:
oh, sorry.

thanks your advise

PR #1772, opened 2017-05-16
WIP - loader: Modify ICD dlopen to use RTLD_NOW

MarkY-LunarG:
Pierre-Loup at Valve caught this.  By using RTLD_NOW on dlopen
it fails if symbols are resolved at dlopen time.  This resolves
an error where an ICD is missing a symbol on a system causing
the loader to fail at attempting to load any ICDs.

Change-Id: Icf43fd36aebfbcb4b08e7e3891570161234118f7


loader/vk_loader_platform.h
```diff
@@ -98,7 +98,10 @@ static inline char *loader_platform_dirname(char *path) { return dirname(path);
 // Dynamic Loading of libraries:
 typedef void *loader_platform_dl_handle;
 static inline loader_platform_dl_handle loader_platform_open_library(const char *libPath) {
-    return dlopen(libPath, RTLD_LAZY | RTLD_LOCAL);
+    // When loading the library, we need to make sure we load it with RTLD_NOW to force
```
karl-lunarg: Maybe change the comment to:

Use RTLD_NOW to force the discovery of any unresolved references in the loaded module now instead of later, when they can cause problems that are harder to diagnose.

Issue #1756, opened 2017-05-10
Execution of secondary command buffer in the initial state crashes validation


martty:
If you allocate and execute a secondary command buffer without ever recording into it, the validation code looks into VkCommandBufferInheritanceInfo inside the VkCommandBufferBeginInfo during vkCmdExecute without checking the pointer for null (core_validation.cpp:9457), causing a crash.

Per the spec: "Any given element of pCommandBuffers must be in the pending or executable state.", so this is definitely an error, but I think there should be a diagnostic for this case instead of the crash.

mark-lunarg:
@martty  -- what version of the layers are you using (from an SDK maybe)?  I tried a test case and get a validation error that says I have to call EndCommandBuffer on the secondary CB before calling CmdExecuteCommands (and no crash).  This is on Master.  I'll look into adding a null ptr check in any case.

martty:
Thanks for looking into it. It is on Master, but it seems since then the line numbers moved, I am sorry for linking it incorrectly. The crash happens on https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/f25d461667ca2db55147d2be49f179945edf24db/layers/core_validation.cpp#L9568 .
Does the Primary CB have an active RenderPass? I seems to me from the code that the dereference is only hit if it does.

mark-lunarg:
Fixed by PR #1773.  Thanks @martty for the report and information!

PR #1773, opened 2017-05-16
Github issue #1756, Fix pInheritanceInfo null ptr dereference

mark-lunarg:
Trying to execute an unrecorded secondary command buffer in an active renderpass resulted in a nullptr dereference. Added null check along with a warning for unrecorded command buffers.

We found no prohibition on executing empty command buffers, so if somebody knows differently we can upgrade the warning.

_CHECK IT OUT_:  MikeS found a cool and useful feature -- when in the 'Files changed' view, adding "?w=1" to the end of the URL in your browser will tell Github to ignore whitespace, making diffs with indent changes MUCH easier to understand.  It'll help with this one.


tests/layer_validation_tests.cpp
```diff
@@ -18962,6 +18962,39 @@ TEST_F(VkLayerTest, CommandQueueFlags) {
     }
 }
 
+TEST_F(VkLayerTest, ExecuteEmptySecondaryCB) {
+    TEST_DESCRIPTION("Attempt vkCmdExecuteCommands with a secondary command buffer that has not been recorded");
+
+    ASSERT_NO_FATAL_FAILURE(Init());
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_WARNING_BIT_EXT, "is unrecorded and contains no commands");
```
chrisforbes: This should be a hard error. The spec says, in VU for CmdExecuteCommands: 

> Any given element of pCommandBuffers must be in the _pending_ or _executable_ state.

A command buffer that has never been recorded, or has been reset, is in the _initial_ state.

(A command buffer in the _pending_ or _executable_ state but which happens to contain no commands is a different beast.)

layers/core_validation.cpp
```diff
@@ -3930,6 +3930,12 @@ static bool validateCommandBufferState(layer_data *dev_data, GLOBAL_CB_NODE *cb_
     if (CB_RECORDED != cb_state->state) {
         if (CB_INVALID == cb_state->state) {
             skip |= ReportInvalidCommandBuffer(dev_data, cb_state, call_source);
+        }
+        if (CB_NEW == cb_state->state) {
+            skip |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_WARNING_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
```
chrisforbes: See complaint in test.

tests/layer_validation_tests.cpp
chrisforbes: We should probably align the state naming in CV with what the spec uses.

mark-lunarg: Sweet, will make the update in the test and check, and look at a follow on commit to update the state naming.

tests/layer_validation_tests.cpp
```diff
@@ -18962,6 +18962,53 @@ TEST_F(VkLayerTest, CommandQueueFlags) {
     }
 }
 
+TEST_F(VkLayerTest, ExecuteEmptyCBs) {
```
chrisforbes: Unrecorded?

PR #1775, opened 2017-05-16
layers: is_extension_added_token(VK_*_MAX_ENUM) returns false

### LINKED ISSUE #1770: is_extension_added_token() is true for all VK_*_MAX_ENUM values
The parameter_validation layer uses is_extension_added_token() checks for inputs larger than ExtEnumBaseValue to see whether an enum value was added by an extension (and can thus safely be outside of [VK_xxx_BEGIN_RANGE  ... VK_xxx_END_RANGE]. However, the special value VK_xxx_MAX_ENUM is always defined to be 0x7FFFFFFF, and is thus treated as an extension-added token. This means MAX_ENUM can be passed to *any* Vulkan enum parameter without triggering a validation failure (and in some cases leading to an immediate driver crash).

An easy fix would be to add a check for value < 0x7FFFFFFF in is_extension_added_token(); assuming we can count on that as the value of MAX_ENUM in all cases. A safer fix would be for validate_ranged_enum() to take a new "max_value" parameter that can be specified per-test, but that would involve a lot of extra work. Any other ideas?

cdwfs:
Also implements the VkSamplerAddressMode variant in terms of the generic
variant, and adds a test to make sure that path is covered.

(Fixes #1770)

### DIFF
--- a/layers/parameter_validation_utils.h
+++ b/layers/parameter_validation_utils.h
@@ -82,9 +82,12 @@ const std::string UnsupportedResultString = "Unhandled VkResult";
 // See Appendix C.10 "Assigning Extension Token Values" from the Vulkan specification
 const uint32_t ExtEnumBaseValue = 1000000000;
 
+// The value of all VK_xxx_MAX_ENUM tokens
+const uint32_t MaxEnumValue = 0x7FFFFFFF;
+
 template <typename T>
 bool is_extension_added_token(T value) {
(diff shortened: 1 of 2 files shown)


cdwfs:
Thanks; will merge once I get a chance to run it through the build farm.

chrisforbes:
Seeing some odd failures in the farm on ShieldTV and PixelC.

PR #1783, opened 2017-05-18
layers: Add NotTestable as a valid testname entry

daveh-lunarg:
Modify spec.py to add the keyword NotTestable as a valid test name.
Use for VU checks that are implemented but which have no reasonable
method of testing.


layers/spec.py
```diff
@@ -256,7 +256,7 @@ def genDB(self, db_file):
         db_lines.append("# <error_enum>%s<check_implemented>%s<testname>%s<api>%s<errormsg>%s<note>" % (self.delimiter, self.delimiter, self.delimiter, self.delimiter, self.delimiter))
         db_lines.append("# error_enum: Unique error enum for this check of format %s<uniqueid>" % validation_error_enum_name)
         db_lines.append("# check_implemented: 'Y' if check has been implemented in layers, or 'N' for not implemented")
-        db_lines.append("# testname: Name of validation test for this check, 'Unknown' for unknown, or 'None' if not implmented")
+        db_lines.append("# testname: Name of validation test for this check, 'Unknown' for unknown, 'None' if not implmented, or 'NotTestable' if cannot be implmented")
```
chrisforbes: implemented (presumably copy/paste from existing error in this line)

chrisforbes:
I'm curious which checks you plan to make NotTestable?

layers/spec.py
daveh-lunarg: I'll implment the spelling correction.

e.g., VU 01217, 'extent.depth must be a multiple of the compressed texel block depth'. All supported compressed formats have block depth of 1, so no test can currently be written to 
fail this VU.

daveh-lunarg: Why can't I reply to the other comment?

In any case, an example of NotTestable:
VU 01217, 'extent.depth must be a multiple of the compressed texel block depth'. All supported compressed formats have block depth of 1, so no test can currently be written to 
fail this VU.

PR #1785, opened 2017-05-18
layers: Add NotTestable to the ...stats.py script

daveh-lunarg:
I added the 'NotTestable' label to spec.py but neglected to update vk_validation_stats.py, causing commits that use NotTestable to fail the stats check on the local CI.  This corrects that omission.

Change-Id: I06befac273829cf54c1ba113be31da9ba3fee190


daveh-lunarg:
I think this is tiny and uncontroversial, so going to push it.

PR #1786, opened 2017-05-18
Better handle invalidation of primary CB due to secondary having been rerecorded

### LINKED ISSUE #1780: Validate resetting a secondary command buffer before the primary buffer is submitted
Hi, I just spent a few days chasing down a bug with the lovely folks making RenderDoc.

The issue was that I had recorded secondary command buffers for draw commands, called vkCmdExecuteCommands with my primary buffer, but before I submitted the primary buffer, I reset and re-recorded in the executed secondary buffers.

More specifically, I did a depth prepass with the secondary buffers, then called vkCmdExecuteCommands, then vkCmdNextSubpass, and then did a G-Buffer pass with the same reset buffers

The result was that the old commands were not executed at all.  The depth prepass never happened, as the contents were overwritten with the G-Buffer draw calls.

Anyway, this screams that validation should have caught it, so I'd appreciate this in there.

chrisforbes:
Tidies up various bad behavior around invalidation & secondary command buffer in-flight accounting.

- The global inflight list is gone.
- Cascaded invalidation from resource -> secondary -> primary works
- Test quality improved
- secondary CB implicit or explicit reset causes invalidation of primary CBs (Fixes #1780)

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -3096,12 +3096,19 @@ bool ValidateCmdQueueFlags(layer_data *dev_data, GLOBAL_CB_NODE *cb_node, const 
     return false;
 }
 
+static char const * GetCauseStr(VK_OBJECT obj) {
+    if (obj.type == kVulkanObjectTypeDescriptorSet)
+        return "destroyed or updated";
+    if (obj.type == kVulkanObjectTypeCommandBuffer)
+        return "destroyed or rerecorded";
+    return "destroyed";
+}
+
 static bool ReportInvalidCommandBuffer(layer_data *dev_data, GLOBAL_CB_NODE *cb_state, const char *call_source) {
     bool skip = false;
     for (auto obj : cb_state->broken_bindings) {
         const char *type_str = object_string[obj.type];
-        // Descriptor sets are a special case that can be either destroyed or updated to invalidate a CB
-        const char *cause_str = (obj.type == kVulkanObjectTypeDescriptorSet) ? "destroyed or updated" : "destroyed";
+        const char *cause_str = GetCauseStr(obj);
         skip |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
                         HandleToUint64(cb_state->commandBuffer), __LINE__, DRAWSTATE_INVALID_COMMAND_BUFFER, "DS",
                         "You are adding %s to command buffer 0x%p that is invalid because bound %s 0x%" PRIxLEAST64 " was %s.",
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -4009,20 +4009,17 @@ static bool validatePrimaryCommandBufferState(layer_data *dev_data, GLOBAL_CB_NO
 
     skip |= validateResources(dev_data, pCB);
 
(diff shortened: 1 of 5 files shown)


layers/core_validation.cpp
```diff
@@ -5474,8 +5447,8 @@ VKAPI_ATTR void VKAPI_CALL FreeCommandBuffers(VkDevice device, VkCommandPool com
         auto cb_node = GetCBNode(dev_data, pCommandBuffers[i]);
         // Delete CB information structure, and remove from commandBufferMap
         if (cb_node) {
-            dev_data->globalInFlightCmdBuffers.erase(cb_node->commandBuffer);
             // reset prior to delete for data clean-up
+            // TODO: fix this, it's insane.
```
tobine: With a quick look I believe there's just a couple calls to clearing cb bindings and maybe mem ref cleanup that can be put in a common function and called from here and reset.

layers/core_validation.cpp
```diff
@@ -143,8 +143,6 @@ struct layer_data {
 
     DeviceExtensions device_extensions = {};
     unordered_set<VkQueue> queues;  // All queues under given device
-    // Global set of all cmdBuffers that are inFlight on this device
-    unordered_set<VkCommandBuffer> globalInFlightCmdBuffers;
```
tobine: Great work, Chris. This is a vestige from prior to when CBs had an in_use so nice to see it gone. Great ROI on this CL.

layers/core_validation.cpp
```diff
@@ -5515,7 +5509,6 @@ static bool PreCallValidateDestroyCommandPool(layer_data *dev_data, VkCommandPoo
 
 static void PostCallRecordDestroyCommandPool(layer_data *dev_data, VkCommandPool pool, COMMAND_POOL_NODE *cp_state) {
     // Must remove cmdpool from cmdpoolmap, after removing all cmdbuffers in its list from the commandBufferMap
-    clearCommandBuffersInFlight(dev_data, cp_state);
```
tobine: Did you kill this function? I believe its only 2 uses go away in this CL.

layers/core_validation_types.h
```diff
@@ -666,9 +666,9 @@ struct GLOBAL_CB_NODE : public BASE_NODE {
     // Track images and buffers that are updated by this CB at the point of a draw
     std::unordered_set<VkImageView> updateImages;
     std::unordered_set<VkBuffer> updateBuffers;
-    // If cmd buffer is primary, track secondary command buffers pending
-    // execution
-    std::unordered_set<GLOBAL_CB_NODE *> secondaryCommandBuffers;
+    // If primary, the secondary command buffers we will call.
+    // If secondary, the primary command buffers we will be called by.
+    std::unordered_set<GLOBAL_CB_NODE *> linkedCommandBuffers;
```
tobine: nice

layers/core_validation.cpp
chrisforbes: The only place we lose slightly with this change is in the handling of events. Without the global in-flight list, we have to loop over all the CBs. I'm going to nuke that code sometime soon though.

layers/core_validation.cpp
chrisforbes: I mean the explicit cleanup and then explicit delete. resetCB is weirder than it needs to be because it tangles together both CB identity and a particular CB recording.

layers/core_validation.cpp
chrisforbes: Yes, this goes away.

PR #1789, opened 2017-05-19
WIP: layers: Use CB invalidation mechanism for VU135.

chrisforbes:
Makes more sense to use command buffer invalidation to drive this than
maintaining parallel logic and only picking this up at draw time.

Drops some additional TODOs in around areas that need some more work to
make this nice.

Note: Looking for feedback here, this isn't ready -- we make the validation message in this case much less clear by this change.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -3258,7 +3258,6 @@ static void resetCB(layer_data *dev_data, const VkCommandBuffer cb) {
         pCB->drawData.clear();
         pCB->currentDrawData.buffers.clear();
         pCB->vertex_buffer_used = false;
-        pCB->primaryCommandBuffer = VK_NULL_HANDLE;
         // If secondary, invalidate any primary command buffer that may call us.
         if (pCB->createInfo.level == VK_COMMAND_BUFFER_LEVEL_SECONDARY) {
             invalidateCommandBuffers(dev_data,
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -9550,6 +9540,12 @@ VKAPI_ATTR void VKAPI_CALL CmdExecuteCommands(VkCommandBuffer commandBuffer, uin
                         pCommandBuffers[i], pCB->commandBuffer);
                     pCB->beginInfo.flags &= ~VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT;
                 }
+                if (pSubCB->linkedCommandBuffers.size()) {
+                    // TODO: adjust message support so we can report an accurate error message for this.
```
tobine: Yeah, change looks good and fulfills the old TODO but currently killing VAL_ERR_00135 which is in Command Buffer Submission portion of spec. Need to preserve that check. It's similar to 133, which is handled in validateCommandBufferSimultaneousUse() just before the killed 135 check. Not immediately obvious to me the best way to preserve 135.

mark-lunarg:
This looks fine, though I have nothing constructive to add just now.

layers/core_validation.cpp
chrisforbes: From private conversation with @tobine: may be better to remove this VU from the spec as it's redundant with the 'causes the command buffer to become invalid' language.

chrisforbes:
Need to rework this a bit for the cleaned up spec language.

mark-lunarg:
What needs to be done here -- are the spec concerns still valid?

mark-lunarg:
@chrisforbes,  this repository will close for **write access** on **Sunday, 5/13/2018**.  If it is pushed before that time it will be present in the follow-on Vulkan-ValidationLayers repository on Monday, otherwise a new PR will be required in the new repo.

PR #1792, opened 2017-05-19
WIP - loader/tests: Fix realloc handling

MarkY-LunarG:
Fix how the loader handles realloc calls.  Instead of always overriding
the pointer, check to see if the allocation succeeds.

Also, set the paths in our run_all_tests.ps1 to global paths.

Finally, fix a spelling error in the test names.

Change-Id: I7b2e0d246dbe6353d29fb56367397c5d8767aaa2


MarkY-LunarG:
Ok, renamed title.

PR #1794, opened 2017-05-19
Split unrecognized value tests

chrisforbes:
Way too much going on in this test, and no shared setup so easy to split.

Fix a bug introduced by Cort's recent change, and disable the final case until we figure out why it's not working on our arm64 targets.

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -991,9 +991,7 @@ TEST_F(VkLayerTest, InvalidStructPNext) {
     m_errorMonitor->VerifyFound();
 }
 
-TEST_F(VkLayerTest, UnrecognizedValue) {
-    TEST_DESCRIPTION("Specify unrecognized Vulkan enumeration, flags, and VkBool32 values");
-
+TEST_F(VkLayerTest, UnrecognizedValueOutOfRange) {
     ASSERT_NO_FATAL_FAILURE(Init());
 
     m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT,
@@ -1006,6 +1004,10 @@ TEST_F(VkLayerTest, UnrecognizedValue) {
     VkFormatProperties format_properties;
     vkGetPhysicalDeviceFormatProperties(gpu(), static_cast<VkFormat>(8000), &format_properties);
     m_errorMonitor->VerifyFound();
+}
+
+TEST_F(VkLayerTest, UnrecognizedValueBadMask) {
+    ASSERT_NO_FATAL_FAILURE(Init());
 
     m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "contains flag bits that are not recognized members of");
     // Specify an invalid VkFlags bitmask value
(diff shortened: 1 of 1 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -991,31 +991,47 @@ TEST_F(VkLayerTest, InvalidStructPNext) {
...
     // Expected to trigger an error with parameter_validation::validate_flags
     VkImageFormatProperties image_format_properties;
-    vkGetPhysicalDeviceImageFormatProperties(gpu(), VK_FORMAT_R8G8B8A8_UNORM, VK_IMAGE_TYPE_2D, VK_IMAGE_TILING_OPTIMAL,
-                                             static_cast<VkImageUsageFlags>(1 << 25), 0, &image_format_properties);
+    vkGetPhysicalDeviceImageFormatProperties(gpu(),
+                                             VK_FORMAT_R8G8B8A8_UNORM,
+                                             VK_IMAGE_TYPE_2D,
+                                             VK_IMAGE_TILING_OPTIMAL,
+                                             static_cast<VkImageUsageFlags>(1
+                                                 << 25),
```
cdwfs: Weird formatting here, but I guess clang-format knows best.

tests/layer_validation_tests.cpp
```diff
@@ -991,31 +991,47 @@ TEST_F(VkLayerTest, InvalidStructPNext) {
...
-                                         "enumeration tokens");
+                                             "range of the core VkFormat "
+                                             "enumeration tokens");
     // Specify an invalid VkFormat value
     // Expected to trigger an error with
     // parameter_validation::validate_ranged_enum
     VkFormatProperties format_properties;
-    vkGetPhysicalDeviceFormatProperties(gpu(), static_cast<VkFormat>(8000), &format_properties);
+    vkGetPhysicalDeviceFormatProperties(gpu(),
+                                        static_cast<VkFormat>(8000),
```
cdwfs: Would it be safer to use something like VK_FORMAT_END_RANGE+1 rather than a literal that might theoretically be valid someday? Granted, it's highly unlikely we'll ever have 8000 valid formats...

tests/layer_validation_tests.cpp
```diff
@@ -1050,7 +1070,6 @@ TEST_F(VkLayerTest, UnrecognizedValue) {
     if (!supports_mirror_clamp) {
         printf("             mirror_clamp_to_edge Extension not supported, skipping tests\n");
         address_mode = VK_SAMPLER_ADDRESS_MODE_CLAMP_TO_EDGE;
-        return;
```
cdwfs: d'oh :(

tests/layer_validation_tests.cpp
```diff
@@ -1079,6 +1098,11 @@ TEST_F(VkLayerTest, UnrecognizedValue) {
     sampler_info.anisotropyEnable = 3;
     vkCreateSampler(m_device->device(), &sampler_info, NULL, &sampler);
     m_errorMonitor->VerifyFound();
+}
+
+#if 0
+TEST_F(VkLayerTest, UnrecognizedValueMaxEnum) {
```
cdwfs: As we discussed in another channel, while it's still not clear why this test is failing on some drivers, it might be safer anyway to reimplement it as a call to vkGetPhysicalDeviceFormatProperties() with VK_FORMAT_MAX_ENUM. I'll fix this in a separate PR.

tests/layer_validation_tests.cpp
chrisforbes: Fixed. clang-format had a fight with my editor.

PR #1796, opened 2017-05-19
More CmdCopyImage valid usage checks

daveh-lunarg:
Adds most (16) of the remaining VU checks for CmdCopyImage, and 2 new tests to provoke them.  There is a lot of overlap among these VUs and the image transfer granularity checks, so most of the tests unavoidably provoke an ITG check along with the target VU. 

Note to Cody: these fail on our Nexus Player for the same reason we discussed last week. Can you verify they're ok on your CI farm? I'll advertise more widely for reviewers once I know this isn't breaking stuff on your end.


daveh-lunarg:
Cody isn't seeing the failure on G's Nexus Player, so putting this out for general review.

PR #1800, opened 2017-05-22
scripts:Update Validation IDs for spec 1.0.50

tobine:
I'm attempting to migrate to the new VUIDs that are integrated into the
spec this week so this will hopefully be the last update of this type.

Migration plan is:
-Update existing VUID database to include new, integrated, string VUIDs
-Update VUID db to include core/ext data from json
-Update links/messages in VUIDs
-Move all old VUIDs in the source to use new VUIDs

Command to perform this current update was:
python spec.py -update -remap 3076-2049,2

And I manually restored 911 & 912 which are in the core spec but not in
the extension spec.


krOoze:
Would you mind ignoring, or manually restoring 768, 769 and 2931 to point to a correct struct (i.e. `VkImageSubresourceRange`)

tobine:
@krOoze Do you mean the links are bad? Almost all of the link are now updated to point directly to the VU text in the spec which have individual anchor tags as of last week.
I looked at the three you noted and I believe they're ok with this PR.

If they aren't, I've got a follow-on PR coming tomorrow or Thurs that is overhauling all of the IDs and links so please re-confirm after that PR and let me know if anything needs updating.

krOoze:
@tobine Yeah the links; they are not bad, just point to an unrelated struct. This change just overwrote what I manually changed in my last PR.

Is fine if it is soon to be changed to the VUID anchors anyway....

PR #1804, opened 2017-05-23
external: update external revisions

jeremy-lunarg:


demos/cube.vert
```diff
@@ -29,10 +29,6 @@ layout(std140, binding = 0) uniform buf {
 
 layout (location = 0) out vec4 texcoord;
 
-out gl_PerVertex {
-        vec4 gl_Position;
```
chrisforbes: This is /REQUIRED/ for an upcoming glslang change!?-- you mean to say we cannot redeclare gl_PerVertex any more? How can that possibly work-- redeclaration to add qualifiers (invariant, etc) is important!

jeremy-lunarg: https://github.com/KhronosGroup/glslang/issues/904
Thanks Chris! I didn't catch that.

PR #1805, opened 2017-05-23
android: Switch to the LLVM libc++ runtime

cnorthrop:
This includes support for std::to_string, so we can also remove the overrides.

Requires switching our shaderc build over too, which is done with a custom Application.mk.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,9 +45,9 @@ before_install:
     # Install the Android NDK.
     if [[ "$VULKAN_BUILD_TARGET" == "ANDROID" ]]; then
       export ARCH=`uname -m`
-      wget http://dl.google.com/android/repository/android-ndk-r13b-linux-${ARCH}.zip
-      unzip -u -q android-ndk-r13b-linux-${ARCH}.zip
-      export ANDROID_NDK_HOME=`pwd`/android-ndk-r13b
+      wget http://dl.google.com/android/repository/android-ndk-r16b-linux-${ARCH}.zip
+      unzip -u -q android-ndk-r16b-linux-${ARCH}.zip
+      export ANDROID_NDK_HOME=`pwd`/android-ndk-r16b
       export JAVA_HOME="/usr/lib/jvm/java-8-oracle"
       export PATH="$ANDROID_NDK_HOME:$PATH"
     fi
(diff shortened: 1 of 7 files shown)


krOoze:
Google seems to consider it not stable enough yet.
https://developer.android.com/ndk/guides/cpp-support.html
"The NDK's libc++ is not stable. Not all the tests pass, and the test suite is not comprehensive."

While messing with this, would it make sense to also try switch `-std=c++14`?

Why is shaderc `android-23` while the rest is `android-22`?

mark-lunarg:
@cnorthrop, is this one toast?  Anything we can do to move it along, or should we close this MR for now?

cnorthrop:
@stroyan has done the most recent work here and can let us know if we should close this one or replace it.

stroyan:
Times have changed.  As of NDK r16 libc++ is the preferred STL.
In the next release it will be the default STL.
The time has come to take this PR.

https://android.googlesource.com/platform/ndk/+/ndk-release-r16/CHANGELOG.md

> 
libc++ is out of beta and is now the preferred STL in the NDK. Starting in r17, libc++ is the default STL for CMake and standalone toolchains. If you manually selected a different STL, we strongly encourage you to move to libc++.

tobine:
@cnorthrop can you rebase off latest master?

krOoze:
@stroyan The CI servers are at r13 though, where it is still beta. Is there any reason to use old NDK, or should it be updated here?

cnorthrop:
Okay, rebased to master, also updated the Travis config.  We'll need to update our respective test farms and downstream repos to NDK r16b.  I'll do that on our end now.

Tony-LunarG:
Ok, I've updated our farm system - LGTM

cnorthrop:
Test run failed internally here.  Both packaging problems and new test failures.  Will work with Tony to get the packing fixed first, then focus on new test failures.

cnorthrop:
Farms are updated, packaging working.

Seeing some failures when running VkLayerValidationTests executable on rooted devices when running more than 23 tests.  Since that doesn't affect most users and running individual tests is fine, let's move forward with this PR.  We can use r15c on our farm until we have a chance to root cause the issue, which will be a high priority.

cnorthrop:
cherry-picked to master

PR #1806, opened 2017-05-23
loader: Fix extension surface params

lenny-lunarg:
This fixes surfaces not getting unwrapped properly in loader codegen, specifically in the `VK_KHR_get_surface_capabilities2` extension. Note that the extension's commands could still fail when using two physical devices, one which does support the extension, and one which doesn't.

### DIFF
--- a/scripts/loader_extension_generator.py
+++ b/scripts/loader_extension_generator.py
@@ -865,19 +865,42 @@ class LoaderExtensionOutputGenerator(OutputGenerator):
             return_prefix = '    '
             base_name = ext_cmd.name[2:]
             has_surface = 0
+            update_structure_surface = 0
+            update_structure_string = ''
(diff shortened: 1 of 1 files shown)


scripts/loader_extension_generator.py
```diff
@@ -865,19 +865,42 @@ def CreateTrampTermFuncs(self):
             return_prefix = '    '
             base_name = ext_cmd.name[2:]
             has_surface = 0
+            update_structure_surface = 0
```
mark-lunarg: Yikes! If there's no way to handle these generically (!), each of these state and accumulator variables should at least have a comment saying what they mean or what they hold.

lenny-lunarg:
Rebased to master with commit a79e092401645f342fef7969ee2a6963f891bf4e

PR #1808, opened 2017-05-23
layers: Add PVRTC formats to format utils

daveh-lunarg:
Minor housekeeping, updated vk_format_utils to classify PVRTC formats correctly.


daveh-lunarg:
Block extents added, thanks @chrisforbes.  Now I'm a little suspicious of the VkFormatCompatibilityClass for these formats, will hold off on this PR till I get that sorted.

daveh-lunarg:
Pushed.  I doubt these will get much use, but at least they'll be handled consistently if so.

PR #1809, opened 2017-05-24
Tobin vuid spec integration

tobine:
This is the VUID migration we've been pushing for since last year.

This change uses the same general format for our accounting in validation layers, but all of the ENUMs now use 8-digit HEX to represent the errors, and the mapping from a VUID string to the HEX value can be seen in vuid_mapping.py script.

The vuid_mapping.py script can be used by code-gen scripts to determine which ID relates to a given error and then errors IDs should be easily integrated into existing codegen (parameter_validation.cpp).

I've also added the file "layers/old_vuid_to_new_mapping.txt" to allow for simple reference and mapping between the legacy VUIDs and the new values. All of the code references should already be updated, but this is for ISVs, tools, issues, etc, that may need to map from the old values to the new.

There's a bit of extra functionality left over in spec.py from the previous ID method as well as from the migration itself so I'll have some follow-on commits to clean that up and streamline the new process for updating the IDs each week based on the json file.


jeremy-lunarg:
Seeing a bunch of errors in vkvalidatelayerdoc.sh. Is this to be expected?

mark-lunarg:
Speaking of which, @jeremy-lunarg, you now need to rename your VALIDATION_ERROR_0nnnn layer validation tests.

tobine:
vkvalidatelayerdoc.sh are not expected as I had it clean locally prior to testing. I'll investigate and fix. No need to rename tests as those were renamed as part of migration.

tobine:
I just pulled master, did a clean build, and ran vkvalidatelayerdoc.sh and it's passing. Can you double-check clean build, @jeremy-lunarg and if you're still seeing errors file an issue and assign to me?

jeremy-lunarg:
Mark just set me straight. It does pass, but I'm still seeing a lot of error spew. I'm told this is working as intended.

mark-lunarg:
Yeah, it'd better to suppress these warnings unless in a verbose or debug mode.

PR #1811, opened 2017-05-24
scripts:Remove cruft from spec.py script

tobine:
Kill a bunch of old stuff from spec.py now that we can use json file
for updating/tracking the VUIDs.

Update the error DB & header to be sorted by the new IDs.

Also add step to VU txt parsing to remove any backslashes which will
break compilation of vk_validation_error_messages.h.


mark-lunarg:
🥇

PR #1813, opened 2017-05-24
layers:Pipe in disable for shader validation

tobine:
Allow VK_VALIDATION_CHECK_SHADERS_EXT enum to disable the shaderc-based
shader validation in core_validation.

There were a few other pieces of internal shader validation and
tracking that were tied to the shader_validation bit but I re-enabled
those bits for now as we lose the active descriptor binding tracking if
those features are disabled. That would lead to a various other
features that need to be disabled to avoid crashes. Specifically the
connection between command buffers and active samplers, buffer, images
that are present in descriptor sets.

To test this I ran layer_validation_tests with the shaders disabled
and most tests work fine with only the ones testing internal shader
validation pieces failing as expected.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -2816,7 +2816,7 @@ static bool verifyPipelineCreateState(layer_data *dev_data, std::vector<PIPELINE
                         validation_error_map[VALIDATION_ERROR_096005ee]);
     }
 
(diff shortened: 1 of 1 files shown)


mark-lunarg:
Haven't tried it here, but wondered if y'all checked that running with the VK_NV_glsl_shader extension still works with these state updates re-enabled?

tobine:
@mark-lunarg I did not run any test with VK_NV_glsl_shader extension

PR #1814, opened 2017-05-25
Revert glslang version for KhronosGroup/glslang#904

chrisforbes:


chrisforbes:
Preferring to roll forward as glslang is fixed.

PR #1815, opened 2017-05-25
tests: Fix (and split) ExecutedUnrecordedCBs test

chrisforbes:
- We were passing RENDER_PASS_CONTINUE_BIT in the begin info for the
primary CB, but not continuing a renderpass. This is nonsense, and
crashes some drivers.

- Split up the test and use helpers to make the subtests concise.


cnorthrop:
This alleviates one of the three failures (so far) on our skylake system.  Will let you know when other platforms finish testing.

PR #1819, opened 2017-05-25
Misc framework fixes

chrisforbes:
This is a collection of test simplifications, bugfixes, and dropping of bits of framework that weren't providing any value.


tobine:
This all looks good to me and it clearing out reams of stuff that simply isn't used. We have favored closer-to-raw Vulkan code in our tests since the beginning and I think that some/most of this code was put in with a possible desire to make things more OO but none of that work has gone anywhere.
I'm reserving approval for a bit as I didn't write any of the code that's being yanked out so I'll pull in a few LunarG folks for their review/opinion.

mark-lunarg:
@Tony-LunarG, @mikew-lunarg, please have a gander.

PR #1824, opened 2017-05-26
loader: Finish VK_KHR_get_pysical_device_properties2 emulation

lenny-lunarg:
These changes finish the loader emulation for the `VK_KHR_get_physical_device_properties2` extension. This does not include emulation for other extensions, except where structs from other extensions are passed to the same Vulkan commands.

The emulation solves the problem where a user has an ICD which supports the extension, and another that doesn't. Without the emulation, the loader crashes when calling functions from this extension on a physical device that uses the unsupported extension. Since it's an instance extension, there's no way for the user to see if the call would crash beforehand.

### DIFF
--- a/loader/loader.c
+++ b/loader/loader.c
@@ -4909,6 +4909,65 @@ VKAPI_ATTR VkResult VKAPI_CALL terminator_CreateDevice(VkPhysicalDevice physical
         }
     }
 
+    // Handle loader emulation for structs that are not supported by the ICD:
+    // Presently, the emulation leaves the pNext chain alone. This means that the ICD will receive items in the chain which
+    // are not recognized by the ICD. If this causes the ICD to fail, then the items would have to be removed here. The current
+    // implementation does not remove them because copying the pNext chain would be impossible if the loader does not recognize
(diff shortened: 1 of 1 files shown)


loader/extension_manual.c
```diff
@@ -244,18 +302,23 @@ VKAPI_ATTR void VKAPI_CALL terminator_GetPhysicalDeviceFormatProperties2KHR(VkPh
...
+        // Emulate the call
         loader_log(
             icd_term->this_instance, VK_DEBUG_REPORT_INFORMATION_BIT_EXT, 0,
             "vkGetPhysicalDeviceFormatProperties2KHR: Emulating call in ICD \"%s\" using vkGetPhysicalDeviceFormatProperties",
             icd_term->scanned_icd->lib_name);
+
+        // Write to the VkFormatProperties2KHR struct
+        icd_term->dispatch.GetPhysicalDeviceFormatProperties(phys_dev_term->phys_dev, format, &pFormatProperties->formatProperties);
+
         if (pFormatProperties->pNext != NULL) {
```
mark-lunarg: Someday, it'd be nice to pull this (and things like this) out into a helper function.

Issue #1812, opened 2017-05-24
VkLayerTest.CopyImageTypeExtentMismatch dies if KHR_maintenance1 is not supported


chrisforbes:
Like so:

```
$ build/tests/vk_layer_validation_tests --gtest_filter='*CopyImageTypeExtentMismatch'
Note: Google Test filter = *CopyImageTypeExtentMismatch
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from VkLayerTest
[ RUN      ] VkLayerTest.CopyImageTypeExtentMismatch
Unexpected: loader_validate_device_extensions: Extension 0 not found in enabled layer list extensions.
Unexpected: vkCreateDevice:  Failed to validate extensions in list
../tests/vktestbinding.cpp:295: Failure
Failed
Assertion: `vkCreateDevice(phy_.handle(), &info, __null, &dev) == VK_SUCCESS'
Segmentation fault (core dumped)
```

Blindly including the extension string causes device creation to fail, and things go poorly after that.

chrisforbes:
Either:
- create another device for the test guts, which tries to enable the extension iff we've established that it exists, (preferred?) or

- Split the KHR_maintenance1 cases into a separate test where they can be skipped together if the extension isn't supported.

chrisforbes:
Possibly ties into #1827 a bit

daveh-lunarg:
The intention here was as Chris' first bullet: check for existence of KHR_maintenance1, enable it if offered, and modify the test behavior to match.  This test fails on our Android farm just as Chris listed above. I talked to Cody before pushing the PR, and thought I understood that the test was passing on Google's farm, so assumed it was a local issue here. The check/enable code comes verbatim from the positive test Maintenance1Tests() - which I presume is failing the same way?

In any case the KHRm1 extension is not essential to that test, only there to include a couple new VUs that I've proposed should be removed from the spec as redundant (although feedback is that they're more likely to remove the original overlapping VU and keep these). I can solve this particular issue by stripping it out of the test.

I actually like Chris' second bullet idea, move all KHRm1 awareness into a segregated separate set of tests (better yet, in another file).

chrisforbes:
`Maintenance1Tests` worked because it skipped the test completely before trying to use the device it failed to create.

cnorthrop:
[...]

PR #1826, opened 2017-05-26
loader: Clean up a few messages

MarkY-LunarG:
Fix a few loader debug messages when an extension isn't valid.
The previous messages were confusing and didn't indicate what
was actually wrong.

Change-Id: Iccd0a4b2cfd673187cb7e92fe9cec23deb058ff2

### DIFF
--- a/loader/loader.c
+++ b/loader/loader.c
@@ -4477,8 +4477,7 @@ VkResult loader_validate_layers(const struct loader_instance *inst, const uint32
         prop = loader_get_layer_property(ppEnabledLayerNames[i], list);
         if (NULL == prop) {
             loader_log(inst, VK_DEBUG_REPORT_ERROR_BIT_EXT, 0,
-                       "loader_validate_layers: Layer %d does not exist in "
-                       "the list of available layers",
+                       "loader_validate_layers: Layer %d does not exist in the list of available layers",
                        i);
             return VK_ERROR_LAYER_NOT_PRESENT;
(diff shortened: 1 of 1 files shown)


MarkY-LunarG:
This is related to (but does not resolve) issue #1812.  The issue resulted in an error message which was not very useful.  The actual problem is likely a test bug, but there's no way they would know that with the original loader debug message.

loader/loader.c
```diff
@@ -4546,8 +4546,8 @@ VkResult loader_validate_instance_extensions(const struct loader_instance *inst,
         if (!extension_prop) {
             // Didn't find extension name in any of the global layers, error out
             loader_log(inst, VK_DEBUG_REPORT_ERROR_BIT_EXT, 0,
-                       "loader_validate_instance_extensions: Extension %d "
-                       "not found in enabled layer list extensions.",
+                       "loader_validate_instance_extensions: Extension %d not found in extensions supported for available "
+                       "ICDs or enabled layers.",
```
lenny-lunarg: This still sounds a little awkward to me. How about something like:

`Extension %d is not supported by available ICDs or enabled layers`

MarkY-LunarG:
Ok. Fixed.  Although, I also found a bug in the usage of "VK_LOADER_DISABLE_INST_EXT_FILTER" which should have allowed unknown instance extensions to be passed along during vkCreateInstance.

PR #1828, opened 2017-05-27
More assorted test cleanup

chrisforbes:


chrisforbes:
This isn't quite right. Will fix it when I get back into the office.

PR #1829, opened 2017-05-29
layers: Add VU offset checks to BlitImage

daveh-lunarg:
Added 10 VU checks to CmdBlitImage around the offset values in the
VkImageBlit struct.  Cleaned up a few tests that were tripping
these checks inadvertently. Added BlitImageOffsets test.

Also updated the appended error message number to match the new
8-char hex format.

Change-Id: I75d319cce6bbe96e8686e0bad3c4a0d58922b7a4


cnorthrop:
This checks out on Android.

PR #1832, opened 2017-05-30
build: Speed up AppVeyor builds.

karl-lunarg:
- Build only VS 2015
- Cache the external directory


krOoze:
@karl-lunarg `Build.md` should perhaps be updated to say VS 2015 is, with this change, the main supported version.

cnorthrop:
LGTM, although I haven't spent any time looking at Appveyor yet.  Would love to get Android NDK build in there too, but the Linux version has been sufficient.

karl-lunarg:
@krOoze The VS versions used in CI builds aren't meant to convey supported versions.  In this case, we're just taking one out of CI to reduce the loads on the CI systems.  

LunarG builds and tests the SDKs using VS 2013, and most of our internal CI machines build with VS 2013 as well.

It is better to let BUILD.md describe the supported versions anyway.  And today it correctly indicates VS 2013.  (Although it could use some updating to say that VS 2013 Community Edition is sufficient and that VS 2015 and VS 2017 builds work as well.)

PR #1835, opened 2017-05-31
PV flag bits cleanup

mikes-lunarg:
Add code gen for parameters of type Vk\*FlagBits and remove unused  Vk\*FlagBits validation functions from parameter_validation.cpp


tests/layer_validation_tests.cpp
```diff
@@ -17799,9 +17802,8 @@ TEST_F(VkLayerTest, ImageFormatLimits) {
     m_errorMonitor->VerifyFound();
     image_create_info.arrayLayers = 1;
 
-    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "is not supported by format");
-    int samples = imgFmtProps.sampleCounts >> 1;
-    image_create_info.samples = (VkSampleCountFlagBits)samples;
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_09e0078e);
```
chrisforbes: There's a VU that makes this a guaranteed error due to the use of VK_IMAGE_TILING_LINEAR-- is that what you were going for, or trying to hit something that's unsupported by driver choice?

mikes-lunarg: I believe the original intent was to test the case when sample is outside the range returned by vkGetPhysicalDeviceImageFormatProperties.  You're correct that I'm cheating a little bit because section 32.4.1 of the spec forces the supported sample count to be VK_SAMPLE_COUNT_1_BIT and saves me from searching for a unsupported format/sample combination.

PR #1836, opened 2017-05-31
Move Parameter Validation extension dependency checks into codegen

### LINKED ISSUE #1810: Add Validation Checks for Extension Dependencies
Many extensions require that other extensions must be enabled.  Add validation checks to enforce these requirements.

mark-lunarg:
Each API call is now checked for any extension dependencies and validates that they were all enabled.  These checks were moved from explicit checks in the cpp file to code-generated checks in parameter_validation.h.

Also added many missing/disabled calls to PV api-validation functions.

Fixes #1810.

Later we can add pNext chain extension dependency checks, and move device_extensions.h into codegen as well.

### DIFF
--- a/layers/parameter_validation.cpp
+++ b/layers/parameter_validation.cpp
@@ -4018,9 +4018,6 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateSwapchainKHR(VkDevice device, const VkSwapc
     std::unique_lock<std::mutex> lock(global_lock);
     debug_report_data *report_data = device_data->report_data;
 
-    skip |= require_device_extension(device_data, device_data->enables.khr_swapchain, "vkCreateSwapchainKHR",
-                                     VK_KHR_SWAPCHAIN_EXTENSION_NAME);
-
     skip |= parameter_validation_vkCreateSwapchainKHR(device_data, pCreateInfo, pAllocator, pSwapchain);
 
     if (pCreateInfo != nullptr) {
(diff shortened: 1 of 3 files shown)


layers/parameter_validation.cpp
```diff
@@ -144,6 +117,23 @@ static const VkLayerProperties global_layer = {
     "VK_LAYER_LUNARG_parameter_validation", VK_LAYER_API_VERSION, 1, "LunarG Validation Layer",
 };
 
+template <typename T>
+bool ValidateRequiredExtensions(const T layer_data, const std::string &api_name, const std::vector<std::string> &required_extensions) {
```
chrisforbes: Keep the * in the parameter type, rather than deducing it as part of T. As written, `layer_data` ends up being `layer_data * const` rather than `layer_data const *` (ptr is const, not the object)

layers/parameter_validation_utils.h
```diff
@@ -29,11 +29,43 @@
...
+struct layer_data {
+    debug_report_data *report_data = nullptr;
+    // Map for queue family index to queue count
+    std::unordered_map<uint32_t, uint32_t> queueFamilyIndexMap;
+    VkPhysicalDeviceLimits device_limits = {};
+    VkPhysicalDeviceFeatures physical_device_features = {};
+    VkPhysicalDevice physical_device = VK_NULL_HANDLE;
+    VkDevice device = VK_NULL_HANDLE;
+    DeviceExtensions enables;
+    std::unordered_set<std::string> enabled_extensions;
```
chrisforbes: Now have two parallel mechanisms for tracking the enabled extensions? It should be possible to tweak the code generation to not need a mess of strings.

mark-lunarg: Yeah, they need to merge somewhere along the line.  In the xml, each extension is assigned an ID.  Codegen could create an enum with a validation-specific generated name we could use in place of the strings AND the names for the bools in device_extensions.h.  Is that preferable in the short-term?

mark-lunarg: @chrisforbes, talked this over with @mikes-lunarg and  we plan a follow-on patch set  to codegen device_extensions.h and standardize on that mechanism.

chrisforbes: Plan gets a :+1: from me :)

layers/parameter_validation.cpp
mark-lunarg: Thanks, will fix!

jeremy-lunarg: Nice catch Chris. :smiley:

mark-lunarg:
Rebased and pushed to Master.

PR #1838, opened 2017-06-02
CodeGen device_extensions.h and unify extension enables

mark-lunarg:
device_extensions.h is now vk_extension_helper.h and is codegened.  The main changes in this patch series are that the device_extensions structure is a subclass of instance_extensions, and has a copy of that data.  All of the enabled-extension checks in the layers now use these data structures.


tobine:
@mark-lunarg I ran doc validator on this branch and not seeing the bug you reference. It's passing with what appears to be a typical list of warnings. Still reviewing the change overall.

mark-lunarg:
The first commit fixes that bug, you'd need to remove that to duplicate on Windows.

tests/layer_validation_tests.cpp
```diff
@@ -19445,6 +19445,24 @@ TEST_F(VkLayerTest, ExecuteUnrecordedPrimaryCB) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(VkLayerTest, ExtensionNotEnabled) {
+    TEST_DESCRIPTION("Validate that using an API from an unabled extension returns an error");
```
tobine: unabled->not enabled (or disabled)

scripts/helper_file_generator.py
```diff
@@ -529,6 +552,82 @@ def GenerateSafeStructHeader(self):
...
+            for ext_name, ifdef in extension_dict.items():
+                bool_name = ext_name.lower()
+                bool_name = re.sub('_extension_name', '', bool_name)
+                struct += '    bool %s;\n' % bool_name
+            struct += '\n'
+            if type == 'Instance':
+                struct += '    void InitFromInstanceCreateInfo(const VkInstanceCreateInfo *pCreateInfo) {\n'
+            else:
+                struct += '    void InitFromDeviceCreateInfo(const InstanceExtensions *instance_extensions, const VkDeviceCreateInfo *pCreateInfo) {\n'
+            struct += '        using E = %sExtensions;\n' % type
```
mikes-lunarg: Is the "using E" shorthand still useful for a generated file?

layers/core_validation.cpp
```diff
@@ -3605,7 +3605,8 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDevice(VkPhysicalDevice gpu, const VkDevice
     device_data->physical_device = gpu;
 
     device_data->report_data = layer_debug_report_create_device(instance_data->report_data, *pDevice);
-    device_data->device_extensions.InitFromDeviceCreateInfo(pCreateInfo);
+    device_data->device_extensions.InitFromDeviceCreateInfo(&instance_data->extensions, pCreateInfo);
```
mikes-lunarg: Consider renaming to device_data->extensions as it now contains both device and instance extensions.

scripts/parameter_validation_generator.py
```diff
@@ -212,6 +215,17 @@ def endFile(self):
         # C-specific
         self.newline()
 
+        # Search validation source for extension tags and replace them with the correct extension names
+        commands_text = '\n'.join(self.validation)
+        for extension, extension_name in self.extension_names.items():
+            ext_target = "REPLACE_EXT_%s_REPLACE_EXT" % extension
```
mikes-lunarg: We already talked about this, storing intermediate data in the generated source feels a little hacky

scripts/helper_file_generator.py
```diff
@@ -529,6 +552,82 @@ def GenerateSafeStructHeader(self):
...
+            if type == 'Instance':
+                extension_dict = self.instance_extension_info
+                struct += 'struct InstanceExtensions { \n'
+            else:
+                extension_dict = self.device_extension_info
+                struct += 'struct DeviceExtensions : public InstanceExtensions { \n'
+            for ext_name, ifdef in extension_dict.items():
+                bool_name = ext_name.lower()
+                bool_name = re.sub('_extension_name', '', bool_name)
+                struct += '    bool %s;\n' % bool_name
```
tobine: You could use default member initializer here set the default values of these bools to false and then skip the assignments to false below. Would save some lines of code which never hurts but isn't a huge deal since it's codegen.

scripts/parameter_validation_generator.py
mark-lunarg: Fixed, pulled the info straight from the XML.

scripts/helper_file_generator.py
mark-lunarg: Removed using directive.

layers/core_validation.cpp
mark-lunarg: Considered and done.

layers/vk_validation_stats.py
```diff
@@ -170,6 +170,7 @@ def __init__(self, source_file_list):
         # 1790 is a special case that provides an exception when an extension is enabled. No specific error is flagged, but the exception is handled so add it here
```
mark-lunarg: Check.

tests/layer_validation_tests.cpp
mark-lunarg: Wow, 'unabled'.

scripts/helper_file_generator.py
mark-lunarg: Yes, great idea, thanks!

Issue #1787, opened 2017-05-18
Memory leaks in vkCreateInstance with standard validation layer enabled


polarina:
The following program demonstrates memory leaks in vkCreateInstance when VK_LAYER_LUNARG_standard_validation is enabled.

Tested against official Linux build of VulkanSDK 1.0.49.0 on Debian Stretch. When `enabledLayerCount` is set to to zero in the sample program, no leaks are detected by either Valgrind or LeakSanitizer.

With Valgrind:
Compile: `gcc vkleak.c -lvulkan -g`
Run: `valgrind ./a.out`

With LeakSanitizer:
Compile: `gcc vkleak.c -lvulkan -g -fsanitize=address`
Run: `./a.out`

```
#include <vulkan/vulkan.h>

int main()
{
	VkInstance instance;
	VkResult result;

	result = vkCreateInstance(
		&(VkInstanceCreateInfo) {
			.sType = VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO,
			.enabledLayerCount = 1,
			.ppEnabledLayerNames = (const char *[]) {
				"VK_LAYER_LUNARG_standard_validation",
			},
		},
		NULL,
		&instance);

	if (result == VK_SUCCESS)
	{
		vkDestroyInstance(instance, NULL);
	}

	return result != VK_SUCCESS;
}
```

Output from Valgrind:
```
$ valgrind --leak-check=full ./a.out 
==31678== Memcheck, a memory error detector
==31678== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==31678== Using Valgrind-3.12.0.SVN and LibVEX; rerun with -h for copyright info
==31678== Command: ./a.out
==31678== 
==31678== 
==31678== HEAP SUMMARY:
==31678==     in use at exit: 176,055 bytes in 994 blocks
==31678==   total heap usage: 26,759 allocs, 25,765 frees, 80,452,011 bytes allocated
==31678== 
==31678== 456 bytes in 1 blocks are definitely lost in loss record 904 of 918
==31678==    at 0x4C2C21F: operator new(unsigned long) (vg_replace_malloc.c:334)
==31678==    by 0x1113F55B: ???
==31678==    by 0x1112968B: ???
==31678==    by 0x1148049F: ???
==31678==    by 0x117E8172: ???
[...]

PR #1841, opened 2017-06-04
Fix a number of memory leaks in validation layers

polarina:
These commits address the leaks reported in #1787 and some more.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5300,6 +5300,7 @@ static void PostCallRecordDestroyPipeline(layer_data *dev_data, VkPipeline pipel
                                           VK_OBJECT obj_struct) {
     // Any bound cmd buffers are now invalid
     invalidateCommandBuffers(dev_data, pipeline_state->cb_bindings, obj_struct);
+    delete getPipelineState(dev_data, pipeline);
     dev_data->pipelineMap.erase(pipeline);
 }
 
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -3020,15 +3020,15 @@ static void freeDescriptorSet(layer_data *dev_data, cvdescriptorset::DescriptorS
 // Free all DS Pools including their Sets & related sub-structs
 // NOTE : Calls to this function should be wrapped in mutex
 static void deletePools(layer_data *dev_data) {
-    if (dev_data->descriptorPoolMap.size() <= 0) return;
-    for (auto ii = dev_data->descriptorPoolMap.begin(); ii != dev_data->descriptorPoolMap.end(); ++ii) {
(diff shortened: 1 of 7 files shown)


layers/core_validation.cpp
```diff
@@ -5305,6 +5305,7 @@ static void PostCallRecordDestroyPipeline(layer_data *dev_data, VkPipeline pipel
                                           VK_OBJECT obj_struct) {
     // Any bound cmd buffers are now invalid
     invalidateCommandBuffers(dev_data, pipeline_state->cb_bindings, obj_struct);
+    delete getPipelineState(dev_data, pipeline);
```
chrisforbes: Cases like this we should really be using unique_ptr or holding the state by value, but fixing the memory leak is a good step.

chrisforbes:
Bad news, this crashes everywhere but nvidia/linux and mesa-skylake/linux.

polarina:
@chrisforbes What an unfortunate coincidence that those two platforms are the ones I tested on. I'll investigate and get back to you. Thanks for testing.

polarina:
@chrisforbes I see that there's an `gh1841` branch with these commits. I assume the tests were ran against a build from that branch? If so, can you smoke test them with #1842 merged on top? That pull request fixes a heap-use-after-free bug in the tests that the changes in 2fd398e0b0030bada3b021c7ab4f048f1631a205 trigger.

chrisforbes:
@polarina Yes, that branch was what I ran. I will test your new change now.

chrisforbes:
@polarina This PR rebased onto master (which includes #1842) still segfaults in `VkLayerTest.SparseResidencyImageCreateUnsupportedTypes` on Android.

chrisforbes:
@polarina This crash now occurs on the skylake/mesa platform as well, which should make your debugging efforts easier.

chrisforbes:
@polarina some valgrind fixes on `gh1841b`; they don't help with this, however. Disabling those two tests gets us through complete test runs on all targets, so should help narrow this down.

polarina:
@chrisforbes I have found and fixed the cause of these crashes, I hope.

GetLayerDataPtr stores the key and a pointer to newly allocated memory in an unordered_map, `layer_data_map`, and returns that allocation on later calls that give the same `data_key`. Only /some/ of the `vkDestroyInstance` and `vkDestroyDevice` functions in the layers erase the key from `layer_data_map` (and some even attempt to erase from the wrong map).

Since we now `delete` these allocations, later calls to `new` may return a pointer that has been deleted but still lingers on in the `layer_data_map`. When that happens, very bad things happen from there on.

Can you test again, please?

chrisforbes:
Much better :+1:

PR #1842, opened 2017-06-04
Fix a few bugs in layer tests

polarina:


tests/layer_validation_tests.cpp
```diff
@@ -12665,7 +12665,8 @@ TEST_F(VkLayerTest, VertexBufferInvalid) {
         // Create and bind a vertex buffer in a reduced scope, which will cause
         // it to be deleted upon leaving this scope
         const float vbo_data[3] = {1.f, 0.f, 1.f};
-        VkVerticesObj draw_verticies(m_device, 1, 1, sizeof(vbo_data), 3, vbo_data);
+        VkVerticesObj draw_verticies(m_device, 1, 1,
```
chrisforbes: Sigh. Good catch.

PR #1843, opened 2017-06-05
Rework extension string handling in test framework

daveh-lunarg:
Reorganize layer/extension string handling in the test framework.
Make extension string vectors members of the framework base class, and
remove overloads that passed them as parameters. Added validation
of layers & extension immediately before CreateInstance/CreateDevice.

Added framework helper functions InstanceLayerSupported(),
InstanceExtensionSupported() and DeviceExtensionSupported() to
verify availablility of layer/extension names.

Modified positive test Maintenance1Tests as an example of how to
use for a device extension.

A whole-file clang-format turned up a lot of line-wrap changes, not 
related to the code I'd changed. Included here as a separate commit.

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -2227,8 +2227,8 @@ TEST_F(VkLayerTest, InvalidUsageBits) {
     m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "Image should have VK_IMAGE_USAGE_TRANSFER_DST_BIT");
     m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "Buffer should have VK_BUFFER_USAGE_TRANSFER_SRC_BIT");
 
-    vkCmdCopyBufferToImage(m_commandBuffer->handle(), buffer.handle(), image.handle(),
-                           VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, 1, &region);
+    vkCmdCopyBufferToImage(m_commandBuffer->handle(), buffer.handle(), image.handle(), VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, 1,
+                           &region);
     m_errorMonitor->VerifyFound();
 }
 
@@ -3165,8 +3165,8 @@ TEST_F(VkLayerTest, ImageSampleCounts) {
         m_errorMonitor->SetDesiredFailureMsg(
             VK_DEBUG_REPORT_ERROR_BIT_EXT,
             "was created with a sample count of VK_SAMPLE_COUNT_2_BIT but must be VK_SAMPLE_COUNT_1_BIT");
-        vkCmdBlitImage(m_commandBuffer->handle(), src_image.handle(), VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL,
-                       dst_image.handle(), VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL, 1, &blit_region, VK_FILTER_NEAREST);
+        vkCmdBlitImage(m_commandBuffer->handle(), src_image.handle(), VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL, dst_image.handle(),
+                       VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL, 1, &blit_region, VK_FILTER_NEAREST);
         m_errorMonitor->VerifyFound();
         m_commandBuffer->end();
     }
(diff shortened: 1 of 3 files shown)


tests/vkrenderframework.cpp
```diff
@@ -61,42 +62,108 @@ VkRenderFramework::VkRenderFramework()
...
+VkPhysicalDevice VkRenderFramework::gpu() {
+    EXPECT_NE((VkInstance)0, inst);  // Invalid to request gpu before instance exists
+    return objs[0];
 }
 
-void VkRenderFramework::InitFramework(std::vector<const char *> instance_layer_names,
-                                      std::vector<const char *> instance_extension_names,
-                                      std::vector<const char *> device_extension_names, PFN_vkDebugReportCallbackEXT dbgFunction,
-                                      void *userData) {
+// Return true if layer name is found and spec+implementation values are >= reqeusted values
```
courtney-g: typo: reqeusted => requested

tests/vkrenderframework.cpp
```diff
@@ -61,42 +62,108 @@ VkRenderFramework::VkRenderFramework()
...
+
+    for (auto &it : layer_props) {
+        if (0 == strncmp(name, it.layerName, VK_MAX_EXTENSION_NAME_SIZE)) {
+            return ((it.specVersion >= spec) && (it.implementationVersion >= implementation));
+        }
+    }
+    return false;
+};
+
+// Return true if extension name is found and spec value is >= reqeusted spec value
```
courtney-g: typo: reqeusted => requested

tests/vkrenderframework.cpp
```diff
@@ -61,42 +62,108 @@ VkRenderFramework::VkRenderFramework()
...
+    uint32_t layer_count = 0;
+    std::vector<VkLayerProperties> layer_props;
+
+    VkResult res = vkEnumerateInstanceLayerProperties(&layer_count, NULL);
+    if (VK_SUCCESS != res) return false;
+    if (0 == layer_count) return false;
+
+    layer_props.resize(layer_count);
+    res = vkEnumerateInstanceLayerProperties(&layer_count, layer_props.data());
+    if (VK_SUCCESS != res) return false;
```
courtney-g: It is possible for the Enumerate functions to return VK_INCOMPLETE. That's not an error. It's also expected to be quite rare. LoaderAndValidation layer demos cube.c and vulkaninfo.c show examples of handling that return code.

tests/vkrenderframework.cpp
```diff
@@ -61,42 +62,108 @@ VkRenderFramework::VkRenderFramework()
...
+};
+
+void VkRenderFramework::InitFramework(PFN_vkDebugReportCallbackEXT dbgFunction, void *userData) {
     // Assert not already initialized
     ASSERT_EQ((VkInstance)0, inst);
 
+    // Remove any unsupported layer names from list
+    for (auto layer = m_instance_layer_names.begin(); layer != m_instance_layer_names.end();) {
+        if (!InstanceLayerSupported(*layer)) {
+            layer = m_instance_extension_names.erase(layer);
```
courtney-g: Shouldn't we output a warning or something here?

tests/vkrenderframework.cpp
```diff
@@ -61,42 +62,108 @@ VkRenderFramework::VkRenderFramework()
...
+        } else {
+            ++layer;
+        }
+    }
+
+    // Remove any unsupported instance extension names from list
+    for (auto ext = m_instance_extension_names.begin(); ext != m_instance_extension_names.end();) {
+        if (!InstanceExtensionSupported(*ext)) {
+            ext = m_instance_extension_names.erase(ext);
```
courtney-g: This should probably abort. In general, I think it an extension we need isn't available then it's a fatal error for the validation tests.

tests/vkrenderframework.cpp
```diff
@@ -176,7 +240,16 @@ void VkRenderFramework::GetPhysicalDeviceFeatures(VkPhysicalDeviceFeatures *feat
 }
 
 void VkRenderFramework::InitState(VkPhysicalDeviceFeatures *features, const VkCommandPoolCreateFlags flags) {
-    m_device = new VkDeviceObj(0, objs[0], device_extension_names, features);
+    // Remove any unsupported device extension names from list
+    for (auto it = m_device_extension_names.begin(); it != m_device_extension_names.end();) {
+        if (!DeviceExtensionSupported(objs[0], *it)) {
+            it = m_device_extension_names.erase(it);
```
courtney-g: I believe this should be a fatal error as we only ask for extensions needed for proper operation of the tests.

tests/layer_validation_tests.cpp
```diff
@@ -355,54 +355,62 @@ class VkLayerTest : public VkRenderFramework {
...
-         * to be enabled at create instance time.
-         */
+        m_instance_layer_names.clear();
+        m_instance_extension_names.clear();
+        m_device_extension_names.clear();
+
+        // Add default instance extensions to the list
+        m_instance_extension_names.push_back(VK_EXT_DEBUG_REPORT_EXTENSION_NAME);
+        m_instance_extension_names.push_back(VK_KHR_SURFACE_EXTENSION_NAME);
+#ifdef _WIN32
```
daveh-lunarg: I moved the unconditional line from VkRenderFramework::InitFramework() to here, but didn't notice the same extension was also added conditionally below.  Agreed it makes no sense, and I'll remove the unconditional line if it doesn't break anything.

tests/vkrenderframework.cpp
daveh-lunarg: Thx!

tests/vkrenderframework.cpp
daveh-lunarg: As I understand the spec, VK_INCOMPLETE can only be returned when the provided layer_count is smaller than the actual number of layer strings to be returned - which can't happen here. But, it would not be a surprise if my understanding is wrong.  Is there another case?

daveh-lunarg:
W.r.t validating the list of layers & extensions immediately before calling CreateInstance/Device, I'd argue that we don't want to abort in the framework. The framework should just do it's best to enable the requested extensions, and communicate what it's done back to the test, which can then decide to close gracefully if needed.  Part of the impetus behind this was to support a test that should run either with or without an extension, but needs to change it's behavior if the extension was enabled.

With this patch layers/extensions can be validated before being pushed to the m_*_names vectors, or they can be pushed without validating and then the m_*_name vectors examined after Init() to see which layers/extensions were enabled. In either case, the framework's Create* calls are protected against invalid inputs. Since I can't think of a concrete example for the second pattern that isn't also addressable by the first, I can be convinced that at least a warning is in order when culling a string from the list during Init*.

tests/vkrenderframework.cpp
courtney-g: It's possible, though rare, that the count could change from the first call to the second. Since each call is independent of the other, this can happen if the json files are changed between those calls.
E.g. something was installed that included new layers and/or extensions.

daveh-lunarg:
Removed the unconditional WSI-related extensions and put some guards in that when unsupported layers or extensions are culled will fail the test.

PR #1844, opened 2017-06-06
CMakeLists: only install headers when loader is being installed

abelal:
There are scenarios where various components such as
Vulkan-LoaderAndValidationLayers, VulkanTools etc
are installed side by side on the same system. The
header deployment gets in to a conflicting situation
in such setups.
Now we conditionalize the deployment so that the package
building the loader will install the headers as well.

Signed-off-by: Awais Belal <awais_belal@mentor.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -307,7 +307,7 @@ if(NOT WIN32)
     endif()
 endif()
 
-if(UNIX)
+if(UNIX AND BUILD_LOADER)
     install(DIRECTORY "${PROJECT_SOURCE_DIR}/include/vulkan" DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 
 # uninstall target


karl-lunarg:
The `make install` target was really only implemented and intended to be used in this LVL repo.

The problem is that the CMake code added to implement the install target in LVL migrates into the VulkanTools and VulkanSamples repos, where it is basically untested and unsupported.

The `make install` functionality in VT does mostly the same thing that it does in LVL.  It isn't going to install any of the programs/files that are specific to VT.

Would it be possible for you to simply not run `make install` on the VT repo?

We could probably proceed by fixing up the install target in VT (and VS) to actually install the programs that are unique to these repos.  But that requires some additional CMake work in VT and VS.

Another approach is to remove the install target from the VT and VS repos (until they are properly implemented).

What is the "conflicting situation" that caused you to open this issue?

abelal:
@karl-lunarg, I am integrating the Vulkan pieces for a project in Yocto build environment. Everything gets done automatically so

> Would it be possible for you to simply not run make install on the VT repo?

is not an option.
The idea behind making this change was that BUILD_LOADER would only be set to ON when building LVL and OFF when building VT or VS and that would accomplish deployment of these headers only from one source (LVL) as

> target in LVL migrates into the VulkanTools and VulkanSamples repos

so we won't have to make any changes to VS or VT.
Regarding the real conflict, in Yocto environment, no two packages can install the same set of files on the RFS. I guess you see what I am pointing at now. The above install target deploys the same headers from LVL and VS which generates a conflict and such builds fail automatically.
All that said, I do not see how this change could harm any of the components/targets and why we might be reluctant with it.

karl-lunarg:
I've been having trouble duplicating your exact results.  Could you please post your exact cmake and make incantations?

That being said, what I believe is happening is that when you say BUILD_LOADER=OFF, nearly all the targets in the repos won't build because nearly all of them depend on the loader.  The exception is the header files, because they are just files and don't depend on the loader.  So, when you build VT with BUILD_LOADER=OFF, only the header files are available to install after the main build steps run (and fail) and that's why the conflicts occur only with the header files.  Your VT build essentially built nothing.

Setting BUILD_LOADER=OFF "just happens" to build nothing due to the loader dependencies in all of the other components.  If we were to introduce another component into the repo that builds OK when the loader is not built (because it has no loader dependency), then that component would install when doing a make install and we'd be back in the same situation you are in now, even with your proposed fix.  This is because all the repos would install this new component and you would get a new conflict.

The root cause of all this is VT and VS installs the same files LVL does (assuming the install CMake code came into VT and VS from LVL correctly), AND your "RFS" filesystem cannot tolerate that.

I think it would be more reasonable to add a CMake variable to the VT and VS repos, something like, INSTALL_LVL_FILES with a default of ON.  We would then wrap the install-related CMake code for LVL components in a test on this new variable.   Then, we could add new CMake code to actually install the components that are specific to VT and VS (e.g., vktrace).

You do realize that when you do a make install of VT today that you are installing NO new components?  And if you build/install VT with BUILD_LOADER=OFF (with your fix), you are simply installing nothing?  What is the point with you even building VT and VS under these conditions?

If you'd like to proceed as I have outlined above, please close this issue and open new ones in VT and VS asking to implement "make install" for components that those repos generate and also the INSTALL_LVL_FILES var, which can be used to prevent these repos from installing LVL content on top of a previous LVL install.  (This is useful for situations other than yours since sometimes the VT and VS lag LVL)

karl-lunarg:
Superseded by pull request #1861.

PR #1858, opened 2017-06-11
Replace characters “ and ” with "

Nekotekina:
They may cause problems if the current codepage doesn't have them. For instance, in my case, VS issues a warning, and the warning is treated as an error.


Nekotekina:
Is it ok this way?

mark-lunarg:
@Nekotekina, yes, that will do nicely.  Thanks for the fix!

Issue #1846, opened 2017-06-07
Bad memory behavior in ExceedMemoryAllocationCount


chrisforbes:
Valgrind is very unhappy about this test:

```
[ RUN      ] VkLayerTest.ExceedMemoryAllocationCount
==69840== Invalid write of size 8
==69840==    at 0x4E4A11F: loader_phys_dev_ext_gpa (loader.c:3922)
==69840==    by 0x4E4EDDC: trampolineGetProcAddr (gpa_helper.h:175)
==69840==    by 0x4E4EDDC: vkGetInstanceProcAddr (trampoline.c:64)
==69840==    by 0x46DD84: VkLayerTest_ExceedMemoryAllocationCount_Test::TestBody() (layer_validation_tests.cpp:3048)
==69840==    by 0x50AC292: HandleSehExceptionsInMethodIfSupported<testing::Test, void> (gtest.cc:2078)
==69840==    by 0x50AC292: void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (gtest.cc:2114)
==69840==    by 0x50A33F6: testing::Test::Run() (gtest.cc:2151)
==69840==    by 0x50A349D: testing::TestInfo::Run() (gtest.cc:2326)
==69840==    by 0x50A35A4: testing::TestCase::Run() (gtest.cc:2444)
==69840==    by 0x50A3857: testing::internal::UnitTestImpl::RunAllTests() (gtest.cc:4311)
==69840==    by 0x50A3B13: HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (gtest.cc:2078)
==69840==    by 0x50A3B13: HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (gtest.cc:2114)
==69840==    by 0x50A3B13: testing::UnitTest::Run() (gtest.cc:3925)
==69840==    by 0x441813: RUN_ALL_TESTS (gtest.h:2297)
==69840==    by 0x441813: main (layer_validation_tests.cpp:24849)
==69840==  Address 0xa06cec8 is 664 bytes inside an unallocated block of size 643,472 in arena "client"
==69840== 
==69840== Invalid read of size 8
==69840==    at 0x4E4A128: loader_phys_dev_ext_gpa (loader.c:3924)
[...]

PR #1862, opened 2017-06-12
loader: Fix unknown physical device extension handling

lenny-lunarg:
The physical device extension handling is currently broken on Windows 32-bit builds because the current implementation relies on compiler optimizations (tail-call optimization, to be specific) to optimize out the stack frame creation, so that the arguments get passed from the caller to the implementation, without the loader touching them.

This pull request uses assembly code to avoid touching the parameters that get passed on to the next function in the call chain. I was not able to find a way to use C, since C gives no way to prevent a stack frame from being created (even with inline assembly I couldn't find a way to do this, since I couldn't define a global symbol in visual studio from inline assembly). If the project is built for a platform where cmake can't find an appropriate assembler, then the loader will fall back on the C code (which may or may not work), while outputting a warning when CMake is run. The project uses MASM on windows (included in Visual Studio) and GAS on linux (included with GCC), so additional dependencies shouldn't be necessary. It is possibly that clang might not use the assembly code, but I haven't tested that.

This code only fixes the trampoline code for physical device extensions, because this is needed for a test that @Tony-LunarG has written. Similar fixes are likely needed for the terminator code, as well as unknown logical device extensions, but that is beyond the scope of this pull request.


lenny-lunarg:
Also, we had initially thought that this fix might solve issue #1846, but we're still seeing that problem, so it looks like there's more bad things going on there. This pull request doesn't change #1846 for the better or worse.

PR #1866, opened 2017-06-12
layers:Remove includes of vk_safe_struct.cpp

tobine:
After looking at these includes of vk_safe_struct.cpp I believe it was
just a hack that worked and has persisted in the code. This change
pulls the source into the layers that depend on it at build time and is
a better solution, in my opinion.

The one quirk is that the we now generate vk_struct.cpp in the base dir
and in the layers dir. Since the file is required by some layer shared
libs I had to generate in the layers dir, but there are also some
dependencies on the file in the top-level dir that I didn't care to
investigate further so I just genrate it in two places.
I'm sure this could be cleaned up if someone with more cmake knowledge
and/or patience than me cares to look into it further.


build-android/jni/Android.mk
```diff
@@ -38,6 +38,7 @@ LOCAL_SRC_FILES += $(SRC_DIR)/layers/descriptor_sets.cpp
 LOCAL_SRC_FILES += $(SRC_DIR)/layers/buffer_validation.cpp
 LOCAL_SRC_FILES += $(SRC_DIR)/layers/shader_validation.cpp
 LOCAL_SRC_FILES += $(SRC_DIR)/layers/vk_layer_table.cpp
+LOCAL_SRC_FILES += $(LAYER_DIR)/include/vk_safe_struct.cpp
```
chrisforbes: Don't much like .cpp files in include/

tobine:
I updated android codegen to put vk_safe_struct.cpp into "common" dir instead of "include" and use it from there in layer libs.

chrisforbes:
@stroyan or we can move this down a level into a static lib, and have the layers depend on that?

mark-lunarg:
@stroyan, any reply to @chrisforbes?

mark-lunarg:
@tobine, @stroyan, any update here, or can we just get by with things the way they are now?

stroyan:
It would be nice to get back to using a proper add_library rule for using vk_safe_struct.cpp.
Looking at the current https://cmake.org/Wiki/CMake_FAQ description of this problem I see
that there is an alternative workaround of making one of the layers depend on the other.
Then the two layers that use vk_safe_struct.cpp will not be built in parallel.
Then visual studio will never clobber vk_safe_struct.cpp.
That could be accomplished by adding one more rule to layers/CMakeLists.txt in the section
starting with- # Core validation has additional dependencies
Add-

# Dependency on unique_objects is only to prevent parallel build of both layers.
# Visual Studio has trouble with clobbering vk_safe_struct.cpp if both build in parallel.
add_dependencies(VkLayer_core_validation VkLayer_unique_objects)

I don't have a system available right now to confirm if visual studio does the right thing with that additional rule.

tobine:
Updated with WA suggested by @stroyan and re-testing.

mark-lunarg:
@tobine,  this repository will close for **write access** on **Sunday, 5/13/2018**.  If it is pushed before that time it will be present in the follow-on Vulkan-ValidationLayers repository on Monday, otherwise a new PR will be required in the new repo.

PR #1868, opened 2017-06-13
Make DescriptorSetLayout data persistent for DescriptorSets

mark-lunarg:
DescriptorSetLayout data can be destroyed before any associated DescriptorSets.  Changed pointers-to-DSLayout-data to copies, at least for the time-being.

Also added a positive test for this case.


mark-lunarg:
@chrisforbes, I created #1869 for changing this over to get rid of the copies.

PR #1870, opened 2017-06-13
Fix SC confusion on per-patch interface blocks

chrisforbes:
A couple of things were broken here:
- Patch decoration appears at member level, and was ignored.
- If the block itself has a Location decoration, we assumed it could not be a block

### DIFF
--- a/layers/shader_validation.cpp
+++ b/layers/shader_validation.cpp
@@ -455,18 +455,19 @@ static spirv_inst_iter get_struct_type(shader_module const *src, spirv_inst_iter
     }
 }
 
-static void collect_interface_block_members(shader_module const *src, std::map<location_t, interface_var> *out,
+static bool collect_interface_block_members(shader_module const *src, std::map<location_t, interface_var> *out,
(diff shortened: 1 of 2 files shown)


chrisforbes:
Fixed commit message

layers/shader_validation.cpp
```diff
@@ -587,7 +597,8 @@ static std::map<location_t, interface_var> collect_interface_by_location(shader_
             // this path for the interface block case, as the individual members of the type are decorated, rather than
             // variable declarations.
 
-            if (location != -1) {
+            if (builtin != -1) continue;
+            else if (!collect_interface_block_members(src, &out, blocks, is_array_of_verts, id, type, is_patch, location)) {
                 // A user-defined interface variable, with a location. Where a variable occupied multiple locations, emit
```
tobine: Is this comment still valid?

chrisforbes: It is. `collect_interface_block_members` bails out in the not-a-block case

PR #1876, opened 2017-06-15
LayerValidationTest Extension enabled vs supported

### LINKED ISSUE #1827: Tests: fix general confusion about supported vs enabled extensions
Various tests check for the physical device supporting an extension to gate some piece of logic, but don't bother to actually enable it.

mark-lunarg:
Fixes #1827.  For any test that looked to see if an extension was supported before using it, added enabling of that extension.  One test required separation to handle enabled/disabled test cases.  Also removed dead code from LVTs.


tests/layer_validation_tests.cpp
```diff
@@ -1017,45 +1017,29 @@ TEST_F(VkLayerTest, UnrecognizedValueBadFlag) {
...
+    // Make sure using VK_SAMPLER_ADDRESS_MODE_MIRROR_CLAMP_TO_EDGE doesn't trigger a false positive.
+    ASSERT_NO_FATAL_FAILURE(InitFramework(myDbgFunc, m_errorMonitor));
+    if (DeviceExtensionSupported(gpu(), VK_KHR_SAMPLER_MIRROR_CLAMP_TO_EDGE_EXTENSION_NAME)) {
+        m_device_extension_names.push_back(VK_KHR_SAMPLER_MIRROR_CLAMP_TO_EDGE_EXTENSION_NAME);
     }
-    VkSamplerAddressMode address_mode = VK_SAMPLER_ADDRESS_MODE_MIRROR_CLAMP_TO_EDGE;
-    if (!supports_mirror_clamp) {
-        printf("             mirror_clamp_to_edge Extension not supported, skipping tests\n");
-        address_mode = VK_SAMPLER_ADDRESS_MODE_CLAMP_TO_EDGE;
+    else {
```
mikes-lunarg: Did you run clang-format?

tests/layer_validation_tests.cpp
```diff
@@ -1017,45 +1017,29 @@ TEST_F(VkLayerTest, UnrecognizedValueBadFlag) {
...
+        printf("             VK_NV_DEDICATED_ALLOCATION_EXTENSION_NAME Extension not supported, skipping test\n");
```
mikes-lunarg: Error message should reference VK_KHR_sampler_mirror_clamp_to_edge

Tony-LunarG: Is this the right extension name?

tests/layer_validation_tests.cpp
```diff
@@ -24279,124 +23610,104 @@ TEST_F(VkPositiveLayerTest, Maintenance1Tests) {
 TEST_F(VkLayerTest, DuplicateValidPNextStructures) {
     TEST_DESCRIPTION("Create a pNext chain containing valid strutures, but with a duplicate structure type");
 
-    ASSERT_NO_FATAL_FAILURE(Init());
+    ASSERT_NO_FATAL_FAILURE(InitFramework(myDbgFunc, m_errorMonitor));
+    if (DeviceExtensionSupported(gpu(), VK_NV_DEDICATED_ALLOCATION_EXTENSION_NAME)) {
+        m_device_extension_names.push_back(VK_NV_DEDICATED_ALLOCATION_EXTENSION_NAME);
```
Tony-LunarG: Is there a non-vendor specific extension we could use here (and below)?

mark-lunarg: Not going to tackle this just now, but it's a great idea.

PR #1877, opened 2017-06-16
layers: Fix layer table leaks

jozefkucia:
Fixes memory leaks produced by VK_LAYER_LUNARG_api_dump.


mark-lunarg:
Pushed to Master.   Thanks!

PR #1879, opened 2017-06-16
build: Do not overwrite CMAKE_ASM-ATT_FLAGS

jozefkucia:
Allows passing flags to assembler in order to build 32-bit targets
on 64-bit platforms.

A 32-bit build can be produced by passing -DCMAKE_ASM-ATT_FLAGS=--32
to cmake.

32-bit builds are broken since commit 748a4c9c986b90006202c345568f292f77bf6a63


jozefkucia:
32-bit builds are broken since commit 748a4c9c986b90006202c345568f292f77bf6a63

PR #1881, opened 2017-06-16
Enable control flow guard compiler & linker option on Windows binaries

scyganINTEL:
Enable control flow guard compiler & linker option on Windows binaries

    This enables the 'Control Flow Guard' security feature, which prevents memory
    corruption vulnerabilites by placing a restriction on where application can
    execute from.

    The change is applied to both VulkanRT and it's NSIS installer.

    WARN:  This change requires re-patching and rebuillding NSIS with attached patch.
    WARN#2: This change requires Visual Studion 2015 or vs140 toolset.

    Control Flow Guard documentation:
    https://msdn.microsoft.com/en-us/library/windows/desktop/mt637065(v=vs.85).aspx


lenny-lunarg:
This one shouldn't get merged for the time being, because it will require that we update or internal SDK build system. Once that's done then, it should be fine.

lenny-lunarg:
This PR was merged in commit 85af915e7af2179748fcaa77cc6d8a5b92585085. I modified the commit slightly before I pushed it to master to do two things:
- The commit message was tweaked to look more like our usual format
- I removed the `/guard:cf` flags when building with older visual studio version, so we can still support VS2013 without a bunch of unrecognized flag warnings

scyganINTEL:
Unfortunately, there is a CMake bug (regression) that may disable CFG even with above change.

**CMake 3.6.x should be used to get CFG-enabled binaries.**  More recent versions (3.7, 3.8) are currently broken. 

The bugfix is tracked in https://gitlab.kitware.com/cmake/cmake/merge_requests/1018  and it should land into 3.9.x.

PR #1882, opened 2017-06-16
Require Python 2.7 instead of Python 3

cdwfs:
The other day, I had some CMake issues while building the LVL project on a new Windows workstation; it complained that Python 3 was required, and I only have 2.7 installed. I've already had to jump through some hoops in the past in this area, as while the layers need Python 3 to build, running "git clang-format" on my patches invokes a script that's only compatible with Python 2.x. I usually just install both and dance between them as necessary, but today I decided to figure out if the layers *really* need Python 3.  And as far as I can tell, they don't; only a few scripts in the "smoketest" demo complained at build time, and they seemed relatively easy to fix.

Please take a look and confirm that I've fixed the two specific Python 2.x issues in a way that won't cause problems with Python 3, or that I'm not missing any other unexpected consequences of using an older version. Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ set(MAJOR "1")
 string(TOLOWER ${API_NAME} API_LOWERCASE)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
-find_package(PythonInterp 3 REQUIRED)
+find_package(PythonInterp 2.7 REQUIRED)
 
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
     set(FALLBACK_CONFIG_DIRS "/etc/xdg" CACHE STRING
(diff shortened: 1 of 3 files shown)


cdwfs:
That AppVeyor failure looks relevant; I'll investigate once I regain access to the build farm.

krOoze:
@cdwfs You should also update the `BUILD.md` documentation, which states Python3 is needed.

cdwfs:
> Can you use this instead please
> #!/usr/bin/env python2

Forgive my ignorance, but would that mean that Python-3-only systems would be actively unsupported? Or does Python 3 install a "python2" alias as well, to provide backwards-compatible behavior? I can totally understand the danger of not specifying a specific target version, but on the other hand my goal with this PR was ideally to _expand_ the range of supported Python versions, not to replace one hard requirement with a mutually exclusive one.

cdwfs:
After further consideration I'm not sure there's any need for this patch; my main motivation was the need to keep Python 2.7 around for git-clang-format, and that's apparently been [fixed](https://github.com/llvm-mirror/clang/commit/7150b8bbaf060023fb2725f21478a95ecfa4fc28) for the next Clang release. Aside from that, I'm fine with requiring Python 3 to build/test the layers.

mark-lunarg:
👍 💯

PR #1886, opened 2017-06-16
tests: Skip test for 14a004dc if all offsets are valid

### LINKED ISSUE #1781: VkLayerTest.VALIDATION_ERROR_01412 fails on AMD

cdwfs:
Also fixes some unexpected VU errors in this test that were
exposed when the main test accidentally passed. Fixes #1781.

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -8979,7 +8979,31 @@ TEST_F(VkLayerTest, VALIDATION_ERROR_14a004dc) {
         "Test VALIDATION_ERROR_14a004dc: offset must be less than or equal to "
         "VkPhysicalDeviceLimits::maxVertexInputAttributeOffset");
 
-    ASSERT_NO_FATAL_FAILURE(Init());
+    if (InstanceLayerSupported("VK_LAYER_LUNARG_device_profile_api")) {
+        m_instance_layer_names.push_back("VK_LAYER_LUNARG_device_profile_api");
+    }
+
+    ASSERT_NO_FATAL_FAILURE(InitFramework(myDbgFunc, m_errorMonitor));
+
+    uint32_t maxVertexInputAttributeOffset = 0;
+    {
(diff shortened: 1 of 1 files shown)


Tony-LunarG:
This looks like a good candidate for the device_profile_api layer (see ExceedMemoryAllocationCount for a lone example).  This looks good for now, but we may switch it over to use the layer at some point.

cdwfs:
I fixed a few additional VU errors, and took a shot at using LUNARG_device_profile_api to artificially lower the offset limit. Seems to work on my end, but please let me know if I'm using the layer incorrectly.

tests/layer_validation_tests.cpp
```diff
@@ -8979,7 +8979,28 @@ TEST_F(VkLayerTest, VALIDATION_ERROR_14a004dc) {
...
+        // Attempt to artificially lower maximum offset
+        PFN_vkSetPhysicalDeviceLimitsEXT fpvkSetPhysicalDeviceLimitsEXT =
+            (PFN_vkSetPhysicalDeviceLimitsEXT)vkGetInstanceProcAddr(instance(), "vkSetPhysicalDeviceLimitsEXT");
+        if (!fpvkSetPhysicalDeviceLimitsEXT) {
+            printf("             All offsets are valid & device_profile_api not found; skipped.\n");
+            return;
+        }
+        default_props.limits.maxVertexInputAttributeOffset = default_props.limits.maxVertexInputBindingStride - 2;
+        fpvkSetPhysicalDeviceLimitsEXT(gpu(), &default_props.limits);
+        maxVertexInputAttributeOffset = default_props.limits.maxVertexInputAttributeOffset;
```
chrisforbes: Not a fan of this being called 'default_props' when you mangle it.

cdwfs: Renamed, & moved it out of scope when it's no longer in use, to prevent confusion with the final Device's props.

PR #1891, opened 2017-06-19
Another round of housekeeping stuff

chrisforbes:
Big change here is adjusting the CB state machine to no longer rely on last_cmd to distinguish two cases of `CB_INVALID`.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -6014,7 +6014,7 @@ VKAPI_ATTR void VKAPI_CALL CmdClearColorImage(VkCommandBuffer commandBuffer, VkI
 
     bool skip = PreCallValidateCmdClearColorImage(dev_data, commandBuffer, image, imageLayout, rangeCount, pRanges);
     if (!skip) {
-        PreCallRecordCmdClearImage(dev_data, commandBuffer, image, imageLayout, rangeCount, pRanges, CMD_CLEARCOLORIMAGE);
+        PreCallRecordCmdClearImage(dev_data, commandBuffer, image, imageLayout, rangeCount, pRanges);
         lock.unlock();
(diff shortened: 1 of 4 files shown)


layers/core_validation.cpp
```diff
@@ -5007,7 +5019,7 @@ VKAPI_ATTR VkResult VKAPI_CALL BeginCommandBuffer(VkCommandBuffer commandBuffer,
                             "vkBeginCommandBuffer(): Cannot call Begin on command buffer (0x%p"
                             ") in the RECORDING state. Must first call vkEndCommandBuffer(). %s",
                             commandBuffer, validation_error_map[VALIDATION_ERROR_16e00062]);
-        } else if (CB_RECORDED == cb_node->state || (CB_INVALID == cb_node->state && CMD_END == cb_node->last_cmd)) {
+        } else if (CB_RECORDED == cb_node->state || CB_INVALID_COMPLETE == cb_node->state) {
```
tobine: So the extra state is getting us the cleaner check here? Seems clearer/simpler.

layers/core_validation_types.h
```diff
@@ -802,7 +801,6 @@ bool ValidateCmdQueueFlags(layer_data *dev_data, GLOBAL_CB_NODE *cb_node, const
 bool ValidateCmd(layer_data *my_data, GLOBAL_CB_NODE *pCB, const CMD_TYPE cmd, const char *caller_name);
 bool insideRenderPass(const layer_data *my_data, GLOBAL_CB_NODE *pCB, const char *apiName, UNIQUE_VALIDATION_ERROR_CODE msgCode);
 void SetImageMemoryValid(layer_data *dev_data, IMAGE_STATE *image_state, bool valid);
-void UpdateCmdBufferLastCmd(GLOBAL_CB_NODE *cb_state, const CMD_TYPE cmd);
```
tobine: This CL really demonstrates the goodness of killing last_cmd

PR #1893, opened 2017-06-19
demos: Update cubepp

jeremy-lunarg:
Modify correct uniform buffer for next frame.
Fix fence sync after new throttling scheme.
Select alpha based on capabilities in cube.
Add uniform buffer per swapchain image to cube.
Cube updated to use triple buffering.

Change-Id: Ia929bde0e6c3c2129dc9164a65696af1ee0c0329


demos/cube.cpp
```diff
@@ -438,24 +441,23 @@ struct Demo {
...
+                // demo->swapchain is out of date (e.g. the window was resized) and
+                // must be recreated:
+                resize();
+            } else if (result == vk::Result::eSuboptimalKHR) {
+                // swapchain is not as optimal as it could be, but the platform's
+                // presentation engine will still present the image correctly.
+            } else {
+                VERIFY(result == vk::Result::eSuccess);
+            }
+        } while (result != vk::Result::eSuccess);
```
Tony-LunarG: I like this better than cube's loop and will probably steal it

PR #1895, opened 2017-06-20
Tobin gh1869 ds shared ptr

### LINKED ISSUE #1869: DescriptorSets: Use shared_ptr for DescriptorSetLayout data instead of copies
DescriptorSets need access to DSLayout state, which was handled via pointers (problematic if DSLayouts were destroyed) but is now handled by making copies of the DSLayout data for each descriptorSet.  This could be better managed by using shared_ptrs.

tobine:
Fixes #1869 

Migrate DescriptorSetLayout references to use shared_ptr.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -709,12 +709,13 @@ FRAMEBUFFER_STATE *GetFramebufferState(const layer_data *dev_data, VkFramebuffer
     return it->second.get();
 }
 
-cvdescriptorset::DescriptorSetLayout const *GetDescriptorSetLayout(layer_data const *dev_data, VkDescriptorSetLayout dsLayout) {
+std::shared_ptr<cvdescriptorset::DescriptorSetLayout> const GetDescriptorSetLayout(layer_data const *dev_data,
+                                                                                   VkDescriptorSetLayout dsLayout) {
     auto it = dev_data->descriptorSetLayoutMap.find(dsLayout);
     if (it == dev_data->descriptorSetLayoutMap.end()) {
         return nullptr;
     }
-    return it->second.get();
+    return it->second;
 }
 
 static PIPELINE_LAYOUT_NODE const *getPipelineLayout(layer_data const *dev_data, VkPipelineLayout pipeLayout) {
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -149,7 +149,7 @@ struct layer_data {
     unordered_map<VkCommandPool, COMMAND_POOL_NODE> commandPoolMap;
     unordered_map<VkDescriptorPool, DESCRIPTOR_POOL_STATE *> descriptorPoolMap;
     unordered_map<VkDescriptorSet, cvdescriptorset::DescriptorSet *> setMap;
-    unordered_map<VkDescriptorSetLayout, std::unique_ptr<cvdescriptorset::DescriptorSetLayout>> descriptorSetLayoutMap;
+    unordered_map<VkDescriptorSetLayout, std::shared_ptr<cvdescriptorset::DescriptorSetLayout>> descriptorSetLayoutMap;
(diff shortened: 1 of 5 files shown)


layers/core_validation.cpp
```diff
@@ -4532,8 +4532,8 @@ static bool PreCallValidateCreateDescriptorSetLayout(layer_data *dev_data, const
 
 static void PostCallRecordCreateDescriptorSetLayout(layer_data *dev_data, const VkDescriptorSetLayoutCreateInfo *create_info,
                                                     VkDescriptorSetLayout set_layout) {
-    dev_data->descriptorSetLayoutMap[set_layout] = std::unique_ptr<cvdescriptorset::DescriptorSetLayout>(
-        new cvdescriptorset::DescriptorSetLayout(create_info, set_layout));
+    dev_data->descriptorSetLayoutMap[set_layout] =
+        std::shared_ptr<cvdescriptorset::DescriptorSetLayout>(new cvdescriptorset::DescriptorSetLayout(create_info, set_layout));
```
chrisforbes: std::make_shared!

layers/descriptor_sets.cpp
```diff
@@ -312,21 +313,21 @@ cvdescriptorset::AllocateDescriptorSetsData::AllocateDescriptorSetsData(uint32_t
     : required_descriptors_by_type{}, layout_nodes(count, nullptr) {}
 
 cvdescriptorset::DescriptorSet::DescriptorSet(const VkDescriptorSet set, const VkDescriptorPool pool,
-                                              const DescriptorSetLayout *layout, const layer_data *dev_data)
+                                              std::shared_ptr<DescriptorSetLayout> layout, const layer_data *dev_data)
```
chrisforbes: Consider passing shared_ptr by ref when you don't need the temporary to be an owner (but you will make a copy of it elsewhere)

layers/descriptor_sets.cpp
```diff
@@ -389,8 +390,8 @@ static std::string string_descriptor_req_view_type(descriptor_req req) {
 }
 
 // Is this sets underlying layout compatible with passed in layout according to "Pipeline Layout Compatibility" in spec?
-bool cvdescriptorset::DescriptorSet::IsCompatible(const DescriptorSetLayout *layout, std::string *error) const {
-    return layout->IsCompatible(&p_layout_, error);
+bool cvdescriptorset::DescriptorSet::IsCompatible(std::shared_ptr<DescriptorSetLayout> const layout, std::string *error) const {
```
chrisforbes: This should be a non-owning ptr.

layers/descriptor_sets.h
```diff
@@ -110,7 +110,7 @@ class DescriptorSetLayout {
     bool HasBinding(const uint32_t binding) const { return binding_to_index_map_.count(binding) > 0; };
     // Return true if this layout is compatible with passed in layout,
     //   else return false and update error_msg with description of incompatibility
-    bool IsCompatible(const DescriptorSetLayout *, std::string *) const;
+    bool IsCompatible(std::shared_ptr<DescriptorSetLayout> const, std::string *) const;
```
chrisforbes: Non-owning!

layers/descriptor_sets.h
```diff
@@ -370,7 +363,7 @@ class DescriptorSet : public BASE_NODE {
     // Perform a CopyUpdate whose contents were just validated using ValidateCopyUpdate
     void PerformCopyUpdate(const VkCopyDescriptorSet *, const DescriptorSet *);
 
-    const DescriptorSetLayout *GetLayout() const { return &p_layout_; };
+    std::shared_ptr<DescriptorSetLayout> const GetLayout() const { return p_layout_; };
```
chrisforbes: This was previously a ptr to a const DescriptorSetLayout. Now it's a const shared_ptr to a (nonconst) DescriptorSetLayout. Are you sure that's what you want?

Same in a few other places.

tobine:
@chrisforbes yep, thanks, I'll clean these up

PR #1896, opened 2017-06-20
GH1889, check for CLAMP_TO_EDGE ext enable

### LINKED ISSUE #1889: No warning if using VK_SAMPLER_ADDRESS_MODE_MIRROR_CLAMP_TO_EDGE without enabling VK_KHR_sampler_mirror_clamp_to_edge
Present in 1.0.51 with VK_LAYER_LUNARG_standard_validation.

mark-lunarg:
Using this addressing mode in CreateSampler should generate an error if the extension is not enabled.
Also added layer validation test for this check and updated the error database.

Fixes #1889.


tests/layer_validation_tests.cpp
```diff
@@ -1053,6 +1053,40 @@ TEST_F(VkLayerTest, UnrecognizedValueBadBool) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(VkLayerTest, MirrorClampToEdgeNotEnabled) {
+    TEST_DESCRIPTION("Validation should catch using CLAMP_TO_EDGE addressing mode if the extension is not enabled.");
+
+    ASSERT_NO_FATAL_FAILURE(InitFramework(myDbgFunc, m_errorMonitor));
+    if (!DeviceExtensionSupported(gpu(), VK_KHR_SAMPLER_MIRROR_CLAMP_TO_EDGE_EXTENSION_NAME)) {
```
mikes-lunarg: Is this check really necessary if we aren't going to enable the extension anyway?

layers/parameter_validation.cpp
```diff
@@ -2471,6 +2471,20 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateSampler(VkDevice device, const VkSamplerCre
             skip |= validate_ranged_enum(report_data, "vkCreateSampler", "pCreateInfo->borderColor", "VkBorderColor",
                                          VK_BORDER_COLOR_BEGIN_RANGE, VK_BORDER_COLOR_END_RANGE, pCreateInfo->borderColor);
         }
+
+        // If any of addressModeU, addressModeV or addressModeW are VK_SAMPLER_ADDRESS_MODE_MIRROR_CLAMP_TO_EDGE, the
+        // VK_KHR_sampler_mirror_clamp_to_edge extension must be enabled
+        if (((pCreateInfo->addressModeU == VK_SAMPLER_ADDRESS_MODE_MIRROR_CLAMP_TO_EDGE) ||
+             (pCreateInfo->addressModeV == VK_SAMPLER_ADDRESS_MODE_MIRROR_CLAMP_TO_EDGE) ||
+             (pCreateInfo->addressModeW == VK_SAMPLER_ADDRESS_MODE_MIRROR_CLAMP_TO_EDGE)) &&
+            (!device_data->extensions.vk_khr_sampler_mirror_clamp_to_edge)) {
```
tobine: It's a bit of a philosophical exercise, but I'm thinking I favor having the extension check first in the if block. That way it will early-out when the extension is supported, which I'm guessing is the common case. Otherwise all three mode checks are always performed.

mark-lunarg:
Updated and pushed to Master.

PR #1903, opened 2017-06-21
build: Exit with return code in update_external bat

Tony-LunarG:
Update_external_sources could fail and the scripts that call it would continue using whatever external sources were already built.  This change makes the failure more obvious.

### DIFF
--- a/update_external_sources.bat
+++ b/update_external_sources.bat
@@ -217,13 +217,12 @@ goto:finish
 :error
 echo.
 echo Halting due to error
+set errorCode=1
 goto:finish
 
 :finish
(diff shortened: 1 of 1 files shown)


update_external_sources.bat
```diff
@@ -217,13 +217,12 @@ goto:finish
...
 echo Halting due to error
+set errorCode=1
 goto:finish
 
 :finish
 if not "%cd%\" == "%BUILD_DIR%" ( cd %BUILD_DIR% )
-endlocal
-goto:eof
-
+exit /b %errorCode%
```
MarkY-LunarG: I thought we did this before and it caused problems when building from the command line.

PR #1905, opened 2017-06-21
Move layer scripts into scripts directory

mark-lunarg:
Simplifies sharing code among python modules, keeps everything together.

- Moved python scripts out of layers dir into scripts dir
- Updated relative paths for input/output files in layers dir
- Updated doc validation tests to use correct paths


scripts/spec.py
```diff
@@ -25,8 +25,8 @@
 #############################
 
 
-out_filename = "vk_validation_error_messages.h" # can override w/ '-out <filename>' option
-db_filename = "vk_validation_error_database.txt" # can override w/ '-gendb <filename>' option
+out_filename = "..//layers//vk_validation_error_messages.h" # can override w/ '-out <filename>' option
```
mikes-lunarg: Why the double slashes?

mark-lunarg: Leftover from some experimentation, I'll change it back.  Thanks!

mark-lunarg:
Updated, rebased and pushed to Master.

PR #1913, opened 2017-06-23
Set up Infrastructure for and wire in first set of VUIDS for generated validation layers

mark-lunarg:
- Modify parameter_validation codegen to compute VUID tags
- Use slightly modified vuid_mapping,py to convert tag to a VUID number
- Validate that vuid number against the existing set of good tags in `vk_validation_error_messages.h`
- Add VUID arg to the sType validation routine and calls
- Update error database

Since we do the check against the good messages, this works even though the spec may be incomplete -- when it does get updated the source will pick up the new VUIDS automatically.

This picks up around 70 VUs.


scripts/vuid_mapping.py
```diff
@@ -470,6 +470,16 @@
 'vkUpdateDescriptorSetWithTemplateKHR' : 413,
 'vkUpdateDescriptorSets' : 414,
 'vkWaitForFences' : 415,
+'VkPhysicalDeviceProperties2KHR' : 416,
```
tobine: :+1:

scripts/vuid_mapping.py
```diff
@@ -6,7 +6,6 @@
     import urllib.request as urllib2
 except ImportError:
     import urllib2
-from bs4 import BeautifulSoup
```
tobine: :+1:

scripts/vuid_mapping.py
```diff
@@ -926,7 +926,7 @@ def convertVUID(vuid_string):
 #        print ("ERROR: Uniqueid %d for string id %s is a duplicate!" % (uniqueid, vuid_string))
 #        print (" TODO: Figure out what caused the dupe and fix it")
         #sys.exit()
-    print ("Storing uniqueid %d for unique string %s" % (uniqueid, vuid_string))
+    # print ("Storing uniqueid %d for unique string %s" % (uniqueid, vuid_string))
```
tobine: :+1:

scripts/vk_validation_stats.py
```diff
@@ -164,8 +172,26 @@ def read(self):
...
+        if len(self.generated_source_files) > 0:
+            qualified_paths = []
+            for source in self.generated_source_files:
+                for build_dir in self.generated_source_dirs:
+                    filepath = '../%s/layers/%s' % (build_dir, source)
+                    if os.path.isfile(filepath):
+                        qualified_paths.append(filepath)
+                        continue
+            if len(self.generated_source_files) != len(qualified_paths):
+                print("Error: Unable to locate one or more of the following source files in the build, dbuild, or release directories")
```
tobine: Should do a join on generated_source_directories rather than include the static text here in case search dirs change in the future.

scripts/parameter_validation_generator.py
```diff
@@ -159,14 +161,29 @@ def __init__(self,
...
         self.StructType = namedtuple('StructType', ['name', 'value'])
         self.CommandParam = namedtuple('CommandParam', ['type', 'name', 'ispointer', 'isstaticarray', 'isbool', 'israngedenum',
                                                         'isconst', 'isoptional', 'iscount', 'noautovalidity', 'len', 'extstructs',
                                                         'condition', 'cdecl'])
         self.CommandData = namedtuple('CommandData', ['name', 'params', 'cdecl', 'extension_type'])
         self.StructMemberData = namedtuple('StructMemberData', ['name', 'members'])
+
+        self.vuid_file = None
+        # Cover cases where file is built from scripts directory, Lin/Win, or Android build structure
+        vuid_filename_locations = [
```
tobine: Does this handle out-of-source build?

mark-lunarg:
Bump.

scripts/parameter_validation_generator.py
mark-lunarg: I had hoped so.  If you have a suggestion to make this more location proof please share -- I don't like it either.

scripts/vk_validation_stats.py
mark-lunarg: @tobine, can you provide a bit more detail?

tobine: print("Error: Unable to locate one or more of the following source files in the %s directories" % (", ".join(generated_layer_source_directories)))

scripts/parameter_validation_generator.py
mark-lunarg: For now, the script will fail gracefully if the file is not found.

mark-lunarg:
Updated and pushed to Master.

Issue #1880, opened 2017-06-16
Add a way to name objects for validation messages


Jasper-Bekkers:
Hi,

Right now often you'll get validation errors like this `Cannot submit cmd buffer using image (0x9a) [sub-resource: aspectMask 0x1 array layer 0, mip level 0], with layout ...`. In our engine we have a policy of naming all objects we create for debugging purposes so it would be extremely useful to pass these along to the validation layers for easier debugging so we can get a message like `Cannot submit cmd buffer using image (LinearRenderTarget1) [sub-resource: aspectMask 0x1 array layer 0, mip level 0], with layout ...` instead.

Tony-LunarG:
Addressed by  1f53907

Jasper-Bekkers:
Does this also fix up all validation warnings to print the named object
instead of the id?

Tony-LunarG:
It does to an extent.  Anything logged through debug_report_log_message will prepend a string "SrcObject name = <name of the primary object>".  Other objects referenced in the message will still only have their handles printed.

Jasper-Bekkers:
Cool, that's already a pretty nice improvement over what was done before. Though it was something that you could easily do app side as well by just keeping the name somewhere and matching it against the object passed back to you in the debug callback. 

However the other objects referenced in the message are also very important (case in point: the message that started this big report), so I would like to include the other referenced objects within the scope of this issue as well.

mark-lunarg:
@Jasper-Bekkers, this family of capabilities is meant to be addressed through a new debug report extension which is currently under development.  It will allow the output of multiple objects through the main interface (among many other things).  This change is what we can reasonably do in the short-term.  Thanks for prodding us to get this in!

Jasper-Bekkers:
@mark-lunarg what's the status of that new extension?

mark-lunarg:
Still winding it's way through Khronos towards ratification.  Hey @Jasper-Bekkers, if you compile a summary of the messages you'd like to see reworked, we can probably get that done sooner rather than later.  Even after the new extension becomes available work will still need to be done to expand the message output and have neither the bandwidth nor inclination to reformat every validation layer error message.  Point out the ones which'll help you and we'll work to get this issue closed.

Jasper-Bekkers:
[...]

PR #1915, opened 2017-06-26
layers: Add names to objects and print in validation

Tony-LunarG:
Addresses GH issue #1880 and covers core validation, object tracker and parameter validation.  Also includes a test to verify that the name given to an object is referenced in an error message.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -10131,6 +10131,41 @@ VKAPI_ATTR VkResult VKAPI_CALL GetDisplayPlaneCapabilitiesKHR(VkPhysicalDevice p
     return result;
 }
 
+VKAPI_ATTR VkResult VKAPI_CALL DebugMarkerSetObjectNameEXT(VkDevice device, VkDebugMarkerObjectNameInfoEXT *pNameInfo) {
+    std::unique_lock<std::mutex> lock(global_lock);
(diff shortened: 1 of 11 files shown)


layers/core_validation.cpp
```diff
@@ -10131,6 +10131,41 @@ VKAPI_ATTR VkResult VKAPI_CALL GetDisplayPlaneCapabilitiesKHR(VkPhysicalDevice p
...
+            std::make_pair<uint64_t, std::string>((uint64_t &&)pNameInfo->object, pNameInfo->pObjectName));
+    } else {
+        device_data->report_data->debugObjectNameMap->erase(pNameInfo->object);
+    }
+    lock.unlock();
+    VkResult result = device_data->dispatch_table.DebugMarkerSetObjectNameEXT(device, pNameInfo);
+    return result;
+}
+
+VKAPI_ATTR VkResult VKAPI_CALL DebugMarkerSetObjectTagEXT(VkDevice device, VkDebugMarkerObjectTagInfoEXT *pTagInfo) {
```
mark-lunarg: This looks good -- I surprised we took so long to add support for this feature.  Do we need to hook these four other functions?  Nothing is done with them but I suppose we could do something later.

Tony-LunarG: I hooked them to "implement" the complete extension.  It seemed like if the layer was going to advertise the extension, it should supply all of the entrypoints

PR #1923, opened 2017-06-29
Github 1544: build: Initial fixes for MinGW-w64 on Windows.

DragoonX6:
These are initial fixes for MSYS2 MinGW-w64 on Windows.
It should also build with GCC without problems, although I haven't tested that yet.

A lot of the tests are failing and some of the layer tests even segfault.

This is the output of the failing tests, but to debug it you would have to build them yourself, unless you don't mind me providing binaries.
https://gist.github.com/DragoonX6/d9a846a5a995df698bf787c0b82c8053

I've tried taking a look at the segfaulting test in a debugger, but I couldn't work out why this crashed.
The samples do actually work though.

Please review the code so we can pin down what's actually causing the tests to fail.


DragoonX6:
Updated with some more fixes to get it building again, and rebased to master. Please review.

karl-lunarg:
The Travis-CI builds are still failing.  (I cleared the cache and restarted it to get around an apparent caching problem.)  It looks like the Linux builds are trying to use `xcopy`.  Please fix that.

Also, your 6f7c765 commit seems to introduce a lot of unneeded whitespace changes.  You might want to drop that commit and make another one without any unintended whitespace changes.

DragoonX6:
Looks like those whitespace changes slipped through unintendedly.

I'll get to it once I'm back from work.

DragoonX6:
@karl-lunarg Fixed the white space issue and fixed the build.

Appveyor build is still broken because of a bug in MSVC, which is an easy fix, but out of scope for this PR.

DragoonX6:
@karl-lunarg ping.

MarkY-LunarG:
@karl-lunarg  is on travel.  He'll get to this as soon as he can.

DragoonX6:
Alright, do you have any idea when he's going to be back? I'll try to rebase with master often so I don't diverge too much.

tests/CMakeLists.txt
```diff
@@ -126,12 +128,18 @@ add_dependencies(vk_layer_validation_tests
    VkLayer_parameter_validation
 )
 
-if (WIN32)
+if (WIN32 AND MSVC)
```
MarkY-LunarG: Nit: I'd rather see:

if (WIN32)
   if (MSVC)
   else()
   endif()
endif()

This grouping seems a little messy and the if formats I suggest would be a little cleaner.

loader/asm_offset.c
```diff
@@ -23,8 +23,10 @@
 #include <stdio.h>
 #include "loader.h"
 
-#if !defined(_MSC_VER) || (_MSC_VER >= 1900)
+#if (!defined(_MSC_VER) || (_MSC_VER >= 1900)) && !defined(__MINGW64__)
 #define SIZE_T_FMT "%-8zu"
+#elif defined(__MINGW64__)
```
MarkY-LunarG: Might be cleaner to move this first, like:
```
#ifdef __MINGW64__
#define SIZE_T_FMT "%-8llu"
#elif !defined(_MSC_VER) || (_MSC_VER >= 1900)
#define SIZE_T_FMT "%-8zu"
```

loader/CMakeLists.txt
```diff
@@ -134,19 +134,29 @@ add_custom_target(loader_gen_files DEPENDS
...
-    endforeach()
+    endif()
 
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_CRT_SECURE_NO_WARNINGS")
     # Build dev_ext_trampoline.c with -O2 to allow tail-call optimization.
     # Build other C files with normal options
     # setup two Cmake targets (loader-norm and loader-opt) for the different compilation flags
     separate_arguments(LOCAL_C_FLAGS_DBG WINDOWS_COMMAND ${CMAKE_C_FLAGS_DEBUG})
-    set(CMAKE_C_FLAGS_DEBUG " ")
+    #set(CMAKE_C_FLAGS_DEBUG " ")
```
MarkY-LunarG: Why'd you remove this?

loader/CMakeLists.txt
```diff
@@ -134,19 +134,29 @@ add_custom_target(loader_gen_files DEPENDS
...
+        foreach(configuration in CMAKE_C_FLAGS_DEBUG CMAKE_C_FLAGS_MINSIZEREL CMAKE_C_FLAGS_RELEASE CMAKE_C_FLAGS_RELWITHDEBINFO
+                                 CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELEASE CMAKE_CXX_FLAGS_RELWITHDEBINFO)
+            if(${configuration} MATCHES "/MD")
+                string(REGEX REPLACE "/MD" "/MT" ${configuration} "${${configuration}}")
+            endif()
+        endforeach()
+    endif()
+
+    if(MINGW)
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11")
```
MarkY-LunarG: Why'd you make this change here?  The base CMakelists.txt sets this to c99, won't it conflict?

include/vulkan/vk_sdk_platform.h
```diff
@@ -23,6 +23,8 @@
 #define VK_SDK_PLATFORM_H
 
 #if defined(_WIN32)
+// Silence redefinition warnings on MinGW-w64
+#undef NOMINMAX
```
MarkY-LunarG: Why not wrap the #define NOMINMAX in an #ifndef NOMINMAX check so you don't have the redefinition?

MarkY-LunarG:
I believe he'll be back either Thursday or Friday, but I don't know when he'll have a chance to get to this since he is actively working another project.

DragoonX6:
That's okay, the earliest I can probably do after Friday is Monday, so no rush.

karl-lunarg:
The Appveyor problem needs to be fixed as well.

DragoonX6:
It's a bug with visual studio that's technically out of scope for this PR.  
The bug is that VS thinks that `_strdup` is deprecated and that you should use `_strdup`, this warning is meant to fire for `strdup` instead.

I can fix it in this PR, but I think a separate PR would be more appropriate.

Also there's a caching problem with Appveyor, because if you look at the build log it's obviously failing, but because it's restoring the build from cache it appears to be working.

include/vulkan/vk_sdk_platform.h
DragoonX6: That would work too, this was the shorter solution.

loader/asm_offset.c
DragoonX6: Agreed, I'll fix it.

tests/CMakeLists.txt
DragoonX6: Alright, I'll change it.

loader/CMakeLists.txt
DragoonX6: Because it got rid of the flags that give me the necessary debug information I needed to debug asm_offset.c.

loader/CMakeLists.txt
```diff
@@ -156,11 +166,13 @@ if (WIN32)
     add_library(loader-opt OBJECT ${OPT_LOADER_SRCS})
     add_dependencies(loader-opt generate_helper_files loader_gen_files loader_asm_gen_files)
 
-    target_compile_options(loader-opt PUBLIC "$<$<CONFIG:DEBUG>:${LOCAL_C_FLAGS_REL}>")
+    #target_compile_options(loader-opt PUBLIC "$<$<CONFIG:DEBUG>:${LOCAL_C_FLAGS_REL}>")
```
DragoonX6: Because for some reason this causes the C flags to be passed to GNAS. It wouldn't compile with `-O3` on its command line, I commented it out because I don't know how to work around this.

loader/CMakeLists.txt
```diff
@@ -170,10 +182,14 @@ if (WIN32)
         file(TO_NATIVE_PATH ${CMAKE_CURRENT_BINARY_DIR}/$<CONFIGURATION>/${API_LOWERCASE}-${MAJOR}.dll COPY_SRC_PATH)
         file(TO_NATIVE_PATH ${CMAKE_CURRENT_BINARY_DIR}/../demos/$<CONFIGURATION>/ COPY_DST_PATH)
         file(TO_NATIVE_PATH ${CMAKE_CURRENT_BINARY_DIR}/../tests/$<CONFIGURATION>/ COPY_DST_TEST_PATH)
-    else()
+    elseif(MSVC)
```
DragoonX6: This is for the case where the MSVC compiler is used, but msbuild is not used.  
For example the generator is nmake or Ninja.

I would hope the above would actually match on `if(MSBUILD)`, but afaik that's not present in CMake.

loader/CMakeLists.txt
DragoonX6: Oh, I see. I fixed the underlying issue instead.

DragoonX6:
Updated to conform with the review.

DragoonX6:
Would porting the MASM code to YASM/NASM be a viable alternative? It does introduce an extra dependency, but having less code to maintain (in contrast to an extra GAS asm file) sounds like a plus.

lenny-lunarg:
I had though about using other assemblers back when we introduced this (NASM was the one I had in mind), but the main reason we didn't do it was because of the extra dependency. The other reason is that even with another assembler, writing cross platform assembly still isn't particularly easy. And there is a certain simplicity to just separating the code for the two platforms entirely.

So I think the answer is that while we could use NASM/YASM/whatever, we'd rather not switch to that, partly because of the dependency issue, and partly because we know that the current assemblers work fine for our needs. Also, while there's more code with the current setup, it's pretty low maintenance code (unless we try to port to other platforms) so maintenance hasn't been much of an issue.

DragoonX6:
I meant using NASM/YASM for Windows, not for both Windows and Linux. Basically we'd stay with 2 files, but just swap out the MASM assembler.

lenny-lunarg:
Well for Windows it really comes down to dependencies. The vast majority of people who work with our software on Windows are using Visual Studio. Consequently, it makes sense to default to the assembler that ships with Visual Studio. I'm just curious, what's your objection to Visual Studio? Is it because it's closed source or something?

DragoonX6:
My qualms with Visual Studio are mostly with the (un)install process, and the standard compliancy of the compiler. The install size is also somewhat of an issue, but it's not that big of an issue anymore.  
My current code base depends on GNU ld, so it's a bit hard for me to switch until I either write a new CRT, or port an existing CRT that supports my use case, or modify all the external libraries I use.

I also chose for MinGW-w64 because porting my code from Linux to Windows was easier using that compiler, and it also generates slightly faster code. Apart from that, I'm really attached to clang as a compiler, so there's that too.

I get the argument about the dependency, but let's be honest, if I were to port the MASM code to GAS for MinGW-w64 builds (unlikely, since my asm knowledge is next to none), it would be largely unmaintained, since nobody else uses the setup or tests for it. I think porting the code to YASM/NASM would widen the Vulkan userbase, while keeping the burden on you lower.  
I get that the most users use VS, but they also use the SDK instead of going through the whole process of compiling it from source. Having it work for MinGW-w64 would enable packagers and the MSYS2 project to release builds for their distributions, taking away the burden of packaging it for this setup from you guys.

lenny-lunarg:
I don't think that maintaining NASM code would be any easier than maintaining MASM code. It would make MinGW builds easier, but it would make Visual Studio builds harder. Also, while you're right that most Visual Studio users go through the SDK, it's still pretty common for people to get the code from the repo. Driver vendors often want to test with the latest code, and that's only in the repo. Also, if developers find a problem in the SDK, they'll sometimes move to the repo where we can get a fix more quickly. And in an open source project, there's always other random people using the repo for a number of reasons.

I'm also not really concerned about maintaining this assembly code (even if we have another platform), because the code actually does very little. It just unwraps one or two objects and calls down the loader's call chain. On top of that, this code is very unlikely to need to change, since it only depends on a couple of Vulkan conventions (the first argument has to be a dispatchable object). However, I'm busy enough that I don't have time to write this new code (actually, it's the testing that would really eat up time).

If you don't want to write up the assembly code for GAS/Windows, then I see two potential solutions. One is that we just use the C code that may or may not work, This isn't going to cause you problems unless you're writing new extensions. The other option is that there's an assembler called [JWasm](https://sourceforge.net/projects/jwasm/) which claims to be MASM-compatible. I should warn you I know nothing about this assembler, but it seems like you may be able to use that to assemble to the MASM code. Personally, I think simply falling back on the C code seems like the best solution, but you seem like you're not crazy about that idea (though it is currently what we do with non-x86 builds).

DragoonX6:
It's not that I don't want to write the assembly code, it's that I lack the knowledge to.  
I'm not sure how long it would take me to get the required knowledge for me to port it, but I doubt it's a short time frame.

I guess I can give JWasm a try, but I have a feeling it won't be that easy to get it in.

mark-lunarg:
@lenny-lunarg, @DragoonX6, any updates here, or has this MR outlived its usefulness?

lenny-lunarg:
I don't have any updates here. Right now it looks like the PR would have to be brought up to date to merge it. I'd be willing to take a PR if it is brought up to master and if the assembly code is either updated for MinGW, or is removed from MinGW builds and replaced with the existing C code. I'm going to close this since it appears dead, but if @DragoonX6 wants to update this PR I can reopen it when that happens.

DragoonX6:
I sadly don't have the time to work on this in the near future. And if the PR were to be merged, the test failures would have to be addressed too.
Perhaps somebody else is willing to pick this up.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

Issue #1926, opened 2017-07-01
Some comments in C header files are not C-style


grim210:
To be able to compile my Vulkan project with GCC flag -std=c90, all comments in the vulkan header files must conform to using C-style comments.  Comments that
`// look like this`
will not compile.  However, if they
`/* look like this */`
it will compile perfectly.  My use case is probably rare, but compiling with clean C90 makes me feel warm inside.

mark-lunarg:
@grim210, outside of the loader, we don't have a bunch of these files.  Can you enumerate the ones that are giving you trouble?  Our coding standards enforce //-style comments universally, except in copyright notices or file headers and currently make no exceptions for the older C standards.

MarkY-LunarG:
Some of the files aren't owned by us, the come from Khronos.  The others, as @mark-lunarg points out, is based on our coding standard.

Looking at the top-level CMakeLists.txt file, -std=c99 should be automatically set.  Are you not using CMake?

mark-lunarg:
As I mentioned in the PR, I'm closing this issue for now, at least until the main Vulkan working group updates their header files.

PR #1927, opened 2017-07-01
headers: Enforce C-style comments

grim210:
When compiling with -std=c90, GCC complains about C++ style comments.
These changes do no affect readability of the header files and
ensures that the headers can be included in any C90-conformant code.

### DIFF
--- a/include/vulkan/vk_icd.h
+++ b/include/vulkan/vk_icd.h
@@ -1,7 +1,7 @@
-//
-// File: vk_icd.h
-//
 /*
+ * File: vk_icd.h
+ *
+ *
  * Copyright (c) 2015-2016 The Khronos Group Inc.
  * Copyright (c) 2015-2016 Valve Corporation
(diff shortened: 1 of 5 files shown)


grim210:
Fixes issue #1926

MarkY-LunarG:
vulkan.h and vk_platform.h come directly from Khronos and we will not directly modify them here.  If this is a problem, you should submit a comment on the KhronosGroup/Vulkan-Docs repo.

Please remove those before we look at the rest.

mark-lunarg:
Sorry, I didn't notice this PR when I responded to the issue.  Unless the vulkan header files change, I'm not excited about making these changes in this repo for the remainder of the files, especially for the header files that get shared among C/C++ projects.  If the main Vulkan project will change their headers then we will revisit updating our files to match.  I apologize for the inconvenience, and thanks for bringing it to our attention.

PR #1932, opened 2017-07-06
loader: Do not enumerate extensions from disabled implicit layers

jfroy:
This patch adds a check to vkEnumerateInstanceExtensionProperties to
omit extensions provided by disabled implicit layers. Per section 31.2
(1.0.51), "When pLayerName parameter is NULL, only extensions provided
by the Vulkan implementation or by implicitly enabled layers are
returned." Arguably, implicit layers that have been disabled are not
"implicitly enabled".


loader/trampoline.c
```diff
@@ -143,9 +143,10 @@ LOADER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkEnumerateInstanceExtensionPropert
         }
         loader_scanned_icd_clear(NULL, &icd_tramp_list);
 
-        // Append implicit layers.
+        // Append enabled implicit layers.
         loader_implicit_layer_scan(NULL, &instance_layers);
         for (uint32_t i = 0; i < instance_layers.count; i++) {
+            if (!loader_is_implicit_layer_enabled(NULL, &instance_layers.list[i])) continue;
```
MarkY-LunarG: Nit: Our coding standard requires braces.  Can you put the continue on a new line and open brace on the end of this and close brace after?

jfroy: Sure can.

mark-lunarg:
@jfroy, @MarkY-LunarG, these changes look fine to me as well, but we'll need a clean rebase (without the merge) in order to take this automatically.  Or, Mark can do this manually -- it's up to you guys.

jfroy:
I can certainly rebase.

PR #1940, opened 2017-07-13
Installer: add an option to hide VulkanRT installation

scyganINTEL:
Add switch /H, that allows to not enlist VulkanRT in
Programs&Features window.


scyganINTEL:
I believe hw makers would want to hide pre-installed software components.  At the same time, if bundled with an app, the installation should be fully controllable by end user.

So the IHV case maybe a bit different from ISV - hence I proposed added a switch, so the behavior can be configured.

davidlunarg:
We're leaning to always installing hidden. As part of that change, we would install all RTs in C:\ProgramFilesx86\VulkanRT, to avoid cluttering up C:\ProgramFilesx86 should lots of RTs be installed. Each version installed would have it's own folder in C:\ProgramFilesx86\VulkanRT. Stay tuned for a more firm proposal.

lenny-lunarg:
After talking about this, we decided that the best solution to the issue here was what @davidlunarg described, where we hide the runtime from "Programs and Features" on all installations. Since that change is a bit different, I'm closing this issue since this PR doesn't quite do what we decided on.

The fix to hide the runtime in "Programs and Features" is on master as commit 41cad1474cabf606437b2556bc4a0f945ceca43e. That change supersedes this PR.

PR #1943, opened 2017-07-14
Tobin pipe barrier vus

tobine:
An update to existing pipeline barrier validation along with a couple new checks and tests.


chrisforbes:
Really not happy with the SetUnexpectedError. Can we get a subpass self-dependency declared in those tests so they aren't doing multiple things wrong?

tobine:
Included CL in #1945 to clean-up InvalidBarriers test.

PR #1945, opened 2017-07-17
Tobin further pipe barrier vus

tobine:
Add two more CmdPipelineBarrier checks and associated tests. Remove previously-added unexpected errors from InvalidBarriers. Did this with a bool in the framework that allows the default renderpass to be created with a self-dependency in subpass 0.


layers/core_validation.cpp
```diff
@@ -6211,99 +6215,102 @@ static bool ValidateBarriers(const char *funcName, VkCommandBuffer cmdBuffer, Vk
...
+                            HandleToUint64(cb_state->commandBuffer), __LINE__, VALIDATION_ERROR_1b80093a, "DS",
                             "%s: As the Image Barrier for image 0x%" PRIx64
                             " is being executed within a render pass instance, oldLayout must equal newLayout yet they are "
                             "%s and %s. %s",
                             funcName, HandleToUint64(mem_barrier->image), string_VkImageLayout(mem_barrier->oldLayout),
                             string_VkImageLayout(mem_barrier->newLayout), validation_error_map[VALIDATION_ERROR_1b80093a]);
             }
-            skip |= ValidateMaskBitsFromLayouts(dev_data, cmdBuffer, mem_barrier->srcAccessMask, mem_barrier->oldLayout, "Source");
-            skip |= ValidateMaskBitsFromLayouts(dev_data, cmdBuffer, mem_barrier->dstAccessMask, mem_barrier->newLayout, "Dest");
+            skip |= ValidateMaskBitsFromLayouts(device_data, cb_state->commandBuffer, mem_barrier->srcAccessMask,
```
chrisforbes: This is asking for a further change to ValidateMaskBitsFromLayouts...

layers/core_validation.cpp
```diff
@@ -6177,7 +6177,7 @@ static bool ValidateRenderPassPipelineBarriers(layer_data *device_data, const ch
                             rp_handle, __LINE__, VALIDATION_ERROR_1b80092c, "CORE",
                             "%s: Barrier dstStageMask(0x%X) is not a subset of VkSubpassDependency dstStageMask(0x%X) of "
                             "subpass %d of renderPass 0x%" PRIx64 ". %s",
-                            funcName, src_stage_mask, sub_src_mask, cb_state->activeSubpass, rp_handle,
+                            funcName, dst_stage_mask, sub_dst_mask, cb_state->activeSubpass, rp_handle,
```
chrisforbes: OK :)

tests/layer_validation_tests.cpp
```diff
@@ -10091,6 +10091,12 @@ TEST_F(VkLayerTest, InvalidBarriers) {
     m_errorMonitor->VerifyFound();
     vkCmdEndRenderPass(m_commandBuffer->handle());
 
+    // Add a token self-dependency for this test to avoid unexpected errors
+    //  Since we re-init RenderTarget w/ self-dep, save initial default RP & FB for explicit clean-up
+    auto save_fb = m_framebuffer;
+    auto save_rp = m_renderPass;
```
chrisforbes: Yuck. This test function is doing far too much if this kind of thing looks attractive.

tests/vkrenderframework.cpp
```diff
@@ -417,6 +418,18 @@ void VkRenderFramework::InitRenderTarget(uint32_t targets, VkImageView *dsBindin
     rp_info.pAttachments = attachments.data();
     rp_info.subpassCount = 1;
     rp_info.pSubpasses = &subpass;
+    VkSubpassDependency subpass_dep = {};
+    if (m_addRenderPassSelfDependency) {
+        // Add a subpass self-dependency to subpass 0 of default renderPass
+        subpass_dep.srcSubpass = 0;
+        subpass_dep.dstSubpass = 0;
+        subpass_dep.srcStageMask = (VK_PIPELINE_STAGE_ALL_COMMANDS_BIT << 1) - 1;
+        subpass_dep.dstStageMask = (VK_PIPELINE_STAGE_ALL_COMMANDS_BIT << 1) - 1;
```
chrisforbes: This isn't really valid, although we're not enforcing it yet. See constraints in 6.6.1.

layers/core_validation.cpp
tobine: I checked it out. The cmdBuffer is only used as the object param for error messages so there's not another cb_state look-up. We could make it cb_state down the chain and only deref in the error chase, which would be a small gain for non-error case. Is that worthwhile?

tests/layer_validation_tests.cpp
tobine: Yeah, pulled the non-self-dep case into RenderPassBarrierConflicts test so all tests in InvalidBarriers now use the default subpass self-dependency.

tests/vkrenderframework.cpp
tobine: After a couple iterations I've updated this to just use all framebuffer-space flags.

layers/core_validation.cpp
chrisforbes: Not a big deal to leave this as-is then. Looks much better after last round of changes.

PR #1950, opened 2017-07-19
layers: Update subresource range VUIDs

krOoze:
Update subresource range with the new VUIDs and associated changes.


tests/layer_validation_tests.cpp
```diff
@@ -17377,99 +17592,107 @@ TEST_F(VkLayerTest, ImageViewSubresourceRangeTests) {
...
-    vkCreateImageView(m_device->handle(), &imgViewInfo, NULL, &imgView);
-    m_errorMonitor->VerifyFound();
-    imgViewInfo.viewType = VK_IMAGE_VIEW_TYPE_2D;
-    imgViewInfo.subresourceRange.layerCount = 1;
+    // Try layerCount = 0
+    {
+        m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                                             "vkCreateImageView: if pCreateInfo->viewType is VK_IMAGE_TYPE_2D_ARRAY, "
+                                             "pCreateInfo->subresourceRange.layerCount must be >= 1");
+        m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_0ac00b92);
```
krOoze: This test fails. I don't really understand why. If I comment the first `SetDesiredFailureMsg` it seems to work (with that error being listed as unexpected).

krOoze: This is because debug callback returns VK_TRUE (i.e. does not propagate to the next layer in chain). For now I "fixed" it with not expecting the second error in test and let `parameter_validation` cover it.

krOoze:
@mark-lunarg Just in case this PR was overlooked: this also needs reviews.

mark-lunarg:
@krOoze, thanks for the reminder.  In a short glance it seemed like some VUs and error messages got removed -- maybe a rebase glitch?

krOoze:
@mark-lunarg They should no longer be in the spec and were superseded by those other new VUs. So I just removed them. Or are those situations supposed to be handled differently?

mark-lunarg:
@krOoze, yeah, it's kind of murky at best.  In practice, no VUs should be added or removed from the database file (or error message file) manually.  When official header updates are handled, @tobine has provided us with a nifty tool that we use to do this automatically while assuring that the VUID lifetimes are managed consistently.  Outside of these header updates,  **only** the '`check_implemented`' and '`testname`' fields of the database file should be modified.

krOoze:
@mark-lunarg Well, the header update should have removed them, no?
So, I will just undelete them then. And mark them unimplemented instead. Or what do you advise?

mark-lunarg:
@krOoze, exactly, so if you rebased on Master they should already be gone.  Looks like there aren't any conflicts so it's all OK with me, but normally just changing the Y's to N's should be plenty good.

krOoze:
@mark-lunarg They are still there in master. Besides this is already based after relevant header update (I wouldn't be able to use the new VUIDs otherwise).

mark-lunarg:
Update test name to fix doc validator on Linux and pushed to master:

```
commit e25e057866af155bd456003120aa6155530a0e5f
Author: Petr Kraus <petr_kraus@email.cz>
Date:   Tue Aug 8 21:46:02 2017 +0200

    layers: PR1950, Update subresource range VUIDs

    With updated test names in database.
```

PR #1954, opened 2017-07-20
Don't complain about missing FS outputs when corresponding attachment writemask is 0

chrisforbes:


layers/shader_validation.cpp
```diff
@@ -768,8 +768,10 @@ static bool validate_vi_against_vs_inputs(debug_report_data const *report_data,
 }
 
 static bool validate_fs_outputs_against_render_pass(debug_report_data const *report_data, shader_module const *fs,
-                                                    spirv_inst_iter entrypoint, VkRenderPassCreateInfo const *rpci,
+                                                    spirv_inst_iter entrypoint, PIPELINE_STATE const *pPipeline,
```
tobine: Can kill camelCase on "pPipeline" name

layers/shader_validation.cpp
```diff
@@ -803,8 +803,13 @@ static bool validate_fs_outputs_against_render_pass(debug_report_data const *rep
                             "fragment shader writes to output location %d with no matching attachment", it_a->first.first);
             it_a++;
         } else if (!b_at_end && (a_at_end || it_a->first.first > it_b->first)) {
-            skip |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0, __LINE__,
+            // Only complain if there are unmasked channels for this attachment. If the writemask is 0, it's acceptable for the
+            // shader to not produce a matching output.
+            if (pPipeline->attachments[it_b->first].colorWriteMask != 0) {
+                skip |=
+                    log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0, __LINE__,
```
tobine: General ?
Should we add a Vulkan Spec VU for something like this? Would it even make sense in Vulkan spec or is SPIR-V spec the correct place? If SPIR-V spec then are we fine just living with no VUID for cases such as this?

chrisforbes: You can rule out the SPIR-V spec. This is interface matching behavior in the Vulkan environment.

PR #1959, opened 2017-07-21
Tobin gh1951 more cmd pipe vus

tobine:
This handles a couple of the trickier CmdPipelineBarrier cases, making sure renderPass barriers match up with renderPass and framebuffer contents.
Also various clean-up and some additional simpler checks.
All the appropriate database updates and new tests as well.


layers/core_validation.cpp
```diff
@@ -6164,16 +6164,18 @@ static bool ValidateRenderPassPipelineBarriers(layer_data *device_data, const ch
...
+        skip |= log_msg(device_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_RENDER_PASS_EXT,
+                        rp_handle, __LINE__, VALIDATION_ERROR_1b800928, "CORE",
+                        "%s: Barriers cannot be set during subpass %d of renderPass 0x%" PRIx64
+                        " with no self-dependency specified. %s",
+                        funcName, active_subpass, rp_handle, validation_error_map[VALIDATION_ERROR_1b800928]);
     } else {
         assert(rp_state->subpass_to_dependency_index[cb_state->activeSubpass] != -1);
-        const auto &sub_dep = rp_state->createInfo.pDependencies[rp_state->subpass_to_dependency_index[cb_state->activeSubpass]];
+        const auto &sub_dep = rp_state->createInfo.pDependencies[rp_state->subpass_to_dependency_index[active_subpass]];
+        const auto &sub_desc = rp_state->createInfo.pSubpasses[active_subpass];
```
chrisforbes: Introducing `sub_desc` here seems like the bit to call out.

layers/core_validation.cpp
```diff
@@ -6247,19 +6247,53 @@ static bool ValidateRenderPassPipelineBarriers(layer_data *device_data, const ch
...
             for (const auto &fb_attach : fb_state->attachments) {
                 if (img_bar_image == fb_attach.image) {
                     image_match = true;
+                    attach_index = index_count;
+                    break;
                 }
+                index_count++;
             }
-            if (!image_match) {
+            if (image_match) {  // Make sure subpass is referring to matching attachment
```
chrisforbes: All of this out into a separate function? `ValidateRenderPassPipelineBarriers` is getting ridiculously large.

layers/core_validation.cpp
```diff
@@ -6247,6 +6247,24 @@ static bool ValidateRenderPassPipelineBarriers(layer_data *device_data, const ch
                                 funcName, i, img_dst_access_mask, sub_dst_access_mask, cb_state->activeSubpass, rp_handle,
                                 validation_error_map[VALIDATION_ERROR_1b800930]);
             }
+            // Verify that a framebuffer image matches barrier image
+            const auto &fb_state = GetFramebufferState(device_data, cb_state->activeFramebuffer);
```
chrisforbes: `cb_state->activeFramebuffer` can be null inside a renderpass if this is a secondary CB. Need a test case for this to ensure we don't explode. All the actual checks in that case must be deferred to CmdExecuteCommands, the first point when you know the actual framebuffer that is used in the renderpass instance.

tests/layer_validation_tests.cpp
```diff
@@ -3923,8 +3923,26 @@ TEST_F(VkLayerTest, RenderPassBarrierConflicts) {
...
                          &img_barrier);
     m_errorMonitor->VerifyFound();
+    vkCmdEndRenderPass(m_commandBuffer->handle());
+    // Have image barrier layouts mis-match subpass attachment layout
+    // Create RP where subpass has wrong image layout
+    ref.layout = VK_IMAGE_LAYOUT_GENERAL;
+    VkRenderPass rp_badlayout;
+    rpci.dependencyCount = 1;
+    rpci.pDependencies = &dep;
+    err = vkCreateRenderPass(m_device->device(), &rpci, nullptr, &rp_badlayout);
```
chrisforbes: I'm not sure why you need the extra renderpass here -- can't you just use a barrier layout that doesn't match the renderpass used above?

tests/layer_validation_tests.cpp
```diff
@@ -3940,8 +3940,49 @@ TEST_F(VkLayerTest, RenderPassBarrierConflicts) {
...
+    VkRenderPassCreateInfo rpci2 = {VK_STRUCTURE_TYPE_RENDER_PASS_CREATE_INFO, nullptr, 0, 2, attach2, 1, subpasses, 1, &dep};
+    VkRenderPass rp_badsubindex;
+
+    err = vkCreateRenderPass(m_device->device(), &rpci2, nullptr, &rp_badsubindex);
+    ASSERT_VK_SUCCESS(err);
+
+    VkImageObj image2(m_device);
+    image2.InitNoLayout(32, 32, 1, VK_FORMAT_R8G8B8A8_UNORM, VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, VK_IMAGE_TILING_OPTIMAL, 0);
+    VkImageView imageView2 = image2.targetView(VK_FORMAT_R8G8B8A8_UNORM);
+    // re-use imageView from start of test
```
chrisforbes: Apart from /this/ tiny bit of entanglement with the previous state, this could be its own test?

tests/layer_validation_tests.cpp
tobine: Issue I ran into when attempting that was that image usage flag only allowed GENERAL or CB_OPTIMAL layouts and couldn't set GENERAL layout for both img_barrier layouts without getting separate layout error.
I'll revisit this now that I refactored generic layout check to be after renderPass barrier check.

tests/layer_validation_tests.cpp
tobine: subpasses & dep structs were also used from above but I agree that it's a bit cluttered. Can certainly pull out into separate test which will just make everything a bit more clear.

tobine:
Added some CLs to address feedback

PR #1966, opened 2017-07-24
Output strings instead of indices in physical device feature error messages

### LINKED ISSUE #1924: Improve output of VkPhysicalDeviceFeature struct error
With the new device simulation layer, it will be common to get errors like this:

While calling vkCreateDevice(), requesting feature #21 in VkPhysicalDeviceFeatures struct, which is not available on this device.

This is not as helpful as a message as it could be.  Basically this is a feature request to do what is in the TODO line here:

https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/core_validation.cpp#L2004

mark-lunarg:
Fixes #1924.  Tacked a generated helper function onto the enum string helper header.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -2009,12 +2009,11 @@ static bool ValidateRequestedFeatures(instance_layer_data *instance_data, const 
     uint32_t total_bools = sizeof(VkPhysicalDeviceFeatures) / sizeof(VkBool32);
     for (uint32_t i = 0; i < total_bools; i++) {
         if (requested[i] > actual[i]) {
-            // TODO: Add index to struct member name helper to be able to include a feature name
             skip |= log_msg(instance_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT,
                             VK_DEBUG_REPORT_OBJECT_TYPE_PHYSICAL_DEVICE_EXT, 0, __LINE__, DEVLIMITS_INVALID_FEATURE_REQUESTED, "DL",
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -2012,9 +2012,9 @@ static bool ValidateRequestedFeatures(instance_layer_data *instance_data, const
             // TODO: Add index to struct member name helper to be able to include a feature name
```
mikes-lunarg: TODONE

scripts/helper_file_generator.py
```diff
@@ -364,6 +364,24 @@ def GenerateEnumStringConversion(self, groupName, value_list):
...
     #
+    # Tack on a helper which, given an index into a VkPhysicalDeviceFeatures structure, will print the corresponding feature name
+    def DeIndexPhysDevFeatures(self):
+        pdev_members = None
+        for name, members, ifdef in self.structMembers:
+            if name == 'VkPhysicalDeviceFeatures':
+                pdev_members = members
+                break
+        deindex = '\n'
+        deindex += 'static const char * IndexToPhysDevFeatureString[] = {\n'
```
mikes-lunarg: You could scope IndexToPhysDevFeatureString inside GetPhysDevFeatureString

layers/core_validation.cpp
mark-lunarg: Ta-da.

scripts/helper_file_generator.py
mark-lunarg: Yes, nicer.

PR #1968, opened 2017-07-25
tests: Return errorcode on Windows if layer tests fail

cnorthrop:
The call to validate docs was stomping on the layer validation test result.


cnorthrop:
cherry-picked to master

PR #1970, opened 2017-07-25
Tighten up parameter_validation Enum checking

mark-lunarg:
To validate enums, PV checked that a value was between BEGIN and END, OR that it was greater than 1,000,000,000.  Any value greater than this number would pass.  New version checks against a list of explicit values.


mark-lunarg:
Great idea @mikes-lunarg, will add this in a future patch series.

PR #1981, opened 2017-08-02
Fix incorrect VUID for negative viewport height

mark-lunarg:
Layers used the wrong VUID for the checks for negative viewport height if the extensions were not enabled.

Fixed existing test, added a new test, and updated the database, and removed special case from validation_stats script.

From a LunarXchange issue, # 692.


mark-lunarg:
BTW, what's up with 1500099c anyhow?  It's the strangest VUID yet and needs to be fixed in the spec:

"If the VK_AMD_negative_viewport_height extension is enabled, height can also be negative."

tobine:
Does it make sense to just kill 1500099c? Seems like the non-extension case is handled with 15000998 & the extension case is handled with 1500099a.

mark-lunarg:
Right now 1500099a technically accounts only for maintenance1. The right thing to do is to probably change the wording of 99c so that it matches 99a, but only applies when the AMD extension is enabled.

mark-lunarg:
Shall I open an MR to have the 1500099c VUID wording changed?

tobine:
Yes, please, thanks! I agree with what you said 

> The right thing to do is to probably change the wording of 99c so that it matches 99a, but only applies when the AMD extension is enabled.

PR #1985, opened 2017-08-05
layers: Fix swapchain extent check

cdwfs:
This fixes a regression introduced -- by me :( -- in commit 583b0c41dfd0876d0481920f2e6d8d87433bc1d7. The swapchain's imageExtent must always be within the surface's min/max extent, regardless of the surface's current extent.


cdwfs:
I notice there's no test cases for vkCreateSwapchainKHR(); is that on the roadmap somewhere?

PR #1988, opened 2017-08-07
Search for ICD JSON in display adapter and software component registry keys

scyganINTEL:
This change extends the functionality of searching for ICD JSONs by adding
registry locations specific to given display adapter and software components
associated with this display adapter.

The exact locations in registry are queried using Windows public PnP
Configuration Manager API[1].

This change is required, as previous ICD locations (constant path
in "HKLM/Software") may be unreachable for drivers and their installers on
Windows RS3[2].

Similar change is being made for OpenCL[2]

[1]https://msdn.microsoft.com/en-us/library/windows/hardware/ff549713.aspx
[2]https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/21


lenny-lunarg:
These changes look good to me. I'm going to run the changes through our internal test farm to make sure they pass just fine, but they look good from reading over them.

loader/LoaderAndLayerInterface.md
```diff
@@ -1922,6 +1922,27 @@ the value is 0, then the loader will attempt to load the file.  In this case,
 the loader will open the first and last listings, but not the middle.  This
 is because the value of 1 for vendorb_vk.json disables the driver.
 
+Additionaly the loader will scan through registry keys specific to Display
+Adapters and all SoftwareComponents assocatied with these adapters for the
+locations of JSON manifest files. These keys are reclecting HKR device hive
```
jjulianoatnv: Should that word be s/reclecting/reflecting/ ?

loader/LoaderAndLayerInterface.md
```diff
@@ -1922,6 +1922,27 @@ the value is 0, then the loader will attempt to load the file.  In this case,
...
+   HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Class\{Adapter GUID}\000X\VulkanDriverName
+   HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Class\{Adapter GUID}\000X\VulkanDriverNameWow
+   
+   HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Class\{SoftwareComponent GUID}\000X\VulkanDriverName
+   HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Class\{SoftwareComponent GUID}\000X\VulkanDriverNameWow
+```
+If any of these keys exsist and is of type REG_SZ, the loader will open the JSON
+manifest file specified by the key valye. Each value must be a full absolute
+path to a JSON manifest file.
+
```
jjulianoatnv: Is the intention that the driver's inf file specify the string value of the new REG_SZ registry values using the %13% variable in the inf script language? This would make the new Vulkan registry values behave similarly to existing DX and OpenGL registry values, and it makes sense to me.

loader/loader.c
```diff
@@ -2746,15 +2991,24 @@ static VkResult loader_get_manifest_files(const struct loader_instance *inst, co
...
+        }
+
+        if ((VK_SUCCESS != reg_result && VK_SUCCESS != regHKR_result) || NULL == reg) {
             if (!is_layer) {
                 loader_log(inst, VK_DEBUG_REPORT_ERROR_BIT_EXT, 0,
                            "loader_get_manifest_files: Registry lookup failed "
                            "to get ICD manifest files.  Possibly missing Vulkan"
                            " driver?");
-                if (VK_SUCCESS == reg_result || VK_ERROR_OUT_OF_HOST_MEMORY == reg_result) {
+                if (VK_SUCCESS == reg_result || VK_ERROR_OUT_OF_HOST_MEMORY == reg_result ) {
```
jjulianoatnv: The change in this line adds incorrect whitespace and should be removed.

loader/loader.c
```diff
@@ -2746,15 +2991,24 @@ static VkResult loader_get_manifest_files(const struct loader_instance *inst, co
...
                     res = reg_result;
+                } else if (VK_SUCCESS == regHKR_result || VK_ERROR_OUT_OF_HOST_MEMORY == regHKR_result ) {
```
jjulianoatnv: wrong whitespace again

jjulianoatnv: I'd like regHKR_result to take precedence. Is there a reason that the old reg_result takes precedence?

loader/LoaderAndLayerInterface.md
```diff
@@ -1922,6 +1922,27 @@ the value is 0, then the loader will attempt to load the file.  In this case,
...
+"VulkanDriverNameWow" for 32-bit Windows on Windows compatibility) key can be
+present, what should look like:
+```
+   HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Class\{Adapter GUID}\000X\VulkanDriverName
+   HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Class\{Adapter GUID}\000X\VulkanDriverNameWow
+   
+   HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Class\{SoftwareComponent GUID}\000X\VulkanDriverName
+   HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Class\{SoftwareComponent GUID}\000X\VulkanDriverNameWow
+```
+If any of these keys exsist and is of type REG_SZ, the loader will open the JSON
```
jjulianoatnv: Should this be REG_MULTI_SZ instead of REG_SZ? The analogous DX and OpenGL registry values are REG_MULTI_SZ.

loader/loader.c
```diff
@@ -401,6 +407,245 @@ VKAPI_ATTR VkResult VKAPI_CALL vkSetDeviceDispatch(VkDevice device, void *object
...
+            loader_log(inst, VK_DEBUG_REPORT_ERROR_BIT_EXT, 0,
+                "loaderGetDeviceRegistryFiles: Failed to allocate space for display device names.");
+            result = VK_ERROR_OUT_OF_HOST_MEMORY;
+            return false;
+        }
+    } while (CM_Get_Device_ID_List(displayGUID, pDeviceNames, deviceNamesSize, flags) == CR_BUFFER_SMALL); 
+    
+    if (pDeviceNames) {
+
+        for (char *deviceName = pDeviceNames; *deviceName; deviceName += strlen(deviceName) + 1) {
```
jjulianoatnv: Is this looping over all all subkeys of HKLM\SYSTEM\CurrentControlSet\Control\Video\ ? If there is a way to extract a device's registry location under this folder from using D3DKMTQueryAdapterInfo, then that might be a more direct and preferred approach. I don't know if QAI exposes the registry path, and I can't tell from only looking over MSDN documentation.

loader/LoaderAndLayerInterface.md
scyganINTEL: done.

loader/loader.c
scyganINTEL: done.

loader/loader.c
scyganINTEL: whitespace: done

loader/LoaderAndLayerInterface.md
scyganINTEL: Indeed D3D &OGL use REG_MULTI_SZ, but only D3D seems to make use of it and store mulitple strings here. OGL uses only one string.

Unless someone **really** needs to have multiple JSON paths per key, I would like to use single string (REG_SZ) for VK, just to make things simpler.

loader/LoaderAndLayerInterface.md
scyganINTEL: Yes!

loader/loader.c
scyganINTEL: It's looping over pre-filtered list of devices returned from `CM_Get_Device_ID_List`  that are PRESENT (`CM_GETIDLIST_FILTER_PRESENT`) and their GUID matches the display GUID.

The purpose of the loop here is to gather all active devices from **all vendors**, however it's usually quite short: on typical setup you would get just 2 runs (discrete GPU + processor GPU).

D3DKMTQueryAdapterInfo is too direct, as it queries concrete adapter, while we would like to get list of all adapters here.

loader/loader.c
scyganINTEL: I have no strong feelings on the precedence, so I have flipped the order.

loader/LoaderAndLayerInterface.md
jjulianoatnv: REG_SZ makes sense.

loader/loader.c
jjulianoatnv: I think this is OK, however I think it might have to change again in the future if scenarios materialize where the app process is not allowed to iterate over this amount of the registry.

For the D3DKMT method, I was actually thinking:
```
foreach hAdapter returned by one of the adapter enumeration functions
    D3DKMTQueryAdapterInfo(hAdapter)
    D3DKMTCloseAdapter(hAdapter)
```

The adapter enumeration functions are a mess. On win7 I think one has to use EnumDisplayDevices and D3DKMTOpenAdapterFromHdc. But that fails to find an adapter that has no displays attached to the desktop. Starting with win8 there is D3DKMTEnumAdapters. In the WDK headers there is a "2" version of this one that AFAICT is not actually documented on MSDN, although it is documented in the WDK header.

I guess it's easier to go with the current PR's registry scanning method, and continue to use the registry scanning method as long as the required registry access is available.

Also, I was not able to find a QueryAdapterInfo function that returns the desired registry path, so I guess my idea is a non-starter anyway.

This is a long winded way of saying that I approve of the registry scanning method in this PR.

lenny-lunarg:
I merged this pull request to master as commits 9beaa801945469eebaae6feb2cb3ac1f85cea292 and b69cb52f7c39c8f6faeed7e172f90c8c4f307043. I also added two commits. One of these prevents the same json manifest from being read twice, and the other changes some of your wanring messages to info messages. I changed the messages because our internal tests were failing on warnings, and because some of the messages looked like they really should have been info.

Just one note for future reference. We do have some [coding conventions](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/CONTRIBUTING.md#coding-conventions-and-formatting) for the repo. I didn't try to enforce them in the code, because we've been notoriously lax about the conventions in the loader and I'm not going to hold you to standards we don't hold ourselves to, but I generally do try to format the commit messages according to those conventions. This is why I modified your commit messages. The general idea is that the first line should look like:

```
loader: Some message here
```

and be no longer than 50 characters. Any other lines after that should be a maximum of 70 characters.

But thanks for the pull request. This should be good to have it taken care of for the future.

scyganINTEL:
Thanks Lenny for all the work here!

jjulianoatnv:
I don't understand the purpose of softwareComponentGUID. Why do you check for this?

scyganINTEL:
@jjulianoatnv This is to look for ICDs in so-called SotwareComponent-s (https://docs.microsoft.com/en-us/windows-hardware/drivers/install/adding-software-components-with-an-inf-file)

If the loader fails to find ICDs in adapter tree, it tries to find ICDs in keys related to  child software components of this adapter.

PR #1995, opened 2017-08-09
Tobin gh1992 consecutive descriptor updates

### LINKED ISSUE #1992: layers: Incorrect handling of consecutive descriptor updates
The validation layer `VK_LAYER_LUNARG_core_validation` currently handles consecutive descriptor updates incorrectly.

As of b87dda8, this validation layer accepts consecutive descriptor updates. However, it does not update the internal state correctly, and causes false positives/negatives in later calls.

For example, let's say you have a descriptor set with the following layout:

|             |   element = 0    |   element = 1    |
| ----------- | ---------------- | ---------------- |
| binding = 0 | (Uniform Buffer) | (Uniform Buffer) |
| binding = 1 | (Uniform Buffer) |                  |

After being updated with the following `VkWriteDescriptorSet`

```
VkWriteDescriptorSet {
    ...
    .dstBinding = 0,
    .dstArrayElement = 0,
    .descriptorCount = 3,
}
```

its contents would look like this:

|             |   element = 0    |   element = 1    |
| ----------- | ---------------- | ---------------- |
| binding = 0 | `pBufferInfo[0]` | `pBufferInfo[1]` |
| binding = 1 | `pBufferInfo[2]` | ⊥                |

However, the current implementation of [`DescriptorSet::PerformWriteUpdate`] (which updates an internal shadow copy of a descriptor set for later validation) does this incorrectly. Specifically, it starts over reading the input descriptor list from the first element whenever proceeding to the next binding. Therefore, it updates its internal state to the following:

|             |     element = 0      |   element = 1    |
| ----------- | -------------------- | ---------------- |
| binding = 0 | `pBufferInfo[0]`     | `pBufferInfo[1]` |
| binding = 1 | **`pBufferInfo[0]`** | ⊥                |

As a result,

- No validation errors are reported if `pBufferInfo[2]` was destroyed after the update, and then there was a use of this descriptor set.
- A false validation error occurs if `pBufferInfo[0]` was destroyed but the corresponding descriptor element was overwritten using a valid buffer.

[`DescriptorSet::PerformWriteUpdate`]: https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/b42b2f0f35df5143dcc982217434cf0d42a45afe/layers/descriptor_sets.cpp#L612

tobine:
Fixes #1992 

When a consecutive descriptor update would roll-over to the next binding, the index into the update array was being incorrectly reset to 0. This fixes the bug an adds a test case to verify the fix.


tests/layer_validation_tests.cpp
```diff
@@ -5322,6 +5322,161 @@ TEST_F(VkLayerTest, WriteDescriptorSetIntegrityCheck) {
...
+    layout_binding[0].pImmutableSamplers = NULL;
+
+    layout_binding[1].binding = 1;
+    layout_binding[1].descriptorType = VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER;
+    layout_binding[1].descriptorCount = 1;
+    layout_binding[1].stageFlags = VK_SHADER_STAGE_ALL;
+    layout_binding[1].pImmutableSamplers = NULL;
+
+    VkDescriptorSetLayoutCreateInfo ds_layout_ci = {};
+    ds_layout_ci.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_CREATE_INFO;
```
chrisforbes: OneShotDescriptorSet!

tobine: Ah yes, I had a sneaky feeling I was forgetting something.
I mean, I wanted to help promote OneOffDescriptorSet so I intentionally excluded it so I could push the follow-on CL that I just did as a promotional tool to show everyone how great OneOffDescriptorSet is...
Anyway, mea culpa. CL to update test is pushed and I won't forget again. (but remind me again if I do)

PR #1996, opened 2017-08-09
Parameter Validation Layer Rework

mark-lunarg:
Changed the codegen to autogenerate validation functions for any new APIs without user intervention. Manual updates are only necessary to extend PV functionality.

- parameter_validation.h is now checked into the layers directory
- parameter_validation.cpp is now generated
- parameter_validation_utils.cpp is new and contains any custom (non-generated) or housekeeping functions.
- parameter_validation_utils.h is deleted
- additionally, add newly-covered VUIDs into database


tobine:
Something fundamental is broken with updated parameter validation layer on Android. The crash occurs in parameter validation layer on the first test (RequiredParameter) when the loader calls into vkEnumerateInstanceLayerProperties().

layers/parameter_validation_utils.cpp
```diff
@@ -0,0 +1,2519 @@
...
+                                            __LINE__, VALIDATION_ERROR_1082b00b, LayerName,
+                                            "vkCreateGraphicsPipelines: parameter pCreateInfos[%d].pTessellationState->sType must "
+                                            "be VK_STRUCTURE_TYPE_PIPELINE_TESSELLATION_STATE_CREATE_INFO. %s",
+                                            i, validation_error_map[VALIDATION_ERROR_1082b00b]);
+                        }
+
+                        if (pCreateInfos[i].pTessellationState->patchControlPoints == 0 ||
+                            pCreateInfos[i].pTessellationState->patchControlPoints >
+                                device_data->device_limits.maxTessellationPatchSize) {
+
```
tobine: The first android crash occurs here when physicalDevice is null. Should this function begin similar to the [object_tracker](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/object_tracker_utils.cpp#L611) version where the first thing checked is like line 2234 below to see if the pLayerName matches our name?

mark-lunarg:
@tobine, thanks, now passes on our farm.

mark-lunarg:
@tobine, OK to go?

tobine:
Yep, looks good to go now.

PR #1997, opened 2017-08-09
layers: Fix VS2017 complaining about OOBA

krOoze:
Replaced std::stringstream with std::ostringstream (except one occurence
of std::istringstream).

<s>Somehow that fixes VS2017's RangeCheck instrumentation complaining about
out of bounds access when running in debug mode.</s> Nope, false alert. Just a rebuild seems to fix it.


krOoze:
Actually just "rebuild" within the IDE suffices.

I build with `build_windows_targets.bat 64` and then in the VS 2017 IDE select the `vk_layer_validation_tests` as default, select the "Debug" configuration and "Start Debugging".
And it makes a RangeCheck exception on some `stringstream::operator<<`.
Doing "rebuild" on that project fixes it. Weird.

Maybe I will try to replace the `stringstream`s with regular `string`s later...

mark-lunarg:
I've been encountering this lately as well and it's driving me nuts.  @krOoze, these `stringstream `refs have been in for almost two years, any idea why this has recently started happening?

krOoze:
@mark-lunarg Isn't it. Was last straw for me too. Though it does it for me for as long as I remember. I usually work around this by using non-debug mode, or by `--gtest_filter`.
Could you guess a commit where this didn't happen? It would help with bisection and finding the offending commit could be very informative.

>  these stringstreamrefs have been in for almost two years, any idea why this has recently started happening?

No idea. Could even be red herring. It is simply outside my debugging skill/experience.

And just Rebuild fixes it. Perhaps messing with the build script could reveal something... And perhaps it is just some obscure VS bug.

krOoze:
@mark-lunarg PS:
I gave it another try and found the "reason":
The Control Flow Guard tries to check indirect call to `std::basic_stringbuf<char,std::char_traits<char>,std::allocator<char> >::overflow` but aparently considers the fpointer invalid.

mark-lunarg:
@krOoze, Sweet!  Next time you're in town we'll buy you a beer.  Talked to @lenny-lunarg who handled the original Control Flow Guard PR, and we think we could disable this feature for validation layers leaving it in place for the loader and VulkanInfo, which get shipped as part of the run-time.  Thanks again.

krOoze:
@mark-lunarg I also get the suspicion that there is something wrong with the build.

This seems to happen when Debug is built after the Realease build (which is what the `build_windows_targets.bat` does).

Also even after `build_windows_targets.bat` selecting Debug in the IDE builds some things. Which it probably shouldn't (already built). It probably means the two targets somehow poison each other.

krOoze:
Closing this for now, it is only taking place &mdash; the commited stuff doesn't work. I will just make an Issue with my findings.

PR #1998, opened 2017-08-10
Implement VK_DBG_LAYER_ACTION_BREAK

### LINKED ISSUE #1991: VK_DBG_LAYER_ACTION_BREAK not working
Enabling VK_DBG_LAYER_ACTION_BREAK didn't worked for me.
Also I don't see it used in sources.

mark-lunarg:
This feature, though documented and allowed for, was apparently unimplemented.  

Also fixed a documentation error that would keep VK_DBG_LAYER_ACTION_DEBUG_OUTPUT from working correctly.

@cnorthrop, any suggestions on ensuring the non-Windows break will work on Android?

Fixes #1991.


cnorthrop:
The signal handling should be very similar if not identical to Linux, so whatever you do to test that, it should work.  If it doesn't, we can look at it.

mark-lunarg:
It worked fine in GDB and produced a good call-stack.

cnorthrop:
Ah, okay, I presumed you were going to hook up a regression test, but maybe that's not possible when testing a breakpoint.  We should be able to try this out by hand with gdbserver, but no need to hold up the PR.

mark-lunarg:
We've begun planning for a follow-on for testing the debugreport stuff including the layer settings file options (recognizing the issues with testing debugbreaks).

PR #1999, opened 2017-08-11
tests: Add PushDescriptorUnboundSetTest test

jozefkucia:
Make sure that no errors are generated when `vkCmdBindDescriptorSets()`
is not called for descriptor set layouts created with
`VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR`. 

The test currently fails. See issue #2000


tests/layer_validation_tests.cpp
```diff
@@ -20611,6 +20611,164 @@ TEST_F(VkPositiveLayerTest, PushDescriptorNullDstSetTest) {
...
+    } else {
+        printf("             Push Descriptors Extension not supported, skipping tests\n");
+        return;
+    }
+    ASSERT_NO_FATAL_FAILURE(InitState());
+    ASSERT_NO_FATAL_FAILURE(InitViewport());
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+    m_errorMonitor->ExpectSuccess();
+
+    if (m_device->props.limits.maxBoundDescriptorSets < 2) {
```
chrisforbes: You don't need this. This limit is guaranteed to be at least 4.

chrisforbes: Do you actually need the non-push set in this test? It adds a lot of noise, and it appears your test would work fine with just the push set.

jozefkucia: When I do not call `vkCmdBindDescriptorSets()` then the false positive validation error message from issue #2000 is not triggered. Thus, the non-push set is needed to test this particular scenario.

mark-lunarg:
@jozefkucia, we'll push this test along with a fix.  Thanks for the test -- this is how EVERYONE should submit issues!

jozefkucia:
Merged as part of #2040

PR #2003, opened 2017-08-15
demos: Fix disappearing objects in smoketest

karl-lunarg:
Fixes LunarG/VulkanSamples#150

When an object reaches the end of its "curve" path
it receives a new curve.  To ease the transition from
the old curve to the new curve, the origin of the new
curve is the ending point of the old curve.  However,
this can result in the starting point of the new curve
being out of the view.  This results in objects slowly
leaving the view.

Add code to move the origin of the new curve back inside
the original view bounds only if the end of the old curve
is out of the view bounds.  This does not add any additional
jarring because any object getting a new origin is already
outside the view anyway.

Change-Id: I50e88be453f31b9c37878f63c97d4c7c8154e718


karl-lunarg:
Force-pushed a fix to address build failure on Windows.

Issue #614, opened 2016-06-02
Validation doesn't fire when setting dynamic state with non-dynamic pipeline bound


baldurk:
This is basically just the validation not catching the case I mentioned in KhronosGroup/Vulkan-Docs#235.

When calling `vkCmdSetViewport` etc, a pipeline must already be bound that uses them as dynamic. It's undefined to call these functions when the currently bound pipeline is non-dynamic.

chrisforbes:
The spec still seems pretty vague about this area. There's the new VU block, but what happens to previously bound dynamic state when binding a new pipeline? Are the intersection of the old and new pipelines' dynamic states still valid after the second CmdBindPipeline?

baldurk:
I think that will need clarification from inside Khronos. My understanding would be that yes as you suggest the intersection would be valid. The text from chapter 9 is:

> A pipeline object is bound to the device state in command buffers. Any pipeline object state that is marked as dynamic is not applied to the device state when the pipeline is bound. Dynamic state not set by binding the pipeline object can be modified at any time and persists for the lifetime of the command buffer, or until modified by another dynamic state command or another pipeline bind. No state, including dynamic state, is inherited from one command buffer to another. Only dynamic state that is required for the operations performed in the command buffer needs to be set. For example, if blending is disabled by the pipeline state then the dynamic color blend constants do not need to be specified in the command buffer, even if this state is marked as dynamic in the pipeline state object. If a new pipeline object is bound with state not marked as dynamic after a previous pipeline object with that same state as dynamic, the new pipeline object state will override the dynamic state. Modifying dynamic state that is not set as dynamic by the pipeline state object will lead to undefined results.

That seems to suggest that binding a pipeline inherits any dynamic state that was previously bound, sets any non-dynamic state it contains, and any subsequent dynamic state sets calls are only valid for whatever the current pipeline has as dynamic.

The spec in this section though only specifically calls out what happens for binding a non-dynamic pipeline after a dynamic one. Not explicitly mentioning dynamic-to-dynamic or static-to-dynamic.

chrisforbes:
Fixed by various changes around 15 August.

PR #2004, opened 2017-08-15
Untangle static/dynamic state validity tracking a bit

chrisforbes:
This doesn't actually solve #614, but it gets us closer to sanity.

Things still to do here:
- Untangle checks from side effects in CmdBindPipeline
- Untangle compute vs graphics cases (they're not really the same at all)
- Implement the additional check for #614.

Need to consider what to do with the viewport & scissor state for #614 -- we currently handle them differently in order to get per-viewport and per-scissor tracking working properly.


layers/core_validation.cpp
```diff
@@ -5130,10 +5130,18 @@ VKAPI_ATTR void VKAPI_CALL CmdBindPipeline(VkCommandBuffer commandBuffer, VkPipe
...
         // TODO: VALIDATION_ERROR_18000612 VALIDATION_ERROR_18000616
 
-        PIPELINE_STATE *pipe_state = getPipelineState(dev_data, pipeline);
+        auto pipe_state = getPipelineState(dev_data, pipeline);
+        if (VK_PIPELINE_BIND_POINT_GRAPHICS == pipelineBindPoint) {
+            auto old_pipe_state = cb_state->lastBound[pipelineBindPoint].pipeline_state;
+            if (old_pipe_state) {
+                cb_state->status &= ~MakeStaticStateMask(old_pipe_state->graphicsPipelineCI.ptr()->pDynamicState);
+            }
+            if (pipe_state) {
```
tobine: This "if" check shouldn't be needed as bad pipeline would have triggered in PV. Similar check below is from long, long ago when CV was much more conservative and that check should be able to be killed in favor of same check in PV.

layers/core_validation.cpp
```diff
@@ -5132,12 +5133,10 @@ VKAPI_ATTR void VKAPI_CALL CmdBindPipeline(VkCommandBuffer commandBuffer, VkPipe
 
         auto pipe_state = getPipelineState(dev_data, pipeline);
         if (VK_PIPELINE_BIND_POINT_GRAPHICS == pipelineBindPoint) {
-            auto old_pipe_state = cb_state->lastBound[pipelineBindPoint].pipeline_state;
-            if (old_pipe_state) {
-                cb_state->status &= ~MakeStaticStateMask(old_pipe_state->graphicsPipelineCI.ptr()->pDynamicState);
-            }
+            cb_state->status &= ~cb_state->static_status;
```
tobine: Yeah, this is good. Extra mask saves the function call.

layers/core_validation.cpp
```diff
@@ -5201,19 +5201,20 @@ VKAPI_ATTR void VKAPI_CALL CmdSetLineWidth(VkCommandBuffer commandBuffer, float
     if (pCB) {
         skip |= ValidateCmdQueueFlags(dev_data, pCB, "vkCmdSetLineWidth()", VK_QUEUE_GRAPHICS_BIT, VALIDATION_ERROR_1d602415);
         skip |= ValidateCmd(dev_data, pCB, CMD_SETLINEWIDTHSTATE, "vkCmdSetLineWidth()");
-        pCB->status |= CBSTATUS_LINE_WIDTH_SET;
 
-        PIPELINE_STATE *pPipeTrav = pCB->lastBound[VK_PIPELINE_BIND_POINT_GRAPHICS].pipeline_state;
-        if (pPipeTrav != NULL && !isDynamic(pPipeTrav, VK_DYNAMIC_STATE_LINE_WIDTH)) {
-            skip |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_WARNING_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
+        if (pCB->static_status & CBSTATUS_LINE_WIDTH_SET) {
```
tobine: Looks like most of the other CmdSet*() dynamic state functions are missing the check for similar VU. Please follow-on with those check added or open an issue to track them to completion.

chrisforbes: I'm adding all the missing cases now.

PR #2005, opened 2017-08-15
Tobin synch model

tobine:
This is a partially complete initial implementation of synchronization object modeling in validation. Initially it will only trigger warnings for suspected WaR and RaW conflicts. There are known holes in coverage, so if people object to having any warnings at all for the initial release, I'm open to just disabling the callback initially.
My primary interest at this point it getting review, fixing code/arch issues, and landing some of the initial framework before the PR gets even bigger than it already is.

For a general idea of what I'm going for here, see this [design doc](https://docs.google.com/document/d/1i4BpZ7Uv_UjL0DLNXagnCCxH30BHHll0D7IakiuSDNg/edit) for a walk-through of the basic algorithm.
This initial version has many known holes including but not limited to events, semaphores, fences, host synch, inter-cmd-buffer race conditions, secondary cmd buffer execution, EXTERNAL deps, and renderPass dependencies which we do already have some checks for but not within this framework.


tobine:
Updated test and windows now happy

chrisforbes:
@tobine, is this PR superceded now?

PR #2008, opened 2017-08-16
android: Updates for NDK r15

cnorthrop:
Removes app_dummy from apps that use native activity.

For more details refer to: https://github.com/android-ndk/ndk/issues/381

@ggfan, this mirrors changes you've made downstream.


cnorthrop:
cherry-picked to master

PR #2010, opened 2017-08-16
vkjson: Add support for Android

cnorthrop:
This commit adds `vkjson` and `vkjson_info` to the Android build.
It is only generated as an executable for devices with root
access, but it could be expanded to an APK with some more
boilerplate work.

See updates to `BUILD.md` for how to build and run on the device.

This work is primarily to support downstream validation of the
`device_simulation` layer on Android.


cnorthrop:

PR #2001, opened 2017-08-11
tests: Clean up PushDescriptorNullDstSetTest

jozefkucia:

PR #2011, opened 2017-08-17
Track descriptor bindings for PushDescriptorSets

### LINKED ISSUE #2000: False validation errors for VK_KHR_push_descriptor
Validation layers produce `VkPipeline 0x10 uses set #0 but that set is not bound` errors for push descriptor sets. See  #1999 for a test case which reproduces the issue.

mark-lunarg:
Fixes #2000.

Hook up PushDescriptorSets in core validation so that their bindings can be accounted for.  This PR includes commits from two other PRs ( #1999 and #2001) as they directly relate.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5577,6 +5577,14 @@ VKAPI_ATTR void VKAPI_CALL CmdBindDescriptorSets(VkCommandBuffer commandBuffer, 
                                                       dynamicOffsetCount, pDynamicOffsets);
 }
 
+VKAPI_ATTR void VKAPI_CALL CmdPushDescriptorSetKHR(VkCommandBuffer commandBuffer, VkPipelineBindPoint pipelineBindPoint,
+                                                   VkPipelineLayout layout, uint32_t set, uint32_t descriptorWriteCount,
+                                                   const VkWriteDescriptorSet *pDescriptorWrites) {
+    layer_data *device_data = GetLayerDataPtr(get_dispatch_key(commandBuffer), layer_data_map);
+    device_data->dispatch_table.CmdPushDescriptorSetKHR(commandBuffer, pipelineBindPoint, layout, set, descriptorWriteCount,
+                                                        pDescriptorWrites);
+}
+
 VKAPI_ATTR void VKAPI_CALL CmdBindIndexBuffer(VkCommandBuffer commandBuffer, VkBuffer buffer, VkDeviceSize offset,
                                               VkIndexType indexType) {
     bool skip = false;
@@ -10645,6 +10653,7 @@ static const std::unordered_map<std::string, void*> name_to_funcptr_map = {
     {"vkDestroyDescriptorUpdateTemplateKHR", (void*)DestroyDescriptorUpdateTemplateKHR},
     {"vkUpdateDescriptorSetWithTemplateKHR", (void*)UpdateDescriptorSetWithTemplateKHR},
     {"vkCmdPushDescriptorSetWithTemplateKHR", (void*)CmdPushDescriptorSetWithTemplateKHR},
+    {"vkCmdPushDescriptorSetKHR", (void*)CmdPushDescriptorSetKHR},
     {"vkCreateSwapchainKHR", (void*)CreateSwapchainKHR},
     {"vkDestroySwapchainKHR", (void*)DestroySwapchainKHR},
     {"vkGetSwapchainImagesKHR", (void*)GetSwapchainImagesKHR},
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5369,168 +5369,212 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
     if (!skip) dev_data->dispatch_table.CmdSetStencilReference(commandBuffer, faceMask, reference);
 }
 
-VKAPI_ATTR void VKAPI_CALL CmdBindDescriptorSets(VkCommandBuffer commandBuffer, VkPipelineBindPoint pipelineBindPoint,
-                                                 VkPipelineLayout layout, uint32_t firstSet, uint32_t setCount,
-                                                 const VkDescriptorSet *pDescriptorSets, uint32_t dynamicOffsetCount,
-                                                 const uint32_t *pDynamicOffsets) {
-    bool skip = false;
-    layer_data *dev_data = GetLayerDataPtr(get_dispatch_key(commandBuffer), layer_data_map);
-    unique_lock_t lock(global_lock);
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -5570,11 +5570,24 @@ VKAPI_ATTR void VKAPI_CALL CmdBindDescriptorSets(VkCommandBuffer commandBuffer,
...
+}
+
+static void PostCallRecordCmdPushDesriptorSetKHR(layer_data *device_data, VkCommandBuffer commandBuffer,
+                                                 VkPipelineBindPoint pipelineBindPoint, VkPipelineLayout layout, uint32_t set,
+                                                 uint32_t descriptorWriteCount, const VkWriteDescriptorSet *pDescriptorWrites) {
+    auto cb_state = GetCBNode(device_data, commandBuffer);
+    VkDescriptorSet desc_set;
+    // Loop through descwrites, binding each desc set in turn
+    for (uint32_t i = 0; i < descriptorWriteCount; i++) {
+        desc_set = pDescriptorWrites[i].dstSet;
```
jozefkucia: That won't work in a general case. `dstSet` should be ignored for push descriptors and can be invalid.

The spec says:

> Each element of pDescriptorWrites is interpreted as in VkWriteDescriptorSet, except the dstSet
> member is ignored.

jozefkucia: It could be a good idea to set `descriptor_write.dstSet` to `VK_NULL_HANDLE` before calling `vkCmdPushDescriptorSetKHR()` in `PushDescriptorUnboundSetTest`. This would catch this problem I guess.

mark-lunarg:
Eh, need to stew over the binding implications, pulling for now.

layers/core_validation.cpp
```diff
@@ -5369,168 +5369,212 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
-                // Verify that set being bound is compatible with overlapping setLayout of pipelineLayout
-                if (!verify_set_layout_compatibility(descriptor_set, pipeline_layout, set_idx + firstSet, error_string)) {
-                    skip |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT,
-                                    VK_DEBUG_REPORT_OBJECT_TYPE_DESCRIPTOR_SET_EXT, HandleToUint64(pDescriptorSets[set_idx]),
-                                    __LINE__, VALIDATION_ERROR_17c002cc, "DS",
-                                    "descriptorSet #%u being bound is not compatible with overlapping descriptorSetLayout "
-                                    "at index %u of pipelineLayout 0x%" PRIxLEAST64 " due to: %s. %s",
-                                    set_idx, set_idx + firstSet, HandleToUint64(layout), error_string.c_str(),
-                                    validation_error_map[VALIDATION_ERROR_17c002cc]);
+static void PostCallRecordCmdBindDescriptorSets(layer_data *device_data, GLOBAL_CB_NODE *cb_state,
```
mark-lunarg: In this PR they're both after.  Doesn't really matter I guess, I could move them before since both routines return void.

layers/core_validation.cpp
```diff
@@ -5369,168 +5369,212 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
+                                                VkPipelineBindPoint pipelineBindPoint, VkPipelineLayout layout, uint32_t firstSet,
+                                                uint32_t setCount, const VkDescriptorSet *pDescriptorSets,
+                                                uint32_t dynamicOffsetCount, const uint32_t *pDynamicOffsets) {
+    uint32_t total_dynamic_descriptors = 0;
+    string error_string = "";
+    uint32_t last_set_index = firstSet + setCount - 1;
+    if (last_set_index >= cb_state->lastBound[pipelineBindPoint].boundDescriptorSets.size()) {
+        cb_state->lastBound[pipelineBindPoint].boundDescriptorSets.resize(last_set_index + 1);
```
mark-lunarg: The pushdescriptor path didn't call the validation routine, so it was necessary.  This'll probably go away anyhow.

mark-lunarg:
@tobine , in the comment above, I read "store the layout & set off in the lastbound struct off of the cmd buffer and **dynamically allocate & update the set** at push descriptors time if the push layout changes." Does this mean to dynamically allocate a descriptor set and update it at push time without the application's knowledge, or just modifying the binding validation to take this information into account?

layers/core_validation.cpp
```diff
@@ -5570,11 +5570,24 @@ VKAPI_ATTR void VKAPI_CALL CmdBindDescriptorSets(VkCommandBuffer commandBuffer,
...
     lock.unlock();
-    if (!skip)
-        PostCallRecordCmdBindDescriptorSets(device_data, cb_state, pipelineBindPoint, layout, firstSet, setCount,
-                                            pDescriptorSets, dynamicOffsetCount, pDynamicOffsets);
-    device_data->dispatch_table.CmdBindDescriptorSets(commandBuffer, pipelineBindPoint, layout, firstSet, setCount, pDescriptorSets,
-                                                      dynamicOffsetCount, pDynamicOffsets);
+    if (!skip) {
+        device_data->dispatch_table.CmdBindDescriptorSets(commandBuffer, pipelineBindPoint, layout, firstSet, setCount,
+                                                          pDescriptorSets, dynamicOffsetCount, pDynamicOffsets);
+        PostCallRecordCmdBindDescriptorSets(device_data, cb_state, pipelineBindPoint, layout, firstSet, setCount, pDescriptorSets,
```
mark-lunarg: Right, I'll add this to the layers doc as well.  Updated stuff to match and fixed the locks/unlocks.

tobine:
I'm thinking have a separate push_set in LAST_BOUND_STATE that is dynamically allocated when vkCmdPushDescriptorSetKHR() is initially called and freed/reallocated if vkCmdPushDescriptorSetKHR() is called with conflicting layout. This matched the specs description of push descriptors:

> Push descriptors are a small bank of descriptors whose storage is internally managed by the command buffer 

I view the application's call of vkCmdPushDescriptorSetKHR() to be its indication that it wants this set using (including allocated if it's not already) so we're not really doing it without the app's knowledge, it's just validation matching what spec indicates is probably also occurring in the driver.

tobine:
FYI for any interested 3rd parties, I spoke to @mark-lunarg to clarify, but when I was suggesting allocating a descriptor set I meant an instance of our internal DescriptorSet class, not actually calling down to the driver to allocate a descriptor set.

chrisforbes:
Proposed tracking of bindings sounds pretty sensible to me. There may be some snags in our descriptor set handling that assume there to be a backing set, though...

PR #2026, opened 2017-08-24
Tobin synch event model

tobine:
This is building on and replacing #2005 
This includes further commits to handle inter-CB memory access conflicts.


chrisforbes:
As mentioned offline, this has brutal perf impact. Frame times in `smoketest --validate -s` are more than 3x higher with this series compared to master.

tobine:
Replaced by #2049

PR #2027, opened 2017-08-25
layers: Introduce some unique_ptr into pipeline creation

chrisforbes:


chrisforbes:
Pushing on the basis of tobine@'s earlier approval; second commit is just more of the same for the compute side.

PR #2033, opened 2017-08-26
WIP: support using cygwin git

jjulianoatnv:
Adds support for syncing with cygwin git using the .sh script and --no-build,
then building with a Visual Studio Developer Command Prompt using the .bat script and --no-sync.

Also improves the usage text for both scripts, and aligns their command line options to be similar.

The .bat script's --all option is removed, as it is no longer needed. Passing it no arguments now does the same thing that the --all option used to do.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -212,7 +212,7 @@ for the Properties->Details tab of the loader vulkan-1.dll file that is built.
 Build all Windows targets after installing required software and cloning the Loader and Validation Layer repo as described above by completing the following steps in a "Developer Command Prompt for VS2013" window (Note that the update\_external\_sources script used below builds external tools into predefined locations. See **Loader and Validation Layer Dependencies** for more information and other options):
 ```
 cd Vulkan-LoaderAndValidationLayers  # cd to the root of the cloned git repository
-update_external_sources.bat --all
+update_external_sources.bat
 build_windows_targets.bat
 ```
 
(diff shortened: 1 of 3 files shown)


jjulianoatnv:
The WIP can be removed after it has been tested on linux.

I have tested the behavior on Windows. I do not have an install of linux on which to test. I would appreciate if someone else tests the changes on linux. Thanks.

jjulianoatnv:
The CI failed because it passes the argument --all, which I removed because it is redundant with passing no arguments.

What's the right thing here? Change the CI, or restore the --all parameter?
I don't know how to change the CI as part of my pull request.

mark-lunarg:
Let's leave the `--all` parameter to keep other folks scripts from breaking.  The contributing.md doc has some guidelines for contributions including the commit message formats.  If you can add the **all** parameter back in I'll test on Linux and run it through our CI systems.

jjulianoatnv:
Thanks. I pushed an additional commit that restores the --all option.

What should I use for the initial word of the commit messages' subject lines? Maybe "build"?

To fix commit messages, do I change them locally then do a forced push to my branch?

mark-lunarg:
Wow, thanks Jeff!  Yeah, 'build:' is perfect.  Exactly right on the updated commits -- gitlab will do The Right Thing.  I will check this on Monday.

mark-lunarg:
Fixed commit messages, squashed some changes to avoid errors and pushed to Master.

jjulianoatnv:
@mark-lunarg , thanks for fixing the commit messages before I got to it.

PR #2036, opened 2017-08-28
external: Update commit IDs: glslang, tools, header

mikew-lunarg:
Update to latest commit IDs.


chrisforbes:
What about android?

mikew-lunarg:
I leave those to Cody, as there is coordination w/ shaderc known-good
<https://github.com/google/shaderc/tree/known-good> that I have no control
over.
-- mew

cnorthrop:
Yes, ideally we would update the Android revisions each time known-good moves forward, but I haven't had time to figure out an automated way to do that.

We'd hoped to get desktop and Android on the same update schedule, but LunarG wants to ship top of tree with each SDK update, so we can't achieve that goal.

cnorthrop:
Thinking about this now, perhaps we should just check in the shaderc known-good json file and train our scripts to pull out the revisions we need.  Then we would still have a running log of what toolchain we're using (for reproducible toolchains while bisecting) with a brute simple way of updating it.  Part of the challenge here is just find the time to manually update and test the new toolchain.  It isn't hard, just tedious and recurring.

PR #2040, opened 2017-08-30
Track descriptor bindings for PushDescriptorSets

### LINKED ISSUE #2000: False validation errors for VK_KHR_push_descriptor

mark-lunarg:
Fixes #2000.

Hook up CmdPushDescriptorSets in core validation so that those bindings can be accounted for. This PR includes commits from two other PRs ( #1999 and #2001) as they directly relate.  This is the  newer version of PR #2011.


layers/core_validation.cpp
```diff
@@ -5558,24 +5571,55 @@ VKAPI_ATTR void VKAPI_CALL CmdBindDescriptorSets(VkCommandBuffer commandBuffer,
...
+        bindings[i].descriptorType = pDescriptorWrites[i].descriptorType;
+        bindings[i].stageFlags = 0;
+        bindings[i].pImmutableSamplers = nullptr;
     }
+
+    const VkDescriptorSetLayout desc_set_layout = 0;
+    auto const shared_ds_layout = std::make_shared<cvdescriptorset::DescriptorSetLayout>(&layout_create_info, desc_set_layout);
+    auto new_desc = new cvdescriptorset::DescriptorSet(0, 0, shared_ds_layout, device_data);
+    new_desc->SetPushDescriptor();
+    cb_state->lastBound->push_descriptors[set] = new_desc;
```
jozefkucia: I think this should read `cb_state->lastBound[pipelineBindPoint].push_descriptors[set] = new_desc;`

Otherwise, it works for my use case. Thanks!

mark-lunarg: Thanks @jozefkucia, updated and reordered the commits properly.

PR #2043, opened 2017-09-01
layers: Remove remnants of old perf warning for disturbing sets

chrisforbes:
Half of this was removed, but the surrounding plumbing was left in
place. If we want to quietly add some 'was disturbed' bits so we can
produce a better error at a later draw, that should go on the
record-side of this validation, not here.

Removes the other case that the previous change missed, along with the
vestigial bits of the previous check.

Remove subtest and surrounding scaffolding that tested this


cdwfs:
LGTM if the tests still pass

chrisforbes:
Green on our farm.

PR #2046, opened 2017-09-02
layers: Fix leaking backing sets for push descriptors

chrisforbes:
We'd previously leak any temporary descriptor set left bound to the
pipeline at the end of the command buffer. Rearrange things so we can
use unique_ptr and assure it's always cleaned up correctly.


chrisforbes:
Green on our farm.

layers/buffer_validation.h
```diff
@@ -22,6 +22,7 @@
 
 #include "core_validation_types.h"
 #include "core_validation_error_enums.h"
+#include "descriptor_sets.h"
```
tobine: Is this include needed? Unclear from quick review of code and no note of it in commit msg.

chrisforbes: It's needed because `std::unique_ptr<cvdescriptorset::DescriptorSet>::~unique_ptr` ends up being instantiated, which needs `cvdescriptorset::DescriptorSet` to be complete.

mark-lunarg:
Thanks @chrisforbes, had to run out of town on short notice and didn't want the fix to languish.

PR #2049, opened 2017-09-04
Tobin synch model cleanup and 2nd cbs

tobine:
Replaces #2026 

This includes all of the previous memory race condition checking but greatly improves performance by splitting per-cmd-buffer memory access map into read and write maps. This allows skipping large groups of checks when there's no possibility of conflicts.

I'm still seeing about 15% perf hit on smoketest, which is much better than ~2.5x previously. Nearly all of the additional hit comes from grabbing the memory accesses from the descriptors at draw time. I'll look at speeding that up as well.

If the 15% perf hit and the experimental warning conditions are deemed acceptable, I could land this near it's present form. Main goal at this point is to land some initial code, though, so that these core changes are in place moving forward. To get this in with minimal impact, though, for now I can turn off the warning and disable the call mentioned above that's eating a bit of perf. I think I favor this approach in the short term so let me know if there are other opinions.


tobine:
I pushed a follow-on CL that just shuts off the callback and tests for now, along with disabling Draw-time checks that have ~15% perf impact on smoketest. This is intended to make it safer to land the initial version so that further testing/development can be done before turning on the callback by default.

layers/core_validation.cpp
```diff
@@ -8232,6 +8234,10 @@ VKAPI_ATTR void VKAPI_CALL CmdBeginRenderPass(VkCommandBuffer commandBuffer, con
                     cb_node->queue_submit_functions.push_back(function);
                 }
             }
+            const auto &subpass_desc = render_pass_state->createInfo.pSubpasses[0];
+            for (uint32_t i = 0; i < subpass_desc.inputAttachmentCount; ++i) {
+                // Mark input attachments as read
```
chrisforbes: What's going on here?

tobine: Started something that I then delayed for later and didn't kill this loop. Just pushed update to kill it. Will add MemAccess tracking for renderpass attachments in the future.

tobine:
Replaced by #2066

PR #2052, opened 2017-09-06
layers: add support for VK_AMD_mixed_attachment_samples

MaciejJesionowskiAMD:
This change removes a few validation errors reported when VK_AMD_mixed_attachment_samples is used and adds two new VU messages.


layers/core_validation.cpp
```diff
@@ -1450,6 +1451,27 @@ static bool ValidatePipelineUnlocked(layer_data *dev_data, std::vector<std::uniq
...
+            if (subpass_desc->pColorAttachments[i].attachment != VK_ATTACHMENT_UNUSED) {
+                max_sample_count = std::max(max_sample_count, pPipeline->render_pass_ci.pAttachments[subpass_desc->pColorAttachments[i].attachment].samples);
+            }
+        }
+        if (subpass_desc->pDepthStencilAttachment && subpass_desc->pDepthStencilAttachment->attachment != VK_ATTACHMENT_UNUSED) {
+            max_sample_count = std::max(max_sample_count, pPipeline->render_pass_ci.pAttachments[subpass_desc->pDepthStencilAttachment->attachment].samples);
+        }
+        if (pPipeline->graphicsPipelineCI.pMultisampleState->rasterizationSamples != max_sample_count) {
+            skip |=
+                log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_PIPELINE_EXT,
```
tobine: Would be nice to pretty-print the sample bits value as a string using helper function "string_VkSampleCountFlagBits" as with the check below.

layers/core_validation.cpp
```diff
@@ -1450,6 +1451,27 @@ static bool ValidatePipelineUnlocked(layer_data *dev_data, std::vector<std::uniq
         }
     }
 
+    if (dev_data->extensions.vk_amd_mixed_attachment_samples) {
+        VkSampleCountFlagBits max_sample_count = static_cast<VkSampleCountFlagBits>(0);
+        for (uint32_t i = 0; i < subpass_desc->colorAttachmentCount; ++i) {
+            if (subpass_desc->pColorAttachments[i].attachment != VK_ATTACHMENT_UNUSED) {
+                max_sample_count = std::max(max_sample_count, pPipeline->render_pass_ci.pAttachments[subpass_desc->pColorAttachments[i].attachment].samples);
```
tobine: Could potentially store this in RENDER_PASS_STATE struct at the time CreateRenderPass is called to prevent having to loop over attachments again here. We already grab the "sample_count" var in ValidateRenderpassAttachmentUsage() so could get max there and pass that back to top-level function to be updated in RENDER_PASS_STATE struct.

This is very minor detail for this case since we only hit this code for this particular extension so certainly not required. Something to think about going fwd, though, and at some point we can potentially consolidate/remove various loops that go over these same arrays multiple times.

MaciejJesionowskiAMD: Sounds reasonable. I tried to keep the changes as localized as possible, so maybe let's revisit the idea if similar code appears in other places.

layers/core_validation.cpp
MaciejJesionowskiAMD: Done.

MaciejJesionowskiAMD:
Updated with very basic tests. This code is probably not optimal as there's a wealth of helper objects that I'm not familiar with.

mark-lunarg:
Resolved conflicts and running through our test farm.

mark-lunarg:
Pushed to Master.

mark-lunarg:
Thanks a ton, @MaciejJesionowskiAMD.  Wish everybody did this kind of stuff!

PR #2054, opened 2017-09-06
demos: Add support for headless XCB

abergmeier:
For headless XCB we cannot create a surface.
Simply ignore XCB if connection could not be established.

### DIFF
--- a/demos/vulkaninfo.c
+++ b/demos/vulkaninfo.c
@@ -930,9 +930,9 @@ static void AppCreateXcbWindow(struct AppInstance *inst) {
     inst->xcb_connection = xcb_connect(NULL, &scr);
     int conn_error = xcb_connection_has_error(inst->xcb_connection);
     if (conn_error) {
-        fprintf(stderr, "XCB failed to connect to the X server due to error:%d.\nExiting ...\n", conn_error);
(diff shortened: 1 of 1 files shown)


demos/vulkaninfo.c
```diff
@@ -930,14 +930,15 @@ static void AppCreateXcbWindow(struct AppInstance *inst) {
     if (inst->xcb_connection == NULL) {
         printf("XCB failed to connect to the X server.\nExiting ...\n");
```
chrisforbes: Should fix this error message too, now that we're no longer exiting.

demos/vulkaninfo.c
```diff
@@ -930,14 +930,15 @@ static void AppCreateXcbWindow(struct AppInstance *inst) {
...
         fflush(stdout);
-        exit(1);
+        return;
     }
 
     int conn_error = xcb_connection_has_error(inst->xcb_connection);
     if (conn_error) {
         printf("XCB failed to connect to the X server due to error:%d.\nExiting ...\n", conn_error);
```
chrisforbes: ditto

abergmeier:
Rebase on top of https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/2055 and fixed messages.

mark-lunarg:
Pushed to Master -- thanks!

PR #2055, opened 2017-09-06
demos: Print XCB errors to stderr

abergmeier:
Printing to stdout is weird for error messages.

### DIFF
--- a/demos/vulkaninfo.c
+++ b/demos/vulkaninfo.c
@@ -928,14 +928,14 @@ static void AppCreateXcbWindow(struct AppInstance *inst) {
 
     inst->xcb_connection = xcb_connect(NULL, &scr);
     if (inst->xcb_connection == NULL) {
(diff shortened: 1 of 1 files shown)


demos/vulkaninfo.c
```diff
@@ -928,15 +928,15 @@ static void AppCreateXcbWindow(struct AppInstance *inst) {
 
     inst->xcb_connection = xcb_connect(NULL, &scr);
     if (inst->xcb_connection == NULL) {
```
polarina: This conditional is unnecessary. The function `xcb_connect` is guaranteed to return a non-NULL address, even when dynamic memory allocation fails.

abergmeier:
Should be fine to merge.

mark-lunarg:
Squashed and pushed to Master.  Thanks for the improvements!

PR #2059, opened 2017-09-06
Tobin rp compat refactor

tobine:
We had two separate functions to validate render pass compatibility. This series converges on a single function.


layers/core_validation.cpp
```diff
@@ -7390,7 +7400,7 @@ static bool PreCallValidateCreateFramebuffer(layer_data *dev_data, const VkFrame
 static void PostCallRecordCreateFramebuffer(layer_data *dev_data, const VkFramebufferCreateInfo *pCreateInfo, VkFramebuffer fb) {
     // Shadow create info and store in map
     std::unique_ptr<FRAMEBUFFER_STATE> fb_state(
-        new FRAMEBUFFER_STATE(fb, pCreateInfo, dev_data->renderPassMap[pCreateInfo->renderPass]->createInfo.ptr()));
+        new FRAMEBUFFER_STATE(fb, pCreateInfo, dev_data->renderPassMap[pCreateInfo->renderPass].get()));
```
chrisforbes: This should probably use `GetRenderPassState`

layers/core_validation_types.h
```diff
@@ -763,10 +763,10 @@ class FRAMEBUFFER_STATE : public BASE_NODE {
 public:
     VkFramebuffer framebuffer;
     safe_VkFramebufferCreateInfo createInfo;
-    safe_VkRenderPassCreateInfo renderPassCreateInfo;
+    const RENDER_PASS_STATE *rp_state;
```
chrisforbes: :-1:, Reintroduces a lifetime handling bug. The framebuffer can outlive the renderpass it was created from. From the spec, section 2.3.1:

```
A VkRenderPass object passed as a parameter to create another object is not further accessed by that object after the duration of the command it is passed into
```

layers/core_validation.cpp
```diff
@@ -826,6 +826,138 @@ static bool attachment_references_compatible(const uint32_t index, const VkAttac
...
+    }
+    if (primaryPassCI.pAttachments[primary_attach].format != secondaryPassCI.pAttachments[secondary_attach].format) {
+        skip |= logInvalidAttachmentMessage(dev_data, type1_string, rp1_state, type2_string, rp2_state, primary_attach,
+                                            secondary_attach, "They have different formats.", caller, error_code);
+    }
+    if (primaryPassCI.pAttachments[primary_attach].samples != secondaryPassCI.pAttachments[secondary_attach].samples) {
+        skip |= logInvalidAttachmentMessage(dev_data, type1_string, rp1_state, type2_string, rp2_state, primary_attach,
+                                            secondary_attach, "They have different samples.", caller, error_code);
+    }
+    if (is_multi && primaryPassCI.pAttachments[primary_attach].flags != secondaryPassCI.pAttachments[secondary_attach].flags) {
```
chrisforbes: `is_multi` should go away throughout. That spec language was removed some time ago. Renderpass compatibility is no longer defined differently for multiple subpasses vs a single subpass

chrisforbes:
As discussed offline, some details in the shared_ptr usage, but this looks much better :+1:

PR #2065, opened 2017-09-08
scripts: Check toolchain revisions and clear the cache if needed

cnorthrop:
This script compares tracked revisions against those loaded from Travis-CI cache and clears it if any don't match.

There's something not quite right about the cache being restored.  This clearly became a problem after the Android revisions were updated in #2062 and the Travis-CI build of desktop toolchain couldn't find clang (failed build [log](https://travis-ci.org/KhronosGroup/Vulkan-LoaderAndValidationLayers/jobs/273004378)).

This is a bigger hammer than we'd like, but should be effective.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,9 @@ before_install:
   - if [[ "$VULKAN_BUILD_TARGET" == "ANDROID" ]]; then export JAVA_HOME="/usr/lib/jvm/java-8-oracle"; fi
   - if [[ "$VULKAN_BUILD_TARGET" == "ANDROID" ]]; then export PATH="$ANDROID_NDK_HOME:$PATH"; fi
 
+  # Clear the toolchain contents if revisions have changed
+  - ./scripts/check_toolchain_revisions.sh
+
 script:
   - if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then ./update_external_sources.sh; fi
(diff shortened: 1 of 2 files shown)


mikew-lunarg:
LGTM but I dont have the travis-fu to feel comfortable giving +1

I assume something similar will be needed for AppVeyor?

karl-lunarg:
There's nothing Travis-specific about this script.  Someone can run it in their own workspace and if it doesn't zap the external directory, they know that they can skip running the update script, assuming that they had done so previously.

I'd suggest changing the name of the script and the comments to remove references to Travis, and then we can have the AppVeyor CI call it as well.

karl-lunarg:
Oh, and I guess some of the comments and code would need to be tweaked to get rid of the notion of "cache".  Maybe "current"?

E.g.:
```
function check_revision()
{
  echo Checking cached revision for $1 in $2
  if [ -d $2/.git ]; then
    cache_rev=$(git --git-dir=$2/.git rev-parse HEAD);
  fi
  echo cache_rev for $1 is $cache_rev;
```
goes to:

```
function check_revision()
{
  echo Checking current revision for $1 in $2
  if [ -d $2/.git ]; then
    current_rev=$(git --git-dir=$2/.git rev-parse HEAD);
  fi
  echo current_rev for $1 is $current_rev;
```

cnorthrop:
Okay, I renamed the script and removed mentions of cache (except the branch name) so it is generalized for local use.

cnorthrop:
cherry-picked to master

PR #2067, opened 2017-09-13
layers: Push descriptors validation improvements

jozefkucia:


layers/core_validation.cpp
```diff
@@ -5659,6 +5659,7 @@ static void PreCallRecordCmdPushDescriptorSetKHR(layer_data *device_data, VkComm
     new_desc->SetPushDescriptor();
     cb_state->lastBound[pipelineBindPoint].boundDescriptorSets[set] = new_desc.get();
     cb_state->lastBound[pipelineBindPoint].push_descriptors[set] = std::move(new_desc);
+    delete [] bindings;
```
chrisforbes: In order of preference:

1/ Don't cook up DS layouts on the fly. This whole thing is crazy.
2/ Make bindings be std::vector<VkDescriptorSetLayoutBinding>.
3/ This

layers/descriptor_sets.h
```diff
@@ -383,8 +384,7 @@ class DescriptorSet : public BASE_NODE {
     };
     // Return true if any part of set has ever been updated
     bool IsUpdated() const { return some_update_; };
-    bool IsPushDescriptor() const { return push_descriptor_; };
-    void SetPushDescriptor() { push_descriptor_ = true;  };
+    bool IsPushDescriptor() const { return p_layout_->GetCreateFlags() & VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR; };
```
chrisforbes: This second commit is great!

layers/core_validation.cpp
tobine: The delete fixes outstanding issue so that's better, but agree that we should redo code next to at least kill new/delete.

tobine:
This LGTM. I'm testing now and will approve/merge when verified clean.

layers/core_validation.cpp
tobine: @chrisforbes #2068 does option 1

tobine:
merged

tobine:
Thanks @jozefkucia !

PR #2072, opened 2017-09-18
layers:Pipe state owns render pass ptr

### LINKED ISSUE #2070: Core validation crashes accessing renderpass info after destruction
In `ValidatePipelineDrawtimeState` in [core_validation.cpp](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/core_validation.cpp#L1098-L1103) there's code to check renderpass compatibillity between the currently active renderpass and the renderpass used to create the pipeline, if they're not the same:

```cpp
        if (pCB->activeRenderPass->renderPass != pPipeline->graphicsPipelineCI.renderPass) {
            // renderPass that PSO was created with must be compatible with active renderPass that PSO is being used with
            skip |= validateRenderPassCompatibility(dev_data, "active render pass", pCB->activeRenderPass, "pipeline state object",
                                                    GetRenderPassState(dev_data, pPipeline->graphicsPipelineCI.renderPass), caller,
                                                    rp_error);
        }
```

But this crashes if `GetRenderPassState` returns `NULL`, which happens if the renderpass used to create the pipeline has been destroyed. E.g. if a renderpass is created purely for the sake of creating the pipeline then destroyed immediately afterwards.

That is legal to do and was clarified [In a Vulkan-Docs issue](https://github.com/KhronosGroup/Vulkan-Docs/issues/516) and is now spelled out in the spec Section 2.3.1:

> A VkRenderPass object passed as a parameter to create another object is not further accessed by that object after the duration of the command it is passed into. 

The diff below to cube.c will reproduce the crash if you run with --validate. Remove the added `vkDestroyRenderPass` call to hide the issue by leaving the renderpass alive.

```diff
diff --git a/demos/cube.c b/demos/cube.c
index 3d79b0e..9b29f6f 100644
--- a/demos/cube.c
+++ b/demos/cube.c
@@ -423,7 +423,7 @@ struct demo {
     VkPipelineLayout pipeline_layout;
     VkDescriptorSetLayout desc_layout;
     VkPipelineCache pipelineCache;
-    VkRenderPass render_pass;
+    VkRenderPass render_pass, render_pass_4_pipe;
     VkPipeline pipeline;

     mat4x4 projection_matrix;
@@ -1922,6 +1922,9 @@ static void demo_prepare_render_pass(struct demo *demo) {

     err = vkCreateRenderPass(demo->device, &rp_info, NULL, &demo->render_pass);
     assert(!err);
+
+    err = vkCreateRenderPass(demo->device, &rp_info, NULL, &demo->render_pass_4_pipe);
+    assert(!err);
 }

 //TODO: Merge shader reading
@@ -2137,12 +2140,14 @@ static void demo_prepare_pipeline(struct demo *demo) {
     pipeline.renderPass = demo->render_pass;
     pipeline.pDynamicState = &dynamicState;

-    pipeline.renderPass = demo->render_pass;
+    pipeline.renderPass = demo->render_pass_4_pipe;

     err = vkCreateGraphicsPipelines(demo->device, demo->pipelineCache, 1,
                                     &pipeline, NULL, &demo->pipeline);
     assert(!err);

+    vkDestroyRenderPass(demo->device, demo->render_pass_4_pipe, NULL);
+
     vkDestroyShaderModule(demo->device, demo->frag_shader_module, NULL);
     vkDestroyShaderModule(demo->device, demo->vert_shader_module, NULL);
 }
```

tobine:
Fixes #2070 

Pipeline can outlive the renderPass that it's created with so add
shared_ptr to RENDER_PASS_STATE to PIPELINE_STATE to make sure
renderPass is available for the life of the pipeline.

### DIFF
--- a/layers/core_validation_types.h
+++ b/layers/core_validation_types.h
@@ -547,6 +547,8 @@ class PIPELINE_STATE : public BASE_NODE {
    public:
     VkPipeline pipeline;
     safe_VkGraphicsPipelineCreateInfo graphicsPipelineCI;
+    // Hold shared ptr to RP in case RP itself is destroyed
+    std::shared_ptr<RENDER_PASS_STATE> rp_state;
     safe_VkComputePipelineCreateInfo computePipelineCI;
     // Flag of which shader stages are active for this pipeline
     uint32_t active_shaders;
@@ -565,6 +567,7 @@ class PIPELINE_STATE : public BASE_NODE {
     PIPELINE_STATE()
         : pipeline{},
           graphicsPipelineCI{},
+          rp_state(nullptr),
           computePipelineCI{},
           active_shaders(0),
           duplicate_shaders(0),
@@ -575,7 +578,7 @@ class PIPELINE_STATE : public BASE_NODE {
           render_pass_ci(),
           pipeline_layout() {}
 
-    void initGraphicsPipeline(const VkGraphicsPipelineCreateInfo *pCreateInfo) {
+    void initGraphicsPipeline(const VkGraphicsPipelineCreateInfo *pCreateInfo, std::shared_ptr<RENDER_PASS_STATE> &&rpstate) {
         graphicsPipelineCI.initialize(pCreateInfo);
         // Make sure compute pipeline is null
         VkComputePipelineCreateInfo emptyComputeCI = {};
@@ -599,6 +602,7 @@ class PIPELINE_STATE : public BASE_NODE {
                                                                                      pCBCI->pAttachments + pCBCI->attachmentCount);
             }
         }
+        rp_state = rpstate;
     }
 
     void initComputePipeline(const VkComputePipelineCreateInfo *pCreateInfo) {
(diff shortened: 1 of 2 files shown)


layers/core_validation_types.h
```diff
@@ -599,6 +602,7 @@ class PIPELINE_STATE : public BASE_NODE {
                                                                                      pCBCI->pAttachments + pCBCI->attachmentCount);
             }
         }
+        rp_state = rpstate;
```
tobine: @chrisforbes is this ok or should I move it? I'm thinking since param is rvalue-ref that assignment is ok but wanted to make sure.

chrisforbes: This is fine.

chrisforbes:
Needs test as this is a bit subtle -- I thought we had a case that covered this at one point?

tobine:
Yeah, seems like we had a test but I don't see anything except for FB/RP lifetime case. I did test this change with Baldur's recommended edit to cube and verified that it crashed before the change and works after the change.

I'll create an LVL test to keep it from breaking going fwd.

PR #2073, opened 2017-09-18
layers: Replace searchs with device_profile_layer

jeremyk-lunarg:
Takes advantage of the device_profile_layer to replace searches in tests.

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -114,6 +114,20 @@ VkFormat FindSupportedDepthStencilFormat(VkPhysicalDevice phy) {
     return (VkFormat)0;
 }
 
+VkFormat static GetSupportedDepthStencilFormat(
+    VkPhysicalDevice phy, PFN_vkSetPhysicalDeviceFormatPropertiesEXT &fpvkSetPhysicalDeviceFormatPropertiesEXT,
+    PFN_vkGetOriginalPhysicalDeviceFormatPropertiesEXT &fpvkGetOriginalPhysicalDeviceFormatPropertiesEXT) {
+    VkFormat format = VK_FORMAT_D24_UNORM_S8_UINT;
+
+    VkFormatProperties formatProps;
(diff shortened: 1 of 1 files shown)


jeremyk-lunarg:
@chrisforbes, does this look better and are there any other changes you would like me to make?

tests/layer_validation_tests.cpp
```diff
@@ -13045,11 +13062,19 @@ TEST_F(VkLayerTest, InvalidImageLayout) {
...
+    PFN_vkSetPhysicalDeviceFormatPropertiesEXT fpvkSetPhysicalDeviceFormatPropertiesEXT = nullptr;
+    PFN_vkGetOriginalPhysicalDeviceFormatPropertiesEXT fpvkGetOriginalPhysicalDeviceFormatPropertiesEXT = nullptr;
+
+    // Load required functions
+    if (!LoadDeviceProfileLayer(fpvkSetPhysicalDeviceFormatPropertiesEXT, fpvkGetOriginalPhysicalDeviceFormatPropertiesEXT)) {
+        printf("             device_profile_layer functions not found. Skipped.\n");
         return;
     }
+
+    depth_format = GetSupportedDepthStencilFormat(gpu(), fpvkSetPhysicalDeviceFormatPropertiesEXT, fpvkGetOriginalPhysicalDeviceFormatPropertiesEXT);
```
chrisforbes: This is broken. A few lines later when you create the image, you end up passing VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT down to the driver for this format. At that point you have undefined behavior.

jeremyk-lunarg:
@chrisforbes, I fixed it, but it looks like all that this changes now is adding the function GetSupportedDepthStencilFormat which is never used.

chrisforbes:
Closing for now, then. Not opposed to the general idea if you find cases where it simplifies /and/ is safe.

PR #2077, opened 2017-09-19
tests:Add DestroyPipelineRenderPass

### LINKED ISSUE #2074: Write test to verify no pipeline state / renderpass lifetime issues
See #2070 
Need an LVL test that uses a pipeline after its renderPass has been destroyed to verify no issues in validation.

tobine:
Fixes #2074 

Add test DestroyPipelineRenderPass. This is a positive test that
destroys the renderPass used to create a gfx pipeline and then uses
that pipeline to draw. This makes sure that validation doesn't depend
on the pipeline's renderPass to still be active at the time the
pipeline is used.

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -6363,6 +6363,97 @@ TEST_F(VkLayerTest, InvalidCmdBufferPipelineDestroyed) {
     vkDestroyPipelineLayout(m_device->device(), pipeline_layout, NULL);
 }
 
+TEST_F(VkPositiveLayerTest, DestroyPipelineRenderPass) {
+    TEST_DESCRIPTION("Draw using a pipeline whose create renderPass has been destroyed.");
+    m_errorMonitor->ExpectSuccess();
+    ASSERT_NO_FATAL_FAILURE(Init());
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+
+    VkResult err;
+
+    // Create a renderPass that's compatible with Draw-time renderPass
+    VkAttachmentDescription att = {};
+    att.format = m_render_target_fmt;
+    att.samples = VK_SAMPLE_COUNT_1_BIT;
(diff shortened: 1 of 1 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -6363,6 +6363,149 @@ TEST_F(VkLayerTest, InvalidCmdBufferPipelineDestroyed) {
...
+    err = vkCreateRenderPass(device(), &rp_info, NULL, &rp);
+    ASSERT_VK_SUCCESS(err);
+
+    VkGraphicsPipelineCreateInfo gp_ci = {};
+    gp_ci.sType = VK_STRUCTURE_TYPE_GRAPHICS_PIPELINE_CREATE_INFO;
+    gp_ci.stageCount = 2;
+    gp_ci.pStages = shaderStages;
+    gp_ci.pVertexInputState = &vi_ci;
+    gp_ci.pInputAssemblyState = &ia_ci;
+    gp_ci.pViewportState = &vp_state_ci;
```
chrisforbes: VkPipelineObj?

tests/layer_validation_tests.cpp
```diff
@@ -6363,6 +6363,149 @@ TEST_F(VkLayerTest, InvalidCmdBufferPipelineDestroyed) {
...
+
+    VkRenderPassCreateInfo rp_info = {};
+    rp_info.sType = VK_STRUCTURE_TYPE_RENDER_PASS_CREATE_INFO;
+    rp_info.attachmentCount = 1;
+    rp_info.pAttachments = &att;
+    rp_info.subpassCount = 1;
+    rp_info.pSubpasses = &subpass;
+
+    VkRenderPass rp;
+    err = vkCreateRenderPass(device(), &rp_info, NULL, &rp);
```
chrisforbes: Assume the intent here is to be compatible with the framework's renderpass.

tobine:
Updated to use VkPipelineObj based on review from @chrisforbes . Thanks!

PR #2081, opened 2017-09-20
Merge 1.0.61

chrisforbes:


chrisforbes:
@lenny-lunarg Is there some reason this isn't in master already? Looks clean so far on our farm. I have some other changes I want to land soon which depend on 1.0.61.

lenny-lunarg:
The reason I didn't stick it on master is that @mikes-lunarg usually pushes this stuff back to master after we release an SDK from private material and since he's been at the F2F, I hadn't wanted to push it without first checking with him. There shouldn't be any reason we can't push it to master &mdash; I had just intended to check with Mike before I did it.

We can get him to make sure it's fine for master. The one other thing I need to do is that I'm going to cherry-pick another commit over to the SDK branch and tag that as 1.0.61.1. It's a fix to `vulkan.hpp` which we're going to release a 1.0.61.1 SDK for shortly. Not having that fix won't brake the repo, bit if other people try to use `vulkan.hpp` from the repo. it'll give them trouble.

lenny-lunarg:
Alright, so this might be a little complicated.

You can merge this branch in the way you have it now if you're in a hurry to get this stuff in, but that wouldn't be my first choice. Merging this will not break the repo, but if people are using `vulkan.hpp` from this repo, then the change may break their code. I have a fix for the `vulkan.hpp` problem on the sdk-1.0.61 branch, but the fix requires Visual Studio 2015 on Windows (because 2013 doesn't support `constexpr`). This means that if you pull in the latest commit, our test farm will start failing for the repo and yours might, too. @cnorthrop, what version of Visual Studio are you using for Windows testing? If it's 2013, how long would it take you to upgrade to 2015?

My preferred solution to this would be to hold off on merging in the 1.0.61 content until we can upgrade our test farm to use 2015 (which should take about a day). Then I'd want to pull over both of the commits  that are currently on the sdk-1.0.61 branch, which would break Visual Studio 2013 support.

If either you don't want to break VS2013 support, or you want to get this stuff in quickly, then I'd also be fine with just merging this PR as-is. This will leave us with a messed up `vulkan.hpp`, but it won't cause any problems for our repo.

The basic problem we've been having is that `vulkan.hpp` started using `constexpr`, but that isn't supported by VS2013. As a result, we replaced `constexpr` with `const` for the 1.0.61.0 SDK, so that it would still build with VS2013. However, since `constexpr` implies `inline`, Visual Studio decided to stop inlining these functions, and generated multiple copies of the functions (from including `vulkan.hpp` in multple cpp files). This causes linker errors because the functions were redefined in every cpp file that included `vulkan.hpp`. We could have tried to use `const` and `inline` to replace `constexpr`, but since making our own changes to `vulkan.hpp` were what caused the problem in the first place, we weren't real keen on doing that again.

Let me know if you're alright with waiting a day or two, and if this will cause problems on your test farm.

cnorthrop:
I should be able to upgrade our farm to 2015 tomorrow.

cnorthrop:
We've updated to VS2015 on our side.

lenny-lunarg:
We've updated to VS2015 on our farm as well. I've pushed 1.0.61 commits to master now. I did it by cherry-picking from the `sdk-1.0.61` branch instead of merging this PR because @mark-lunarg didn't like the merge commit.

In any case, I'm closing this PR because it's been de facto merged.

PR #2082, opened 2017-09-20
Tobin gh2080 rp unexpected errors

### LINKED ISSUE #2080: RenderPassIncompatible test have unexpected validation messages
Running `vk_layer_validation_tests` I get:

> [ RUN      ] VkLayerTest.RenderPassIncompatible
Unexpected: vkCmdBeginRenderPass(): RenderPasses incompatible between render pass w/ renderPass 0xe and framebuffer w/ renderPass 0x6 Attachment 0 is not compatible with 4294967295: The second is unused while the first is not.. The spec valid usage text states 'renderPass must be compatible with the renderPass member of the VkFramebufferCreateInfo structure specified when creating framebuffer.' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkRenderPassBeginInfo-renderPass-00904)
Unexpected: vkCmdBeginRenderPass(): RenderPasses incompatible between render pass w/ renderPass 0xe and framebuffer w/ renderPass 0x6 Attachment 0 is not compatible with 0: They have different formats.. The spec valid usage text states 'renderPass must be compatible with the renderPass member of the VkFramebufferCreateInfo structure specified when creating framebuffer.' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkRenderPassBeginInfo-renderPass-00904)
Unexpected: vkCmdDraw(): RenderPasses incompatible between active render pass w/ renderPass 0xe and pipeline state object w/ renderPass 0x6 Attachment 0 is not compatible with 0: They have different formats.. The spec valid usage text states 'The current render pass must be compatible with the renderPass member of the VkGraphicsPipelineCreateInfo structure specified when creating the VkPipeline currently bound to VK_PIPELINE_BIND_POINT_GRAPHICS.' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkCmdDraw-renderPass-00435)
[       OK ] VkLayerTest.RenderPassIncompatible (442 ms)

tobine:
Fixes #2080 

Refactor CmdBeingRenderPass() to only update state if no skipping call down.

Update RenderPassIncompatible test to hit another case and not have unexpected error.

Note that I updated validateSubpassCompatibility() function to return early if skip case is hit. That function only flags a single error ID so I felt the early return is reasonable since it won't kill any unique VUIDs. This also benefits testing, which isn't a reason to change validation, but given that no new errors are skipped I think the change is ok. In cases where a renderPass is incompatible for multiple reasons, this silences repeated callbacks giving the same VUID with a different reason. I can see a case where some users may want all of the callbacks but in general I'm hopeful that seeing the first error will cause users to examine the entire renderPass and fix it completely. Worst case is that a user only fixes the first error and then sees the next error on second run.

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -12424,7 +12424,8 @@ TEST_F(VkLayerTest, NumSamplesMismatch) {
 TEST_F(VkLayerTest, RenderPassIncompatible) {
     TEST_DESCRIPTION(
         "Hit RenderPass incompatible cases. "
-        "Initial case is drawing with an active renderpass that's "
+        "First attempt BeginRenderPass() with incompatible FrameBuffer,"
+        "then attempt to draw with an active renderpass that's "
         "not compatible with the bound pipeline state object's creation renderpass");
     VkResult err;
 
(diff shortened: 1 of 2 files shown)


krOoze:
> Note that I updated validateSubpassCompatibility() function to return early if skip case is hit. That function only flags a single error ID so I felt the early return is reasonable since it won't kill any unique VUIDs. 

I think the users are more likely to be interested in the message text. It won't kill unique IDs, but it will kill unique messages, no?

What is the reasoning behind such change? It does not seem like the ideal solution. I mean "perfect is the enemy of better"; perhaps leave a `TODO` note describing the current vs ideal behavior?

tobine:
@krOoze the early return makes testing easier for renderPass compatibility. It avoids having to flag a bunch of unexpected errors in the test.

With this change, the user will get all of the errors about why input attachments of renderPass cause an incompatibility, then if the user returns "true" to skip the call down, they will miss any potential error messages related to why color or DS attachments are incompatible. If user doesn't skip call down, they'll still see all of the error callbacks.

Not sure what ideal solution is here. I could make an argument that validation should queue up all the text for a single VUID and then make a single call. Alternatively, for a complex case like RP compatibility maybe it's better to make multiple calls for same VUID, or perhaps we need more VUIDs to make identifying the exact error simpler for the user.

Let me know if you have strong thoughts one way. I don't feel too strongly about how I've structured this update. It just makes testing easier, fixes known unexpected errors in tests, and seems reasonable. If users don't return "true" from callback they'll still see all the errors.

chrisforbes:
@tobine, can you rebase this?

tobine:
rebased to master

tests/layer_validation_tests.cpp
```diff
@@ -12426,17 +12429,20 @@ TEST_F(VkLayerTest, RenderPassIncompatible) {
     rpbi.sType = VK_STRUCTURE_TYPE_RENDER_PASS_BEGIN_INFO;
     rpbi.framebuffer = m_framebuffer;
     rpbi.renderPass = rp;
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_12000710);
     vkCmdBeginRenderPass(m_commandBuffer->handle(), &rpbi, VK_SUBPASS_CONTENTS_INLINE);
+    m_errorMonitor->VerifyFound();
+    // Now we want to bind the RenderPass to trigger Draw-time error so allow the error for this call
+    //  The better way to do this to avoid the error would be to create separate FB with compatible RP
```
chrisforbes: I think we should really do this properly.

chrisforbes: From private chat: This error is no good to hide because it exposes the driver to an incorrect call. It would be perfectly reasonable for the driver to explode or subtly corrupt itself as a result of this.

mark-lunarg:
Can/should we close this MR?

krOoze:
#2080 was closed. Is here anything worth preserving?

mark-lunarg:
@tobine, this repository will close for **write access** on **Sunday, 5/13/2018**.  If it is pushed before that time it will be present in the follow-on Vulkan-ValidationLayers repository on Monday, otherwise a new PR will be required in the new repo.

PR #2084, opened 2017-09-22
Implement KHR_image_format_list validation

chrisforbes:


chrisforbes:
Test case yet to come.

layers/buffer_validation.cpp
```diff
@@ -3288,6 +3288,16 @@ bool ValidateImageBarrierSubresourceRange(const layer_data *device_data, const I
                                          subresourceRangeErrorCodes);
 }
 
+template<typename T>
+static T const * GetExtensionStruct(void const *base_struct, VkStructureType stype) {
```
tobine: Nice, should be able to expand use of this. May want to hoist to vk_layer_utils.

layers/buffer_validation.cpp
```diff
@@ -3332,6 +3342,21 @@ bool PreCallValidateCreateImageView(layer_data *device_data, const VkImageViewCr
...
                                     validation_error_map[VALIDATION_ERROR_0ac007f4]);
                 }
+
+                else if (GetDeviceExtensions(device_data)->vk_khr_image_format_list) {
+                    auto format_list = GetExtensionStruct<VkImageFormatListCreateInfoKHR>(image_state->createInfo.ptr(),
+                                                                                          VK_STRUCTURE_TYPE_IMAGE_FORMAT_LIST_CREATE_INFO_KHR);
+                    if (format_list && format_list->viewFormatCount) {
+                        auto end = format_list->pViewFormats + format_list->viewFormatCount;
+                        if (std::find(format_list->pViewFormats, end, view_format) == end) {
+                            skip |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0,
```
tobine: I'd include the image as the object here.

chrisforbes:
This is no good. The codegen for `safe_*` doesn't do the right thing with `pNext`. Will have to fix that first, and then this will be the right thing.

mark-lunarg:
@jzulauf-lunarg, was that you that was taking a look at this pNext issue?

layers/buffer_validation.cpp
jzulauf-lunarg: I've already written this template... 

 lvl_find_in_chain in vk_typemap_helper.h, which doesn't need an stype arge, but pulls the stype from a codegen'd traits for all T with pNext.

The traits don't exist for the safe structs, but could, or an override with an explicit sType could be added, or a SFINAE version could be written to map from the "safe" to "Vk" version through another autogenerated traits.

layers/buffer_validation.cpp
```diff
@@ -3332,6 +3342,21 @@ bool PreCallValidateCreateImageView(layer_data *device_data, const VkImageViewCr
                                     __LINE__, VALIDATION_ERROR_0ac007f4, "IMAGE", "%s %s", ss.str().c_str(),
                                     validation_error_map[VALIDATION_ERROR_0ac007f4]);
                 }
+
+                else if (GetDeviceExtensions(device_data)->vk_khr_image_format_list) {
+                    auto format_list = GetExtensionStruct<VkImageFormatListCreateInfoKHR>(image_state->createInfo.ptr(),
+                                                                                          VK_STRUCTURE_TYPE_IMAGE_FORMAT_LIST_CREATE_INFO_KHR);
```
jzulauf-lunarg: The duplication of the type/stype isn't needed with lvl_find_in_chain.

PR #2085, opened 2017-09-22
layers: Avoid potential NULL pointer dereference

jozefkucia:
The `boundDescriptorSets` vector may be resized in `PreCallRecordCmdBindDescriptorSets()`.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5629,10 +5629,6 @@ static void PreCallRecordCmdPushDescriptorSetKHR(layer_data *device_data, VkComm
     if (set >= cb_state->lastBound[pipelineBindPoint].boundDescriptorSets.size()) {
         cb_state->lastBound[pipelineBindPoint].boundDescriptorSets.resize(set + 1);
         cb_state->lastBound[pipelineBindPoint].dynamicOffsets.resize(set + 1);
-    } else {
-        if (cb_state->lastBound[pipelineBindPoint].boundDescriptorSets[set]->IsPushDescriptor()) {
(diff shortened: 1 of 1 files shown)


layers/core_validation.cpp
```diff
@@ -5629,10 +5629,9 @@ static void PreCallRecordCmdPushDescriptorSetKHR(layer_data *device_data, VkComm
     if (set >= cb_state->lastBound[pipelineBindPoint].boundDescriptorSets.size()) {
         cb_state->lastBound[pipelineBindPoint].boundDescriptorSets.resize(set + 1);
         cb_state->lastBound[pipelineBindPoint].dynamicOffsets.resize(set + 1);
-    } else {
-        if (cb_state->lastBound[pipelineBindPoint].boundDescriptorSets[set]->IsPushDescriptor()) {
-            cb_state->lastBound[pipelineBindPoint].push_descriptors[set] = nullptr;
-        }
+    } else if (cb_state->lastBound[pipelineBindPoint].boundDescriptorSets[set] &&
+               cb_state->lastBound[pipelineBindPoint].boundDescriptorSets[set]->IsPushDescriptor()) {
```
chrisforbes: Why do we even do this? We're about to clobber it a few lines later.

jozefkucia: Right, we shouldn't need this.

chrisforbes:
@jozefkucia this looks good now. If you want to continue prodding at the push descriptor support, there's plenty more to do -- you'll notice we trash the temporary descriptor set every push, and we don't bother to track which push descriptors have actually been updated, so we have no idea what's going on at draw time [and so just skip validating push descriptors at that point entirely]

PR #2091, opened 2017-09-27
layers: Implement vkCreateDescriptorPool() checks

krOoze:
Check explicit validity of `vkCreateDescriptorPool` command:
- `maxSets` > 0
- `pPoolSizes[].descriptorCount` > 0

\+ implement relevant tests


layers/parameter_validation_utils.cpp
```diff
@@ -2344,6 +2344,35 @@ bool pv_vkDebugMarkerSetObjectNameEXT(VkDevice device, const VkDebugMarkerObject
...
+                if (pCreateInfo->pPoolSizes[i].descriptorCount <= 0) {
+                    skip |= log_msg(
+                        device_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DESCRIPTOR_POOL_EXT,
+                        VK_NULL_HANDLE, __LINE__, VALIDATION_ERROR_04a0025c, LayerName,
+                        "vkCreateDescriptorPool(): pCreateInfo->pPoolSizes[%" PRIu32 "].descriptorCount is not greater than 0. %s",
+                        i, validation_error_map[VALIDATION_ERROR_04a0025c]);
+                }
+            }
+        }
+    }
```
tobine: The "else" case here is VALIDATION_ERROR_1f611e01

krOoze:
^ ups, forgot final Google Style uglyfication. Re-pushed.

layers/parameter_validation_utils.cpp
krOoze: Should be checked by the autogenerated part.
The purpose of the `if` here is only to protect against null dereference.

layers/parameter_validation_utils.cpp
```diff
@@ -2378,7 +2407,7 @@ VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vkGetPhysicalDeviceProcAddr(VkInstance
 
 // If additional validation is needed outside of the generated checks, a manual routine can be added to this file
 // and the address filled in here. The autogenerated source will call these routines if the pointers are not NULL.
-void InitializeManualParameterValidationFunctionPointers(void) {
+void InitializeManualParameterValidationFunctionPointers() {
```
krOoze: Unrelated, but I can never resist killing random C-isms.

mark-lunarg:
Thanks @krOoze!

PR #2095, opened 2017-09-29
Android toolchain single ABI

cnorthrop:
This series:
* Updates the Android toolchain script to specify a single ABI (default is still all)
* Specifies a toolchain ABI in the Travis-CI config file
* Adds a 64-bit Android job to run in parallel

End result is a much faster Android toolchain build (~3m30s -> ~42s for arm32 locally), and better coverage in our public CI.


cnorthrop:

PR #2102, opened 2017-10-04
fix double renderpass deletion

chrisforbes:


chrisforbes:
:-1: This is no good. objtrack is confused here -- it thinks the renderpass was destroyed in the earlier call.

chrisforbes:
@tobine, @mark-lunarg this is another case of objtrack not composing well with CV. objtrack has no way of knowing that CV rejected the DestroyRenderPass call.

PR #2103, opened 2017-10-04
External semaphore compatibility fixes

mikes-lunarg:
Track when a semaphore payload is imported and restored, turn off
semaphore validation when using an external payload.

VK_KHR_external_semaphore_fd entry points and test cases to come in a follow-on patch

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -181,6 +181,7 @@ struct layer_data {
     PHYS_DEV_PROPERTIES_NODE phys_dev_properties = {};
     VkPhysicalDeviceMemoryProperties phys_dev_mem_props = {};
     VkPhysicalDeviceProperties phys_dev_props = {};
+    bool external_sync_warning = false;
 };
 
 // TODO : Do we need to guard access to layer_data_map w/ lock?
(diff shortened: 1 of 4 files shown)


chrisforbes:
Hmmn, this one is going to be fun. Give me some time to wrap my head around all the details here.

chrisforbes:
@mikes-lunarg do you want to split out the two refactoring commits into their own pull request and land those, while we're still figuring out what to do with the actual external fence stuff?

mikes-lunarg:
Rebased and added suggestion from @chrisforbes about retiring work early when an external semaphore is signaled

layers/core_validation.cpp
```diff
@@ -2692,8 +2693,18 @@ static void PostCallRecordQueueSubmit(layer_data *dev_data, VkQueue queue, uint3
                     pSemaphore->in_use.fetch_add(1);
                     semaphore_signals.push_back(semaphore);
                 } else {
-                    semaphore_externals.push_back(semaphore);
-                    pSemaphore->in_use.fetch_add(1);
+                    // Retire work up until this submit early, we will not see the wait that corresponds to this signal
+                    early_retire_seq = pQueue->seq + pQueue->submissions.size() + 1;
+                    static bool early_retire_warning = false;
```
chrisforbes: Make this device state.

mikes-lunarg: Moved the warning bool into dev_data

layers/core_validation.cpp
```diff
@@ -2715,6 +2726,10 @@ static void PostCallRecordQueueSubmit(layer_data *dev_data, VkQueue queue, uint3
                                          submit_idx == submitCount - 1 ? fence : VK_NULL_HANDLE);
     }
 
+    if (early_retire_seq) {
+        RetireWorkOnQueue(dev_data, pQueue, early_retire_seq);
```
chrisforbes: Have you thought about whether we need to also call `VerifyQueueStateToSeq` before we do the retirement here? [As a bigger picture thing, I'm not sure the remaining error produced by  `VerifyQueueStateToSeq` is any good in general...]

chrisforbes: I think this is pretty much OK not calling that function, but something to think about.

PR #2109, opened 2017-10-05
Fix doc validator on Windows

mark-lunarg:
vk.xml now includes some wacky unicode characters which end up in the vk_validation_error_database.txt and cause the doc validator to choke.  Added UTF encoding to fix that, and added python3 to the validator script to fix _that_.


mark-lunarg:
@cnorthrop, does this look OK on the Gfarm?

cnorthrop:
/sigh

This has exposed a problem on our side with how we're testing the docs.  It used to work... now I'm getting:
```
[ SKIPPED  ] ./vkvalidatelayerdoc.sh
  To run validation DB checks you can manually execute
  vk_validation_stats.py from the 'scripts' dir of your source tree
```
So go ahead with this PR and I'll get it fixed up soon.

PR #2113, opened 2017-10-06
Loader: Fix OOM in loaderGetDeviceRegistryFiles()

iostrowsINTEL:
loaderGetDeviceRegistryEntry() out of memory result was not handled properly.


iostrowsINTEL:
This should be marked with `bug` label

PR #2115, opened 2017-10-07
demos: Update cube.cpp DISPLAY_KHR code

laanwj:
The physical display rendering code didn't compile due to drift in `vulcan.hpp`. Update for these changes.

Tested on i.MX8M / Vivante GC7000L.

(same change is needed in https://github.com/LunarG/VulkanSamples, originally noticed it there)


laanwj:
I pushed an analogous change to `cube.c`.

mark-lunarg:
Nice catch!  Thanks a bunch, and keep the fixes coming.

PR #2116, opened 2017-10-09
Version 62 Vulkan Header Update

mark-lunarg:
Had to tweak the spec script, but pretty painless OTT.


scripts/spec.py
```diff
@@ -139,6 +139,8 @@ def parseJSON(self):
                             just_txt = just_txt.get_text().strip()
                             unicode_map = {
                             u"\u2019" : "'",
+                            u"\u201c" : "'",
```
mikes-lunarg: \u201c and \u201d should both be translated to double quote

mark-lunarg: Yeah, and that breaks the error database because it ends up with gems like "Error message must not have 'stuff like "_2PLANE" except on Wednesdays.' in any command" and nobody seems happy with three levels of nested quotes.

mark-lunarg: Fixed as @mikes-lunarg suggested, using original double-quotes and escaping them in the database file.  Thanks!

PR #2117, opened 2017-10-09
Remove checks for AccessMasks vs ImageLayouts

### LINKED ISSUE #1146: READ_BIT in srcAccessMask
When I read the [code](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/core_validation.cpp#L8987) we can see that for most layout requires a `READ_BIT`.

However, I think using a READ_BIT for a srcAccessMask is not useful.

Take for example the `VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL`. It means that before the barrier, the image was used in source for transfer.
However, a transfer does not perform any write to the src, thus I do not see WHY should we use a READ_BIT in this case.
If we are reasonning with flush/invalidation cache, the first half of a memory dependency is a flush (since it makes the memory available), that's it, we do not need to have a `READ_BIT`.
However, we still need to have an execution barrier with `srcStage = TRANSFER`.

Is this a bug? Or did I just miss something?

mark-lunarg:
Fixes #1146.  Removes these checks entirely.

I could find no spec language saying that these should now be warnings for HAVING them set, so that's not included in this PR.


chrisforbes:
I'm not convinced about the bigger picture here yet. What's the spec justification for enforcing particular access masks vs layouts? We have `ValidateAccessMaskPipelineStage` a couple layers of calls further up, which does what the spec requires.

mark-lunarg:
@chrisforbes, I also don't see anyplace in the spec where conditions are placed on access masks vs image layouts.  None of these checks have VUs, so maybe we should excise them entirely?

chrisforbes:
:+1: for removing this completely.

tobine:
After a quick spec review, I agree with sentiment to just axe these checks.

PR #2118, opened 2017-10-09
vulkaninfo html output

shannon-lunarg:
Update vulkaninfo to provide html output option.


mark-lunarg:
This has been squashed and pulled into a development branch, and will be pushed to Master shortly.

PR #2122, opened 2017-10-10
loader: GH2121; move private libs in vulkan.pc file

johnzupin:
GH 2121,

Specify the Private Libs to "Libs.private" in the vulkan.pc.in cmake config file.


johnzupin:
Thanks for looking this over @chrisforbes

Issue #2094, opened 2017-09-29
Clean up descriptor copy update validation


tobine:
Src probably shouldn't need to be updated. Need some more spec language and potentially new VUIDs around copy update cases, currently primarily copy/reuse write update VUIDs. Add more copy update testing as well.

PR #2132, opened 2017-10-13
layers: Allow copying nonupdated descriptors

jozefkucia:
Partially addresses  #2094

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -12340,6 +12340,39 @@ TEST_F(VkLayerTest, CopyDescriptorUpdateErrors) {
     vkDestroySampler(m_device->device(), sampler, NULL);
 }
 
+TEST_F(VkPositiveLayerTest, CopyNonupdatedDescriptors) {
+    TEST_DESCRIPTION("Copy nonupdated descriptors");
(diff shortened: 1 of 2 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -12340,6 +12340,38 @@ TEST_F(VkLayerTest, CopyDescriptorUpdateErrors) {
...
+        { 2, VK_DESCRIPTOR_TYPE_SAMPLER, 1, VK_SHADER_STAGE_ALL, nullptr },
+    });
+    OneOffDescriptorSet dst_ds(m_device->device(), {
+        { 0, VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 1, VK_SHADER_STAGE_ALL, nullptr },
+        { 1, VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE, 1, VK_SHADER_STAGE_ALL, nullptr },
+    });
+
+    m_errorMonitor->ExpectSuccess();
+
+    VkCopyDescriptorSet copy_ds_update[2];
```
tobine: Please assign "const uint32_t copy_size = 2;" or some var name, then use that var here and on lines 12362 & 12370.

PR #2141, opened 2017-10-17
external: Update commit IDs and scripts

mikew-lunarg:
Update to top-of-tree glslang, plus the spirv dependencies specified by glslang's known_good.json file.

ALSO: added a top-level .gitattributes file, to force CRLF line termination on .bat files.

Modify the LVL update_external_sources.* scripts to invoke glslang's update_glslang_sources.py script as part of the checkout process.

Enhance LVL's CMakeLists.txt for new SPIRV_TOOLS_OPT_* variables.

Add SPIRV-Tools-opt to linker.

Adjust some comments.

Change-Id: I8e782c15cade4260528ab055361208af137204fa


mikew-lunarg:
This builds and passes locally on Linux, but Windows currently fails with error in upstream update_glslang_sources.py

mikew-lunarg:
Upstream script requires python v2.7

cnorthrop:
I'm not a fan of using python to generate shell commands.  I don't think we're doing that anywhere else in this repo (or downstreams).  We've only used python to generate source files.  I also don't think we're calling any script pulled from external repos.

If the python is simply downloading a known good version of its dependencies, there are much simpler ways to do it.  You could do it with a couple lines in a shell script using [jq](https://stedolan.github.io/jq/#), which is available on all our host platforms:

$ jq '.commits[] | select(.name=="spirv-tools") | .commit' glslang/known_good.json
"99cd25c4139e0dc914ab8a5a3b75e6fed0ad1329"

$ jq '.commits[] | select(.name=="spirv-tools/external/spirv-headers") | .commit' glslang/known_good.json
"2bb92e6fe2c6aa410152fc6c63443f452acb1a65"

And then you could share that version for any other copy of SPIRV-tools, etc.

That being said, we've already discussed this offline and I think you're planning to move forward with this anyway.  I just wanted to log the alternatives here.

update_external_sources.bat
```diff
@@ -300,6 +300,7 @@ REM // ======== Functions ======== //
    cd %GLSLANG_DIR%
    git clone %GLSLANG_GITURL% .
    git checkout %GLSLANG_REVISION%
+   C:\Python27\python.exe .\update_glslang_sources.py
```
karl-lunarg: This isn't "great", but will likely work on most systems.  It might fail where people do non-standard/expected installs of Python.

It might be slightly better to ask the upstream repo to make the first line in the file:

#!/usr/bin/env python2

Not only does this "clearly" state that the file is Python 2, but it also lets you say:

.\update_glslang_sources.py

on Windows and it will pick Python 2.  This does rely on the Python launcher being there, but Python 3, which contains the launcher, is already a prerequisite.

And the above change works on Linux too.  I think that PEP-397 and PEP-394 document all this.

Probably OK to checkin as-is, but you might ask glslang (Greg) to make the above change and then update this later to remove the "C:\Python27\python.exe".

karl-lunarg:
@cnorthrop Your "jq" commands just parse the json and produce the commit IDs.  The update_glslang_sources.py script (coming from the glslang repo, and not LVL) does a lot more than generate a few shell commands.  It would seem like a pain to write the equivalent twice, in BAT and SH languages.

Maybe you need to direct your alternative suggestions to the upstream repo?

Also, adding a tool like "jq" needs to be done carefully.  I'm not against it, but it does add another prerequisite for building the repo which needs to be documented, added to CI machines, etc.

CMakeLists.txt
```diff
@@ -268,12 +281,16 @@ if (WIN32)
     set_target_properties(SPIRV-Tools PROPERTIES
                          IMPORTED_LOCATION       "${SPIRV_TOOLS_LIB}"
                          IMPORTED_LOCATION_DEBUG "${SPIRV_TOOLS_DLIB}")
+    set_target_properties(SPIRV-Tools-opt PROPERTIES
```
karl-lunarg: missing add_library() command for SPIRV-Tools-opt  ?

cnorthrop:
According to @greg-lunarg the script is just downloading known good versions of its dependencies (spirv-tools, spirv-headers).  Since we're already downloading known good versions of those tools, it would be simpler just to get on the same version using the lines I shared (or some other way, that's just one example that I thought was succinct).

Also, looking at the cmake changes, it isn't clear to me that LVL will even link against the files it just downloaded under glslang, so we probably don't even need to run the new upstream script.  Or maybe it becomes a NOP if we change the order of tool downloads.

I'll try to test this stuff out in a sandbox since the .gitattributes changes affect all platforms.

update_external_sources.bat
mikew-lunarg: Suggested to @greg-lunarg for upstream.

CMakeLists.txt
mikew-lunarg: Added add_library()

update_external_sources.bat
mikew-lunarg: @greg-lunarg @cnorthrop Please also convey this suggestion to shaderc.

mikew-lunarg:
The remaining issues have been resolved, and it is passing LunarG and GitHub CI.  Please review and approve.

cnorthrop:
No problems with Android build.

mikew-lunarg:
FWIW This passed previously, the only difference here is rebasing.
The [currently reported failure](https://ci.appveyor.com/project/Khronoswebmaster/vulkan-loaderandvalidationlayers/build/1727/messages) is actually a timeout, not a failure.

mikew-lunarg:
@mark-lunarg understood. 
@cnorthrop and @greg-lunarg  and I met to plan for a more unified dependency solution in the future.
But for now, I'd like to get this in as-is for more exposure.

PR #2142, opened 2017-10-17
Add VK_KHR_external_semaphore_fd entry points and tests

mikes-lunarg:


layers/core_validation.cpp
```diff
@@ -10950,8 +10976,10 @@ static const std::unordered_map<std::string, void*> name_to_funcptr_map = {
     {"vkDestroyDebugReportCallbackEXT", (void*)DestroyDebugReportCallbackEXT},
     {"vkDebugReportMessageEXT", (void*)DebugReportMessageEXT},
     {"vkGetPhysicalDeviceDisplayPlanePropertiesKHR", (void*)GetPhysicalDeviceDisplayPlanePropertiesKHR},
-    {"GetDisplayPlaneSupportedDisplaysKHR", (void*)GetDisplayPlaneSupportedDisplaysKHR},
-    {"GetDisplayPlaneCapabilitiesKHR", (void*)GetDisplayPlaneCapabilitiesKHR},
+    {"vkGetDisplayPlaneSupportedDisplaysKHR", (void*)GetDisplayPlaneSupportedDisplaysKHR},
```
jzulauf-lunarg: This seems like an unrelated cleanup.  Should it have it's own commit?

mikes-lunarg: It has it's own commit, but I added it to this PR because I didn't want to spin up another branch and PR for a typo fix.

PR #2147, opened 2017-10-19
External fence compatibility

mikes-lunarg:
Similar to the external semaphore changes:
- Track when fence objects are exported/imported and skip checks and state tracking while a fence is using an external payload.
- Retire all work up until a signal operation on an external fence and emit a warning about validation that will no longer occur


layers/core_validation.cpp
```diff
@@ -9803,8 +9903,16 @@ static void PostCallRecordAcquireNextImageKHR(layer_data *dev_data, VkDevice dev
                                               VkSemaphore semaphore, VkFence fence, uint32_t *pImageIndex) {
     auto pFence = GetFenceNode(dev_data, fence);
     if (pFence) {
-        pFence->state = FENCE_INFLIGHT;
-        pFence->signaler.first = VK_NULL_HANDLE;  // ANI isn't on a queue, so this can't participate in a completion proof.
+        if (GetDeviceExtensions(dev_data)->vk_khr_external_fence) {
+            // A successful call with external fences enabled acts as a temporary import
+            if (pFence->scope == kSyncScopeInternal) {
```
chrisforbes: Not sure about extending this model to internal fences passed to WSI -- we'd like to be able to prove things someday about whether images are acquired at the point the app tries to use them, and also track whether a WSI fence is inflight.

I know this is what the spec says happens now, but I think this is just making things worse for validation. Is there some detail I'm overlooking that /requires/ you to lose track of WSI fences?

mikes-lunarg: Good point, I think the constraints on temporarily imported WSI fences are well defined enough that we can treat them like signaled internal fences and track when they are inflight.

PR #2149, opened 2017-10-19
Fix for validation of copies between compressed and uncompressed images

### LINKED ISSUE #2145: vkCmdCopyImage and compressed texture destination issue
This seems to be a bug in the spec and the validation layer.

We have code to create a BC1 compressed texture from an R16G16B16A16 texture (RGBA16 for short).

Let's suppose we want to create a 12x12 BC1 compressed texture using a 3x3 RGBA16 texture.  This should work, since the compressed texture block size is 4x4.  Now we perform VkCmdCopyImage:

According to the spec, extent is always given in source image size units, so extent = 3x3.
Source subresource = 3x3.  Dest subresource = 12x12.

The validation layer will say:
vkCmdCopyImage(): pRegion[0] extent width (3) must be a multiple of the compressed texture block width (4), or when added to dstOffset.x (0) must equal the image subresource height (12).

The spec says:
"if the srcImage is compressed, then:
* If extent.width is not a multiple of the compressed texel block width, then (extent.width + dstOffset.x) must equal the image subresource width. [...]
Similarly, if the dstImage is compressed, then:
* If extent.width is not a multiple of the compressed texel block width, then (extent.width + dstOffset.x) must equal the image subresource width. [...]"

This makes no sense, since when the dstImage is compressed, the sizes (extent vs dest subresource width) are flipped, and yet the text isn't flipped.  The dest subresource size is what should be a multiple of the block width, not the extent.

Am I missing something here?

mark-lunarg:
For image-to-image copies, there are special cases for copies between compressed and uncompressed images. These modifications to copy extents are now allowed for in the validation checks.  Also added positive tests for these cases.  Fixes #2145.


tests/layer_validation_tests.cpp
```diff
@@ -20339,6 +20339,48 @@ TEST_F(VkLayerTest, InvalidCreateDescriptorPool) {
 // POSITIVE VALIDATION TESTS
 //
 // These tests do not expect to encounter ANY validation errors pass only if this is true
+
+TEST_F(VkPositiveLayerTest, UncompressedToCompressedImageCopy) {
```
Tony-LunarG: This test failed on NexusPlayer.  Do I need to add the test to the Nexus player don't run list or can the test be changed to accommodate the hardware?  It failed to create an image so I assume it was an unsupported format or feature.

layers/buffer_validation.cpp
```diff
@@ -1508,32 +1523,39 @@ bool ValidateImageCopyData(const layer_data *device_data, const debug_report_dat
                                 validation_error_map[VALIDATION_ERROR_09c0013a]);
             }
 
-            // extent width must be a multiple of block width, or extent+offset width must equal subresource width
+            // Extent width must be a multiple of block width, or extent+offset width must equal subresource width
+            VkExtent3D adj_extent = image_copy.extent;
+            if (!FormatIsCompressed(dst_state->createInfo.format)) {
+                // Copying from uncompressed source. Modify extent to be dst extent * texel block size
```
tobine: Is this supposed to say "Copying to uncompressed dst"?

mark-lunarg: Yup, thanks!

layers/buffer_validation.cpp
```diff
@@ -1508,32 +1523,39 @@ bool ValidateImageCopyData(const layer_data *device_data, const debug_report_dat
...
+                adj_extent.width *= block_size.width;
```
tobine: Should these adjustments be "/="?

tests/layer_validation_tests.cpp
mark-lunarg: @Tony-LunarG, thanks. Added a check for the formats so we won't need an exception.

mark-lunarg:
Incorporated all feedback and ran through the farm one more time.  Thanks all!

tests/layer_validation_tests.cpp
```diff
@@ -20339,6 +20339,54 @@ TEST_F(VkLayerTest, InvalidCreateDescriptorPool) {
...
+        !ImageFormatIsSupported(gpu(), VK_FORMAT_BC1_RGBA_SRGB_BLOCK, VK_IMAGE_TILING_OPTIMAL)) {
+        printf("             Required formats not supported - UncompressedToCompressedImageCopy skipped.\n");
+        return;
+    }
+
+    uncomp_color_image.Init(3, 3, 1, VK_FORMAT_R16G16B16A16_UINT, VK_IMAGE_USAGE_TRANSFER_SRC_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT);
+    comp_color_image.Init(12, 12, 1, VK_FORMAT_BC1_RGBA_SRGB_BLOCK, VK_IMAGE_USAGE_TRANSFER_SRC_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT);
+
+    ASSERT_TRUE(uncomp_color_image.initialized());
+    ASSERT_TRUE(comp_color_image.initialized());
```
tobine: This assert is failing on AMD Vega

> [ RUN      ] VkPositiveLayerTest.UncompressedToCompressedImageCopy
> S:\windows-slave\LVL_windows\build\tests\vktestbinding.cpp(504): error: Failed
> Assertion: `vkCreateImage(dev.handle(), &info, 0, &handle) == VK_SUCCESS'
> S:\windows-slave\LVL_windows\build\tests\layer_validation_tests.cpp(20360): error: Value of: comp_color_image.initialized()
>   Actual: false
> Expected: true
> [  FAILED  ] VkPositiveLayerTest.UncompressedToCompressedImageCopy (436 ms)

mark-lunarg: OK, didn't check for good usage bits.  Fix up in a moment.

PR #2155, opened 2017-10-23
tests: use only 4 sample mode in ImageSampleCounts test

chrisforbes:
Implementations are not required to support 2x or 8x modes, and some
don't. Just use what's required to work.


chrisforbes:
This clears up some unexpected errors on Marlin/Sailfish.

chrisforbes:
Verified on our farm.

PR #2156, opened 2017-10-24
tests: Do not call vkQueueBindSparse() when sparse bindings are not supported

jozefkucia:


tobine:
This LGTM. Since it's a positive test that has other useful calls I think it's fine to just skip unsupported section of the test without an output message.
Do we need a similar guard on TemporaryExternalSemaphore test? I could see that it may still work b/c vkQueueBindSparse() is void and the layers will intercept the call and perform the check regardless of whether or not sparse binding is supported. Probably still best to just skip that check on unsupported devices, to avoid any unwanted behavior (like a crash) in the driver.

jozefkucia:
Yeah, there are more tests with similar issues, e.g. SparseBindingImageBufferCreate, but in practice only this one fails. For other tests the call to a driver is usually skipped and/or VkResult is not checked.

PR #2157, opened 2017-10-24
layers: Validate push descriptor set layout count

jozefkucia:


layers/core_validation_types.h
```diff
@@ -628,15 +628,15 @@ struct LAST_BOUND_STATE {
     // Track each set that has been bound
     // Ordered bound set tracking where index is set# that given set is bound to
     std::vector<cvdescriptorset::DescriptorSet *> boundDescriptorSets;
-    std::vector<std::unique_ptr<cvdescriptorset::DescriptorSet>> push_descriptors;
+    std::unique_ptr<cvdescriptorset::DescriptorSet> push_descriptors;
```
tobine: Would like to see a rename of this to push_descriptor_set. Slightly more clear that it's a single set and that there should only be one push set.

tests/layer_validation_tests.cpp
```diff
@@ -21156,6 +21156,62 @@ TEST_F(VkPositiveLayerTest, EmptyDescriptorUpdateTest) {
...
+    err = vkCreateDescriptorSetLayout(m_device->device(), &ds_layout_ci, NULL, &ds_layouts[1]);
+    ASSERT_VK_SUCCESS(err);
+
+    VkPipelineLayout pipeline_layout;
+    VkPipelineLayoutCreateInfo pipeline_layout_ci = {};
+    pipeline_layout_ci.sType = VK_STRUCTURE_TYPE_PIPELINE_LAYOUT_CREATE_INFO;
+    pipeline_layout_ci.pNext = NULL;
+    pipeline_layout_ci.pushConstantRangeCount = 0;
+    pipeline_layout_ci.pPushConstantRanges = NULL;
+    pipeline_layout_ci.setLayoutCount = sizeof(ds_layouts) / sizeof(*ds_layouts);
```
tobine: Would like to see `sizeof(ds_layouts) / sizeof(*ds_layouts)` assigned to a var so that var can be reused on line 21211 below.

jozefkucia: IMHO it would be better to introduce ARRAY_SIZE() macro.

mark-lunarg:
@jozefkucia, if you wouldn't mind, the first line in commit messages should be right at 50 characters in length -- please see the [CONTRIBUTING.md](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/CONTRIBUTING.md) doc in the repo.

PR #2161, opened 2017-10-26
icd: Adding generated mock icd

tobine:
Initial check-in for mock icd which is being built to allow validation
testing without the need for an actual Vulkan device.

ICD is currently building and passing 324/332 tests. It creates ptr
handles for dispatchable objects and unique id handles for
non-dispatchable objects. It currently has some hard-coded values for
various Get* device-query functions in order to allow forward progress.
The long-term intention is to allow the device configuration to be set
by the test itself.

See the ICD README.md file for more info.

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -94,8 +94,6 @@ foreach (config_file ${LAYER_JSON_FILES_NO_DEPENDENCIES})
 endforeach(config_file)
 
 
-
-
 # Add targets for JSON file install on Linux.
 # Need to remove the "./" from the library path before installing to /etc.
 if(UNIX)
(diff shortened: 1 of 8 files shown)


scripts/mock_icd.py
```diff
@@ -0,0 +1,1080 @@
+#!/usr/bin/python3 -i
+#
+# Copyright (c) 2015-2016 The Khronos Group Inc.
```
mark-lunarg: 2017?

mark-lunarg:
LGTM!  Please rename mock_icd.py to mock_icd_generator.py to match all of the other generators.

layers/CMakeLists.txt
```diff
@@ -94,8 +94,6 @@ foreach (config_file ${LAYER_JSON_FILES_NO_DEPENDENCIES})
 endforeach(config_file)
 
 
-
```
stroyan: This is an unrelated change that could be done in a different commit.

tobine:
Pushed minor updates based on initial review feedback

tobine:
Replaced by #2164

Issue #2096, opened 2017-09-29
Memory validation does not account for external memory


jschnarr-nv:
Doing a blit with a source image using external memory results in this spurious error:

vkCmdBlitImage(): Cannot read invalid region of memory allocation <x> for bound Image object <y>, please fill the memory before using.

This validation (part of VK_LAYER_LUNARG_core_validation) should check for external memory. In this case we have no way of verifying if the memory has been initialized by another API (e.g. D3D) so this warning should be skipped.

PR #2162, opened 2017-10-26
Remove uninitialized warning for external memory

mikes-lunarg:
Fixes GH issue #2096 / LX issue [712](https://vulkan.lunarg.com/issue/view/59b6b9b2e46ffe6c0d3db481)

Treat imported external memory as valid (similar to mapped memory) so we don't warn when an image or buffer bound to the external memory is used as a src.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -358,6 +358,18 @@ static void add_mem_obj_info(layer_data *dev_data, void *object, const VkDeviceM
     assert(object != NULL);
 
     dev_data->memObjMap[mem] = unique_ptr<DEVICE_MEM_INFO>(new DEVICE_MEM_INFO(object, mem, pAllocateInfo));
+
+    if (pAllocateInfo->pNext) {
+        auto struct_header = reinterpret_cast<const GENERIC_HEADER *>(pAllocateInfo->pNext);
+        while (struct_header) {
+            if (VK_STRUCTURE_TYPE_IMPORT_MEMORY_FD_INFO_KHR == struct_header->sType ||
+                VK_STRUCTURE_TYPE_IMPORT_MEMORY_WIN32_HANDLE_INFO_KHR == struct_header->sType) {
+                dev_data->memObjMap[mem]->global_valid = true;
+                break;
+            }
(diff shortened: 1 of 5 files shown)


layers/core_validation.cpp
```diff
@@ -358,6 +358,18 @@ static void add_mem_obj_info(layer_data *dev_data, void *object, const VkDeviceM
     assert(object != NULL);
 
     dev_data->memObjMap[mem] = unique_ptr<DEVICE_MEM_INFO>(new DEVICE_MEM_INFO(object, mem, pAllocateInfo));
+
+    if (pAllocateInfo->pNext) {
+        auto struct_header = reinterpret_cast<const GENERIC_HEADER *>(pAllocateInfo->pNext);
+        while (struct_header) {
```
jzulauf-lunarg: If you merge after #2160 with lvl_find_in_chain instead

tests/vktestbinding.h
```diff
@@ -615,6 +617,13 @@ inline VkMemoryAllocateInfo DeviceMemory::alloc_info(VkDeviceSize size, uint32_t
     return info;
 }
 
+inline VkMemoryAllocateInfo DeviceMemory::get_resource_alloc_info(const vk_testing::Device &dev, const VkMemoryRequirements &reqs,
```
jzulauf-lunarg: why inline this? why move it's implementation into the .h?

layers/core_validation.cpp
```diff
@@ -358,6 +358,18 @@ static void add_mem_obj_info(layer_data *dev_data, void *object, const VkDeviceM
...
+            if (VK_STRUCTURE_TYPE_IMPORT_MEMORY_FD_INFO_KHR == struct_header->sType ||
+                VK_STRUCTURE_TYPE_IMPORT_MEMORY_WIN32_HANDLE_INFO_KHR == struct_header->sType) {
+                dev_data->memObjMap[mem]->global_valid = true;
```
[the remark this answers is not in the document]
mikes-lunarg: Comment updated

tests/vktestbinding.h
mikes-lunarg: Moved back into cpp file

PR #2163, opened 2017-10-26
Fix parameter_validation_generation.py to run under python 2

benvanik:
The 'encoding' arg to open() is python 3 only. The other *_generation.py scripts don't pass the parameter when opening the same file already and are python 2 compatible.


mark-lunarg:
Also, Python3 is a dependency for building this repo.

benvanik:
I just noticed I wasn't building the object_tracker, and found the same thing :)

Darn, the build environment I'm in doesn't support python 3. I'll find some other crazy local workaround. Thanks for taking a look!

mark-lunarg:
Any time!  When you do find a good workaround, please post another PR...

PR #2165, opened 2017-10-27
repo: normalize line termation across repo

mikew-lunarg:
This is a scrub of the entire repo to normalize end-of-line termination to conform to the recently added .gitattributes file.  A mismatch of line termination causes difficult-to-resolve conflicts when rebasing or merging.

The process performed is described at:
https://help.github.com/articles/dealing-with-line-endings/#refreshing-a-repository-after-changing-line-endings

The .gitattributes file is required because some files (*.bat and *.sh) are sensitive to line termination.  The .gitattributes file is stored in the repo, and overrides settings which could be corrupted by a developer's local git configuration.  The .gitattributes file was added in f077c81

This commit completes bcc9fec, which fixed a single file.

If your tree should become wedged on persistent modified files (i.e. if "git checkout master" still has modified files), here's a simple hack to get it clean again:
1. git checkout mikew_before_gitattributes
1. git checkout master


davidlunarg:
LGTM

jzulauf-lunarg:
The process described at the github is certainly correct.  For those with branches based on pulls of master in the incomplete state the process is to check out the affected branch (git status will show the list of affected as magically modified), disable the crlf in .gitattributes, checkout the affect files (s.t. they are no longer modified), checkout out the .gitattributes (undoing the disable), and then rebase to the the fully updated master.

mikew-lunarg:
Removing the .gitattributes would not fix the issue, unless the public history were rewritten such that .gitattributes never existed.  IMO this is a bug in git.

The problem originally being fixed was that LVL was depending upon the git "old system" of handling line termination via .gitconfig which is configured per-user.  Since the EOL selection is not in the repo, essential EOL state (on files that require it) is not guaranteed preserved.  Git's "new system" uses .gitattributes and is stored within the repo.  See http://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/

jzulauf-lunarg:
While removing .gitattributes *doesn't* work, you can comment out all
entries in .gitattributes and manually check-out the stubborn files and
then checkout .gitattributes with the same effect.

John

PR #2173, opened 2017-11-01
Updates to improve build time

cnorthrop:
This series:

- Updates the Android build script to allow skipping the build.  This is required due to how we're using import to build shaderc.  We're building it twice currently.
- Switches Travis-CI to use **ccache** which improves Android and desktop build times.  The improvements aren't uniform, so there is probably more work that could be done here.
- Use multiple cores in Travis-CI build.  I don't recall why we didn't do that, probably oversight.
- Add ccache option to local cmake builds.  This could probably be added to toolchain components too.

From my tests with hot cache, old build times:
```
Android arm32:  10 min 15 sec
Android arm64:  10 min 8 sec
Linux gcc:  9 min 17 sec
Linux clang:  7 min 30 sec
```
New build times:
```
Android arm32:  3 min 28 sec
Android arm64:  4 min 59 sec
Linux gcc:  3 min 17 sec
Linux clang:  5 min 49 sec
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,10 +33,7 @@ android:
     - build-tools-23.0.2
     - android-23
 
-cache:
-  directories:
-    - external
-    - build-android/external
+cache: ccache
 
 before_install:
   # Install the appropriate Linux packages.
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -50,18 +40,17 @@ before_install:
   - if [[ "$VULKAN_BUILD_TARGET" == "ANDROID" ]]; then export JAVA_HOME="/usr/lib/jvm/java-8-oracle"; fi
   - if [[ "$VULKAN_BUILD_TARGET" == "ANDROID" ]]; then export PATH="$ANDROID_NDK_HOME:$PATH"; fi
 
-  # Clear the toolchain contents if revisions have changed
-  - ./scripts/check_toolchain_revisions.sh
```
mikew-lunarg: Will removing this cause stale cache problems?
If not, can the script be deleted?

cnorthrop: I'm hoping that a change to the backing cache technology will avoid whatever problems we were having.  I left the script in so we can quickly add this back if we need to.  I'm not even sure the script clears all the right spots anymore, since we've been pulling in other dependencies.

cnorthrop:

PR #2179, opened 2017-11-03
Replaced fixed path python 2.7 executable calls

SaschaWillems:
This PR replaces the fixed Python 2.7 path (C:\Python27\python.exe) with just the python.exe executable call that will work when the python binary has been added to the path (which is an option in the installer).

The [windows system requirements for building](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/BUILD.md#windows-system-requirements) already state Python 3 as a requirement, so the old (fixed) path doesn't fit anymore. 

Just calling the executable is also more flexible and the scripts that are called work just fine with a recent Python 3.6 version.


mew-cx:
The upstream script that forced the Py2 requirement has been modified to work with Py3, so the need for Py2 is eliminated.

PR #2181, opened 2017-11-03
Tobin mock icd unexpected error cleanup

tobine:
This cleans up most of the unexpected errors when testing with the mock ICD.
Make a few changes to the ICD limits and format properties.
Update a bunch of tests, primarily to make sure buffer allocation size meets requirements.

### DIFF
--- a/scripts/mock_icd_generator.py
+++ b/scripts/mock_icd_generator.py
@@ -133,9 +133,9 @@ static VkPhysicalDeviceLimits SetLimits(VkPhysicalDeviceLimits *limits) {
     limits->viewportBoundsRange[1] = 8191;
     limits->viewportSubPixelBits = 0;
     limits->minMemoryMapAlignment = 64;
-    limits->minTexelBufferOffsetAlignment = 256;
-    limits->minUniformBufferOffsetAlignment = 256;
-    limits->minStorageBufferOffsetAlignment = 256;
+    limits->minTexelBufferOffsetAlignment = 16;
+    limits->minUniformBufferOffsetAlignment = 16;
+    limits->minStorageBufferOffsetAlignment = 16;
     limits->minTexelOffset = -8;
     limits->maxTexelOffset = 7;
     limits->minTexelGatherOffset = -8;
--- a/scripts/mock_icd_generator.py
+++ b/scripts/mock_icd_generator.py
@@ -588,8 +588,12 @@ CUSTOM_C_INTERCEPTS = {
     GetPhysicalDeviceFeatures(physicalDevice, &pFeatures->features);
 ''',
 'vkGetPhysicalDeviceFormatProperties': '''
-    // TODO: Just returning full support for everything initially
-    *pFormatProperties = { 0x00FFFFFF, 0x00FFFFFF, 0x00FFFFFF };
+    if (VK_FORMAT_UNDEFINED == format) {
+        *pFormatProperties = { 0x0, 0x0, 0x0 };
+    } else {
+        // TODO: Just returning full support for everything initially
+        *pFormatProperties = { 0x00FFFFFF, 0x00FFFFFF, 0x00FFFFFF };
+    }
 ''',
 'vkGetPhysicalDeviceFormatProperties2KHR': '''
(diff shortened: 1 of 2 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -7054,7 +7054,8 @@ TEST_F(VkLayerTest, InvalidCmdBufferDescriptorSetImageSamplerDestroyed) {
     // TODO - though the particular error above doesn't re-occur, there are other 'unexpecteds' still to clean up
     vkQueueWaitIdle(m_device->m_queue);
     m_errorMonitor->SetUnexpectedError(
-        "pDescriptorSets must be a pointer to an array of descriptorSetCount VkDescriptorSet handles, each element of which must "
+        "pDescriptorSets must be a valid pointer to an array of descriptorSetCount VkDescriptorSet handles, each element of which "
+        "must "
```
chrisforbes: join this back with the next line?

tests/layer_validation_tests.cpp
```diff
@@ -5695,7 +5695,7 @@ TEST_F(VkLayerTest, InvalidCmdBufferBufferDestroyed) {
 
     VkMemoryAllocateInfo alloc_info = {};
     alloc_info.sType = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO;
-    alloc_info.allocationSize = 256;
+    alloc_info.allocationSize = mem_reqs.size;
```
chrisforbes: :+1: although maybe we should just make this test use a helper that manages the lifetime of a buffer and its backing memory together? This detail isn't interesting for the test, only an opportunity to get it wrong (as we did)

tests/layer_validation_tests.cpp
```diff
@@ -6676,14 +6676,13 @@ TEST_F(VkLayerTest, InvalidCmdBufferDescriptorSetBufferDestroyed) {
     ASSERT_VK_SUCCESS(err);
     // Allocate memory and bind to buffer so we can make it to the appropriate
     // error
+    VkMemoryRequirements memReqs;
+    vkGetBufferMemoryRequirements(m_device->device(), buffer, &memReqs);
     VkMemoryAllocateInfo mem_alloc = {};
     mem_alloc.sType = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO;
     mem_alloc.pNext = NULL;
-    mem_alloc.allocationSize = 1024;
+    mem_alloc.allocationSize = memReqs.size;
```
chrisforbes: :+1:, ditto to previous comment.

tests/layer_validation_tests.cpp
```diff
@@ -11656,6 +11659,17 @@ TEST_F(VkLayerTest, DSUsageBitsErrors) {
         }
     }
     if (image_ci.format == VK_FORMAT_UNDEFINED) {
+        // Cleanup before early return
```
chrisforbes: This test is crazy and we should fix it. I'll file a bug.

scripts/mock_icd_generator.py
```diff
@@ -133,9 +133,9 @@
     limits->viewportBoundsRange[1] = 8191;
     limits->viewportSubPixelBits = 0;
     limits->minMemoryMapAlignment = 64;
-    limits->minTexelBufferOffsetAlignment = 256;
-    limits->minUniformBufferOffsetAlignment = 256;
-    limits->minStorageBufferOffsetAlignment = 256;
+    limits->minTexelBufferOffsetAlignment = 16;
+    limits->minUniformBufferOffsetAlignment = 16;
```
chrisforbes: There is some real HW that requires 256, are we giving up now on being able to run the tests easily across all HW?

tobine: Not giving up, tests should still run easily across all HW. The 256 minOffsetAlignment value led to unexpected errors DSBufferLimitErrors for a slightly convoluted offset calculation in the test that was easier to fix by making the limit smaller than by re-doing the calculation. It's the "sb_offset" value in that test. I'll mark it with a TODO to make sure it gets cleaned-up to run with all possible offset alignment values.

tobine:
Rebased to master

PR #2187, opened 2017-11-07
demos: Remove runaway HTML in plain-text version of vulkaninfo

polarina:


mark-lunarg:
Modified commit message for length and pushed to Master -- Thanks!

PR #2195, opened 2017-11-08
layers: Don't confuse VkResult/VkBool32 in generators.

null77:
In some cases, the generator would return the special error for
validation failed for functions that return a VkBool32. Fix this, and
also some cases of initializing a VkBool32 with VkResult.

Change-Id: Icc071e647f293848d62a922d492ed41dfc5a4be1


null77:
Note that there is probably a warning we could enable to catch this in the CI builder. Can look into that if desired.

PR #2196, opened 2017-11-08
loader/layers: Fix MSVS warnings

null77:
A couple warnings popped up with my integration's build. These are fairly low impact. Hopefully the new enable flags will prevent further regressions (and might catch a few in the CI builder that I didn't fix).


null77:
Looks like there are a couple more warnings. #2195 should fix two of them, but there's one other warning in the layers.

PR #2197, opened 2017-11-08
tests:Add doc validation to travis CI

tobine:
Run doc validation following build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,7 @@ script:
   - if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then ./update_external_sources.sh; fi
   - if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then cmake -H. -Bdbuild -DCMAKE_BUILD_TYPE=Debug; fi
   - if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then make -C dbuild -j $core_count; fi
+  - if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then ( cd dbuild/tests; ./vkvalidatelayerdoc.sh); fi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -46,7 +46,8 @@ script:
   - if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then ./update_external_sources.sh; fi
   - if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then cmake -H. -Bdbuild -DCMAKE_BUILD_TYPE=Debug; fi
   - if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then make -C dbuild -j $core_count; fi
-  - if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then LD_LIBRARY_PATH=dbuild/loader:$LD_LIBRARY_PATH VK_LAYER_PATH=dbuild/layers VK_ICD_FILENAMES=dbuild/icd/VkICD_mock_icd.json dbuild/tests/vk_layer_validation_tests; fi
+  - if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then pushd dbuild/tests;./vkvalidatelayerdoc.sh; fi
```
cnorthrop: Any reason not to have the popd on this line?  Would look a bit cleaner.

tobine: The last thing run on the line determines pass/fail so when popd is the last thing it masks any doc validation failure.

jzulauf-lunarg: I'm generally not a fan of changing the directory during a linked series of step.  what if one were to run the chdir'ing command in a ( ... ) subshell?

( cd dbuild/test; ./vkvalidatelayerdoc.sh); fi

I just tested and the last exit status of the sub-shell is was is current after returning from the subshell.

cnorthrop: Switching to a subshell looks good to me.

PR #2200, opened 2017-11-10
layers: Make imageMap/bufferMap thread safe

jfmarquis:
Add dedicated mutex for imageMap/bufferMap.
Force lock of mutex of imageMap/bufferMap inside scope using
smart pointer. There was already too many places missing a lock to
track by hand; better rely on the compiler to have 100% coverage.
Reduce contention on global_lock when only accessing imageMap/bufferMap.
Remove some locks of global_lock that are now unnecessary.


jfmarquis:
I am really interesting in your opinion on giving each container its own mutex and the use of a scoped object to both lock and access the value. After adding validation to make sure the mutex is locked when it should be I realized there are just too many places that require locking to track by hand; so I made something to make it impossible to forget to lock, while still allowing control of the scope of the lock if needed.
I didn't notice an obvious place to put that new type because I didn't see a lot of files and files are all full of different types.

layers/core_validation_types.h
```diff
@@ -781,6 +782,29 @@ class FRAMEBUFFER_STATE : public BASE_NODE {
...
+    struct unique_lock_value {
+        unique_lock_value(LOCKED_VALUE_TYPE &_value, MUTEX_TYPE &mutex) : value(_value), lock(mutex) {}
+        LOCKED_VALUE_TYPE *operator->() const { return &value; }
+        LOCKED_VALUE_TYPE &operator*() const { return value; }
+
+       private:
+        LOCKED_VALUE_TYPE &value;
+        std::unique_lock<MUTEX_TYPE> lock;
+    };
+    unique_lock_value<VALUE_TYPE> get() { return {value, mutex}; }
```
jzulauf-lunarg: I'm not a huge fan of using semantics with implicit locking, as the risk of coding a deadlock becomes very high.  Imagine two mutex_value a, b;

Because is it not obvious to the coder that a.get() and b.get() are acquiring locks then the coder/maintainer could be unaware that a consistent acquisition order is required.  If:

// Function afirst
auto aref = a.get();
// gobs of code
auto bref = b.get();

and in

// Function bFirst
auto bref = b.get();
// Other gobs of code
auto aref = a.get();

we've written a deadlock, but there's nothing to indicate it.  When one codes with std::*lock and std::*mutex's it's at least clear that lock acquisition is occurring.

layers/core_validation_types.h
```diff
@@ -781,6 +782,29 @@ class FRAMEBUFFER_STATE : public BASE_NODE {
...
+    unique_lock_value<VALUE_TYPE const> get() const { return {value, mutex}; }
+    unique_lock_value<VALUE_TYPE> operator->() { return {value, mutex}; }
```
jzulauf-lunarg: while it's unlikely that someone will create a persistent reference with ->, it's even less clear that a lock is occurring.

layers/core_validation_types.h
jfmarquis: I understand your concerns. My main motivation was to allow access to the value only while locking at the same time to prevent access without locking. To be clear that locking happens, the first level of operator-> would be removed.

If it is preferable to have more granular locking, by having several mutex, the risk of deadlock will be present whether locking is automatic or manual. Do you think it means 'global_lock' should stay unique for everything?

I don't want to rename get() to just lock() because it would lack an equivalent unlock(). I have trouble finding a suitable name that would convey the intention of a lock_guard combined with getting a (smart) pointer to the value.
Should I drop automatic locking instead? It was my impression that manual locking was also error prone, since there are already locks missing at the moment.

Thank you for going through my changes and commenting.

jzulauf-lunarg: I'd say stick with the global lock for now. When we start looking into layer performance, looking into contended lock time is going to be pretty high on the list, and I think this is something to attack deliberately/systematically whenever we do. 

And yes, as soon as we have more than one lock, we're going to have to be careful (and obvious) such that we don't code ourselves into having deadlocks lurking in the code.

jfmarquis: Very well. You can vote to cancel this PR (if that's even a thing).
On my side I will instead attempt to track each missing lock through some kind of validation and add a lock_guard to global_lock.
Is creating an issue first the correct way to exchange on possible solutions before implementing something in a branch and waiting for a PR to chat?
I am only used to teams that are physically together :)

jzulauf-lunarg: An issue is a useful tool to track work to do.  If your looking to attack this yourself, please take a look at other places where the validation layers are using the global lock.  Currently it's not used in a fine grained way, s.t. grabbing it for the duration of a function would be acceptable, unless you are calling *out* of the function to a function that is either a) long, b) doesn't need the lock, or c) may need to grab the lock.  In those cases we lock and unlock the lock_guard.  (for example see the pattern in GetQueryPoolResults in core_validation.cpp, and many other locations)

jfmarquis:
I am closing this because it was done with #2208

PR #2202, opened 2017-11-10
GetQueryPoolResults validation corner-cases

mark-lunarg:
Internal app development uncovered a couple of false-positives related to 'unavailable' timestamp queries.  Removed this erroneous check.   Also added a semi-related VUID check and test.


tobine:
This causes VkLayerTest.QueryPoolInUseDestroyedSignaled to fail.
Cloud based LVL test run paying dividends.

layers/core_validation.cpp
```diff
@@ -3428,7 +3439,8 @@ static bool PreCallValidateGetQueryPoolResults(layer_data *dev_data, VkQueryPool
                     auto cb = GetCBNode(dev_data, cmd_buffer);
                     make_available |= cb->queryToStateMap[query];
                 }
-                if (!(((flags & VK_QUERY_RESULT_PARTIAL_BIT) || (flags & VK_QUERY_RESULT_WAIT_BIT)) && make_available)) {
+                if (!(((flags & VK_QUERY_RESULT_PARTIAL_BIT) || (flags & VK_QUERY_RESULT_WAIT_BIT)) && make_available) &&
+                    !(flags & VK_QUERY_RESULT_WITH_AVAILABILITY_BIT) && (query_count > 1)) {
```
jzulauf-lunarg: I can't figure out what in the spec this second line represents.  What usage are we testing against?

jzulauf-lunarg: I'm not sure these aren't *all* false positives.  The spec allows for queries where neither partial nor waited for results are available with VK_NOT_READY.

"If VK_QUERY_RESULT_WAIT_BIT and VK_QUERY_RESULT_PARTIAL_BIT are both not set then no result values are written to pData for queries that are in the unavailable state at the time of the call, and vkGetQueryPoolResults returns VK_NOT_READY. However, availability state is still written to pData for those queries if VK_QUERY_RESULT_WITH_AVAILABILITY_BIT is set."

mark-lunarg: Surely someone could have a single query and ignore the NOT_READY bit, but in that case the user has enough information to know whether the timestamp has been reached or not.  If WITH_AVAILABILITY is set they also have enough information -- these cases should not generate errors.  I don't think we generate errors for other APIs that rely on a return code to communicate status.  Maybe this is a case for the assistant layer?

mark-lunarg: @jzulauf-lunarg and I talked, and we agree that this entire sub-case should be removed.  Testing now, update to follow.

layers/vk_validation_error_database.txt
```diff
@@ -3207,7 +3207,7 @@ VALIDATION_ERROR_2fa0065c~^~N~^~Unknown~^~vkGetQueryPoolResults~^~VUID-vkGetQuer
...
 VALIDATION_ERROR_2fa00662~^~N~^~Unknown~^~vkGetQueryPoolResults~^~VUID-vkGetQueryPoolResults-dataSize-00817~^~core~^~The spec valid usage text states 'dataSize must be large enough to contain the result of each query, as described here' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkGetQueryPoolResults-dataSize-00817)~^~
-VALIDATION_ERROR_2fa00664~^~N~^~Unknown~^~vkGetQueryPoolResults~^~VUID-vkGetQueryPoolResults-queryType-00818~^~core~^~The spec valid usage text states 'If the queryType used to create queryPool was VK_QUERY_TYPE_TIMESTAMP, flags must not contain VK_QUERY_RESULT_PARTIAL_BIT' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkGetQueryPoolResults-queryType-00818)~^~
+VALIDATION_ERROR_2fa00664~^~Y~^~Unknown~^~vkGetQueryPoolResults~^~VUID-vkGetQueryPoolResults-queryType-00818~^~core~^~The spec valid usage text states 'If the queryType used to create queryPool was VK_QUERY_TYPE_TIMESTAMP, flags must not contain VK_QUERY_RESULT_PARTIAL_BIT' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkGetQueryPoolResults-queryType-00818)~^~
```
tobine: Need a test for this case.

PR #2203, opened 2017-11-13
Move BuildTemplate call inside global lock

mark-lunarg:
Helper routine was being called outside the lock causing freezes in Dota2.


mark-lunarg:
Appveyor failure is due to our nifty 60-minute timeout.

PR #2208, opened 2017-11-15
layers: Add some missings locks of global_lock

jfmarquis:
Add locks of global_lock in BindBufferMemory, BindImageMemory
and PreCallValidateGetImageSubresourceLayout to have safe access
to imageMap and bufferMap.


layers/core_validation.cpp
```diff
@@ -96,9 +96,7 @@ std::string to_string(T var) {
 // This intentionally includes a cpp file
 #include "vk_safe_struct.cpp"
 
-using mutex_t = std::mutex;
-using lock_guard_t = std::lock_guard<mutex_t>;
-using unique_lock_t = std::unique_lock<mutex_t>;
+using lock_guard_t = std::lock_guard<core_validation::mutex_t>;
```
chrisforbes: Can this one move too?

layers/core_validation.cpp
```diff
@@ -96,9 +96,7 @@ std::string to_string(T var) {
 // This intentionally includes a cpp file
 #include "vk_safe_struct.cpp"
 
-using mutex_t = std::mutex;
```
jzulauf-lunarg: since we are keeping the locks in this file, we should keep the definitions here as well.

layers/core_validation.cpp
```diff
@@ -210,7 +208,7 @@ void ValidateLayerOrdering(const TCreateInfo &createInfo) {
 }
 
 // TODO : This can be much smarter, using separate locks for separate global data
-static mutex_t global_lock;
```
jzulauf-lunarg: If all locking is done at the API level, the global lock can remain static.

layers/buffer_validation.cpp
```diff
@@ -4068,7 +4068,11 @@ bool PreCallValidateGetImageSubresourceLayout(layer_data *device_data, VkImage i
                         validation_error_map[VALIDATION_ERROR_2a6007ca]);
     }
 
-    IMAGE_STATE *image_entry = GetImageState(device_data, image);
+    IMAGE_STATE *image_entry;
+    {
+        core_validation::unique_lock_t lock(core_validation::global_lock);
```
jzulauf-lunarg: this is a gap in the contributors documentation, but our convention is for PreCallValidate.* calls is to acquire and release the lock at the API entry, i.e. in the calling context, unlocked before the "dispatch_table", at least as long as we're using a single global lock. (as an example see core_validation.cpp::CmdResolveImage)

layers/core_validation_types.h
```diff
@@ -61,6 +61,7 @@
 #include <vector>
 #include <memory>
 #include <list>
+#include <mutex>
```
jzulauf-lunarg: not needed, given the above.

layers/core_validation_types.h
```diff
@@ -786,6 +787,11 @@ struct DeviceExtensions;
 
 // Fwd declarations of layer_data and helpers to look-up/validate state from layer_data maps
 namespace core_validation {
+using mutex_t = std::mutex;
```
jzulauf-lunarg: not needed given the above

layers/core_validation_types.h
```diff
@@ -786,6 +787,11 @@ struct DeviceExtensions;
...
+using unique_lock_t = std::unique_lock<mutex_t>;
```
jzulauf-lunarg: not  needed given the above

layers/core_validation.cpp
```diff
@@ -3861,7 +3859,11 @@ static void PostCallRecordBindBufferMemory(layer_data *dev_data, VkBuffer buffer
 VKAPI_ATTR VkResult VKAPI_CALL BindBufferMemory(VkDevice device, VkBuffer buffer, VkDeviceMemory mem, VkDeviceSize memoryOffset) {
     layer_data *dev_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
     VkResult result = VK_ERROR_VALIDATION_FAILED_EXT;
-    auto buffer_state = GetBufferState(dev_data, buffer);
+    BUFFER_STATE *buffer_state;
```
jzulauf-lunarg: LGTM

layers/core_validation.cpp
jzulauf-lunarg: the deletion part of this change shouldn't be needed if the global_lock is staying contained to this compilation unit.

jfmarquis:
I have the equivalent fixes using the locking convention done, which puts everything back as private to core_validation. I took a look at some locks that are inside Pre or Post functions instead of in the API, to remove double locks in some cases but will not go as far as planned, because it is somewhat of a pandora's box for a bug fix.
I just need to mention that while the locking is mostly done the right way, there are still cases that require unlock+lock inside Pre functions when they use the dispatch table. 
Will commit tomorrow.

jfmarquis:
Sorry about the back and forth. At least the PR shows a single changed file, so I guess the commit would still look clean. The original commit descriptor and PR description still apply perfectly, too.

If squashing into a single change is a git feature, I don't know how to do such a thing. (I am not being sarcastic)

jzulauf-lunarg:
Sorry this has stretched out for you.  Really appreciate your patience and persistence.

To squash, assuming your topic branch is named "image-buffer-maps-lock" and your local master is named "master" (and that the two changes we're interested in are the only ones on that branch)

$ git checkout image-buffer-maps-lock
$ git rebase -i master 

This will open up an editor listing your changes, where you change 'pick' to 'squash' in the second change.  You save this, and git brings up a merger of the two commit messages. Edit this to a good single commit message, and save.

https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history

The, force push the rebased branch to github.

and more details here https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

Feel free to ping me directly if I can be of help.

jfmarquis:
Squashing done. Sorry for the delay.

Issue #1685, opened 2017-04-22
Validation layers doesn't check for NULL when provided count is greater then 0


RippeR37:
Many Vulkan commands and structures takes some arguments in pairs: count and pointer to an array.
I've noticed however, that validation layers doesn't check if pointer is not NULL if count is greater then 0.

For some fields, pointers can be set to NULL, so it can't be simply checking all `pSomethingSomething` for NULLs. However such checks might be useful as sometimes something is added over a time and some fields might not have been updated with others which might lead to crash or god knows what...

One of such places is `VkPipelineViewportStateCreateInfo` structure which has pairs:
* `viewportCount` and `pViewports`
* `scissorCount` and `pScissors`

On my system (Windows 7 x64, Nvidia GTX 660M, latest drivers) it lead to crash on `vkCreateGraphicsPipelines`. Usually it will be easy to find, but in such big struct like the ones for pipelines it might take some time...

jeremy-lunarg:
> On my system (Windows 7 x64, Nvidia GTX 660M, latest drivers) it lead to crash on `vkCreateGraphicsPipelines`.

Can you double check that the parameter_validation layer is loaded? If the corresponding state is not dynamic, it should validate the `pViewports` and `pScissors` pointers during `vkCreateGraphicsPipelines`. We have tests for these errors in `layer_validation_tests.cpp` (`VALIDATION_ERROR_02110` and `VALIDATION_ERROR_02111`).

krOoze:
Your specific count&ndash;array pair example is one of the more complicated ones. (The spec VU has to be hand-made for this case and the relation is further complicated by other variables.)

But for other arrays I think the  KhronosGroup/Vulkan-Docs#491 . (The count > 0 + nullptr seem to be currently allowed &mdash; but maybe it shouldn't be)

jeremy-lunarg:
With respect to `pViewports` and `pScissors`, a `nullptr` can make it through the validation layers and into the driver if the state is dynamic, the count is between `1` and `VkPhysicalDeviceLimits::max*` inclusive, and the pointer is null. In which case, I believe the driver should ignore the pointer as stated in the spec.

The issue raised by @krOoze is relevant and may affect the VUs.

RippeR37:
@jeremy-lunarg I was using standard LunarG meta-layer in 1.0.42.2 VulkanSDK so I assume it is.
I didn't use any dynamic state there, it was simple demo.

[...]

PR #2209, opened 2017-11-16
layers: Rewrite ViewportState checks and tests

krOoze:
- deal with #1685
- report correct object type
- improve? msg texts a bit
- add extension structs to `pNext` check
- use correct printf token where possible
- make tests to match  changes + improve coverage (`nullptr` +
`multiViewport`)
- remove viewport state (and other stuff) from tests that do not use it
- merge some tests together that are too related
- remove unnecessary dynamic state from `PSOLineWidthInvalid`)
- correct invalid `VkViewport` in some places
- correct missing/null `pViewport` and `pScissors` in some places
- update and correct the VUID database


layers/vk_validation_error_database.txt
```diff
@@ -1503,15 +1503,15 @@ VALIDATION_ERROR_14e09005~^~Y~^~Unknown~^~vkCreateViSurfaceNN~^~VUID-VkViSurface
...
-VALIDATION_ERROR_150009a4~^~N~^~Unknown~^~vkCmdSetViewport~^~VUID-VkViewport-minDepth-01234~^~!(VK_EXT_depth_range_unrestricted)~^~The spec valid usage text states 'minDepth must be between 0.0 and 1.0, inclusive' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkViewport-minDepth-01234)~^~
-VALIDATION_ERROR_150009a6~^~N~^~Unknown~^~vkCmdSetViewport~^~VUID-VkViewport-maxDepth-01235~^~!(VK_EXT_depth_range_unrestricted)~^~The spec valid usage text states 'maxDepth must be between 0.0 and 1.0, inclusive' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkViewport-maxDepth-01235)~^~
+VALIDATION_ERROR_15000996~^~Y~^~ViewportAndScissorBoundsChecking~^~vkCreateGraphicsPipelines,vkCmdSetViewport~^~VUID-VkViewport-width-01227~^~core~^~The spec valid usage text states 'width must be greater than 0.0 and less than or equal to VkPhysicalDeviceLimits::maxViewportDimensions[0]' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkViewport-width-01227)~^~
```
mark-lunarg: @krOoze, the fourth value in the database is generated each time we rebuild the DB from the spec, so editing these is unnecessary and could possibly cause problems (not sure about this).  This field is unused right now and is often misleading in that the generator picks up whatever the current section is for the VU which, as you have discovered, is sometimes inaccurate.  We should probably remove it at some point.

krOoze: It is potentially useful info though. It should be possible to generate this accurately from the `vk.xml`.

krOoze: It should be reset back now.

I also flicked `N` the `VkViewport` checks as I deferred implementing them in `vkCreateGraphicsPipelines`.

layers/vk_validation_error_database.txt
```diff
@@ -1503,13 +1503,13 @@ VALIDATION_ERROR_14e09005~^~Y~^~Unknown~^~vkCreateViSurfaceNN~^~VUID-VkViSurface
...
-VALIDATION_ERROR_1500099a~^~Y~^~ViewportBoundsCheckingWithNVHExtensionEnabled~^~vkCmdSetViewport~^~VUID-VkViewport-height-01229~^~(VK_KHR_maintenance1)~^~The spec valid usage text states 'height must be greater than or equal to -VkPhysicalDeviceLimits::maxViewportDimensions[1] and less than or equal to VkPhysicalDeviceLimits::maxViewportDimensions[1]' (https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VUID-VkViewport-height-01229)~^~
+VALIDATION_ERROR_15000996~^~N~^~ViewportAndScissorBoundsChecking~^~vkCmdSetViewport~^~VUID-VkViewport-width-01227~^~core~^~The spec valid usage text states 'width must be greater than 0.0 and less than or equal to VkPhysicalDeviceLimits::maxViewportDimensions[0]' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkViewport-width-01227)~^~
```
tobine: Doc validator is complaining that the 6 checks below are still in source. Did you change the PR?

    VALIDATION_ERROR_1500099e
    VALIDATION_ERROR_15000998
    VALIDATION_ERROR_150009a0
    VALIDATION_ERROR_150009a2
    VALIDATION_ERROR_15000996
    VALIDATION_ERROR_1500099a

krOoze: As said above, I changed them to N as they are only partially implemented.

What is the purpose of this "doc validator"?
I think it has it bit backwards. I think it is a problem if it says `Y` on something that is not fully implemented, not other way around. That is assuming purpose of this flag is to track progress, and inform app-makers what is and isn't (fully) checked.

krOoze: I could just implement them though, I just run out of free time when I committed (and the PR seemed good enough to me)...

krOoze: OK, reverted this, so it makes script happy.

tobine: The doc validator checks the state of the database against the actual code. It's primary purpose is to make sure that checks that are in the code are reflected as "Y" in the database. Secondarily it flags issues if "Y" checks don't appear in the code, and also makes sure that test names in the database appear in the test file.

All of its checks are a bit naive, as it just scrubs the source for VUID strings and checks against DB. I agree that items should only be "Y" in the DB if they're completely implemented. That means that in order for doc validator to be happy, if you only partially implement a check, then its VUID should not appear in the code (on a non-commented line) until it's fully implemented and DB can be switched to "Y".

tobine:
This LGTM and is passing testing so I'll merge.
In general this is alot of code for a single CL. In the future, if possible, it would be good to split all of the test changes across multiple CLs to simplify review.

mark-lunarg:
Hey @tobine, something in this (we think) last patch set is causing Nvidia Linux to crash when running DynViewportAndScissorMismatch, can you take a look?

tobine:
Yes I'm looking at this. Crashing on pipeline creation. Also crashing on Nvidia PixelC w/ Android OS.

krOoze:
@tobine That's odd, the set up should be very similar to the other tests. Do they crash too?

krOoze:
Note to self:
- The `VkPipelineRasterizationStateCreateInfo` is weird (`lineWidth?!)
- gp_ci is missing the `shader_stages_size`
- `VkDescriptorSetLayout`s should not be necessary for the tests

PR #2212, opened 2017-11-16
Tobin devsim lvl testing

tobine:
This fixes up a number of test and ICD issues that I found while testing the mock ICD with different devsim profiles.
It adds in a pre-build DevSim layer along with its json file and 7 initial device profiles from an assortment of vendors.
The Travis-CI regression now runs the LVL test suite 8 times. Once on default mock ICD and then once on each of the 7 different json device profiles.
For tests that require the device profile layer, that will take the place of the devsim layer.
Eventually we can enhance the device profile layer and migrate all tests to just use that but for now this gives us an initial assortment of coverage over some simulated profiles.


tobine:
merged

PR #2215, opened 2017-11-16
Make ExternalMemory test more robust

### LINKED ISSUE #2206: ExternalMemory tests fail on AMD
`ExternalMemory` layer positive test fails on AMD (Radeon Software 17.11.1):

> tests\vktestbinding.cpp(403): error: Failed
> Assertion: `vkAllocateMemory(dev.handle(), &info, 0, &handle) == VK_SUCCESS'
> C:\Dev\_GitHub\Vulkan-LoaderAndValidationLayers\tests\vktestbinding.cpp(483): error: Failed
> Assertion: `vkBindBufferMemory(device(), handle(), mem.handle(), mem_offset) == VK_SUCCESS'
> C:\Dev\_GitHub\Vulkan-LoaderAndValidationLayers\tests\layer_validation_tests.cpp(368): error: Failed
> Expected to succeed but got error: vkBindBufferMemory: required parameter memory specified as VK_NULL_HANDLE
> tests\layer_validation_tests.cpp(368): error: Failed
> Expected to succeed but got error: vkCmdCopyBuffer(): VkBuffer object 0x2 used with no memory bound. Memory should be bound by calling vkBindBufferMemory(). The spec valid usage text states 'If srcBuffer is non-sparse then it must be bound completely and contiguously to a single VkDeviceMemory object' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkCmdCopyBuffer-srcBuffer-00119)

mikes-lunarg:
- Use dedicated allocation when required by the external memory implementation
- Add required pNext structs for all buffers and memory
- Relax the requirement that external memory be host visible for this test
- Verify that the export/import memory is actually linked

Also fixes #2206

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -25579,7 +25579,7 @@ TEST_F(VkPositiveLayerTest, ClearDepthStencilWithValidRange) {
 }
 
 TEST_F(VkPositiveLayerTest, ExternalMemory) {
-    TEST_DESCRIPTION("Perform a copy from a buffer backed by external memory");
+    TEST_DESCRIPTION("Perform a copy through a pair of buffers linked by external memory");
 
 #ifdef _WIN32
     const auto extension_name = VK_KHR_EXTERNAL_MEMORY_WIN32_EXTENSION_NAME;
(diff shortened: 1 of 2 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -25659,25 +25697,46 @@ TEST_F(VkPositiveLayerTest, ExternalMemory) {
...
+    vkCmdPipelineBarrier(m_commandBuffer->handle(), VK_PIPELINE_STAGE_TRANSFER_BIT, VK_PIPELINE_STAGE_TRANSFER_BIT, 0, 1,
+                         &mem_barrier, 0, nullptr, 0, nullptr);
+    vkCmdCopyBuffer(m_commandBuffer->handle(), buffer_import.handle(), buffer_output.handle(), 1, &copy_info);
     m_commandBuffer->end();
     m_commandBuffer->QueueCommandBuffer();
 
+    // Verify output buffer
+    auto output_mem = (uint8_t *)buffer_output.memory().map();
+    for (int i = 0; i < buffer_size; i++) {
+        ASSERT_TRUE(output_mem[i] == (i & 0xFF));
```
mikes-lunarg: @tobine I dont' think this assert is going to play nice with #2212 Are you maintaining an exclusion list or should I just pull it out?

tobine: Hmm, this makes it through default mock testing. Let me test a branch with this change on top of 2212 and see if there are issues before we push either.

tobine:
merged

PR #2217, opened 2017-11-17
tests:Fix DynViewportAndScissorMismatch crash

tobine:
This test was crashing on Nvidia drivers.
The crash seems to be a driver bug as the workaround is to not destroy
a descriptor set layout prior to creating gfx pipelines whose
pipeline_layout object was created from the given descriptor set layout

@krOoze note workaround for NV crash


krOoze:
@tobine is the `VkDescriptorSetLayout` necessary at all?

krOoze:
@tobine The same code is in `PSOViewportStateMultiViewportTests` and `PSOViewportStateTests`. Those do not crash?

tobine:
@krOoze both of the tests you referenced expect errors when attempting to create gfx pipeline so they don't call down to the driver.

krOoze:
@tobine  Oh!

PR #2218, opened 2017-11-17
Consolidate spirv-tools/headers locations

mark-lunarg:
The glslang component now pulls down its own versions of spirv-tools and spirv-headers, leaving LVL with multiple copies.  Reworked makefiles, scripts, and docs to use the glslang versions and avoided downloading/building the duplicates.


update_external_sources.bat
```diff
@@ -85,14 +58,19 @@ REM // ======== Parameter parsing ======== //
          goto:parameterLoop
       )
 
+      if "%1" == "--all" (
```
mikew-lunarg: should --spriv-tools and -s also go here?   As it stands now, they'll (perhaps confusingly) trigger "Unrecognized option" where they were once legal.

update_external_sources.bat
```diff
@@ -85,14 +58,19 @@ REM // ======== Parameter parsing ======== //
...
+         shift
+         goto:parameterLoop
+      )
+
       echo.
       echo Unrecognized option "%1"
       echo.
       echo usage: update_external_sources.bat [options]
       echo.
       echo   Available options:
```
mikew-lunarg: perhaps add here that --all -s --spirv-tools are all deprecated?

update_external_sources.sh
```diff
@@ -137,7 +84,6 @@ do
       echo "Usage: update_external_sources.sh [options]"
       echo "  Available options:"
       echo "    -g | --glslang      # enable glslang component"
-      echo "    -s | --spirv-tools  # enable spirv-tools component"
```
mikew-lunarg: as above.

update_external_sources.bat
karl-lunarg: You'll need to keep --all and have it do what it did before, less the spirv-tools stuff.  This is because people and CI machines may still use --all and expect it to build whatever is there, even if it is just one module.  If you change it as shown here, then people and CI machines will have to change needlessly.

I'm on the fence about -s and --spirv-tools doing anything.  I think I slightly favor printing a deprecated message, although I'd think that not many people used these options.

BUILD.md
```diff
@@ -389,34 +389,30 @@ adb shell am start -a android.intent.action.MAIN -c android-intent.category.LAUN
 The [Qt Creator IDE](https://qt.io/download-open-source/#section-2) can open a root CMakeList.txt as a project directly, and it provides tools within Creator to configure and generate Vulkan SDK build files for one to many targets concurrently, resolving configuration issues as needed. Alternatively, when invoking CMake use the -G Codeblocks Ninja option to generate Ninja build files to be used as project files for QtCreator
 
 - Follow the steps defined elsewhere for the OS using the update\_external\_sources script or as shown in **Loader and Validation Layer Dependencies** below
-- Open, configure, and build the gslang and spirv-tools CMakeList.txt files
+- Open, configure, and build the gslang CMakeList.txt files
```
karl-lunarg: "gslang" is consistently wrong in this general area and needs changing to glslang.

It may be worth working in a sentence someplace saying that the glslang repo now gives you spirv-tools and headers as well.  Maybe something as simple as adding this to the end of this line:  (The glslang repo contains spirv-headers and spirv-tools.)

BUILD.md
```diff
@@ -389,34 +389,30 @@ adb shell am start -a android.intent.action.MAIN -c android-intent.category.LAUN
...
 
 ## Loader and Validation Layer Dependencies
-gslang and SPIRV-Tools repos are required to build and run Loader and Validation Layer components. They are not git sub-modules of Vulkan-LoaderAndValidationLayers but Vulkan-LoaderAndValidationLayers is linked to specific revisions of gslang and spirv-tools. These can be automatically cloned and built to predefined locations with the update\_external\_sources scripts. If a custom configuration is required, do the following steps:
+The gslang repo is required to build and run Loader and Validation Layer components. They are not git sub-modules of Vulkan-LoaderAndValidationLayers but Vulkan-LoaderAndValidationLayers is linked to a specific revision of gslang. This can be automatically cloned and built to predefined locations with the update\_external\_sources scripts. If a custom configuration is required, do the following steps:
```
karl-lunarg: Missed a "They".

BUILD.md
```diff
@@ -389,34 +389,30 @@ adb shell am start -a android.intent.action.MAIN -c android-intent.category.LAUN
...
 
-1) clone the repos:
+1) clone the repo:
 
     git clone https://github.com/KhronosGroup/glslang.git
-    git clone https://github.com/KhronosGroup/SPIRV-Tools.git
```
karl-lunarg: Removing these docs may be a little questionable.  We're removing the documentation for how to do the custom spirv, but a person could still do it, even if glslang pulls it in.  Lines 217 and 218 in the top-level CMakeLists.txt file are still there to support this activity.  And there are the CUSTOM_PATH CMake variables that are probably still in there as well.  glslang will build with its spirv, but the rest of LVL could still be built with a custom spirv.  The revision files for spirv are going away, but then if someone is doing a custom spirv, they might be using a different revision anyway.

All that being said, I suppose that taking this out is better.   If a person wants to tweak spirv, they can do it within the glslang repo.

CMakeLists.txt
```diff
@@ -212,11 +212,10 @@ find_path(GLSLANG_SPIRV_INCLUDE_DIR SPIRV/spirv.hpp HINTS "${EXTERNAL_SOURCE_ROO
                                                     "${CMAKE_SOURCE_DIR}/../glslang"
                                               DOC "Path to SPIRV/spirv.hpp")
 
-find_path(SPIRV_TOOLS_INCLUDE_DIR spirv-tools/libspirv.h HINTS "${EXTERNAL_SOURCE_ROOT}/spirv-tools/include"
-                                                   "${EXTERNAL_SOURCE_ROOT}/SPIRV-Tools/include"
+find_path(SPIRV_TOOLS_INCLUDE_DIR spirv-tools/libspirv.h HINTS "${EXTERNAL_SOURCE_ROOT}/glslang/External/spirv-tools/include"
+                                                   "${EXTERNAL_SOURCE_ROOT}/glslang/External/SPIRV-Tools/include"
                                                    "${CMAKE_SOURCE_DIR}/../spirv-tools/include"
                                                    "${CMAKE_SOURCE_DIR}/../SPIRV-Tools/include"
```
karl-lunarg: 217 and 218 can come out, per above discussion.

CMakeLists.txt
```diff
@@ -212,11 +212,10 @@ find_path(GLSLANG_SPIRV_INCLUDE_DIR SPIRV/spirv.hpp HINTS "${EXTERNAL_SOURCE_ROO
...
-                                                   "${EXTERNAL_SOURCE_ROOT}/source/spirv-tools/external/include"
```
karl-lunarg: I agree this can be removed, and I think it is just from a really old way of doing things.

BUILD.md
```diff
@@ -389,34 +389,30 @@ adb shell am start -a android.intent.action.MAIN -c android-intent.category.LAUN
...
 *non windows*
 
     git checkout `cat [path to Vulkan-LoaderAndValidationLayers]\glslang_revision` [in glslang repo]
-	git checkout `cat [path to Vulkan-LoaderAndValidationLayers]\spirv-tools_revision` [in spriv-tools repo]
 
-3) Configure the gslang and spirv-tools source trees with cmake and build them with your IDE of choice
+3) Configure the gslang source tree with cmake and build it with your IDE of choice
 
-4) Enable the CUSTOM\_GSLANG\_BIN\_PATH and CUSTOM\_SPIRV\_TOOLS\_BIN\_PATH options in the Vulkan-LoaderAndValidationLayers cmake configuration and point the GSLANG\_BINARY\_PATH and SPIRV\_TOOLS\_BINARY\_PATH variables to the correct location
+4) Enable the CUSTOM\_GSLANG\_BIN\_PATH option in the Vulkan-LoaderAndValidationLayers cmake configuration and point the GSLANG\_BINARY\_PATH variable to the correct location
```
karl-lunarg: I think you have to keep these.   If a person sets up a custom glslang with a recent glslang repo, they will get spirv in the custom location as well and may wish to point at the binaries in the custom location (under glslang).

CMakeLists.txt
karl-lunarg: or rather changed to
```
"${CMAKE_SOURCE_DIR}/../glslang/External/spirv-tools/include"
"${CMAKE_SOURCE_DIR}/../glslang/External/SPIRV-Tools/include"

Are both the spirv-tools and SPIRV-Tools forms needed, even under glslang/External?

update_external_sources.bat
mark-lunarg: I was under the impression that '--all' still does the same thing as before -- glslang gets built regardless.  I will add messages (similar to the new one for --all) for -s and -spirv-tools.

update_external_sources.bat
mark-lunarg: If they use the a deprecated option they'll get a message. Seems unnecessary to add stuff in the usage for these guys.  Will update BUILD.md though.

update_external_sources.sh
mark-lunarg: See above.

CMakeLists.txt
mark-lunarg: K.

BUILD.md
mark-lunarg: Right.

BUILD.md
mark-lunarg: Fixed.

BUILD.md
mark-lunarg: Fixed.

update_external_sources.bat
karl-lunarg: No, if I am reading your change correctly, specifying only "--all" just skips to the next parameter.  It should set the flag to build all components, and in this case, there is only one: glslang.

karl-lunarg:
The AppVeyor build took 50 mins instead of 60-ish, so that should help the timeout problem.  Nice!

Issue #2180, opened 2017-11-03
Add validation for sample rate shading


SaschaWillems:
There doesn't seem to be any validation for enabling sample rate shading. As per spec the validation should check the value of the ```VkPipelineMultisampleStateCreateInfo``` ```sampleShadingEnable``` member against the ```sampleRateShading``` device feature.

tobine:
FYI this is VALIDATION_ERROR_10000620 in DB.

jzulauf-lunarg:
Fix merged in #2220

Issue #2159, opened 2017-10-25
pNext chain not correctly validated


MaciejJesionowskiAMD:
I think there's a mistake in the code that verifies pNext chains in structures. It seems to recognize which structure types can be attached there, but it doesn't actually use them with the verification function. This is an example with `EXT_sample_locations`:

> ParameterValidation(WARN): object: 0x0 type: 0 location: 538 msgCode: 268551181: vkCreateGraphicsPipelines: value of pCreateInfos[0].pMultisampleState->pNext must be NULL. The spec valid usage text states 'Each pNext member of any structure (including this one) in the pNext chain must be either NULL or a pointer to a valid instance of VkPipelineCoverageModulationStateCreateInfoNV, VkPipelineCoverageToColorStateCreateInfoNV, or **VkPipelineSampleLocationsStateCreateInfoEXT**' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkPipelineMultisampleStateCreateInfo-pNext-pNext) This warning is based on the Valid Usage documentation for version 62 of the Vulkan header. It is possible that you are using a struct from a private extension or an extension that was added to a later version of the Vulkan header, in which case your use of pCreateInfos[0].pMultisampleState->pNext is perfectly valid but is not guaranteed to work correctly with validation enabled

The structure `VkPipelineSampleLocationsStateCreateInfoEXT` is definitely known to the layer, but it's not used by parameter validation:

parameter_validation_utils.cpp#L1171
```c++
skip |= validate_struct_pnext(
    report_data,
    "vkCreateGraphicsPipelines",
    ParameterName("pCreateInfos[%i].pMultisampleState->pNext", ParameterName::IndexVector{i}),
    NULL,  // const char *allowed_struct_names <-- it's empty
    pCreateInfos[i].pMultisampleState->pNext,
    0,     // size_t allowed_type_count
    NULL,
    GeneratedHeaderVersion,
    VALIDATION_ERROR_1001c40d);
```

jzulauf-lunarg:
Fix merged in #2220

PR #2220, opened 2017-11-20
Fixes for pipeline multi sample issue 2159 and 2180

jzulauf-lunarg:
Fixes for https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/issues/2180 and https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/issues/2159

Includes a common code refactor of the CreateGraphicsPipeline setup code for layer_validation_tests.cpp, used by the new cases, but presented here for review/feedback/refinement before refactoring the 20+ other copies in extant test cases to use it.


tests/layer_validation_tests.cpp
```diff
@@ -173,13 +176,18 @@ struct NearestGreaterThan {
 };
 template <typename T, typename Comp>
 T NearestNonEqual(const T value, const Comp &comp) {
```
jzulauf-lunarg: The previous version had some numerical corner cases uncovered when expanding the usage.  This is simpler and while it has an iterative fallback (for further missed corners), I couldn't find a test case that uses it.  There are some ULP differences between this and the prior implementation that don't matter to layer validation testing.

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
         return pool_ != VK_NULL_HANDLE && layout_ != VK_NULL_HANDLE && set_ != VK_NULL_HANDLE;
     }
 };
+
+struct CreatePipelineHelper {
```
jzulauf-lunarg: This refactors out common setup code for testing CreateGraphicsPipelines.  For typical usage see OneshotTest below

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+        }
+        if (implicit_destroy && (pipeline_ != VK_NULL_HANDLE)) {
+            vkDestroyPipeline(layer_test_->device(), pipeline_, nullptr);
+            pipeline_ = VK_NULL_HANDLE;
+        }
+        err = vkCreateGraphicsPipelines(layer_test_->device(), pipeline_cache_, 1, &gp_ci_, NULL, &pipeline_);
+        return err;
+    }
+    template <typename Test, typename OverrideFunc, typename Error>
+    static void OneshotTest(Test *test, OverrideFunc &info_override, const VkFlags flags, Error error, bool positive_test = false) {
```
jzulauf-lunarg: The usage here is similar to the rest of the framework, init, twiddle, init some more, test.  Oneshot works for all the cases where simple twiddling with the create info structs (and the things they point to) in info_override is sufficient.

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+        helper.CreateGraphicsPipeline();
+
+        if (positive_test) {
+            test->Monitor()->VerifyNotFound();
+        } else {
+            test->Monitor()->VerifyFound();
+        }
+    };
+};
+namespace chain_util {
```
jzulauf-lunarg: More common code removal for a pattern in the pNext test added below

tests/layer_validation_tests.cpp
```diff
@@ -9044,6 +9347,84 @@ TEST_F(VkLayerTest, InvalidPipelineCreateState) {
...
+    VkPhysicalDeviceFeatures device_features = {};
+    ASSERT_NO_FATAL_FAILURE(GetPhysicalDeviceFeatures(&device_features));
+    device_features.sampleRateShading = VK_FALSE;
+
+    ASSERT_NO_FATAL_FAILURE(InitState(&device_features));
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+
+    // Cause the error by enabling sample shading...
+    auto set_shading_enable = [](CreatePipelineHelper &helper) { helper.pipe_ms_state_ci_.sampleShadingEnable = VK_TRUE; };
+    CreatePipelineHelper::OneshotTest(this, set_shading_enable, VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_10000620);
```
jzulauf-lunarg: Typical OneshotTest usage.  The info_override is a simple one-liner.

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+    ~CreatePipelineHelper() {
+        VkDevice device = layer_test_->device();
+        vkDestroyPipelineCache(device, pipeline_cache_, nullptr);
+        vkDestroyPipeline(device, pipeline_, nullptr);
+        vkDestroyPipelineLayout(device, pipeline_layout_, nullptr);
+        vkDestroyDescriptorSetLayout(device, ds_layout_, nullptr);
+        vkDestroyDescriptorPool(device, ds_pool_, nullptr);
+    }
+
+    void InitDescriptorPoolInfo() {
```
chrisforbes: Reuse OneOffDescriptorSet?

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+    VkPipelineDynamicStateCreateInfo dyn_state_ci_ = {};
+    std::vector<VkPipelineShaderStageCreateInfo> shader_stages_;
+    VkPipelineRasterizationStateCreateInfo rs_state_ci_ = {};
+    VkPipelineColorBlendAttachmentState cb_attachments_ = {};
+    VkPipelineColorBlendStateCreateInfo cb_ci_ = {};
+    VkGraphicsPipelineCreateInfo gp_ci_ = {};
+    VkPipelineCacheCreateInfo pc_ci_ = {};
+    VkPipeline pipeline_ = VK_NULL_HANDLE;
+    VkPipelineCache pipeline_cache_ = VK_NULL_HANDLE;
+    std::unique_ptr<VkShaderObj> vs_;
```
chrisforbes: Why separate VS/FS members? The tests you want to replace use a mix of stages.

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +956,266 @@ struct OneOffDescriptorSet {
...
+
+struct CreatePipelineHelper {
+   public:
+    std::vector<VkDescriptorPoolSize> ds_pool_sizes_;
+    VkDescriptorPoolCreateInfo ds_pool_ci_ = {};
+    VkDescriptorPool ds_pool_ = VK_NULL_HANDLE;
+    std::vector<VkDescriptorSetLayoutBinding> dsl_bindings_;
+    VkDescriptorSetLayoutCreateInfo ds_layout_ci_ = {};
+    VkDescriptorSetLayout ds_layout_ = VK_NULL_HANDLE;
+    VkDescriptorSet descriptorSet_ = VK_NULL_HANDLE;
```
tobine: descriptor_set_

tests/layer_validation_tests.cpp
jzulauf-lunarg: Looking further into the test cases, VkPipelineObj should be something I'm using instead as well.

tests/layer_validation_tests.cpp
jzulauf-lunarg: Most are VS/FS only, but looking at adapting/replacing the helper with use of the VkPipelineObj (which I didn't notice before), and that supports more stages AFAICT

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+        //    VkPipelineVertexInputStateCreateInfo
+        //    VkPipelineInputAssemblyStateCreateInfo
+        //    VkPipelineViewportStateCreateInfo
+        //    VkPipelineRasterizationStateCreateInfo
+        //    VkPipelineMultisampleStateCreateInfo
+        //    VkPipelineColorBlendStateCreateInfo
+        gp_ci_.sType = VK_STRUCTURE_TYPE_GRAPHICS_PIPELINE_CREATE_INFO;
+        gp_ci_.pNext = nullptr;
+        gp_ci_.flags = VK_PIPELINE_CREATE_DISABLE_OPTIMIZATION_BIT;
+        gp_ci_.pVertexInputState = &vi_ci_;
```
krOoze: `stageCount` and `pStages` not set here. Should at least note that.

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+        gp_ci_.pInputAssemblyState = &ia_ci_;
+        gp_ci_.pTessellationState = nullptr;
+        gp_ci_.pViewportState = &vp_state_ci_;
+        gp_ci_.pRasterizationState = &rs_state_ci_;
+        gp_ci_.pMultisampleState = &pipe_ms_state_ci_;
+        gp_ci_.pDepthStencilState = nullptr;
+        gp_ci_.pColorBlendState = &cb_ci_;
+        // If pDynamicState is pointing to dyn_state_ci_, it will late bind to dyn_states_ in LateBindCreatePipeline
+        gp_ci_.pDynamicState = &dyn_state_ci_;
+        gp_ci_.renderPass = layer_test_->renderPass();
```
krOoze: rest seems mostly explicitly initialized here. should say `subpass = 0`?

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+        InitDescriptorPoolInfo();
+        InitDescriptorSetInfo();
+        InitInputAndVertexInfo();
+        InitMultisampleInfo();
+        InitPipelineLayoutInfo();
+        InitViewportInfo();
+        InitDynamicStateInfo();
+        InitShaderInfo();
+        InitRasterizationInfo();
+        InitBlendStateInfo();
```
krOoze: These should preferably be in order they appear in `VkGraphicsPipelineCreateInfo`.

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+        InitGraphicsPipelineInfo();
+        InitPipelineCacheInfo();
+    }
+
+    void InitState() {
+        VkResult err;
+        ds_pool_ci_.poolSizeCount = ds_pool_sizes_.size();  // Late bind from vector, to allow modification
```
krOoze: Should be protected from 0 size.

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+        ds_pool_ci_.pPoolSizes = ds_pool_sizes_.data();
+        err = vkCreateDescriptorPool(layer_test_->device(), &ds_pool_ci_, NULL, &ds_pool_);
+        ASSERT_VK_SUCCESS(err);
+
+        ds_layout_ci_.bindingCount = dsl_bindings_.size();  // Late bind from vector, to allow modifications
```
krOoze: Should be protected from 0 size.

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+        InitBlendStateInfo();
+        InitGraphicsPipelineInfo();
+        InitPipelineCacheInfo();
+    }
+
+    void InitState() {
+        VkResult err;
+        ds_pool_ci_.poolSizeCount = ds_pool_sizes_.size();  // Late bind from vector, to allow modification
+        ds_pool_ci_.pPoolSizes = ds_pool_sizes_.data();
+        err = vkCreateDescriptorPool(layer_test_->device(), &ds_pool_ci_, NULL, &ds_pool_);
```
krOoze: does not really allow to set `ds_pool_`. But `ds_pool_` is public! Should only create if it is `VK_NULL_HANDLE` ?

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+
+        // Handle value in info struct, need to late bind here
+        alloc_info_.descriptorPool = ds_pool_;
+        err = vkAllocateDescriptorSets(layer_test_->device(), &alloc_info_, &descriptorSet_);
+        ASSERT_VK_SUCCESS(err);
+
+        err = vkCreatePipelineLayout(layer_test_->device(), &pipeline_layout_ci_, NULL, &pipeline_layout_);
+        ASSERT_VK_SUCCESS(err);
+
+        err = vkCreatePipelineCache(layer_test_->device(), &pc_ci_, NULL, &pipeline_cache_);
```
krOoze: I wonder if it helps anything...

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
         return pool_ != VK_NULL_HANDLE && layout_ != VK_NULL_HANDLE && set_ != VK_NULL_HANDLE;
     }
 };
+
+struct CreatePipelineHelper {
+   public:
+    std::vector<VkDescriptorPoolSize> ds_pool_sizes_;
+    VkDescriptorPoolCreateInfo ds_pool_ci_ = {};
+    VkDescriptorPool ds_pool_ = VK_NULL_HANDLE;
+    std::vector<VkDescriptorSetLayoutBinding> dsl_bindings_;
```
krOoze: could just be private static array?
the user can use his own vector, or modify the default array through the `ds_layout_ci_.pBindings`.

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+        ASSERT_VK_SUCCESS(err);
+
+        err = vkCreatePipelineLayout(layer_test_->device(), &pipeline_layout_ci_, NULL, &pipeline_layout_);
+        ASSERT_VK_SUCCESS(err);
+
+        err = vkCreatePipelineCache(layer_test_->device(), &pc_ci_, NULL, &pipeline_cache_);
+        ASSERT_VK_SUCCESS(err);
+    }
+
+    void LateBindPipelineInfo() {
```
krOoze: Should be private? The `CreateGraphicsPipeline` seems to call it implicitly anyway, so probably no need to call it explicitly

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+            dyn_state_ci_.pDynamicStates = dyn_states_.data();
+        }
+    }
+
+    VkResult CreateGraphicsPipeline(bool implicit_destroy = true, bool do_late_bind = true) {
+        VkResult err;
+        if (do_late_bind) {
+            LateBindPipelineInfo();
+        }
+        if (implicit_destroy && (pipeline_ != VK_NULL_HANDLE)) {
```
krOoze: `vkDestroy*` should be able to destroy `VK_NULL_HANDLE`. Though there used to be some driver bugs, so probably better.

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+        gp_ci_.stageCount = shader_stages_.size();
+        gp_ci_.pStages = shader_stages_.data();
+        if (gp_ci_.pDynamicState == &dyn_state_ci_) {
+            // Late bind the dyn_states_ array
+            dyn_state_ci_.dynamicStateCount = dyn_states_.size();
+            dyn_state_ci_.pDynamicStates = dyn_states_.data();
+        }
+    }
+
+    VkResult CreateGraphicsPipeline(bool implicit_destroy = true, bool do_late_bind = true) {
```
krOoze: `implicit_destroy` is bit confusing name. I did not really expect it to do that. Maybe `destroy_previous_pipeline`. Even so it would be weird the old pipeline to just get leaked.
Maybe just create the pipelines into a stack?

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+        return err;
+    }
+    template <typename Test, typename OverrideFunc, typename Error>
+    static void OneshotTest(Test *test, OverrideFunc &info_override, const VkFlags flags, Error error, bool positive_test = false) {
+        CreatePipelineHelper helper(test);
+        helper.InitInfo();
+        info_override(helper);
+        helper.InitState();
+
+        test->Monitor()->SetDesiredFailureMsg(flags, error);
```
krOoze: Does this work?
The positive tests usually set something like `m_errorMonitor->ExpectSuccess();`.

tests/layer_validation_tests.cpp
```diff
@@ -946,6 +958,297 @@ struct OneOffDescriptorSet {
...
+        shader_stages_[0] = vs_->GetStageCreateInfo();
+        shader_stages_[1] = fs_->GetStageCreateInfo();
+    }
+
+    void InitRasterizationInfo() {
+        rs_state_ci_.sType = VK_STRUCTURE_TYPE_PIPELINE_RASTERIZATION_STATE_CREATE_INFO;
+        rs_state_ci_.pNext = nullptr;
+        rs_state_ci_.flags = 0;
+        rs_state_ci_.depthClampEnable = VK_FALSE;
+        rs_state_ci_.rasterizerDiscardEnable = VK_FALSE;
```
krOoze: This may be very commonly used setting. Might be worth to make some shortcut for setting `VK_TRUE`.

tests/layer_validation_tests.cpp
jzulauf-lunarg: Converting to std::nextafter

krOoze: @jzulauf-lunarg From experience, it wont pass the Android CI here. You migt have to use the C version of the function.

layers/vk_validation_error_database.txt
```diff
@@ -1143,12 +1143,12 @@ VALIDATION_ERROR_0fe1c40d~^~Y~^~Unknown~^~vkCreatePipelineLayout~^~VUID-VkPipeli
...
+VALIDATION_ERROR_1001c40d~^~Y~^~InvalidPipelineSamplePNext~^~vkCmdSetViewport~^~VUID-VkPipelineMultisampleStateCreateInfo-pNext-pNext~^~core~^~The spec valid usage text states 'Each pNext member of any structure (including this one) in the pNext chain must be either NULL or a pointer to a valid instance of VkPipelineCoverageModulationStateCreateInfoNV, VkPipelineCoverageToColorStateCreateInfoNV, or VkPipelineSampleLocationsStateCreateInfoEXT' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkPipelineMultisampleStateCreateInfo-pNext-pNext)~^~implicit, TBD in parameter validation layer.
```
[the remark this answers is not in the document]
jzulauf-lunarg: There are improvements needed that will best be done in enhancements to code generation

jzulauf-lunarg:
Can I get a thumbs up or a set of requested changes?

mark-lunarg:
LGTM.

PR #2222, opened 2017-11-20
Implement viewport checks

krOoze:
- Implement checks of `VkViewport` checks in `vkCreateGraphicsPipelines` and `vkCmdSetViewport`
- Implement checks of other params of `vkCmdSetViewport`
- \+ relevant tests


layers/parameter_validation_utils.cpp
```diff
@@ -903,6 +903,171 @@ bool pv_vkCreateImageView(VkDevice device, const VkImageViewCreateInfo *pCreateI
...
+        // f can be converted to uint32_t with well-defined behavior (trunc)
+        return {true, static_cast<uint32_t>(f), f == truncf(f)};
+    };
+
+    // width
+    bool width_healthy = true;
+    const auto u32_w = float_to_u32(viewport.width);
+    const auto max_w = device_data->device_limits.maxViewportDimensions[0];
+    const bool w_within_limits =
+        viewport.width > 0.0 && u32_w.success && u32_w.result <= max_w && (u32_w.result < max_w || u32_w.exact);
```
jzulauf-lunarg: By preference comparisons should be enclosed in () for clarity.  (a >b) && (c !=d)

jzulauf-lunarg: Given "width must be greater than 0.0 and less than or equal to VkPhysicalDeviceLimits::maxViewportDimensions[0]"  It looks like the last clause is looking for the case where f is out of range but trunc f is in range, but given the precision limits, this is likely to have equal false negatives and positives.

The spec is having us compare a uint32_t limit with a floating point value that is only accurate to +/-256 (for IEEE single precision) at uint32 max.  I'd think it may be simpler to promote the limit to float and compare, and no worse at the boundary case. (actually it's worse than that, the spec only requires a floating point that is accurate to "1 part in 1.0E5" -- or roughly +/-20,000 at uint32_t max)

layers/parameter_validation_utils.cpp
```diff
@@ -903,6 +903,171 @@ bool pv_vkCreateImageView(VkDevice device, const VkImageViewCreateInfo *pCreateI
...
+
+    // Note: for numerical correctness, float comparisons must be formed in a way to assume NaN (comparison always false).
+    //       VkPhysicalDeviceLimits::maxViewportDimensions is uint32_t, not float -> careful.
+    struct ConvProduct {
+        bool success;
+        uint32_t result;
+        bool exact;
+    };
+    const auto float_to_u32 = [](const float f) -> ConvProduct {
+        if (!(f >= 0.0)) return {false};  // negative or NaN -> failed conversion
```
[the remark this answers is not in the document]
krOoze: correct.
shouldn't matter though; afaik `double` is a superset of `float`. It is one of the healthier implicit conversions in C++.

layers/parameter_validation_utils.cpp
```diff
@@ -903,6 +903,171 @@ bool pv_vkCreateImageView(VkDevice device, const VkImageViewCreateInfo *pCreateI
...
+
+        const float one_past_max_u32 = exp2f(32.0f);      // should be absolute precission
```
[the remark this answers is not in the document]
krOoze: You cannot convert arbitrary `uint32_t` into `float` precisely. Especially UINT32_MAX. Some lowers bits would get sliced.

layers/parameter_validation_utils.cpp
```diff
@@ -903,6 +903,171 @@ bool pv_vkCreateImageView(VkDevice device, const VkImageViewCreateInfo *pCreateI
...
+        if (f >= one_past_max_u32) return {false};        // more than fits uint32_t -> failed conversion
```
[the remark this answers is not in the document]
krOoze: I don't follow?

layers/parameter_validation_utils.cpp
krOoze: Correct.
But since I am using a local helper function anyway I could perhaps just include it in the `u32_w.success`.
I don't see any false negatives and positives. As said above the function is supposed to have absolute precission for the purposes of the validation check.

layers/parameter_validation_utils.cpp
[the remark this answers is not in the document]
jzulauf-lunarg: all integer values from 294967040.0 <   and < 4294967808.0 mapped to the floating point value 4294967296.0.  This means that you are treating a floating point value that equally well represents
(294967040.0 ,   4294967295], as being overflow.  The test should be for f > one_past_max_u32, with the equality case mapping to 4294967295 (uint32max).

layers/parameter_validation_utils.cpp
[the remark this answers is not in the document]
jzulauf-lunarg: I think we're in agreement in principle, but given the lack of precision at uint32_max... the cast is as good as the exp2f.  The representation is only accurate to +/-256 integer values. thus max + 1 maps to the same bit pattern as max (since max is a pow2)

casting UINT32_MAX to float yields 4294967296.0
the next lower representable integer is : 4294967040.0, integers > that value casts to 4294967296.0 
the next higher representable integer is: : 4294967808.0, integers < that value cast to 4294967296.0

layers/parameter_validation_utils.cpp
krOoze: PS: I have opposite preference. IMO there's no sensible ambiguity of operator precedence; writing parentheses here would be as anoying as writing parentheses after an assignment operator.

jzulauf-lunarg: Given that there is nothing one can do to test for exact bounds when the ULP integer differences are aliases in floating point (for values > 2^24, roughly), I'm suggesting that converting all values to the common lower precision and then bounds testing would make more sense, especially when fraction width and height values are valid. 

Trunc'ing in the case where the values for the limits are fully representable in float actually fails to detect the case were trunc(width) <= max but width > max.

mark-lunarg: +1 for the parentheses aiding readability and clarity.

layers/parameter_validation_utils.cpp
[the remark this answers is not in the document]
krOoze: Technically the results happens to be the same, but `static_cast<float>(std::numeric_limits<uint32_t>::max())` feels like a hack. (Also ugly.)
`exp2f(32.0f)` is nice and explicit.

layers/parameter_validation_utils.cpp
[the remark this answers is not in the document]
krOoze: Right. That's why am I converting the `float` to `uint32_t`. Not other way around.

layers/parameter_validation_utils.cpp
krOoze: @mark-lunarg You ask, and so you shall receive.

krOoze: @jzulauf-lunarg 
> Given that there is nothing one can do

How so? I believe I did **something**.
I converted the `width` (`float`) to `uint32_t` (exactly, per 4.9 of the C++14 draft specification) + detection of remainder. Now I can determine (exactly) if the given `float` is within the given `uint32_t` limit.

layers/parameter_validation_utils.cpp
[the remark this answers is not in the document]
krOoze: >since max is a pow2

no, max is 2<sup>N</sup> - 1.
Buuut, max+1 is 2<sup>N</sup>, which is what this code leverages.

layers/parameter_validation_utils.cpp
jzulauf-lunarg: I don't believe you can.  For values above 16.8 million float(x+1.F) == x, this means that you cannot determine the value j from the value i in:

unsigned i = unsigned(float(j));

All we can say is that j is of a given set of values.  For all positive values < ~16.7M there is a one to one correspondence j to i, past that it becomes N to 1.  For the case we are looking at UINT32_MAX the correlation of j to i is about 500 to 1.  There are about 250 values < UINT32_MAX that have the same float point value as exp2f(32.f).  It is as valid to say that (within limits of precision) exp2f(32.f) represents the values <, ==,  and > UINT32_MAX.  A round-trip cast to and from an uint64t would all yield UINT32_MAX, even for values more that 100 < the boundary.   There simply isn't the precision to make the comparison you want, because the states you wish to discriminate between all have the same representation in float.   You're thus left with ambiguity.

Consider as the alternative  instead casting the uint32 limiting value to float.  Any comparison with another float is unambiguous. Potentially precision has been lost (most definitely so for values > 16.77M), but the same loss of precision affects both values.

If I set my viewport width to float(UINT32_MAX) and my maxViewportDimensions[0] is also (UINT32_MAX) the <= comparison will be correct.  For any representable state > float(UINT32_MAX) the test will fail.

That's not to say that (float)((long long)UINT32_MAX +1LL) won't also pass, the floating point state is also identify with (float)UINT32_MAX.  

This will be true for any point in between 16.77M and UINT32_MAX, values compared to a float of the limit will be <= the limit for all values that cast to the same floating point value.

Note that we cannot do any better by promoting to uint32_t, because we cannot do so with recognizing the inherent ambiguity -- that is to say that a exp2f(32.f) == (float)UINT32_MAX

krOoze: IMO you keep knocking some kind of straw-man.
Let's try it this way: I will give you a code snippet, and you tell me on which line it conceptually fails and why.

```
// check if `value` is less than or equal to `limit` -- numeric hell edition
bool is_in_limit( float value, uint32_t limit ) {
  if( is_NAN(value) ) return false;
  if( value <= 0.0f ) return true;

  // both 32 and 2^32 are 2^N -> safe and exactly UINT32_MAX + 1
  float u32_max_plus_one = exp2f(32.0f);

  // value clearly > UINT32_MAX .999999999
  if( value >= u32_max_plus_one ) return false; 

  // well defined by C++ standard if the truncated part <= UINT32_MAX -> truncates toward zero
  uint32_t value_u32 = static_cast<uint32_t>(value);

  if( value_u32 < limit ) return true; // obviously fits limit
  if( value_u32 > limit ) return false; // obviously outside limit

  // ambiguous if before truncation was value == limit, or value == limit.000something
  if( value_u32 == limit ){
    if( truncf(value) == value ) return true; // truncation remainder == 0.0f
    else return false; // truncation remainder > 0.0f
  }
}

File: layers/parameter_validation_utils.cpp (hunk 4 above)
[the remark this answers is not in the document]
Comment (jzulauf-lunarg) [reply layers/parameter_validation_utils.cpp]: Two more values to consider:

float f3 = 4294967295.0f;  // i.e. static_cast<float>(UINT32_MAX)
float f4 = 4294967041.0f; // least value that casts to the same representation as exp2f(32.0f)

Both are inside of the limit, but have the same representation as f2 and thus fail.
limit and limit+1 are all the same floating point value, but the code assumes that value only represents limit+1, which it does *exactly* but neither precisely, nor uniquely.  It represents limit+1, +/- ~256.

File: layers/parameter_validation_utils.cpp (hunk 1 above)
Comment (jzulauf-lunarg) [reply layers/parameter_validation_utils.cpp]: if(value >= u32_max_plus_one) return false.
The comparison fails for values <= UINT32_MAX because 
`(float)UINT32_MAX == u32_max_plus_one` evaluates to *true*

So passing `static_cast<float>(UINT32_MAX)` to is_in_limit would return false.

Comment (krOoze) [reply layers/parameter_validation_utils.cpp]: Ah I see.
That is the users numerical problem, not mine.
I get a `float`, I test a `float`.

I cannot be responsible for what the user **meant** the value to be. I only test what the value **is**.
If he screws integral-floating conversion, that is his error.
If he assigns a float (base10) literal that cannot be represented exactly and the compiler rounds it up -> also his error.

Comment (jzulauf-lunarg) [reply layers/parameter_validation_utils.cpp]: If the user specifies:

'viewport.width = static_cast<float>(device_physical_limit.maxViewportDimensions[0])'

This should *never* cause a validation error.

The problem (if there is one) lies in the standard specifying the quantity and the limit at to different precision levels.  It is *not* a valid validation_layer check to report an error in such a case.

The code mistakes what the value *is* with *is precisely*.  The representation of that *exact* value is not precise at the granularity needed to perform the comparison -- it represents a range of values on both the pass and fail sides of the test.  The limit check failure is an artifact of assuming a *precision* that doesn't exist. 

If you were to compare this in the float domain, no such ambiguity (nor false errors) would exist, as we are doing the comparison at the highest precision that does exist.

Comment (krOoze) [reply layers/parameter_validation_utils.cpp]: > viewport.width = static_cast<float>(device_physical_limit.maxViewportDimensions[0])
This should never cause a validation error.

Why not?

Counter:
should
`viewport.width = static_cast<float>(device_physical_limit.maxViewportDimensions[0] + 1)`
work sometimes?

> The problem (if there is one) lies in the standard specifying the quantity and the limit at two different precision levels.

I agree. That's something the standard could improve.
Though I think you mean "accuracy". As said above we cannot play mind-reader, and guess what the user meant by the value.

> viewport.width = static_cast<float>(device_physical_limit.maxViewportDimensions[0])

The user is performing a dangerous operation. It is his problem, not ours. He needs to read the 4.9.2 section of the C++14 specification. We can only assume he knows what is he doing and hope he will not shoot his leg off.

Besides that is a slipery slope.
What if he does
` viewport.width = ((float)maxViewportDimensions[0] / 3.0f + 1.0f) * 3.0f +3.0f`
or
`viewport.width = float(maxViewportDimensions[0] - 1) +1.0f`

Should we still protect him from those mistakes? Would drivers even be so nice (I don't think they would).

There are pretty much two interpretation of the spec's limit:
1) the float has to be under the uint limit under all situations
2) The limit is converted to float with rounding away from zero and the float has to be under that number

I think 1) is more sane, even if it may sometimes be harder to work with for the user.

Comment (jzulauf-lunarg) [reply layers/parameter_validation_utils.cpp]: "Unlikely to be caught in the wild" is certainly true.  a 16.7M wide viewport would belong to a giga-pixel monitor with any reasonable aspect ratio... unless somebody is computing a viewport way out into the guardband.  The largest limits seen in the wild is 32K ... http://vulkan.gpuinfo.org/listlimits.php

So, as it's an academic question currently, I'd advise the simplest implementation (cast limiting value to float and compare). I think an issue needs to be opened with the spec to clarify this for the future.

For the spec question: I think the operable  principle is:

"should a limit ever be specified at higher precision than the limited value and if so when given limits of  different precision and range from the limit value, should validation ever flag as an error:

some_struct.lower_precision_field = static_cast<lp_type>(higher_precision_limit);

where the requirement is lower_precision_field <= higher_precision_limit"

One answer is "yes, because always want to catch:

lower_precision_field = static_cast<lp_type>(higher_precision_limit + delta) 

as an error ."

However, if the only "error" we are catching is when the two lower_precision values are identical, and we are flagging values identical to valid values as invalid, isn't this the definition of "false positive".

Comment (krOoze) [reply layers/parameter_validation_utils.cpp]: So reading the specification:

- `maxViewportDimensions` seems intended to be the maximal dimensions of a scratchpad the implementation can make, so the results should **always** fit inside this limit

The vp transformation formula given is
x<sub>f</sub> = (p<sub>x</sub>  / 2) x<sub>d</sub>  + o<sub>x</sub> 

I.e.
fb.x = (vp.width / 2) * x<sub>d</sub> + vp.x + vp.width / 2
x<sub>d</sub> is NDC, so [-1.0f, 1.0f]

Which could be something like:
fb.x = (vp.width * 0.5f * (x<sub>d</sub> + 1.0f ) + vp.x

So the scratchpad transformation would be
sp.x = (vp.width * 0.5f * (x<sub>d</sub> + 1.0f )

For x<sub>d</sub> -1.0f we get sp.x = 0.0, which converts well to 0. Good.
For x<sub>d</sub> 1.0f we get sp.x = vp.width. If the vp.width == exactly UINT32_MAX + 1, then it would be converted to UINT32_MAX + 1, and so wouldn't fit in the scratchpad (even if it happens to be the same as `static_cast<float>(UINT32_MAX)`).
The implementation would have to clamp given vp.width to work correctly, which defeats the purpose of having a VU.

This leads me to believe my implementation is the right way to do this.
But it couldn't hurt to ask the Vulkan specifiers...

Comment (krOoze) [reply layers/parameter_validation_utils.cpp]: > So, as it's an academic question currently, I'd advise the simplest implementation (cast limiting value to float and compare).

That's fine with me.
I can also add an `assert` to make sure.

**Comment by krOoze** [comment]
Rebased, implemented suggestions, improved `height` checks readability.

**Comment by krOoze** [comment]
Disclaimer: Probably not worth rewieving now, because I have PR in Vulkan that will invalidate all of those VUIDs, and this will have to be modified.

**Comment by krOoze** [comment]
Can I has update to 1.0.67, so I can continue?

**Comment by krOoze** [comment]
Updated with 1.0.67 changes.

File: layers/parameter_validation_utils.cpp [hunk 5]
```diff
@@ -962,6 +962,138 @@ bool pv_vkCreateImageView(VkDevice device, const VkImageViewCreateInfo *pCreateI
...
+    const bool negative_height_enabled =
+        device_data->extensions.vk_khr_maintenance1 || device_data->extensions.vk_amd_negative_viewport_height;
+    const auto max_h = device_data->device_limits.maxViewportDimensions[1];
+
+    if (!negative_height_enabled && !(viewport.height > 0.0f)) {
+        height_healthy = false;
+        skip |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, object_type, object, __LINE__, VALIDATION_ERROR_15000dd8,
+                        LayerName, "%s: %s.height (=%f) is not greater 0.0. %s", fn_name, param_name, viewport.height,
+                        validation_error_map[VALIDATION_ERROR_15000dd8]);
+    } else if (!f_lte_u32(fabsf(viewport.height), max_h)) {
```
jzulauf-lunarg: std::abs or std::fabs?

layers/parameter_validation_utils.cpp
```diff
@@ -962,6 +962,138 @@ bool pv_vkCreateImageView(VkDevice device, const VkImageViewCreateInfo *pCreateI
...
+    // y
+    bool y_healthy = true;
+    if (!(viewport.y >= device_data->device_limits.viewportBoundsRange[0])) {
+        y_healthy = false;
+        skip |=
+            log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, object_type, object, __LINE__, VALIDATION_ERROR_15000dde, LayerName,
+                    "%s: %s.y (=%f) is less than VkPhysicalDeviceLimits::viewportBoundsRange[0] (=%f). %s", fn_name, param_name,
+                    viewport.y, device_data->device_limits.viewportBoundsRange[0], validation_error_map[VALIDATION_ERROR_15000dde]);
+    } else if (negative_height_enabled && !(viewport.y <= device_data->device_limits.viewportBoundsRange[1])) {
+        skip |=
```
jzulauf-lunarg: y_healthy = false here as well, I think

layers/parameter_validation_utils.cpp
```diff
@@ -962,6 +962,138 @@ bool pv_vkCreateImageView(VkDevice device, const VkImageViewCreateInfo *pCreateI
...
+    if (!(viewport.x >= device_data->device_limits.viewportBoundsRange[0])) {
+        x_healthy = false;
+        skip |=
+            log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, object_type, object, __LINE__, VALIDATION_ERROR_15000ddc, LayerName,
+                    "%s: %s.x (=%f) is less than VkPhysicalDeviceLimits::viewportBoundsRange[0] (=%f). %s", fn_name, param_name,
+                    viewport.x, device_data->device_limits.viewportBoundsRange[0], validation_error_map[VALIDATION_ERROR_15000ddc]);
+    }
+
+    // x + width
+    if (x_healthy && width_healthy) {
```
jzulauf-lunarg: Are there any other examples of preventing validation checks because previous checking has failed.  I can see that this would be less verbose... just don't know where we've done it.

layers/parameter_validation_utils.cpp
```diff
@@ -962,6 +962,138 @@ bool pv_vkCreateImageView(VkDevice device, const VkImageViewCreateInfo *pCreateI
...
+    //       - float comparisons should expect NaN (comparison always false).
+    //       - VkPhysicalDeviceLimits::maxViewportDimensions is uint32_t, not float -> careful.
+
+    const auto f_lte_u32 = [](const float v1_f, const uint32_t v2_u32) {
+        using v1_f_t = decltype(v1_f);
+        assert(isnan(v1_f) || (v1_f < scalbnf(1.0f, std::numeric_limits<v1_f_t>::digits)));
+        const v1_f_t v2_f = static_cast<v1_f_t>(v2_u32);  // not accurate for > radix^digits
+
+        return (v1_f <= v2_f);
+    };
```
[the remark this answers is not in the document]
jzulauf-lunarg: Other than the assert I'm fine with this as is... if you want to warn on ambiguity, (only when it occurs) I'm fine with that but  it's certainly not required.

layers/parameter_validation_utils.cpp
```diff
@@ -962,6 +962,138 @@ bool pv_vkCreateImageView(VkDevice device, const VkImageViewCreateInfo *pCreateI
...
+bool pv_VkViewport(const layer_data *device_data, const VkViewport &viewport, const char *fn_name, const char *param_name,
+                   VkDebugReportObjectTypeEXT object_type, uint64_t object = 0) {
+    bool skip = false;
+    debug_report_data *report_data = device_data->report_data;
+
+    // Note: for numerical correctness
+    //       - float comparisons should expect NaN (comparison always false).
+    //       - VkPhysicalDeviceLimits::maxViewportDimensions is uint32_t, not float -> careful.
+
+    const auto f_lte_u32 = [](const float v1_f, const uint32_t v2_u32) {
```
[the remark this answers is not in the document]
krOoze: I don't expect this function to be used elsewhere, so it is local.

layers/parameter_validation_utils.cpp
krOoze: Last time I tried std:: <cmath> it did not work on the Android build.

layers/parameter_validation_utils.cpp
krOoze: Also makes implementing the tests easier.

layers/parameter_validation_utils.cpp
krOoze: right.

layers/parameter_validation_utils.cpp
```diff
@@ -962,6 +962,138 @@ bool pv_vkCreateImageView(VkDevice device, const VkImageViewCreateInfo *pCreateI
...
+        // minDepth
+        if (!(viewport.minDepth >= 0.0) || !(viewport.minDepth <= 1.0)) {
+            skip |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, object_type, object, __LINE__, VALIDATION_ERROR_150009a4,
+                            LayerName,
+                            "%s: VK_EXT_depth_range_unrestricted extension is not enabled and %s.minDepth (=%f) is not within the "
+                            "[0.0, 1.0] range. %s",
+                            fn_name, param_name, viewport.minDepth, validation_error_map[VALIDATION_ERROR_150009a4]);
+        }
+
+        // minDepth
```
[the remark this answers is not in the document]
krOoze: darn you addictive copy-pasting

layers/parameter_validation_utils.cpp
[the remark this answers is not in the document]
krOoze: ad 1. 4. Right, no, I should just use `float` because the `scalbnf` already assumes it.
ad 2. I kinda hoped this would be temporary. But I guess neither one of us reported it to the spec. AIS previously I would kinda prefer you to do it as I would be one-sided (because I see no ambiguity in comparing `float` to `uint32_t`). Or should I?
ad 3. Same reason as below. Last time I tried C++ version it did not work on the Android CI.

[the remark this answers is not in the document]
krOoze: ad warning: I suppose that is better temporary solution.

layers/parameter_validation_utils.cpp
tobine: @jzulauf-lunarg I know there have been a few cases where we skip follow-on checks after an initial check fails. I don't have specific examples off the top of my head but I know they're out there and this discussion has come up previously.

The agreed-upon general practice is to try to get through all of the checks as much as possible, but allow reasonable exceptions as needed. Certainly any case where a first error (invalid handle) would lead to a crash with further checks the further checks should be skipped. A grayer area is when a first check failing essentially guarantees that a follow-on check will fail. Usually I'd favor redundancy as long as follow-on check is clear. If the follow-on check may cause a user confusion due to known bad state, or if fixing the first check will almost certainly fix any successive checks I'm ok with guards such as this.

I think the "if" guard in this specific case falls into the last category and I'm ok with it.

If the only benefit to such a guard is to simplify testing then I'd generally push for not allowing it. I've been tempted at various times to do such things but I think for the most part I've refrained. The downside is having tests with two (or more) unexpected errors but in some cases this can't be avoided.

krOoze: > or if fixing the first check will almost certainly fix any successive checks I'm ok with guards such as this.

That's not quite the case. It is more like: fixing the first check will almost likely invalidate the previous result of the second check (i.e. the sum will likely be different, but not necessarily valid, after the fix).

>If the only benefit to such a guard is to simplify testing then I'd generally push for not allowing it.

TBH it is my main motivation (next to the spamminess). There would be bit of a combinatorial explosion of the edge cases to test. But, I guess, managable...

krOoze:
Rebased to handle #2394 conflicts. Anything else TBD here? Please review.

layers/parameter_validation_utils.cpp
jzulauf-lunarg: I'm okay with this.. it's not a perfect fit to Tobin's criteria, but if the size and offset are wrong, there's little point handling the range cases.

layers/parameter_validation_utils.cpp
```diff
@@ -2064,88 +2237,46 @@ bool pv_vkCmdSetViewport(VkCommandBuffer commandBuffer, uint32_t firstViewport,
...
+            count_healthy = false;
+            skip |=
+                log_msg(device_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
+                        HandleToUint64(commandBuffer), __LINE__, VALIDATION_ERROR_1e000992, LayerName,
+                        "vkCmdSetViewport: The multiViewport feature is disabled, but viewportCount (=%" PRIu32 ") is not 1. %s",
+                        viewportCount, validation_error_map[VALIDATION_ERROR_1e000992]);
+        }
+    } else {  // multiViewport enabled
+        const uint64_t sum = (uint64_t)firstViewport + viewportCount;
+        if (sum > device_data->device_limits.maxViewports) {
```
jzulauf-lunarg: The VUID check is for the range for sum on [1,  maxViewports] on this only tests on max...

layers/parameter_validation_utils.cpp
```diff
@@ -2064,88 +2237,46 @@ bool pv_vkCmdSetViewport(VkCommandBuffer commandBuffer, uint32_t firstViewport,
...
+    }
 
-            if (viewport.y + viewport.height > limits.viewportBoundsRange[1]) {
-                skip |=
-                    log_msg(device_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0,
-                            __LINE__, VALIDATION_ERROR_150009a2, LayerName,
-                            "vkCmdSetViewport %d: y (%f) + height (%f) exceeds permitted bound (%f). %s", viewportIndex, viewport.y,
-                            viewport.height, limits.viewportBoundsRange[1], validation_error_map[VALIDATION_ERROR_150009a2]);
-            }
+    if (count_healthy && pViewports) {
```
jzulauf-lunarg: I think filtering on just pViewports would be preferable, and not introduce and corner cases -- but okay as is.

layers/parameter_validation_utils.cpp
jzulauf-lunarg: ... but this is somewhat redundant with VALIDATION_ERROR_1e030a1b / VUID-vkCmdSetViewport-viewportCount-arraylength which requires the count to be> 0...  probably good as it is.

krOoze: Yea. I am giving priority to implicit VU over explicit.
Might propose change in the spec in the future. Last time I tried on one case, they didn't like it as it does not match style of the preexisting VUs...

layers/parameter_validation_utils.cpp
krOoze: Agree. It only delays a crash. But we might as well terminate here. The user already got the msg that count is bad.

krOoze: On the other hand it makes implementing test case for `viewportCount = maxViewports + 1` hard. It is impractical to provide sufficiently large `pViewports` (so it might crash and spam validation errors about `VkViewport`s until it does).

krOoze: Updated.
Given that most platforms have `maxViewport` only 16 I instead make the test skip if `maxViewport` happens to be too large to implement the test.

PR #2225, opened 2017-11-22
demos: Fix cubepp to use correct pipeline state for barrier

karl-lunarg:
Effectively apply same fix applied to cube.
(9455b5ed214aac547a9fbe88da3ea52415025f91)
Sliences validation error.

Change-Id: I4ba6a99e65400d171cfc365d7cf40464c189c150


krOoze:
Why does it need `HOST` stage? Should it not use the implicit synchronization, which should make the image available?

chrisforbes:
@krOoze it's true that implicit synchronization is partially redundant with this -- but you do need the layout transition this barrier is performing.

krOoze:
@chrisforbes Yea, but my point is the original code seems correct. Is it just to silence layers? In which case, shouldn't the layers be the ones being fixed?

chrisforbes:
@krOoze the other direction you could take is to set srcAccessMask=0, and leave the src stage as TOP_OF_PIPE.

krOoze:
@chrisforbes Ah, OK. It is only about about matching the stage flag to the access flag here.
Still, "the other direction" seems more sensible to me.
`HOST` stage synchronization is probably not trivial and probably has some cost. In cube demos we probably want to show the sensible technique to the incoming developers.

karl-lunarg:
TBH, I had assumed that the cube code was correct and was just applying the same fix to cubepp.  But this discussion and "the other direction" makes sense.  I'm away this week, but I'll try to update this PR unless someone beats me to it.

PR #2231, opened 2017-11-26
demos: GH1054, make cube shaders inline

### LINKED ISSUE #1054: Embed cube-frag.spv  cube-vert.spv into the cube demo to make it more portible
Is it possible to embed cube-frag.spv  cube-vert.spv into the cube demo so the executable can be run from anywhere, currently it must be loaded when the two shaders are in the current directory

krOoze:
fixes #1054 
Make cube demo shaders inline include instead of loaded from file.


krOoze:
@tobine Right, my bad. I assumed Android uses the cmake build...
Apparently the CI does not build the demos then, since it passes...

`build-android\generated\include` seems to be the proper place for it and needs to be added to `android-generate`.

Alternatively I could just restore the headers, but I kinda like that they are restored from source.

krOoze:
I went with the (original) precompiled way on Android because it is complicated to determine the shader compiler path there.

mark-lunarg:
This LGTus, I'll leave it to @tobine to ensure that Android is happy.

krOoze:
Worked for me, though I used updated NDK, and it needed #2233.

krOoze:
The MSVS CI keeps time-outing (>60 minutes). Maybe the 32 and 64 build needs to be separated...

PR #2232, opened 2017-11-28
Add C++11 compliance compiler flags

krOoze:
- add c++14 standard flags in MSVC versions that support it
- opt out of permissive in MSVC
- add c++11 flag and `pedantic` in other compilers
- should also fix VS 2017 15.5.1 build, which deprecates TR1 that the gtest uses


krOoze:
Ugh. Just screw linux...

mark-lunarg:
@krOoze, what needs to happen with this PR -- these changes don't work at the moment for Linux?

krOoze:
@mark-lunarg I suppose Windows and Linux (more precisely `gcc` and `clang`) are orthogonal here (so PR IMO can be accepted as is), so I provide only MSVC, which seems to work.

Problem on Linux is with the old standard libraries in the Trusty Theanimalorother on CI (and possibly other distros).

Trying to enable `-std=c++14` (actually have to use `c++1y`, because again obsolete packages on Linux) in clang results it [this error](https://bugs.llvm.org/show_bug.cgi?id=18402).

Options I have is:

1) Screw Linux

2) Try one of the suggested workarounds. I failed at that. Particularly, switching to `libc++` didn't work for me.

3) I could at least enable C++11 on linux, which should work. I found it pointless, because without flags it seems to work anyway, and the point is to restrict accidental use of features newer than C++14, which cannot be achieved with that.

mark-lunarg:
OK then, this Looks Good To Me. If you fix up the commit message (we usually label commits like these with `build:` or `cmake:`) I'll push it and we can worry about Linux later!

krOoze:
@mark-lunarg done

jeremy-lunarg:
> the point is to restrict accidental use of features newer than C++14 >

@krOoze I was, perhaps mistakenly, under the impression that we are targeting C++11. Is this changing?

krOoze:
@jeremy-lunarg Good question. No idea &mdash; the local readmes does not seem to mention. There are some `auto`s in there, so definitely we are at least C++11.
C++14 seemed like a good target, because it is the "default" in the current compilers.

jeremy-lunarg:
@krOoze I think the challenge we run into with C++14 is VS 2013 support. Which, IIRC, is the oldest MS compiler we support.

https://msdn.microsoft.com/en-us/library/hh567368.aspx

krOoze:
@jeremy-lunarg I think it was dropped some time ago in favor of 2015.

jeremy-lunarg:
@krOoze Ah! You are correct. VS2015 has some holes as well, but it's an improvement. Should we state that C++14 (with VS2015 restrictions) is our baseline in the readme?

mark-lunarg:
@krOoze, will c++14-specific feature implementations fail on Linux builds with this change?

krOoze:
@jeremy-lunarg Can't hurt.
@mark-lunarg Depends on the distro. The oldLTS used in the CI is probably stuck on g++ 4.8.4 and clang 3.4. Clang seems to have most of the features. The g++ only initial support. They should have finished C++11 though, so perhaps it should be C++11 still after all.

krOoze:
Updated to C++11 flag.

@jeremy-lunarg It seems that is now become Issue https://github.com/KhronosGroup/Vulkan-Ecosystem/issues/8

mark-lunarg:
Eh, @mikes-lunarg and I were talking, and we're not sure the added cmake stuff is worth the payoff on this one.  If it closed the holes entirely, then that's one thing.  I'm going to leave this up to @jeremy-lunarg and @mikes-lunarg, but I'm willing to wait for a more complete solution in the future.

mark-lunarg:
Closing for now -- please reopen if more/better information comes to light.

PR #2167, opened 2017-10-30
Validation cache

chrisforbes:

PR #2235, opened 2017-11-28
loader: fix memory leak for layers with >1 dev_exts

cdwfs:
This was causing failures in the loader's Allocation.Instance test on desktop, while testing the validation_cache branch.


lenny-lunarg:
Good catch and thanks for the fix. The change looks great to me from reading through it. I'm running it through our internal CI now (since that CI has some extra checks compared to appveyor and travis). Assuming this passes those tests I'll approve it and merge it in.

cdwfs:
Thanks! The true test will be to run it once #2167 is merged, and a layer actually exposes >1 device extension. But I'm confident :)

PR #2238, opened 2017-11-28
layers:Add secondary CB warning

### LINKED ISSUE #2237: layers: Issue warning/error if secondary command buffer inheritance info has non-null RenderPass and CONTINUE bit is not set
Currently, in this case, validation passes up to the point when the app starts to record commands. Then commands such as vkCmdDrawArrays will generate errors saying that a RenderPass is not active, even though this is notionally a secondary command buffer (though created badly). Clarifying the error by adding a warning would catch the error earlier.

tobine:
Fixes #2237

If VK_COMMAND_BUFFER_USAGE_RENDER_PASS_CONTINUE_BIT is not set by a
secondary command buffer, then any renderPass specified in the
inheritance info will be ignored. Adding a warning if a non-NULL
renderPass is included in the inheritance info but the
VK_COMMAND_BUFFER_USAGE_RENDER_PASS_CONTINUE_BIT is not set. This is
likely an oversight by the user and the current validation error that
this case leads to (no active renderPass) doesn't directly point to the
root cause.


layers/core_validation.cpp
```diff
@@ -5152,12 +5152,24 @@ VKAPI_ATTR VkResult VKAPI_CALL BeginCommandBuffer(VkCommandBuffer commandBuffer,
...
-                cb_node->activeSubpass = cb_node->beginInfo.pInheritanceInfo->subpass;
-                cb_node->activeFramebuffer = cb_node->beginInfo.pInheritanceInfo->framebuffer;
-                cb_node->framebuffers.insert(cb_node->beginInfo.pInheritanceInfo->framebuffer);
+            if (cb_node->createInfo.level != VK_COMMAND_BUFFER_LEVEL_PRIMARY) {
+                if (cb_node->beginInfo.flags & VK_COMMAND_BUFFER_USAGE_RENDER_PASS_CONTINUE_BIT) {
+                    cb_node->activeRenderPass = GetRenderPassState(dev_data, cb_node->beginInfo.pInheritanceInfo->renderPass);
+                    cb_node->activeSubpass = cb_node->beginInfo.pInheritanceInfo->subpass;
+                    cb_node->activeFramebuffer = cb_node->beginInfo.pInheritanceInfo->framebuffer;
+                    cb_node->framebuffers.insert(cb_node->beginInfo.pInheritanceInfo->framebuffer);
+                } else if (cb_node->beginInfo.pInheritanceInfo->renderPass) {
```
jzulauf-lunarg: this is a handle correct? should this be compared again VK_NULL_HANDLE ?  AFAICT there's probably lots that breaks if VK_NULL_HANDLE is ever non-zero... but typically we do the comparison for handles, but not pointers

jzulauf-lunarg: I'm not sure about this warning being on by default -- compare and contrast...

"VK_COMMAND_BUFFER_USAGE_RENDER_PASS_CONTINUE_BIT specifies that a secondary command buffer is considered to be entirely inside a render pass."

"If the VkCommandBuffer will not be executed within a render pass instance, renderPass is ignored."

As I read it, it is entirely valid for a non-null renderPass in this context.

layers/core_validation.cpp
```diff
@@ -5152,12 +5152,24 @@ VKAPI_ATTR VkResult VKAPI_CALL BeginCommandBuffer(VkCommandBuffer commandBuffer,
...
+                    // This is a user-requested warning. This is a likely case where user forgot to set RP continue bit
+                    skip |=
+                        log_msg(dev_data->report_data, VK_DEBUG_REPORT_WARNING_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_RENDER_PASS_EXT,
+                                HandleToUint64(cb_node->beginInfo.pInheritanceInfo->renderPass), __LINE__,
+                                VALIDATION_ERROR_0280006a, "CORE",
```
jzulauf-lunarg: VALIDATION_ERROR_0280006a says only the converse condition:

'If flags contains VK_COMMAND_BUFFER_USAGE_RENDER_PASS_CONTINUE_BIT, the renderPass member of pInheritanceInfo must be a valid VkRenderPass'

tobine:
@jzulauf-lunarg Warnings have been used in layers for valid behavior that may point to app issue. Note definition of WARNING from spec:

> VK_DEBUG_REPORT_WARNING_BIT_EXT specifies use of Vulkan that may expose an app bug. Such cases may not be immediately harmful, such as a fragment shader outputting to a location with no attachment. Other cases may point to behavior that is almost certainly bad when unintended such as using an image whose memory has not been filled. In general if you see a warning but you know that the behavior is intended/desired, then simply ignore the warning.

layers/core_validation.cpp
tobine: Yes, in general the question for warnings like this is "does the benefit outweigh the cost?" Are there common cases where this warning case will be raised but user fully intended to set renderPass without having CONTINUE_BIT set? I can't think of any offhand, and user raised #2237 due to a bug in their code where this warning would have benefited debug.
Therefore I currently lean towards inclusion of this warning.

layers/core_validation.cpp
tobine: There are no VUIDs for warnings and no standard methodology related to this. This VUID is directly related to the warning so I used it rather than creating a new one-off enum.

PR #1983, opened 2017-08-04
layers: Fix Graphics Pipeline pointers not ignored

krOoze:
Some `VkGraphicsPipelineCreateInfo` pointers must be ignored under some conditions,
but were not in the layers (resulting in crashes despite valid use).

PR #2240, opened 2017-11-29
tests: Clean up VKTriangleTest

krOoze:
- remove uniforms and descriptors and dependence on glm (not used anyway)
- BsoFailSelect into non-mask enum
- add depth attachment if needed by test
- add proper image usage flag if it is going to be cleared
- fixups in PipelineObj create info
- make layout changes in clear destructive
- use SetLayout where appropriate instead of custom code
- might help #1983

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -25,6 +25,7 @@
 #ifdef ANDROID
 #include "vulkan_wrapper.h"
 #else
+#define NOMINMAX
 #include <vulkan/vulkan.h>
 #endif
 
@@ -46,40 +47,23 @@
 #include <limits.h>
 #include <unordered_set>
 
-#define GLM_FORCE_RADIANS
-#include "glm/glm.hpp"
-#include <glm/gtc/matrix_transform.hpp>
-
 //--------------------------------------------------------------------------------------
 // Mesh and VertexFormat Data
 //--------------------------------------------------------------------------------------
-struct Vertex {
-    float posX, posY, posZ, posW;  // Position data
-    float r, g, b, a;              // Color
-};
 
-#define XYZ1(_x_, _y_, _z_) (_x_), (_y_), (_z_), 1.f
-
-typedef enum _BsoFailSelect {
-    BsoFailNone = 0x00000000,
-    BsoFailLineWidth = 0x00000001,
-    BsoFailDepthBias = 0x00000002,
-    BsoFailViewport = 0x00000004,
-    BsoFailScissor = 0x00000008,
-    BsoFailBlend = 0x00000010,
-    BsoFailDepthBounds = 0x00000020,
-    BsoFailStencilReadMask = 0x00000040,
-    BsoFailStencilWriteMask = 0x00000080,
-    BsoFailStencilReference = 0x00000100,
-    BsoFailCmdClearAttachments = 0x00000200,
-    BsoFailIndexBuffer = 0x00000400,
-} BsoFailSelect;
-
-struct vktriangle_vs_uniform {
-    // Must start with MVP
-    float mvp[4][4];
-    float position[3][4];
-    float color[3][4];
+enum BsoFailSelect {
+    BsoFailNone,
+    BsoFailLineWidth,
+    BsoFailDepthBias,
+    BsoFailViewport,
+    BsoFailScissor,
+    BsoFailBlend,
+    BsoFailDepthBounds,
+    BsoFailStencilReadMask,
+    BsoFailStencilWriteMask,
+    BsoFailStencilReference,
+    BsoFailCmdClearAttachments,
+    BsoFailIndexBuffer
 };
 
 static const char bindStateVertShaderText[] =
@@ -111,7 +95,7 @@ VkFormat FindSupportedDepthStencilFormat(VkPhysicalDevice phy) {
             return ds_formats[i];
         }
     }
-    return (VkFormat)0;
+    return VK_FORMAT_UNDEFINED;
 }
 
 // Returns true if *any* requested features are available.
@@ -414,9 +398,9 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL myDbgFunc(VkFlags msgFlags, VkDebugReportO
 
 class VkLayerTest : public VkRenderFramework {
    public:
-    void VKTriangleTest(const char *vertShaderText, const char *fragShaderText, BsoFailSelect failMask);
+    void VKTriangleTest(BsoFailSelect failCase);
     void GenericDrawPreparation(VkCommandBufferObj *commandBuffer, VkPipelineObj &pipelineobj, VkDescriptorSetObj &descriptorSet,
-                                BsoFailSelect failMask);
+                                BsoFailSelect failCase);
 
     void Init(VkPhysicalDeviceFeatures *features = nullptr, const VkCommandPoolCreateFlags flags = 0) {
         InitFramework(myDbgFunc, m_errorMonitor);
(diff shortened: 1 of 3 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -5311,7 +5295,7 @@ TEST_F(VkLayerTest, DynamicDepthBiasNotBound) {
     ASSERT_NO_FATAL_FAILURE(Init());
     // Dynamic depth bias
     m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "Dynamic depth bias state not set for this command buffer");
```
jzulauf-lunarg: With the substantial clean up, why not store the tests as an array of BsoFailSelect, const *char pairs, and then loop over them? 

Or define a test lambda (like the following, which I haven't run through the compiler)
```
   auto testcase= [this] (BsoFailSelect case, const char *error) {
      m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, error);
      VKTriangleTest(case);		 +    VKTriangleTest(BsoFailLineWidth);
};
```
And each test is a oneline

```
testcase(BsoFailLineWidth, "Dynamic line width state not set for this command buffer"); 
testcase(BsoFailViewport, "Dynamic viewport(s) 0 are used by pipeline state object, but were not provided");
 // et. seq.

```

Killer boilerplate/cut&paste code is a good thing if you're already cleaning it up :)

tests/layer_validation_tests.cpp
```diff
@@ -518,107 +503,110 @@ class VkLayerTest : public VkRenderFramework {
...
-        data.position[i][1] = tri_data[i].posY;
-        data.position[i][2] = tri_data[i].posZ;
-        data.position[i][3] = tri_data[i].posW;
-        data.color[i][0] = tri_data[i].r;
-        data.color[i][1] = tri_data[i].g;
-        data.color[i][2] = tri_data[i].b;
-        data.color[i][3] = tri_data[i].a;
-    }
+void VkLayerTest::VKTriangleTest(BsoFailSelect failCase) {
+    assert(m_device); // VKTriangleTest assumes Init() has finished
```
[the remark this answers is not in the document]
krOoze: I am really catching a logical error here. But I suppose there's no harm using the test versions and getting rid of another dependency...

mark-lunarg:
**Note** that each commit should contain one complete bit of the implementation, and that the repo should _build and run successfully_ between each commit in the series.  Given these limitations, definitely the more the better.

tests/layer_validation_tests.cpp
krOoze: Not a bad idea.
TBH though, I am not exactly sure `VKTriangleTest` can be re-played.

tests/vkrenderframework.cpp
```diff
@@ -585,7 +585,10 @@ int VkDescriptorSetObj::AppendSamplerTexture(VkSamplerObj *sampler, VkTextureObj
 
 VkPipelineLayout VkDescriptorSetObj::GetPipelineLayout() const { return m_pipeline_layout.handle(); }
 
-VkDescriptorSet VkDescriptorSetObj::GetDescriptorSetHandle() const { return m_set->handle(); }
+VkDescriptorSet VkDescriptorSetObj::GetDescriptorSetHandle() const {
+    if (m_set) return m_set->handle();
```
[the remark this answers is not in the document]
krOoze: I am using the git-clang-format. Of course I may have forgot, because it needs to be done manually...

krOoze:
@tobine It is not really that bad. But sometimes I wish I could include some kind of "diff guide". Git diff does tend to really show these kinds of rewrites in a chaotic way.

tests/layer_validation_tests.cpp
jzulauf-lunarg: You don't have to replay if the object is declared within the loop body, it is constructed/destructed each iteration.

jzulauf-lunarg:
You are doing a couple of things at once.  1 is an interface/framework cleanup with a lot of lines altered, the other is a test cleanup.  These could be partitioned s.t. interface/framework changes with a broad impact could be done in one commit, and test cleanup along with any interface/framework changes *local* to the test changes could be done in another.

tests/layer_validation_tests.cpp
krOoze: But `InitRenderTarget` is not. I would possibly have to do full TearDown...

mark-lunarg:
@krOoze, don't worry about refactoring this PR -- these are all guidelines that we try to adhere to to make the repo more consistent and easier to maintain.  It's a lot to take in, and you're doing great -- we appreciate all contributions!  Just keep these things in mind for future submissions.

tests/layer_validation_tests.cpp
krOoze: I defer this for another PR, as it is not strictly necessary to do here.

jzulauf-lunarg: I see what you mean.  You could write a VkTriangleTest or VkLayerTest OneshotTest similar to the other helper I wrote to turn each test into a one-liner -- or if you're really feeling adventurous, add a "Reset" path to VkLayerTest .

krOoze: I am not really satisfied with either facility and their partial duplications.
As I thrown the idea in your thread it could use some unification in a systematic way and obsolete all the `VkTriangleTest`s, `OneshotTest`s, pipeline tests and `Init()`s and `InitRenderTarget()`.
But perfect is enemy of good....

krOoze:
Updated with suggestions.

tests/vkrenderframework.cpp
```diff
@@ -1286,67 +1288,34 @@ void VkCommandBufferObj::PipelineBarrier(VkPipelineStageFlags src_stages, VkPipe
...
+        depth_stencil_obj->SetLayout(this, subrange.aspectMask, clear_layout);
 
-        // prepare depth buffer for rendering
-        memory_barrier.image = depthStencilObj->handle();
-        memory_barrier.newLayout = memory_barrier.oldLayout;
-        memory_barrier.oldLayout = VK_IMAGE_LAYOUT_GENERAL;
-        memory_barrier.subresourceRange = dsRange;
-        vkCmdPipelineBarrier(handle(), src_stages, dest_stages, 0, 0, NULL, 0, NULL, 1, pmemory_barrier);
+        VkClearDepthStencilValue clear_value = {depth_clear_value, stencil_clear_value};
+        ClearDepthStencilImage( depth_stencil_obj->handle(), clear_layout, &clear_value, 1, &subrange);
```
tobine: Mali driver is crashing in a couple of tests that now hit this code and didn't previously (VkLayerTest.DynamicStencilReadNotBound & VkLayerTest.DynamicStencilWriteNotBound & DynamicStencilRefNotBound).

I ran a branch with this call commented out to verify. On that branch (tobin_gh2240_with_mali_workaround_hack) those tests don't crash but later on Mali is crashing in DrawTimeImageViewTypeMismatchWithPipeline test, which I haven't looked into further.

For the initial failing tests, I noted that vkCmdClearDepthStencil() will only work on formats that support VK_FORMAT_FEATURE_TRANSFER_DST_BIT_KHR, which we don't make sure of so I thought perhaps this was causing the crash. I'm working on a separate branch (tobin_clear_image_checks) that adds check for this and also makes sure that DS formats we use support that feature.
With those new changes and this PR (tobin_gh2240_with_new_clear_checks), Mali no longer crashes down this call for the 3 tests mentioned above. Instead those tests fail b/c they don't get the correct error due to what I suspect a bug that arises when we can't find desired DS format.

So, anyway for Mali that appears to be some positive progress.

krOoze: @tobine There should probably be validation check for that. That's easy mistake to make.

krOoze: @tobine Isn't this basically driver bug? In core every supported format must support transfer ops.
`VK_FORMAT_FEATURE_TRANSFER_*_BIT_KHR` seems like it is only meant to enable formats that are transfer-only. So, any other feature flag should imply `VK_FORMAT_FEATURE_TRANSFER_*_BIT_KHR`.

tobine: @krOoze the issue may still be a driver bug. Note that I was slightly confused in my initial analysis as I didn't realize until I implemented the checks that VK_FORMAT_FEATURE_TRANSFER_DST_BIT_KHR is only required if VK_KHR_maintenance1 is enabled, which it isn't in this case.
When I initially added that check to the test, it was getting past the test (no segF) because it couldn't find an appropriate DS format, but the test wasn't passing.
So, still not clear why Mali is crashing down CmdClearDepthStencilImage() path.

krOoze: BTW As much as important finding driver bugs is, it should be possible to get rid of the dependency on `vkCmdClear`. I am not sure it does anything important in `VKTriangleTest`, or some of the other tests...

tobine: Yeah, I'm trying a few more hacks to see if anything works but otherwise may just disable, at least for now. Just found that when doing the Clear if I change "VK_REMAINING_MIP_LEVELS" and "VK_REMAINING_ARRAY_LAYERS" to both be "0" that Mali passes the 3 failing tests, so that seems like a driver bug.
I'll probably add that as a workaround and note with a TODO, and once I make sure we have the latest Mali driver I'll follow-up with vendor.
Mali back to crashing on DrawTimeImageViewTypeMismatchWithPipeline so will investigate that more.

cforfang: Mali has a known problem with the VK_REMAINING_* enums on older drivers,
 see https://community.arm.com/graphics/f/discussions/9344/vulkan-driver-issues---workaround-collection-thread/30710, so that might be it.

tobine:
After working around the first 3 Mali crashes, the next two crashing tests are DrawTimeImageViewTypeMismatchWithPipeline & DrawTimeImageMultisampleMismatchWithPipeline.
I've narrowed those crashes down to [CopyImage](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/tests/vkrenderframework.cpp#L989) call when finalizing VkTextureObj. Going through parameters now to see if we're sending some bad state.

tobine:
Just an update that I'm still looking at this. Got past all of the Mali crashes and some IMG ones as well, but still more to investigate.

krOoze:
@tobine Great, thanks for doing the hard part.

tests/layer_validation_tests.cpp
```diff
@@ -635,13 +626,9 @@ void VkLayerTest::VKTriangleTest(const char *vertShaderText, const char *fragSha
 
 void VkLayerTest::GenericDrawPreparation(VkCommandBufferObj *commandBuffer, VkPipelineObj &pipelineobj,
                                          VkDescriptorSetObj &descriptorSet, BsoFailSelect failMask) {
```
tobine: There's a bug here in that failMask is still used as a mask instead of an enum. There's one use of it in this function that was causing depthBoundsTestEnable to be set all the time which causes a crash on some systems. Core validation includes check VALIDATION_ERROR_0f6004ac to make sure depth bounds test feature is enabled, and that's not being triggered so this seems to be a bug.

I've got a branch with a fix for this (as well as a couple other workarounds) that I'm testing and will merge if all is clear.
I'll file a separate issue for a new CTS test.
I'll update the check for VALIDATION_ERROR_0f6004ac to include the VUID and mark is as actually implemented in the DB.

tobine:
Just filed #2269 that includes this PR with a couple CLs from me to get it passing on our farm.

PR #2244, opened 2017-12-02
build: Change AppVeyor config to use multiple jobs.

karl-lunarg:
This is a complete rewrite of the AppVeyor config file.

A separate job is used for each config in the Platform-Configuration
matrix.  This helps get around the problem of the 60-minute limit per job.

The invocation of the update_external_sources script has been
replaced with individual commands to build glslang and spirv-tools
only for the platform configuration selected for each job.
This avoids having to build glslang and spirv-tools 4 times for each job.


karl-lunarg:
This should be good to go.

I tested it on my own fork with AppVeyor to make sure that the scripts work when a cache is restored.
(Doesn't seem to be using cache in this branch - perhaps due to the config file change.)
The scripts can stumble a bit when they find directories present that they would normally create in a totally clean build tree.  But it seems to work ok and at least part of glslang does not rebuild, saving a few minutes per job.

The main point is that this config now uses a separate job for each Platform / Configuration combination, which helps avoid the problem we've been having with the jobs timing out.

cnorthrop:
Karl, this is great, thanks for working on it!

Is there any way we could move some of your changes into the build_windows_targets.bat file?  I like the direction of getting the build going on multiple jobs, but this is introducing a parallel set of instructions that have to be maintained.  Could we update the script to mirror your changes, passing in the platform and config to invoke just one of the four combinations?

I'm not opposed to getting this in as it is as a first step, it saves so much developer time.  I need to add Android at some point too.

cnorthrop:
Adding to previous comment, part of the reason I'd like to get it ported to the script is so we can use this in our our various test farm setups.  Right now I think we both do the full Windows build on a single box.  With an updated script, we could easily farm it out to four boxes.

karl-lunarg:
I did this to immediately address the AppVeyor timeout issue, but agree that some common scripting would help avoid maintaining the code in AppVeyor.  (It got more involved than I wanted it to)

The current `build_windows_targets.bat` script already supports the `32` and `64` arguments, which would allow you to partially parallelize the builds today.  The script could be expanded to handle `Debug` and `Release` arguments in a similar manner to let you split it four ways.

The other factor is the `update_external_sources` script.  The glslang and spirv-tools builds also take a lot of time and this script builds all four combinations with no options for building fewer configs.  In addition, the rest of the LVL build needs the shader compiler and I think it expects to use the 64-bit Release version, independent of the current build config.   So this makes it necessary to build 64-bit Release for all configs.

If the AppVeyor caching facility is working well, this can be less of an issue.  But there are limits on how much cache we can have, and I was bumping up against this 1 GB limit a lot while testing in my fork.  The external folder is 3.4 GB when fully built.  The cache operation compresses it down quite a bit before it is uploaded.  And with this new AppVeyor config, the sizes of the external folder are smaller because there are at most two configs in each, but there are four of them.  The 1 GB limit is for the entire account.  So, I'm anxious to try this in master to see how the caching works out.  (BTW, I think that `update_external_sources.bat` always caused a complete rebuild anyway, so I'm not sure we were benefiting from the cache anyway.)

I think good follow-on steps are:

* Add the Release and Debug args to build_windows_targets.bat to allow non-AppVeyor CI systems to use the script to build a single config.  AppVeyor automatically starts a job for each config today via its "build matrix" feature.

* Rewrite the update_external_sources script in Python so that the same script can be used for both Linux and Windows.  Add options to it to build individual configs.  Then the logic in the AppVeyor instructions can be simplified to just run the script with the selected config (plus the Release 64).

karl-lunarg:
@cnorthrop Can I go ahead and merge this?  I'll keep an eye on the AppVeyor builds over the next few days.

karl-lunarg:
Some results:

Build [1938](https://ci.appveyor.com/project/Khronoswebmaster/vulkan-loaderandvalidationlayers/build/1938): One job.  49 minutes.  Many prior to this one have been failing due to a 60-minute timeout.

Build [1939](https://ci.appveyor.com/project/Khronoswebmaster/vulkan-loaderandvalidationlayers/build/1939): First build with 4 jobs.  Job times: 29+22+16+16= 85 mins.  Caches created for all but the last job.

Build [1940](https://ci.appveyor.com/project/Khronoswebmaster/vulkan-loaderandvalidationlayers/build/1940): Job times: 16+16+7+17 = 57 mins.  There was no (or less) recompiling of glslang in the first three jobs.  There was no cache for the last job, making it take longer.

The cache wasn't created for one of the jobs due to the cache storage being full, and not because of a quota overrun.  Perhaps this condition will ease and a future job will create a cache for this config.  That should reduce the time of that config from 17 to about 7, for a total time of 47, which is comparable to the single-job time.

Although there are no real time savings, there shouldn't be any more timeouts.

PR #2246, opened 2017-12-03
tests: Add layout objs

krOoze:
Add `VkDescriptorSetLayoutObj` and `VkPipelineLayoutObj`.


krOoze:
WIP:
> [ RUN      ] VkLayerTest.VertexBufferInvalid
> Unexpected: vkBindBufferMemory: required parameter buffer specified as VK_NULL_HANDLE

in one of the CI. I am messing with Handles, so I better investigate this...

krOoze:
@tobine I forgot to investigate the test log diff above. If you don't mind, I am gonna rebase and give it a look if it is still there.

tobine:
@krOoze I'll hold off on the merge until you've checked it out w/ rebase and let me know it's ok. Note that your latest push hit the new clang-format check in CI. Just requires a few formatting updates.

krOoze:
@tobine It appears only on the clang CI. Dunno why/how the otherwisely obsolete distro there uses unstable clang 5.0.0.
Also I cannot reproduce it locally with clang 5.0.1. So I am gonna spam the CI a bit to bisect the source before I am comfortable blaming a compiler.

krOoze:
@tobine Seems OK to me. I think the affected test just uses uninitialized variable; i.e. nothing related to this PR.

daveh-lunarg:
This PR breaks our Windows VS2013 build in a couple of ways.

The first is lack of support for 'noexcept'.  There's a work-around in branch **dave-noexcept** to resolve that.  

The second is an error with explicitly-defaulted special member functions, producing this output:
```
c:\users\lunarg\.jenkins\workspace\gerrit-lvl\32-bit\release\tests\vktestbinding.h(145): error C2610: 'vk_testing::internal::NonDispHandle<T>::NonDispHandle(vk_testing::internal::NonDispHandle<T> &&)' : is not a special member function which can be defaulted [C:\Users\LunarG\.jenkins\workspace\Gerrit-LVL\32-bit\Release\build32\tests\vk_layer_validation_tests.vcxproj]
c:\users\lunarg\.jenkins\workspace\gerrit-lvl\32-bit\release\tests\vktestbinding.h(146): error C2610: 'vk_testing::internal::NonDispHandle<T> &vk_testing::internal::NonDispHandle<T>::operator =(vk_testing::internal::NonDispHandle<T> &&)' : is not a special member function which can be defaulted
...
```
Note that the line numbers are after the noexcept patch, they'd be 122/123 without.  My impulse would be to just eliminate the explicit defaults, but I'm not confident enough why they're there to do that myself.  Please review @tobine , @krOoze .

krOoze:
@daveh-lunarg The defaulted constructors are there, because it is much easier than reasoning out if compiler makes implicit non-deleted constructors.

It should be fine removing them as long as it compiles. Alternativelly the can be defined as just calling super.

PR #2247, opened 2017-12-03
Simplify printf format types

krOoze:
- use `long long` for `uint64`
- use `long` for `uint32`
- use standard `z` for `size_t`

It is not really necessary to use the macros as long as the type is same or bigger.
They only seem to trip `clang-format` anyway.


krOoze:
@stroyan Yeah, the `PRINTF_SIZE_T_SPECIFIER` is more of a collateral damage here.

`%ll` is guaranteed to be 64 bit so valid target for `uint64_t`.
Same with `%l` being guaranteed 32 bit so good gor `uint32_t`.

Anyway, one of the compiler does not like this, so I think I am probably gonna close this...

mark-lunarg:
I as well would prefer to leave things as they are, at least for now.

PR #2248, opened 2017-12-04
validation: Add a warning for LOAD_OP_LOAD + LAYOUT_UNDEFINED

jekstrand:
While not technically an error (VK_IMAGE_LAYOUT_UNDEFINED is always
allowed as an initial layout), this combination is almost never what
you actually want.  If nothing else, it's a potential performance
problem because VK_ATTACHMENT_LOAD_OP_LOAD is liable to be more
expensive than VK_ATTACHMENT_LOAD_OP_DONT_CARE.  Give developers a
helpful warning in this case.

This would have caught an actual bug in some of the Sascha Willems
demos as fixed in this PR:

https://github.com/SaschaWillems/Vulkan/pull/400

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -2793,6 +2793,32 @@ bool ValidateLayouts(core_validation::layer_data *device_data, VkDevice device, 
     const debug_report_data *report_data = core_validation::GetReportData(device_data);
     bool skip = false;
 
+    for (uint32_t i = 0; i < pCreateInfo->attachmentCount; ++i) {
+        VkFormat format = pCreateInfo->pAttachments[i].format;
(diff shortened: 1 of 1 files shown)


mark-lunarg:
@karl-lunarg, validation or assistant layer?

karl-lunarg:
This is a good example of something that goes in the assistant layer.

But since the code has already been (nicely) written, I'd suggest accepting this PR and adding this item to the assistant layer list along with a note that the same check is in this PR.  When the assistant layer goes out with this check, we can then think about taking this out of the VLs.  I'd rather risk a little redundancy than miss this useful check.

mark-lunarg:
Excellent.  @jekstrand, if you click through the CLA we'll get this in.

jekstrand:
There you go.  CLA signed.

mark-lunarg:
Thanks a bunch @jekstrand -- normalized commit message and pushed to Master.

PR #2252, opened 2017-12-05
NOMERGE: add some build profiling options

chrisforbes:


chrisforbes:
Sadly this doesn't work -- all the output is currently suppressed.

Issue #2230, opened 2017-11-25
VL seems to not check maxBoundDescriptorSets


Brotcrunsher:
**Overview**: When calling `vkCreatePipelineLayout` with an invalid (too high) `setLayoutCount` inside `VkPipelineLayoutCreateInfo`, no message is printed. Instead the program crashes with an illegal write operation (varying locations).

**Details**: The spec says: 
> setLayoutCount must be less than or equal to VkPhysicalDeviceLimits::maxBoundDescriptorSets
[link](https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VkPipelineLayoutCreateInfo)

Vulkan Version: 1.0.65.0
Tested under Visual Studio 2017 15.4.1
Device: GeForce GTX 1080 Ti
Allowed setLayoutCount: 8
Used setLayoutCount: 9

krOoze:
Probably also reported in #2063.

daveh-lunarg:
This should be resolved with PR #2255.  Closing.

daveh-lunarg:
Re-opened as I've temporarily reverted #2255.  Will be PR #2262 when re-submitted.

daveh-lunarg:
Resolved in PR #2262.

PR #2255, opened 2017-12-05
CreatePipelineLayout device limits VUs

daveh-lunarg:
Adds 6 valid usage checks of device physical limits to VkCreatePipelineLayouts().  Re-organizes the validation code into PreCallValidation, dispatch, and PostCallRecord functions.  Adds 2 tests to provoke the VUs.

This PR resolves issue #2230.


krOoze:
It seems some of it could theoretically be parameter_validation.

layers/core_validation.cpp
```diff
@@ -4853,34 +4905,139 @@ VKAPI_ATTR VkResult VKAPI_CALL CreatePipelineLayout(VkDevice device, const VkPip
...
-                        __LINE__, VALIDATION_ERROR_0fe0024a, "DS",
-                        "vkCreatePipelineLayout() Multiple push descriptor sets found. %s",
+        skip |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0, __LINE__,
+                        VALIDATION_ERROR_0fe0024a, "DS", "vkCreatePipelineLayout() Multiple push descriptor sets found. %s",
                         validation_error_map[VALIDATION_ERROR_0fe0024a]);
     }
+
+    // Validate per-stage descriptor limits for resource types. This invloves calculating the intersection of 5 resource types
+    // with 6 shader stages, then summing these arrays over setLayoutCount layouts.
+    std::valarray<uint32_t> vertex_sum(0U, 5);
```
tobine: I'd like to see a bunch of this work offloaded to the GetDescriptorCountSumsForStage() helper function, which can be renamed as needed. The ultimate goal here is to get a max value for each of the DSL_TYPE_* entries across all shader types. The helper function could return just the vector of those max values. That means the helper would take the set_layouts vector as a parameter, and it would loop over the layouts and calculate the std::max() values before returning the final result.
Does that make sense and seem like a beneficial thing to you?
It's just refactoring code and functionally the same, but it gets almost all of this logic out of the path of where the errors are flagged, which I prefer and I think is simpler to read/understand for new people looking at the code.

tests/layer_validation_tests.cpp
```diff
@@ -364,7 +364,7 @@ class ErrorMonitor {
                 ADD_FAILURE() << "Did not receive expected error '" << desired_msg << "'";
             }
             for (auto desired_id : desired_message_ids_) {
-                ADD_FAILURE() << "Did not receive expected error ENUM '" << desired_id << "'";
+                ADD_FAILURE() << "Did not receive expected error ENUM '0x" << std::hex << desired_id << "'";
```
tobine: :+1:

tests/layer_validation_tests.cpp
```diff
@@ -6239,6 +6239,207 @@ TEST_F(VkLayerTest, WriteDescriptorSetConsecutiveUpdates) {
...
+    vkDestroyDescriptorSetLayout(m_device->device(), ds_layout, NULL);
+
+    // VU 0fe00240 - too many uniform buffer type descriptors in vertex stage
+    //
+    dslb_vec.clear();
+    dslb.binding = 0;
+    dslb.descriptorType = VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER;
+    dslb.descriptorCount = 1 + (max_uniform_buffers / 2);
+    dslb.stageFlags =
+        VK_SHADER_STAGE_VERTEX_BIT | VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT | VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT;
```
tobine: Need to detect if tessellationShader feature is supported before using these flags.

layers/core_validation.cpp
```diff
@@ -4823,12 +4824,64 @@ static bool validatePushConstantRange(const layer_data *dev_data, const uint32_t
     return skip;
 }
 
-VKAPI_ATTR VkResult VKAPI_CALL CreatePipelineLayout(VkDevice device, const VkPipelineLayoutCreateInfo *pCreateInfo,
-                                                    const VkAllocationCallbacks *pAllocator, VkPipelineLayout *pPipelineLayout) {
+// Used by PreCallValiateCreatePipelineLayout.  Sums descriptor entries by type for a given shader stage.
+const int DSL_TYPE_SAMPLERS = 0;
```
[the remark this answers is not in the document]
daveh-lunarg: Yep.

layers/core_validation.cpp
```diff
@@ -4823,12 +4824,64 @@ static bool validatePushConstantRange(const layer_data *dev_data, const uint32_t
...
+                    break;
+                case VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE:
+                case VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER:
+                    psc[DSL_TYPE_SAMPLED_IMAGES] += binding->descriptorCount;
+                    break;
+                case VK_DESCRIPTOR_TYPE_STORAGE_IMAGE:
+                case VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER:
+                    psc[DSL_TYPE_STORAGE_IMAGES] += binding->descriptorCount;
+                    break;
+                case VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER:
```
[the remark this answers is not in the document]
daveh-lunarg: I started there, but I think the more verbose version is clearer and more robust to future changes.

layers/core_validation.cpp
daveh-lunarg: Makes sense.  Agree it's cleaner to segregate the entire calculation from the check.

tests/layer_validation_tests.cpp
daveh-lunarg: Yes, and geometry stage as well.  A new bit of Vulkan trivia for my quiver.

The specific stages aren't actually important to the test, so rather than complicate the test further I'll change to avoid the non-mandatory stages.

daveh-lunarg: @krOoze, I agree in principle that this could be parameter validation, but in this case the use of core_validation's cvdescriptorset class makes it much simpler to do there.

daveh-lunarg:
@krOoze, I agree in principle that this could be parameter validation, but in this case the use of core_validation's cvdescriptorset class makes it much simpler to do there.

daveh-lunarg:
@tobine Changes in. Do you want to take another run through your farm before I push this?

tobine:
@daveh-lunarg thanks! Yes, I'll run this now and approve if all's good.

daveh-lunarg:
@tobine Oops - sorry, missed your reply, jumped the gun.

tobine:
@daveh-lunarg Please revert. VkLayerTest.CreatePipelineLayoutExcessPerStageDescriptors test is failing on Vega. Error:

> [ RUN      ] VkLayerTest.CreatePipelineLayoutExcessPerStageDescriptors
> S:\windows-slave\dev_windows\build\tests\layer_validation_tests.cpp(373): error: Failed
> Did not receive expected error ENUM '0xfe0023e'
> S:\windows-slave\dev_windows\build\tests\layer_validation_tests.cpp(373): error: Failed
> Did not receive expected error ENUM '0xfe00240'
> S:\windows-slave\dev_windows\build\tests\layer_validation_tests.cpp(373): error: Failed
> Did not receive expected error ENUM '0xfe00242'
> S:\windows-slave\dev_windows\build\tests\layer_validation_tests.cpp(373): error: Failed
> Did not receive expected error ENUM '0xfe00244'
> S:\windows-slave\dev_windows\build\tests\layer_validation_tests.cpp(373): error: Failed
> Did not receive expected error ENUM '0xfe00246'
> [  FAILED  ] VkLayerTest.CreatePipelineLayoutExcessPerStageDescriptors (692 ms)

daveh-lunarg:
Reverted on master, re-submitted as PR #2262

PR #2259, opened 2017-12-05
Copy json files for in-tree builds

### LINKED ISSUE #2243: In-tree builds of layers don't work
I've been wondering for a while now why I can never seem to get the layers to configure and build.  I knew *someone* was getting them to build reliably.  Turns out it's because I'm using an in-tree build.  It seems that all of the -json targets are just created for out-of-tree.  I don't know enough cmake to really know how to fix it or I'd submit a PR.

mark-lunarg:
Modified layer and icd cmake files to copy JSON files for in- and out-of-tree builds.  Fixes #2243.

@jekstrand, can you verify that this takes care of the problems you were seeing?


jekstrand:
It doesn't appear to.  Running "ccmake ." from the source tree yields this:

    CMake Error at layers/CMakeLists.txt:109 (add_dependencies):
    Cannot add target-level dependencies to non-existent target
    "VkLayer_core_validation-json".

    The add_dependencies works for top-level logical targets created by the
    add_executable, add_library, or add_custom_target commands.  If you want to
    add file-level dependencies see the DEPENDS option of the add_custom_target
    and add_custom_command commands.

mark-lunarg:
@jekstrand, I guess I was performing the builds differently -- is this happening on Linux or Windows?

mark-lunarg:
@jekstrand, if you have some time, please give this second version a shot.  @karl-lunarg provided some solid improvements and it seems to be OK here.

mark-lunarg:
And sorry, please ignore the vxproj.user changes

jekstrand:
Seems to work now.  It at least builds.

PR #2260, opened 2017-12-06
build: Remove clumsy syntax in Travis-CI config

karl-lunarg:
We were using a clumsy workaround in order to
get good failure information when a Travis-CI build
failed and also to make it stop at the point of
failure.

This commit uses some techniques to workaround
travis-ci/travis-ci#1066, which asserts that
a Travis-CI job should stop when a before_install
or script command fails.  Today, Travis-CI
continues with the next commands.

This was tested with both an intentional build
script error and a compilation error.  In both
cases the Travis-CI job ended at the point of
failure and the logs were clear enough to
diagnose the problem.


karl-lunarg:
@cnorthrop : I think that this config file still avoids all the problems we had that drove us to use the original clumsy syntax.   As I noted in the commit notes, the job quits when there is an error and the logs are still useful.  Anyway, I think this new syntax will make it easier to work with this file in the future.

.travis.yml
```diff
@@ -28,38 +28,60 @@ matrix:
...
 
+# The before_install: and script: steps employ some workarounds discussed in
+# https://github.com/travis-ci/travis-ci/issues/1066 in order to allow coding
+# the commands so that the build fails if any of the commands fail.
+# The default action for Travis-CI is to continue running even if a command fails.
+
 before_install:
   # Install the appropriate Linux packages.
-  - if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then sudo apt-get -y install libxkbcommon-dev libwayland-dev libmirclient-dev libxrandr-dev libx11-xcb-dev; fi
+  - |
```
cnorthrop: Can you explain what this pipe is doing?

.travis.yml
```diff
@@ -28,38 +28,60 @@ matrix:
...
-  - if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then LD_LIBRARY_PATH=dbuild/loader:$LD_LIBRARY_PATH VK_LAYER_PATH=dbuild/layers VK_ICD_FILENAMES=dbuild/icd/VkICD_mock_icd.json VK_DEVSIM_FILENAME=tests/device_profiles/nvidia_tegra_x1.json dbuild/tests/vk_layer_validation_tests --devsim; fi
-  - if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then LD_LIBRARY_PATH=dbuild/loader:$LD_LIBRARY_PATH VK_LAYER_PATH=dbuild/layers VK_ICD_FILENAMES=dbuild/icd/VkICD_mock_icd.json VK_DEVSIM_FILENAME=tests/device_profiles/powervr_rogue_ge8300.json dbuild/tests/vk_layer_validation_tests --devsim; fi
-  - if [[ "$VULKAN_BUILD_TARGET" == "ANDROID" ]]; then pushd build-android; fi
-  - if [[ "$VULKAN_BUILD_TARGET" == "ANDROID" ]]; then ./update_external_sources_android.sh --abi $ANDROID_ABI --no-build; fi
-  - if [[ "$VULKAN_BUILD_TARGET" == "ANDROID" ]]; then ./android-generate.sh; fi
-  - if [[ "$VULKAN_BUILD_TARGET" == "ANDROID" ]]; then USE_CCACHE=1 NDK_CCACHE=ccache ndk-build APP_ABI=$ANDROID_ABI -j $core_count; fi
-  - if [[ "$VULKAN_BUILD_TARGET" == "ANDROID" ]]; then popd; fi
+  - set -e
+  - |
```
cnorthrop: Should this pipe be before set -e as it is before_install?

.travis.yml
karl-lunarg: https://www.jeffgeerling.com/blog/2017/bash-logic-structures-and-conditionals-if-case-loops-etc-travis-ci

.travis.yml
karl-lunarg: TBH, that was a pattern I noticed in the many issues that referenced the travis-ci issue.  I'll revisit.

karl-lunarg: So, it really isn't a pipe.  It is a YAML block scalar header which says to take the following text "as-is".  I think I coded it this way so that we could do:
```
- set -e
- |
  if [[ "$VULKAN_BUILD_TARGET" == "LINUX" ]]; then
    <commands>
  fi
- |     # <- new part
  if [[ "$VULKAN_BUILD_TARGET" == "ANDROID" ]]; then
    <commands>
  fi
```

which sort of breaks things up a little and makes it easier to see different logical parts of the script.

So, if you do end up with more than one block scalar header, it is better to put the set -e at the top in its own command so that it applies to all the blocks.

I'll push a change to make it look like the above.

I also think that the logger echoes everything in the block to the log at once, which is a downside to this that I just noticed.  This results in a command's output being sort of far away from the echo of the command itself.  This can perhaps be mitigated by adding more block scalar headers and/or adding "echo Running tests..." statements.

karl-lunarg: I made the before_install section look like the script section.  Should be good enough now.

PR #2261, opened 2017-12-06
VS2013 compatibility

mikes-lunarg:
Remove all usages of constexpr to restore compatibility with Visual Studio
2013.  Also make a handful of other minor edits where the VS2013
compiler was reporting errors.


scripts/helper_file_generator.py
```diff
@@ -934,19 +931,17 @@ def GenerateTypeMapHelperHeader(self):
         explanatory_comment = '\n'.join((
```
jzulauf-lunarg: For the code generation, I'd prefer the fully functional generation present under a 'if vs_2013_compatible'  check so as to make undoing the damage easier.

PR #2268, opened 2017-12-12
Turn off tr1 namespace deprecation warning

mikes-lunarg:
The tr1 namespace is used by Google Test and has been deprecated
starting with Visual Studio 15.5.  Turn off the deprecation warning
until we can update Google Test to prevent the build from failing.


krOoze:
I also fix it in #2232.
I actualy instead use `GTEST_HAS_TR1_TUPLE=0` to prevent gtest using tr1. It seems to work &mdash; I guess we do not depend on any of the features that need it.

PR #2269, opened 2017-12-13
Tobin tri test cleanup

tobine:
Replaces #2240 

Includes all of @krOoze clean-up work with a couple of minor additions to workaround specific device issues and a minor fix-up.

The CL from @krOoze has already been through lots of review and testing so unless there are strenuous objections to something there no need to comment on those changes. This is primarily to allow a review of my additions which I believe are minor but wanted to put them up for review before merging.


krOoze:
Great! I will rebase #1983 momentarily, and we shall see if we got anywhere with this... 😋

krOoze:
Hmm, you just remove the depth attachments. That kinda sets me back to square one. Besides, it kinda makes the three tests invalid. E.g. without depth attachment, the `VkPipelineDepthStencilStateCreateInfo` is ignored, therefore `stencilTestEnable` is actually disabled, which makes the test cases actually a valid usage.

I guess I need to rewrite it once more. The code still looks like it needs it anyway. The clear (which I assume is the driver problem) should theoretically not be necessary at all.

tobine:
Yes, the clear is the problem. Let me re-test with the DS attachments back in for those 3 tests as I made that workaround before a couple of others and we also just updated the devices in our farm.

PR #2272, opened 2017-12-13
demos: Remove vulkaninfo dependency on DISPLAY

krOoze:
Remove vulkaninfo dependency on DISPLAY and so allow it to continue listing other info,


karl-lunarg:
Tested fix - looks good.  Thanks!

PR #2275, opened 2017-12-14
tests:Re-enable DS attachment for TriTest

tobine:
Three cases were disabling DS attachment to accomodate a driver bug,
but this is actually invalid for the tests.
Re-enabling the DS attachment for those cases as with device updates
and other workarounds, the tests are now working across all devices.


tobine:
@krOoze FYI this is working now

krOoze:
@tobine yay

PR #2276, opened 2017-12-14
DSBufferLimitErrors overflow

mikes-lunarg:
Mostly motivated by some new drivers failing DSBufferLimitErrors: both tests and parameter validation were storing some VkDeviceSize values as uint32_t.  In addition to fixing the overflow issue, also made the DSBufferLimitErrors test a little more robust.


layers/parameter_validation.h
```diff
@@ -153,9 +154,10 @@ bool ValidateGreaterThan(debug_report_data *report_data, const char *api_name, c
     bool skip_call = false;
 
     if (value <= lower_bound) {
-        skip_call |=
-            log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0, __LINE__, 1, LayerName,
-                    "%s: parameter %s must be greater than %d", api_name, parameter_name.get_name().c_str(), lower_bound);
+        std::ostringstream ss;
+        ss << api_name << ": parameter " << parameter_name.get_name() << " must be greater than " << lower_bound;
```
tobine: We should also include the actual "value" in this error message

tests/layer_validation_tests.cpp
```diff
@@ -12281,115 +12281,99 @@ TEST_F(VkLayerTest, DSBufferLimitErrors) {
...
-    ASSERT_VK_SUCCESS(err);
-    // TODO: This was failing on mock w/ 256 offsetAlignment. Update to work for all values.
-    auto sb_offset = (ub_ci.size + sb_mem_reqs.alignment - 1) & ~(sb_mem_reqs.alignment - 1);
-    err = vkBindBufferMemory(m_device->device(), storage_buffer, mem, sb_offset);
-    ASSERT_VK_SUCCESS(err);
+        // Have to bind memory to buffer before descriptor update
+        VkMemoryAllocateInfo mem_alloc = {};
+        mem_alloc.sType = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO;
+        mem_alloc.pNext = NULL;
+        mem_alloc.allocationSize = bci.size;
```
tobine: This value should be set from the size value returned in mem_reqs from vkGetBufferMemoryRequirements() call below.

layers/parameter_validation.h
mikes-lunarg: fixed

tests/layer_validation_tests.cpp
mikes-lunarg: fixed

PR #2282, opened 2017-12-15
scripts: Prefer git.bat on Windows.

null77:
scripts: Prefer git.bat on Windows.
    
    Calling git.bat seems to work in every instance, whereas some
    installations don't have access to "git" directly.

@tobine PTAL


scripts/external_revision_generator.py
```diff
@@ -33,7 +33,9 @@
     output_header_file = sys.argv[3]
     
     # Extract commit ID from the specified source directory
-    commit_id = subprocess.check_output(["git", "rev-parse", "HEAD"], cwd=source_dir).decode('utf-8').strip()
+    # Call git.bat on Windows for compatiblity.
+    git_binary = "git.bat" is os == "nt" else "git"
```
tobine: I think "is" should be "if", yes?

null77:
You were correct Tobin, I clobbered the CL after re-coding it on a second computer because I couldn't authenticate with GitHub on the first. Fixed and seems to work with local testing.

scripts/external_revision_generator.py
null77: Done.

PR #2283, opened 2017-12-15
layers: Work around bugs in old STL versions.

null77:
Some old versions of the C++ STL have a few bugs with advanced features.
This works around two such issues:

- std::unordered_map::const_reference having the wrong type
- an error with an explicit std::vector constructor

@tobine PTAL


layers/parameter_validation_utils.cpp
```diff
@@ -455,7 +455,8 @@ VKAPI_ATTR VkResult VKAPI_CALL vkCreateDevice(VkPhysicalDevice physicalDevice, c
 
         lock.lock();
 
-        validate_result(my_instance_data->report_data, "vkCreateDevice", {}, result);
+        std::vector<VkResult> empty_vec;
+        validate_result(my_instance_data->report_data, "vkCreateDevice", empty_vec, result);
```
krOoze: Should be `const` to preserve the "emptiness" ivariant.

Seems like a `initializer_list` problem though. `validate_result(std::vector<VkResult>());` should work.

Does parameter_validation even still check `VkResult`s? Maybe this can just be deleted?

null77: Not sure, maybe the other reviewers would know?

tobine: validate_result() only warns, and only if it sees an error other than VK_ERROR_VALIDATION_FAILED_EXT. Not sure how often this comes up, but when it does I could see the warnings being helpful. I'm on fence if we should delete the check. @mark-lunarg or @mikes-lunarg opinions?

krOoze: @tobine The search tells me this is the only call-point of `validate_result`. I remember `VkResults` reported (b/c I complained about it), but seems like it was removed when the layer became autogenerated.

mikes-lunarg: I think that checking the return code is useful.  It's not clear to me that its removal from the rest of parameter validation was intentional.  @mark-lunarg Do you remember why it was removed when the PV codegen got reworked?

mark-lunarg: The validate_result checks were hand-generated, and when PV moved to the mostly-code-generated model with the VUID auto-tagging, it was outside that scope of work.  Besides, we received nothing but complaints about warning for normal behavior.  Nobody's missed it, I'd vote for killing this last one.

mikes-lunarg: +1 for going into the assistant layer, that seems like a better fit than parameter validation.  I'm fine with it being removed here for now.

tobine:
Alright, I went ahead and killed the result check in this PR. Merging now.

null77:
Thanks for killing the result check for me.

PR #2284, opened 2017-12-17
layers: Ignore immutable samplers pointer

krOoze:
- ignore `pImmutableSamplers` pointer on `vkCreateDescriptorSetLayout` if type is a non-sampler
- add positive test


mark-lunarg:
LGTM!  Thanks!

PR #2285, opened 2017-12-19
layers: Remove unused variable from core validation.

null77:
This commit was accidentall reverted in dec37fb9172:
layers: Add descriptor limit checks to vkCreatePipelineLayout()

@tobine PTAL. Not sure if there is a way to add clang/GCC warnings to the CI? The tricky part about this particular warning is I think it also shows up in auto-generated code, so would need to be suppressed or fixed in those files, which is a bit tricky.

core_validation.cpp:109:23: error: unused variable 'kSurfaceSizeFromSwapchain' [-Werror,-Wunused-const-variable]


null77:
Fixed the typo in my branch.

PR #2286, opened 2017-12-19
build: Add clang-format checking of pull request changes to CI

karl-lunarg:
Check modified lines in a pull request for correct clang-format.
Note that checking is not performed unless there is a pull request.

Heavily leveraged from https://github.com/KhronosGroup/SPIRV-Tools.

### DIFF
--- a/scripts/check_code_format.sh
+++ b/scripts/check_code_format.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+# Copyright (c) 2017 Google Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Script to determine if source code in Pull Request is properly formatted.
+# Exits with non 0 exit code if formatting is needed.
+#
+# This script assumes to be invoked at the project root directory.
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m' # No Color
+
+FILES_TO_CHECK=$(git diff --name-only master | grep -E ".*\.(cpp|cc|c\+\+|cxx|c|h|hpp)$")
+
+if [ -z "${FILES_TO_CHECK}" ]; then
+  echo -e "${GREEN}No source code to check for formatting.${NC}"
+  exit 0
+fi
+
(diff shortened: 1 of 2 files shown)


tobine:
Looks great. Testing this out w/ #2287

cnorthrop:
I think the clang-format check should come before any build steps, so you'll get a quick rejection.

scripts/check_code_format.sh
```diff
@@ -0,0 +1,37 @@
...
+  exit 0
+fi
+
+FORMAT_DIFF=$(git diff -U0 master -- ${FILES_TO_CHECK} | python ./scripts/clang-format-diff.py -p1 -style=file)
+
+if [ -z "${FORMAT_DIFF}" ]; then
+  echo "All source code in PR properly formatted."
+  exit 0
+else
+  echo "Found formatting errors!"
```
tobine: If you could make this show up in red that would be ideal.

cnorthrop:
I see now the clang-format is a separate non-blocking job.  I'm fine with it going in as is, but we should refactor for quick rejection on format issues.  Otherwise, PRs have to go through two full build cycles to fix formatting issues.

karl-lunarg:
> I think the clang-format check should come before any build steps, so you'll get a quick rejection.

The jobs start in a non-deterministic order, so changing the order in the matrix won't matter.
If the format job starts relatively early, it will finish before the others.
I think a format job failure causes the result icon in the GitHub Checks pane to indicate a failure even before all the jobs finish.
So, you may get a quick indication that way.

I think that the only way to get a quick rejection is to set fast_finish on and set allow_failures for all the jobs but the clang-format job.  I really don't think we want that.

scripts/check_code_format.sh
karl-lunarg: Do you want just this line in red or the entire diff?

karl-lunarg:
> I see now the clang-format is a separate non-blocking job. I'm fine with it going in as is, but we should refactor for quick rejection on format issues. Otherwise, PRs have to go through two full build cycles to fix formatting issues.

This would be super-easy to do.
But if you have a quick rejection for a formatting issue, if there is a compilation error in the PR as well, then the user won't see that on the first push.
The error would show up only after the formatting passes, and that ends up needing at least two build cycles anyway.

Another drawback is that the format checking would be done for each build config, which is redundant.  If there are a lot of formatting diffs, the log could end up with 4 copies of a long diff instead of just one.

cnorthrop:
> But if you have a quick rejection for a formatting issue, if there is a compilation error in the PR as well, then the user won't see that on the first push.
> The error would show up only after the formatting passes, and that ends up needing at least two build cycles anyway.

Yes, but the first pass only takes ~44 seconds, the second takes 5+ minutes. :)

Granted, this is all absolutely dwarfed by Appveyor taking almost an hour, but I envision a future when that problem is fixed too.

If we duplicate the format check to each job, including the Appveyor jobs, there is a smidgen of redundant work, but it would be worth it.

Again, this can be done in the future, doesn't need to hold up this PR.

scripts/check_code_format.sh
karl-lunarg: I added code to display the two positive echo messages in green and this one in red.  I'll push to master when CI clears.

PR #2290, opened 2017-12-20
scripts: Scan CWD in object tracker generator.

null77:
This makes it more easy to locate a header in configs which have the
CWD set to something less obvious. In Chrome's case the build directory
is completely separated from the script directory.

Also reset the CWD after the header is found to keep the build intact.

This is similar to the change to the parameter validation generator.

@tobine PTAL, missed this in the object tracker script when I was fixing the parameter validation script.

### DIFF
--- a/scripts/object_tracker_generator.py
+++ b/scripts/object_tracker_generator.py
@@ -238,6 +238,9 @@ class ObjectTrackerOutputGenerator(OutputGenerator):
         self.valid_vuids = set()       # Set of all valid VUIDs
         self.vuid_file = None
         # Cover cases where file is built from scripts directory, Lin/Win, or Android build structure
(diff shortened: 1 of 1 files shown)


null77:
Anything I can do to help expedite landing this one?

tobine:
There's an [error](https://ci.appveyor.com/project/Khronoswebmaster/vulkan-loaderandvalidationlayers/build/2055/job/1ry2umh4qu7cox8w) in Windows CI. Did you build on Windows locally. Error seems related to a separate change that's already clean so I'm confused.

@null77 can you re-base off latest master and re-push the branch to kick-off CI again? It needs to be CI clean before I can merge.

null77:
@tobine Done, thanks for the tip. Checks are pending.

PR #2291, opened 2017-12-21
tests: Fix use of uninitialized var in VkBufferTest

krOoze:
- Fix use of uninitialized variable VulkanBuffer in `VkBufferTest`
- Separate case to `NULL`, and bogus value handle
- Add fake memory object instead of `SetUnexpectedError`


tests/layer_validation_tests.cpp
```diff
@@ -739,9 +740,32 @@ class VkBufferTest {
...
+            memory_allocate_info.sType = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO;
+            memory_allocate_info.allocationSize = 1;   // fake size -- shouldn't matter for the test
+            memory_allocate_info.memoryTypeIndex = 0;  // fake type -- shouldn't matter for the test
+            vkAllocateMemory(VulkanDevice, &memory_allocate_info, nullptr, &VulkanMemory);
+            AllocateCurrent = true;
+
+            VulkanBuffer = VK_NULL_HANDLE;
+
+            vkBindBufferMemory(VulkanDevice, VulkanBuffer, VulkanMemory, 0);
+        } else if (eBindFakeBuffer == aTestFlag) {
```
tobine: Could combine both of these cases into the "if" check and just switch on aTestFlag to assign VulkanBuffer to appropriate value since all other code is common.

krOoze: Why not. Updated. TBH though it is putting a lipstick on a pig here in this class.
Beside overal bad design, the VkResults are not checked (and even cannot be trivially, because all of this is inside a constructor).

tobine: @krOoze we've already outfitted the pig with a luxurious evening gown, high heels, and a tiara, so a little lipstick won't hurt :)
Perhaps one day we can have a luau and replace the pig with something a little nicer...

mark-lunarg: I like the tiara.

Issue #2207, opened 2017-11-15
Enabled validation layers should give performance warning


krOoze:
Enabled validation layers should give performance warning.

The primary motivation is that I would want a feedback from the implementation, that I enabled the layers properly and my validation environment works. The message could be something like:

> Enabling validation layers: [ordered list of layers or the standard_validation] {for the duration of `vkCreateInstance` and `vkDestroyInstance`}. This will likely have negative performance impact.

The reason for using `VK_DEBUG_REPORT_PERFORMANCE_WARNING_BIT_EXT` is because it technically is a performance drain. (And it is simpler to have this enabled, otherwise it could be `INFO`, but that would require filtering out all the other unwanted `INFO` messages.)

It probably have to be logged twice if the user uses both the temporary and regular debug callback. It is a stretch though to pass the temporary callback as a `PERFORMANCE_WARNING`...

mark-lunarg:
@krOoze, there is already a facility in the loader for something close to this -- you can set  
"VK_LOADER_DEBUG=all" in your environment and it will provide some of the information you want.  

One minor issue related to spitting out a perf-warning with the above message is that each of the layers is independent and shares no data, so each layer would have to output its own message.

mark-lunarg:
Hi again, we talked about it here and were hoping to get a better idea about which problem(s) you are hoping to solve, taking into account that the loader debug environment setting does indicate which layers are loaded.  We're open to the suggestion but need to justify the effort.  Thanks!

krOoze:
Use cases:
- if I run an app I don't know if it is just perfect, or if the layers are simply not working.
- if I switch between layer versions (e.g. release vs debug, or SDK vs master branch) I want to know which I am currently using
- I want to know, if I have layers enabled by accident (let's say I accidentally enable layers in persistent environment variables)

Currenty I have the options:
1) Enable `INFO` and `DEBUG` and sift through several pages of unrelated information. Then unenable it again.
2) Break something on purpose, then unbreak it again. (Doesn't give extended info -- like version.)
3) Enable `INFO` and `DEBUG` and filter the messages manually. (Can't do with environment variable enabled layers)

Those options simply sound inadequate for something that is as basic and simple as this. -> the suggestion above to make an intro `VK_DEBUG_REPORT_PERFORMANCE_WARNING_BIT_EXT` message.

PR #2293, opened 2017-12-21
Tobin khr std val header

tobine:
Initial proposal for vk_standard_validation.h.

The header has a const array with the standard validation stack as well as a vector for use in C++. This updates the demos and tests to use this new method. The documentation is also updates to refer to the vk_standard_validation.h header as the preferred method and is refers to the VK_LAYER_LUNARG_standard_validation as deprecated.
I'm open for discussion on how this is advertised, but just took a first stab as updates that I hope will encourage unification on this new header.
From a technical standpoint, I don't like missing the "vkval" namespace on C, where namespaces aren't available. I tried some struct trick to fake the namespace but couldn't get anything that worked. As it is, I suspect C++ is dominant language for using Vulkan so I can live without the namespace but if anyone has input on this or anything else, let me know.

### DIFF
--- a/demos/cube.c
+++ b/demos/cube.c
@@ -55,6 +55,7 @@
 #endif
 
 #include <vulkan/vk_sdk_platform.h>
+#include <vulkan/vk_standard_validation.h>
 #include "linmath.h"
 
 #include "gettime.h"
@@ -380,7 +381,7 @@ struct demo {
     uint32_t enabled_extension_count;
     uint32_t enabled_layer_count;
     char *extension_names[64];
-    char *enabled_layers[64];
+    char const *enabled_layers[64];
 
     int width, height;
     VkFormat format;
@@ -2921,8 +2922,7 @@ static void demo_run_display(struct demo *demo)
  * Return 1 (true) if all layer names specified in check_names
  * can be found in given layer properties.
  */
-static VkBool32 demo_check_layers(uint32_t check_count, char **check_names,
-                                  uint32_t layer_count,
+static VkBool32 demo_check_layers(uint32_t check_count, char const *const *check_names, uint32_t layer_count,
                                   VkLayerProperties *layers) {
     for (uint32_t i = 0; i < check_count; i++) {
         VkBool32 found = 0;
@@ -2944,20 +2944,10 @@ static void demo_init_vk(struct demo *demo) {
     VkResult err;
     uint32_t instance_extension_count = 0;
     uint32_t instance_layer_count = 0;
-    uint32_t validation_layer_count = 0;
-    char **instance_validation_layers = NULL;
+    uint32_t validation_layer_count = VALIDATION_LAYER_COUNT;
+    char const *const *instance_validation_layers = standard_validation_array;
     demo->enabled_extension_count = 0;
     demo->enabled_layer_count = 0;
-
-    char *instance_validation_layers_alt1[] = {
-        "VK_LAYER_LUNARG_standard_validation"
-    };
-
-    char *instance_validation_layers_alt2[] = {
-        "VK_LAYER_GOOGLE_threading",      "VK_LAYER_LUNARG_parameter_validation",
-        "VK_LAYER_LUNARG_object_tracker", "VK_LAYER_LUNARG_core_validation",
-        "VK_LAYER_GOOGLE_unique_objects"
-    };
 
     /* Look for validation layers */
     VkBool32 validation_found = 0;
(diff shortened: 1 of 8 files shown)


polarina:
Have you considered the name `vkval_standard_validation_array` for C users?

For C++, have you considered `std::array` in place of `std::vector` for the layer names? Some people do not want to pull in a runtime dependency to the C++ standard library.

tobine:
Good things to consider, thanks @polarina

krOoze:
This is possibly harmful. App compiled with old header would not work with newer SDK. This does need to be decided at runtime.

tobine:
Created alternative #2307 that addresses @polarina comments.
@krOoze I understand your point but am not sure how big of a problem this is in reality. I would think, (perhaps naively?) that developers running with validation will pull a new SDK only when they want header updates that will require re-compilation anyway. Also changes to layer stack will be so infrequent that recompilation for layer stack will only be required very infrequently, and can be called out in SDK release notes in rare cases that layer stack changes.

krOoze:
@tobine Perhaps. But in that case this PR does not solve anything, and user can just supply those layers themselves directly without copying them out of arrays or vectors provided in this header.

And it is a functional downgrade. Current situation that pulls the needed layer names at runtime from json seems superior.

tobine:
@krOoze the issue is that initial android loader that shipped on many devices and will live on does not support the json meta-layer technique. The hard-coded header solution is trying to solve an Android problem and unify validation layer enabling between Android and desktop without requiring user to supply complete layer list. Currently on Android users do have to supply complete layer list themselves and this leads to issues when layers are excluded or in the wrong order.

krOoze:
@tobine That's unfortunate, but this looks like it's trying to rob Peter to pay Paul.
What about making a real `VK_LAYER_LUNARG_standard_validation` so then as suggested that would enable the proper layers?

tobine:
@krOoze if by "making a real `VK_LAYER_LUNARG_standard_validation`" you mean making a layer that acts as a mini-loader to pull the correct validation layer stack then that's certainly a possibility and a smoother transition forward than the current solution since it wouldn't require any code changes by desktop developers. I was just putting this header out first to get thoughts since it's simpler to put together. I still may end up doing the "real layer" solution, though.

krOoze:
@tobine Yup, that's what I mean. I am not sure from top of my head if it is possible to do in a layer (theoretically should be possible to layer the `vkCreateInstance`, but the loader if badly implemented may resist, being always the top layer).

While this PR is online, a constructive one:
I may want to perhaps use this as a macro list, i.e. something used like:

```
vector<const char*> myLayers = {
    "VK_LAYER_ACME_super_bomb",
    STANDARD_VALIDATION_LAYERS,
    "VK_LAYER_LUNARG_api_dump"
};
```

cnorthrop:
What if the list of layers was moved out of the header and into a small library that apps link against?  Then it could be updated along with the LunarG SDK (and the Android NDK).

krOoze:
@cnorthrop If it is a static library, then it is exactly the same as the header (with the same problems). If it is a dynamic library then it would create app dependency on the SDK (i.e. the dll has to be in path, and so SDK must be installed for app linked against it to work).

karl-lunarg:
I'm somewhat against doing this (and #2307) for a few reasons, as demonstrated by the possible options.

1) Do nothing.  The list of validation layers doesn't change that often, and when they do, there is usually advance notice given in the release notes for the SDK prior to the change in the layer list.   Discovering and recovering from this change isn't that hard.  Apps coded in the "desktop" way of using standard_validation won't be affected. Only the apps that are coded to supply a list of layers for Android would be affected by a layer change.

2) Add the proposed header file.  This seems like an optimization for the rare case when the layers change, making this optimization less valuable.  It saves manually changing the layer list, but the app needs to be rebuilt anyway.
This proposal also changes the interface to the application.  It suggests stopping the usage of std_val and replacing it with this header.   std_val is already out there in the form of docs, tutorials, and code, so people are used to it.  std_val is also very useful to desktop devs because it makes it easier to turn on validation with env vars.
This option seems better to me if it is advertised as an equal alternative to std_val.  Apps that are desktop-only can stick with std_val.  Apps that are Android-only or Android-desktop could use the header file.  But I'm not sure if having two ways to do all this is a good thing.

3) std_val loads the validation layers.  This is the "mini-loader" proposal mentioned here.  This is a good option because it doesn't change the interface to the application.  It also isolates the solution for an Android problem to the Android implementation of the standard validation layer.  I don't know how hard it would be to implement.  This results in two ways of implementing meta-layers, but both could co-exist and the application interface to both is the same.

I suppose I like the idea of not putting this header in now because there's no urgent need.  Perhaps the Android folks can evaluate how hard it would be to have a layer load other layers.

cnorthrop:
>  The list of validation layers doesn't change that often, and when they do, there is usually advance notice given in the release notes for the SDK prior to the change in the layer list. 

One of the problems we're also trying to solve (in a cross platform manner) is incorrect usage of the existing validation layers, i.e. folks that simply enable `core_validation` by itself (even on desktop), since that is the intuitive thing to do.  There is no feedback whatsoever informing them that the layers don't behave correctly in isolation.

I'm leaning toward simply doing the investment of merging all the validation layers back together so we don't need an explicit ordered list to get correct behavior.  Then `VK_LAYER_LUNARG_standard_validation` will continue to work as is, and will just be translated to `core_validation`, or whatever we choose to name the remaining layer.

krOoze:
> incorrect usage of the existing validation layers, i.e. folks that simply enable core_validation by itself

I was trying to solve that with #2207. I would want an initial feedback message for exactly that reason (and others).

> There is no feedback whatsoever informing them that the layers don't behave correctly in isolation.

They should behave correctly in isolation though (just differently/suboptimally).

Though it might be nice to merge the layers and control the amount of validation by config file or api.

karl-lunarg:
> I'm leaning toward simply doing the investment of merging all the validation layers back together so we don't need an explicit ordered list to get correct behavior. Then VK_LAYER_LUNARG_standard_validation will continue to work as is, and will just be translated to core_validation, or whatever we choose to name the remaining layer.

This seems like a good direction since the original vision of having each VL being more independent has mostly faded.  And we do have the job of adding control over the amount of validation on the list.

tobine:
Closing this. I think eventual correct solution will be merging the validation layers in such a fashion that they're all contained under a single layer name. Fine to just live with what we have until we're there.

PR #2300, opened 2017-12-28
layers: Change target name for devsim json

joey-lunarg:
Fixes duplicate targets in layers / layersvt for VulkanTools

Change-Id: Ieb2577454743ac085153b04b2cbb3fcb25918578

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -86,7 +86,7 @@ else()
                 )
         endforeach(config_file)
         foreach (config_file ${LAYER_BINARY_JSON_FILES})
-            add_custom_target(${config_file}-json ALL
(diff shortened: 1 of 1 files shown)


layers/CMakeLists.txt
```diff
@@ -86,7 +86,7 @@ else()
                 )
         endforeach(config_file)
         foreach (config_file ${LAYER_BINARY_JSON_FILES})
-            add_custom_target(${config_file}-json ALL
+            add_custom_target(${config_file}-bin-json ALL
```
joey-lunarg: Since the filename has to match, add "bin" to layer json targets in /layers to avoid this happening again.

mark-lunarg:
LGTM, once appveyer passes go ahead and push it.  Thanks!

PR #2301, opened 2017-12-29
Performance fixes for large numbers of bound descriptors -- fixes #2221

### LINKED ISSUE #2221: Performance issue: large arrays of texture descriptors (bindless) is very slow
We've just hooked up using a large unflattened array of textures in our shader, for example like this in HLSL:

Texture2D g_tTextureArray[ 4096 ] : layout( set = 1, binding = 3 ) ;

This array is dynamically uniform indexed in the shader.  The descriptor set associated with this array is updated very infrequently (not every frame).  

Turning this rendering path on has caused a massive drop in performance with validation (< 1FPS) in DOTA.  The costs look spread out from profiling, but it basically is from iterating over the entire array in the cases where it validates descriptors.  

I can't quite share a test case yet, but I'm wondering if we could disable validation somehow for large arrays of textures?  Another option might be detecting that the descriptor set is unchanged since previous use and skipping redundant validation?

jzulauf-lunarg:
Cleans up descriptor set validation, by replacing iterative lookups with O(log n), reorganizing data to avoid duplicate lookups, and tracking already validated state (cached validation) -- particularly with an eye toward improving performance for large numbers of descriptors bound. Addresses #2221 

In testing, factors of 3-10 reduction in frame times were observed for use cases with "many" descriptors.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -572,17 +572,22 @@ bool ValidateMemoryIsBoundToBuffer(const layer_data *dev_data, const BUFFER_STAT
 
 // SetMemBinding is used to establish immutable, non-sparse binding between a single image/buffer object and memory object.
 // Corresponding valid usage checks are in ValidateSetMemBinding().
-static void SetMemBinding(layer_data *dev_data, VkDeviceMemory mem, uint64_t handle, VulkanObjectType type, const char *apiName) {
+static void SetMemBinding(layer_data *dev_data, VkDeviceMemory mem, BINDABLE *mem_binding, VkDeviceSize memory_offset,
+                          uint64_t handle, VulkanObjectType type, const char *apiName) {
+    assert(mem_binding);
+    mem_binding->binding.mem = mem;
+    mem_binding->UpdateBoundMemorySet();  // force recreation of cached set
+    mem_binding->binding.offset = memory_offset;
+    mem_binding->binding.size = mem_binding->requirements.size;
+
     if (mem != VK_NULL_HANDLE) {
-        BINDABLE *mem_binding = GetObjectMemBinding(dev_data, handle, type);
-        assert(mem_binding);
         DEVICE_MEM_INFO *mem_info = GetMemObjInfo(dev_data, mem);
         if (mem_info) {
             mem_info->obj_bindings.insert({handle, type});
             // For image objects, make sure default memory state is correctly set
             // TODO : What's the best/correct way to handle this?
             if (kVulkanObjectTypeImage == type) {
-                auto const image_state = GetImageState(dev_data, VkImage(handle));
+                auto const image_state = reinterpret_cast<const IMAGE_STATE *>(mem_binding);
                 if (image_state) {
                     VkImageCreateInfo ici = image_state->createInfo;
                     if (ici.usage & (VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT | VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT)) {
@@ -590,7 +595,6 @@ static void SetMemBinding(layer_data *dev_data, VkDeviceMemory mem, uint64_t han
                     }
                 }
             }
-            mem_binding->binding.mem = mem;
         }
     }
 }
@@ -673,6 +677,7 @@ static bool SetSparseMemBinding(layer_data *dev_data, MEM_BINDING binding, uint6
             mem_info->obj_bindings.insert({handle, type});
             // Need to set mem binding for this object
             mem_binding->sparse_bindings.insert(binding);
+            mem_binding->UpdateBoundMemorySet();
         }
     }
     return skip;
@@ -2890,23 +2895,22 @@ static void PostCallRecordFreeMemory(layer_data *dev_data, VkDeviceMemory mem, D
         log_msg(dev_data->report_data, VK_DEBUG_REPORT_INFORMATION_BIT_EXT, get_debug_report_enum[obj.type], obj.handle, __LINE__,
                 MEMTRACK_FREED_MEM_REF, "MEM", "VK Object 0x%" PRIx64 " still has a reference to mem obj 0x%" PRIx64,
                 HandleToUint64(obj.handle), HandleToUint64(mem_info->mem));
+        BINDABLE *bindable_state = nullptr;
         switch (obj.type) {
-            case kVulkanObjectTypeImage: {
-                auto image_state = GetImageState(dev_data, reinterpret_cast<VkImage &>(obj.handle));
-                assert(image_state);  // Any destroyed images should already be removed from bindings
-                image_state->binding.mem = MEMORY_UNBOUND;
+            case kVulkanObjectTypeImage:
+                bindable_state = GetImageState(dev_data, reinterpret_cast<VkImage &>(obj.handle));
                 break;
-            }
-            case kVulkanObjectTypeBuffer: {
-                auto buffer_state = GetBufferState(dev_data, reinterpret_cast<VkBuffer &>(obj.handle));
-                assert(buffer_state);  // Any destroyed buffers should already be removed from bindings
-                buffer_state->binding.mem = MEMORY_UNBOUND;
+            case kVulkanObjectTypeBuffer:
+                bindable_state = GetBufferState(dev_data, reinterpret_cast<VkBuffer &>(obj.handle));
                 break;
-            }
             default:
                 // Should only have buffer or image objects bound to memory
                 assert(0);
         }
+
+        assert(bindable_state);
+        bindable_state->binding.mem = MEMORY_UNBOUND;
+        bindable_state->UpdateBoundMemorySet();
     }
     // Any bound cmd buffers are now invalid
     invalidateCommandBuffers(dev_data, mem_info->cb_bindings, obj_struct);
@@ -3801,11 +3805,7 @@ static void PostCallRecordBindBufferMemory(layer_data *dev_data, VkBuffer buffer
 
         // Track objects tied to memory
         uint64_t buffer_handle = HandleToUint64(buffer);
-        SetMemBinding(dev_data, mem, buffer_handle, kVulkanObjectTypeBuffer, "vkBindBufferMemory()");
-
-        buffer_state->binding.mem = mem;
-        buffer_state->binding.offset = memoryOffset;
-        buffer_state->binding.size = buffer_state->requirements.size;
+        SetMemBinding(dev_data, mem, buffer_state, memoryOffset, buffer_handle, kVulkanObjectTypeBuffer, "vkBindBufferMemory()");
     }
 }
 
@@ -8947,11 +8947,7 @@ static void PostCallRecordBindImageMemory(layer_data *dev_data, VkImage image, I
 
         // Track objects tied to memory
         uint64_t image_handle = HandleToUint64(image);
-        SetMemBinding(dev_data, mem, image_handle, kVulkanObjectTypeImage, "vkBindImageMemory()");
-
-        image_state->binding.mem = mem;
-        image_state->binding.offset = memoryOffset;
-        image_state->binding.size = image_state->requirements.size;
+        SetMemBinding(dev_data, mem, image_state, memoryOffset, image_handle, kVulkanObjectTypeImage, "vkBindImageMemory()");
     }
 }
 
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -1184,15 +1184,24 @@ static bool ValidateDrawState(layer_data *dev_data, GLOBAL_CB_NODE *cb_node, CMD
                 cvdescriptorset::DescriptorSet *descriptor_set = state.boundDescriptorSets[setIndex];
                 // Validate the draw-time state for this descriptor set
                 std::string err_str;
-                if (!descriptor_set->IsPushDescriptor() &&
-                    !descriptor_set->ValidateDrawState(set_binding_pair.second, state.dynamicOffsets[setIndex], cb_node, function,
-                                                       &err_str)) {
-                    auto set = descriptor_set->GetSet();
-                    result |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT,
-                                      VK_DEBUG_REPORT_OBJECT_TYPE_DESCRIPTOR_SET_EXT, HandleToUint64(set), __LINE__,
-                                      DRAWSTATE_DESCRIPTOR_SET_NOT_UPDATED, "DS",
-                                      "Descriptor set 0x%" PRIx64 " encountered the following validation error at %s time: %s",
-                                      HandleToUint64(set), function, err_str.c_str());
+                if (!descriptor_set->IsPushDescriptor()) {
+                    // For the "bindless" style resource usage with many descriptors, need to optimize command <-> descriptor
+                    // binding validation. Take the requested binding set and prefilter it to eliminate redundant validation checks.
+                    // Here, the currently bound pipeline determines whether an image validation check is redundant...
+                    // for images are the "req" portion of the binding_req is indirectly (but tightly) coupled to the pipeline.
+                    const cvdescriptorset::PrefilterBindRequestMap reduced_map(*descriptor_set, set_binding_pair.second, cb_node,
+                                                                               pPipe);
+                    const auto &binding_req_map = reduced_map.Map();
+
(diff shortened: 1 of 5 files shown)


layers/core_validation.cpp
```diff
@@ -1179,15 +1184,22 @@ static bool ValidateDrawState(layer_data *dev_data, GLOBAL_CB_NODE *cb_node, CMD
...
-                    auto set = descriptor_set->GetSet();
-                    result |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT,
-                                      VK_DEBUG_REPORT_OBJECT_TYPE_DESCRIPTOR_SET_EXT, HandleToUint64(set), __LINE__,
-                                      DRAWSTATE_DESCRIPTOR_SET_NOT_UPDATED, "DS",
-                                      "Descriptor set 0x%" PRIx64 " encountered the following validation error at %s time: %s",
-                                      HandleToUint64(set), function, err_str.c_str());
+                if (!descriptor_set->IsPushDescriptor()) {
+                    // For the "bindless" style resource usage with many descriptors, need to optimize command <-> descriptor
+                    // binding
+                    const cvdescriptorset::PrefilterBindRequestMap reduced_map(*descriptor_set, set_binding_pair.second, cb_node,
```
jzulauf-lunarg: This is the pattern for validation caching... take the requested binding set and prefilter it to eliminate redundant validation checks.

layers/core_validation.cpp
```diff
@@ -1179,15 +1184,22 @@ static bool ValidateDrawState(layer_data *dev_data, GLOBAL_CB_NODE *cb_node, CMD
...
+                                                                               pPipe);
```
jzulauf-lunarg: In this case, the currently bound pipeline determines whether a validation check is redundant... for images are the "req" portion of the binding_req is indirectly (but tightly) coupled to the pipeline -- non-obviously, at least to me.

layers/core_validation.cpp
```diff
@@ -2890,23 +2907,22 @@ static void PostCallRecordFreeMemory(layer_data *dev_data, VkDeviceMemory mem, D
...
                 break;
-            }
             default:
                 // Should only have buffer or image objects bound to memory
                 assert(0);
         }
+
+        assert(bindable_state);
+        bindable_state->binding.mem = MEMORY_UNBOUND;
+        bindable_state->UpdateBoundMemorySet();
```
jzulauf-lunarg: There is some consolidation of code, where new code is needed for each path, and expanding the scope of cut and paste exceeded... reasonable.

layers/core_validation.cpp
```diff
@@ -5369,6 +5381,12 @@ VKAPI_ATTR VkResult VKAPI_CALL BeginCommandBuffer(VkCommandBuffer commandBuffer,
 
     return result;
 }
+static void PostCallRecordEndCommandBuffer(layer_data *dev_data, GLOBAL_CB_NODE *cb_state) {
+    for (auto descriptor_set : cb_state->validated_descriptor_sets) {
+        descriptor_set->ClearCachedValidation(cb_state);
+    }
+    cb_state->validated_descriptor_sets.clear();
+}
```
jzulauf-lunarg: Cached validation is specific to a specific recording of a specific command buffer.

layers/core_validation_types.h
```diff
@@ -196,19 +196,27 @@ class BINDABLE : public BASE_NODE {
...
             for (auto sb : sparse_bindings) {
-                mem_set.insert(sb.mem);
+                bound_memory_set_.insert(sb.mem);
             }
         }
-        return mem_set;
     }
+
+    // Return unordered set of memory objects that are bound
+    const std::unordered_set<VkDeviceMemory> &GetBoundMemory() const { return bound_memory_set_; }
```
jzulauf-lunarg: Instead of creating a set from scratch each call we keep a cached one and return a const &

layers/descriptor_sets.cpp
```diff
@@ -36,51 +36,59 @@ cvdescriptorset::DescriptorSetLayout::DescriptorSetLayout(const VkDescriptorSetL
...
-    // Dyn array indicies are ordered by binding # and array index of any array within the binding
-    //  so we store up bindings w/ count in ordered map in order to create dyn array mappings below
+    binding_type_stats_ = {0, 0, 0};
+    std::set<uint32_t> sorted_bindings;
+    // Create the sorted set and unsorted map of bindings and indices
+    for (uint32_t i = 0; i < binding_count_; i++) {
+        sorted_bindings.insert(p_create_info->pBindings[i].binding);
+    }
+    uint32_t index = 0;
+    binding_to_index_map_.reserve(binding_count_);
```
jzulauf-lunarg: proactively reserve and resize as possible, as the reallocation was visible in profiles.

layers/descriptor_sets.cpp
```diff
@@ -36,51 +36,59 @@ cvdescriptorset::DescriptorSetLayout::DescriptorSetLayout(const VkDescriptorSetL
...
         auto binding_num = bindings_[i].binding;
-        binding_to_index_map_[binding_num] = i;
-        binding_to_global_start_index_map_[binding_num] = global_index;
-        global_index += bindings_[i].descriptorCount ? bindings_[i].descriptorCount - 1 : 0;
-        binding_to_global_end_index_map_[binding_num] = global_index;
-        global_index += bindings_[i].descriptorCount ? 1 : 0;
+        auto final_index = global_index + bindings_[i].descriptorCount;
+        binding_to_global_index_range_map_[binding_num] = IndexRange(global_index, final_index);
+        if (final_index != global_index) {
+            global_start_to_index_map_[global_index] = i;
```
jzulauf-lunarg: this map allows an non-iterative lookup of a binding from a global index... something that was previously potentially O(n^2)

layers/descriptor_sets.cpp
```diff
@@ -102,93 +110,64 @@ bool cvdescriptorset::DescriptorSetLayout::ValidateCreateInfo(debug_report_data
...
 
-// put all bindings into the given set
-void cvdescriptorset::DescriptorSetLayout::FillBindingSet(std::unordered_set<uint32_t> *binding_set) const {
-    for (auto binding_index_pair : binding_to_index_map_) binding_set->insert(binding_index_pair.first);
-}
-
-VkDescriptorSetLayoutBinding const *cvdescriptorset::DescriptorSetLayout::GetDescriptorSetLayoutBindingPtrFromBinding(
-    const uint32_t binding) const {
+// Return valid index or "end" i.e. binding_count_;
+uint32_t cvdescriptorset::DescriptorSetLayout::GetIndexFromBinding(uint32_t binding) const {
```
jzulauf-lunarg: Common code for all binding lookups.

layers/descriptor_sets.cpp
```diff
@@ -102,93 +110,64 @@ bool cvdescriptorset::DescriptorSetLayout::ValidateCreateInfo(debug_report_data
...
+    return index;
 }
-// For the given binding, return end index
-uint32_t cvdescriptorset::DescriptorSetLayout::GetGlobalEndIndexFromBinding(const uint32_t binding) const {
-    assert(binding_to_global_end_index_map_.count(binding));
-    const auto &btgei_itr = binding_to_global_end_index_map_.find(binding);
-    if (btgei_itr != binding_to_global_end_index_map_.end()) {
-        return btgei_itr->second;
+// For the given binding, return the global index range
+const cvdescriptorset::IndexRange &cvdescriptorset::DescriptorSetLayout::GetGlobalIndexRangeFromBinding(
```
jzulauf-lunarg: start and end were frequently looked up in pairs, this cuts lookup cost in half

layers/descriptor_sets.cpp
```diff
@@ -200,16 +179,17 @@ VkSampler const *cvdescriptorset::DescriptorSetLayout::GetImmutableSamplerPtrFro
...
 uint32_t cvdescriptorset::DescriptorSetLayout::GetNextValidBinding(const uint32_t binding) const {
-    uint32_t new_binding = binding;
-    do {
-        new_binding++;
-    } while (GetDescriptorCountFromBinding(new_binding) == 0);
-    return new_binding;
+    auto it = non_empty_bindings_.upper_bound(binding);
+    assert(it != non_empty_bindings_.cend());
+    if (it != non_empty_bindings_.cend()) return *it;
+    return GetMaxBinding() + 1;
```
jzulauf-lunarg: Another iterative binding walk that would show up in profiles is now an O(log N) tree search

layers/descriptor_sets.cpp
```diff
@@ -200,16 +179,17 @@ VkSampler const *cvdescriptorset::DescriptorSetLayout::GetImmutableSamplerPtrFro
...
 }
 // For given index, return ptr to ImmutableSampler array
 VkSampler const *cvdescriptorset::DescriptorSetLayout::GetImmutableSamplerPtrFromIndex(const uint32_t index) const {
-    assert(index < bindings_.size());
```
jzulauf-lunarg: The asserts in "Get" were reduced to the set where no valid answer (like null or 0) could be given

layers/descriptor_sets.cpp
```diff
@@ -319,7 +299,7 @@ cvdescriptorset::AllocateDescriptorSetsData::AllocateDescriptorSetsData(uint32_t
     : required_descriptors_by_type{}, layout_nodes(count, nullptr) {}
 
 cvdescriptorset::DescriptorSet::DescriptorSet(const VkDescriptorSet set, const VkDescriptorPool pool,
-                                              const std::shared_ptr<DescriptorSetLayout const> &layout, const layer_data *dev_data)
+                                              const std::shared_ptr<DescriptorSetLayout const> &layout, layer_data *dev_data)
```
jzulauf-lunarg: Validation caching required non-const access.

layers/descriptor_sets.cpp
```diff
@@ -438,7 +418,7 @@ bool cvdescriptorset::DescriptorSet::ValidateDrawState(const std::map<uint32_t,
                                   << " references invalid buffer " << buffer << ".";
                         *error = error_str.str();
                         return false;
-                    } else {
+                    } else if (!buffer_node->sparse) {
```
jzulauf-lunarg: Sparse resources can only be validated at submit time

layers/descriptor_sets.cpp
```diff
@@ -740,13 +727,69 @@ void cvdescriptorset::DescriptorSet::BindCommandBuffer(GLOBAL_CB_NODE *cb_node,
...
+            continue;
+        }
+        if ((layout_binding->descriptorType == VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC) ||
+            (layout_binding->descriptorType == VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC)) {
+            FilterAndTrackOneBindingReq(binding_req_pair, in_req, out_req, validated.dynamic_buffers, stats.dynamic_buffer_count);
+        } else if ((layout_binding->descriptorType == VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER) ||
+                   (layout_binding->descriptorType == VK_DESCRIPTOR_TYPE_STORAGE_BUFFER)) {
+            FilterAndTrackOneBindingReq(binding_req_pair, in_req, out_req, validated.non_dynamic_buffers,
+                                        stats.non_dynamic_buffer_count);
+        } else {
```
jzulauf-lunarg: Caching criteria differs per type.  If image_layout have changed , the image descriptors need to be validated against them.  This is rather crude, as the changed layouts may not impact the bound descriptors, but the simple "versioning" is a simple "dirt" test.

layers/descriptor_sets.h
```diff
@@ -135,29 +159,38 @@ class DescriptorSetLayout {
...
+    };
+    const BindingTypeStats &GetBindingTypeStats() const { return binding_type_stats_; }
+
    private:
     VkDescriptorSetLayout layout_;
-    std::map<uint32_t, uint32_t> binding_to_index_map_;
-    std::unordered_map<uint32_t, uint32_t> binding_to_global_start_index_map_;
-    std::unordered_map<uint32_t, uint32_t> binding_to_global_end_index_map_;
+    std::set<uint32_t> non_empty_bindings_;  // Containing non-emtpy bindings in numerical order
+    std::unordered_map<uint32_t, uint32_t> binding_to_index_map_;
```
jzulauf-lunarg: Changing the binding_to_index_map_ to "unordered" changes lookups from O(log n) to O(1), at the cost of an extra-set

chrisforbes:
There's a fair bit to process here. Thanks for the detailed notes -- will take a look at this when I'm back in the office.

mark-lunarg:
I would like to suggest that (where it is appropriate) the explanatory review comments you made be added to the source.

layers/descriptor_sets.cpp
```diff
@@ -168,27 +167,19 @@ VkShaderStageFlags cvdescriptorset::DescriptorSetLayout::GetStageFlagsFromBindin
...
 // For the given binding, return start index
-uint32_t cvdescriptorset::DescriptorSetLayout::GetGlobalStartIndexFromBinding(const uint32_t binding) const {
-    assert(binding_to_global_start_index_map_.count(binding));
-    const auto &btgsi_itr = binding_to_global_start_index_map_.find(binding);
-    if (btgsi_itr != binding_to_global_start_index_map_.end()) {
+const cvdescriptorset::IndexRange &cvdescriptorset::DescriptorSetLayout::GetGlobalIndexRangeFromBinding(
+    const uint32_t binding) const {
+    assert(binding_to_global_index_range_map_.count(binding));
+    const static IndexRange kInvalidRange = {0xFFFFFFFF, 0xFFFFFFFF};
+    const auto &btgsi_itr = binding_to_global_index_range_map_.find(binding);
```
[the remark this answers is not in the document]
jzulauf-lunarg: it's an *acronym* (I didn't understand and didn't change it).  btgir_itr seems too FORTRAN (or Welsh :) ) for clarity.  How about range_it?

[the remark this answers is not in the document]
tobine: Yes, "range_it" SGTM.

layers/core_validation_types.h
```diff
@@ -196,19 +196,27 @@ class BINDABLE : public BASE_NODE {
...
-        std::unordered_set<VkDeviceMemory> mem_set;
+
+    std::unordered_set<VkDeviceMemory> bound_memory_set_;
+
+    BINDABLE()
+        : sparse(false), binding{}, requirements{}, memory_requirements_checked(false), sparse_bindings{}, bound_memory_set_{} {};
+
+    // Update the cached set of memory bindings.
+    // Code that changes binding.mem or sparse_bindings must call UpdateBoundMemorySet()
+    void UpdateBoundMemorySet() {
```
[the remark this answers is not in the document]
jzulauf-lunarg: The best choice would be to make binding private and add the correct interfaces to the class.  Does that seem out of scope for this change?  I can clean it up as a subsequent change to the current commits, I can include it in the relevant commit, or I can create a clean-up commit prior to this commit.  Thoughts?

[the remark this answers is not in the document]
tobine: The fact that UpdateBoundMemorySet() was introduced here I'd be fine with the interface update in the CL or as a follow-on CL to clean up. Since the current CL seems pretty solid I'd just go the follow-on CL route.

jzulauf-lunarg:
Updated branch to reflect all review comments, re-validated for performance with same gains in target use case.

layers/descriptor_sets.h
```diff
@@ -401,8 +426,37 @@ class DescriptorSet : public BASE_NODE {
...
+    CachedValidationMap cached_validation_;
+};
+// For the "bindless" style resource usage with many descriptors, need to optimize binding and validation
+class PrefilterBindRequestMap {
+   public:
+    static const uint32_t kManyDescriptors_ = 64;  // TODO base this number on measured data
+    std::unique_ptr<BindingReqMap> filtered_map_;
+    const BindingReqMap &orig_map_;
+
+    PrefilterBindRequestMap(DescriptorSet &ds, const BindingReqMap &in_map, GLOBAL_CB_NODE *cb_state);
```
tobine: I'd like to see a comment here and/or at the corresponding Filter* functions above describing the difference in the two functions. If I'm understanding the code correctly, then the call w/ PIPELINE state is the "meaty" one that helps guard against repeat validation during pre-draw() time. The other call is more for book-keeping between ds<->cb bindings to simply save us from re-binding as much as possible. Yes?

chrisforbes:
I'd got distracted with other things and so didn't have time to read until
now, but have a thumbs-up for this after the fact.

PR #2304, opened 2018-01-02
icd: Move include to header to fix clang build

### LINKED ISSUE #2303: Building with clang/libc++ is broken
When building Vulkan-LoaderAndValidationLayers from latest git master, the build fails with [1].

The problem happens because #include <string> is missing in the generated file. Adding "#include <string>" to the generated icd/mock_icd.h file fixes the issue. The question remains where the include should really go.

[1] https://gist.github.com/elkrejzi/257bdadaebd82c74b1f1caaa59f2b616

karl-lunarg:
Move include <string> to header file to fix
clang build when using libstdc++.

Fixes #2303

Change-Id: Id2fc97be7cb83fa12ee1495ef57177c0a7dad893


karl-lunarg:
@tobine This LGTM.  Can we merge this?

PR #2305, opened 2018-01-02
loader: Allow layer to intercept commands before instance creation

lenny-lunarg:
Currently, layers cannot intercept functions that are called without an instance (`vkEnumerateInstanceExtensionProperties` and `vkEnumerateInstanceLayerProperties`). However, it can be useful for layers to intercept these functions. For example, the devsim layer @mikew-lunarg has in the VulkanTools repo needs to filter out extensions if it is to properly simulate devices.

This PR adds a new mechanism so that layers can intercept these calls. Only implicit layers can do this, since explicit layers are supposed to be explicitly enabled at instance creation while implicit layers are enabled externally before the program runs. A new mechanism to intercept the layers is needed because the current mechanism passes the information during instance creation, and relies on using the pNext chain in `VkInstanceCreateInfo` &mdash; neither of which can be used in the pre-instance functions.

For more information on this mechanism, read through my documentation changes in this PR. If the mechanism isn't clear from my doc changes, then I probably need to revise the docs.

### DIFF
--- a/loader/loader.c
+++ b/loader/loader.c
@@ -2305,6 +2305,28 @@ typedef struct {
     uint16_t patch;
 } layer_json_version;
 
+static inline bool is_valid_layer_json_version(const layer_json_version *layer_json) {
+    // Supported versions are: 1.0.0, 1.0.1, 1.1.0, 1.1.1, and 1.1.2.
+    if ((layer_json->major == 1 && layer_json->minor == 1 && layer_json->patch < 3) ||
+        (layer_json->major == 1 && layer_json->minor == 0 && layer_json->patch < 2)) {
+        return true;
+    }
+    return false;
+}
+
+static inline bool layer_json_supports_layers_tag(const layer_json_version *layer_json) {
+    // Supported versions started in 1.0.1, so anything newer
+    if ((layer_json->major > 1 || layer_json->minor > 0 || layer_json->patch > 1)) {
+        return true;
+    }
+    return false;
+}
+
+static inline bool layer_json_supports_pre_instance_tag(const layer_json_version *layer_json) {
+    // Supported versions started in 1.1.2, so anything newer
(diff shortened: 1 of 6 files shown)


loader/loader.c
```diff
@@ -2305,6 +2305,28 @@ typedef struct {
     uint16_t patch;
 } layer_json_version;
 
+static inline bool is_valid_layer_json_version(const layer_json_version *layer_json) {
+    // Supported versions are: 1.0.0, 1.0.1, and 1.1.0.
+    if ((layer_json->major == 1 && layer_json->minor == 1 && layer_json->patch < 3) ||
+        (layer_json->major == 1 && layer_json->minor == 0 && layer_json->patch < 3)) {
```
baldurk: Note that comment is out of date, this also supports 1.1.1 and 1.1.2, but it also supports 1.0.2 which is not a valid version I don't think?

loader/loader.c
```diff
@@ -5832,3 +5871,129 @@ VkStringErrorFlags vk_string_validate(const int max_length, const char *utf8) {
...
+        for (uint32_t i = 0; i < instance_layers.count; i++) {
+            struct loader_layer_properties *props = &instance_layers.list[i];
+            if (strcmp(props->info.layerName, pLayerName) == 0) {
+                global_ext_list = &props->instance_extension_list;
+                break;
+            }
+        }
+    } else {
+        // Scan/discover all ICD libraries
+        memset(&icd_tramp_list, 0, sizeof(struct loader_icd_tramp_list));
```
mikew-lunarg: sizeof(icd_tramp_list)

lenny-lunarg:
I just pushed up changes to address all of the issues mentioned above. I'll let travis and appveyor run again and then merge, assuming they still pass.

PR #2307, opened 2018-01-02
Tobin vkval std val array

tobine:
This is an alternative to #2293 

Just using an array in both C & C++ and killed vector. This is not quite as clean to instantiate a vector from but not terrible. It also prevents need for including anything from stl and pushes that decision to the app which I think makes more sense.

Note that this solution is set at compile-time so if a developer pulls a new SDK and the layer names or ordering have changed, they're required to re-compile with the updated SDK. My general feeling is that this makes sense as standard practice anyway. If you're developing a vulkan app and using validation, I would think that anytime you pull a new SDK, you're probably doing it to get new header updates that you'll recompile with anyway. If this is not the case then this simplified solution is probably infeasible.

### DIFF
--- a/layers/README.md
+++ b/layers/README.md
@@ -34,13 +34,15 @@ logging callbacks.
 
 Note that some layers are code-generated and will therefore exist in the directory `(build_dir)/layers`
 
-`include/vkLayer.h` - header file for layer code.
+`include/vulkan/vkLayer.h` - header file for layer code.
 
 ### Layer Details
-For complete details of current validation layers, including all of the validation checks that they perform, please refer to the document `layers/vk_validation_layer_details.md`. Below is a brief overview of each layer.
+For complete details of current validation layers, the code is the ultimate source. There is also a text database file that contains all of the valid usage statements from the Vulkan specification and catalogs which checks are currently covered in the validation layers. The database file can be accessed at `layers/vk_validation_error_database.txt`. Below is a brief overview of each layer.
 
 ### Standard Validation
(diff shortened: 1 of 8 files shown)


mark-lunarg:
If we go in this direction, I suggest changing the name of the header file (vk_standard_validation.h) to something more generic.  This scheme makes possible other supported, official, or merely just useful layer groupings and the current header file is unnecessarily limiting.

tobine:
@mark-lunarg Yes, makes sense. Could be vk_layer_groups.h or something more generic like that.

layers/README.md
```diff
@@ -34,13 +34,15 @@ logging callbacks.
...
+"Standard Validation" refers to the recommended stack of validation layers that should be used to validate an application's correct use of Vulkan. To enable the standard validation layer, its recommended that you include `include/vulkan/vk_standard_validation.h` which has const vars that contain the standard validation stack. Usage details are included in that header file, with the concept being that you pass the const vars into VkCreateInstanceInfo when creating the VkInstance. This will work across all platforms.
+
+There is also a deprecated meta-layer that works on some loaders. (name = `VK_LAYER_LUNARG_standard_validation`) - specifying this layer name will cause a supporting loader to load the all of the standard validation layers (listed below) in the following optimal order:  `VK_LAYER_GOOGLE_threading`, `VK_LAYER_LUNARG_parameter_validation`, `VK_LAYER_LUNARG_object_tracker`, `VK_LAYER_LUNARG_core_validation`, and `VK_LAYER_GOOGLE_unique_objects`. Other layers can be specified and the loader will remove duplicates.
```
jzulauf-lunarg: I'm not sure that I like the concept of deprecating the meta layer simply because it is not supported in the Android loader.  For all other platforms, the mechanism of specifying a single meta layer by name in registry or environment is the existing workflow for enabling validation without requiring direct application support, and needs to be retained for those platforms that support it.

tobine:
Closing this. I think eventual correct solution will be merging the validation layers in such a fashion that they're all contained under a single layer name. Fine to just live with what we have until we're there.

PR #2308, opened 2018-01-03
VS2013 Compatibility Round 2

mikes-lunarg:
Restore compatibility with Visual Studio 2013:
- Put noexcept specifier behind a common macro
- Provide move constructors and assignment operators because we cannot rely on the compiler to generate them
- Update BUILD.md to list VS2013 as supported again

TODO:
- Update AppVeyor configuration to run VS2013 -- in progress

### DIFF
--- a/tests/vktestbinding.cpp
+++ b/tests/vktestbinding.cpp
@@ -19,7 +19,8 @@
  * Author: Tony Barbour <tony@LunarG.com>
  */
 
-#include "vktestbinding.h"
+#include "test_common.h"    // NOEXCEPT macro (must precede vktestbinding.h)
+#include "vktestbinding.h"  // Left for clarity, no harm, already included via test_common.h
 #include <algorithm>
 #include <assert.h>
 #include <iostream>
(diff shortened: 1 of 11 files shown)


tests/vktestbinding.cpp
```diff
@@ -19,7 +19,8 @@
  * Author: Tony Barbour <tony@LunarG.com>
  */
 
-#include "vktestbinding.h"
+#include "test_common.h"    // NOEXCEPT macro (must precede vktestbinding.h)
```
jzulauf-lunarg: doesn't that usually imply that test_common.h should be being included in vktestbinding.h?

tests/vktestbinding.cpp
```diff
@@ -19,7 +19,8 @@
...
+#include "vktestbinding.h"  // Left for clarity, no harm, already included via test_common.h
```
jzulauf-lunarg: actually it confused rather than clarified for me.

daveh-lunarg: The inclusion tree is complicated by the desire to maintain a single NOEXCEPT macro definition that's shared by layers and tests.  test_common.h already pulled in a shared layer/test header, as well as vktestbinding.h in the correct order, so it's really the only header required within vktestbinding.cpp.  But since it's odd to have a foo.cpp that doesn't include it's corresponding foo.h, I left vktestbinding.h there after test_common.h as it causes no harm. 

If its just sowing confusion, then it can be deleted or at least commented out.

PR #2311, opened 2018-01-03
Update android-build/cmake scripts to build with Android Studio 3.0.0+

ggfan:
this is mainly to fix https://forums.khronos.org/showthread.php/13651-android-studio-issues?p=42974#post42974
@cnorthrop @tobine please take a look and add other people review too; and if ok, also please merge( I do not have write access)

### DIFF
--- a/build-android/cmake/README.md
+++ b/build-android/cmake/README.md
@@ -1,7 +1,9 @@
-Build Validation Layers with Android CMake
-=========================================
-Gradle's CMake library project in this directory builds layers into AAR;
-the project could be directly added into application's gradle projects
+Build Validation Layers with Android CMake Plugin
+=================================================
+Gradle project in this directory builds layers into AAR.
+The project could be directly added into application's gradle projects.
+[Android Studio 3.0.0+](https://developer.android.com/studio/index.html)
+IS required: earlier versions only publish release libs by default.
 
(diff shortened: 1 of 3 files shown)


build-android/cmake/README.md
```diff
@@ -1,7 +1,10 @@
...
-=========================================
-Gradle's CMake library project in this directory builds layers into AAR;
-the project could be directly added into application's gradle projects
+Build Validation Layers with Android CMake Plugin
+=================================================
+Gradle project in this directory builds layers into AAR.
+The project could be directly added into application's gradle projects.
+[Android Studio 3.0.0+](https://developer.android.com/studio/index.html)
+IS required: earlier versions only publish release libs [by default]
+(http://tools.android.com/tech-docs/new-build-system/user-guide#TOC-Library-Publication).
```
stroyan: This link now leads to a page that says "This doc is deprecated" and goes to https://developer.android.com/studio/build/index.html after ten seconds.

ggfan: thanks for catching it up: I will have to delete them all and android studio 3.0.0+ is the only option

cnorthrop:
This LGTM and is working with some follow up CI changes we plan to add for gradle coverage.  Thanks!

PR #2313, opened 2018-01-03
Add device/instance extension lists to extension helper header file

mark-lunarg:
These were previously generated in unique objects and used for the extension white-list.  The same information will be needed in other layers, so moved this definition to a common header file and updated U-O to use the new definitions.


scripts/helper_file_generator.py
```diff
@@ -660,6 +660,15 @@ def GenerateExtensionHelperHeader(self):
             struct += '    }\n'
             struct += '};\n'
             struct += '\n'
+            # Output reference lists of instance/device extension names
```
jzulauf-lunarg: Fits in nicely here.

PR #2317, opened 2018-01-04
layers: Fix loading on Linux

y-novikov:
Since every layer defines global vkGetDeviceProcAddr and other symbols,
Linux loader will use the symbol from the first library loaded,
by default, which will be Vulkan loader's symbol in our case,
instead of the local to current library symbol.

This causes loader's vkCreateInstance to get called multiple times
during loader_create_instance_chain and hang locking loader_lock.

So, fix this by assigning VkNegotiateLayerInterface members
to point to locally defined symbols instead of their global wrappers.


y-novikov:
@tobine, @cnorthrop, PTAL

mark-lunarg:
@y-novikov, the guys will have a look at the PR.  Please update the commit message to match the rest of the repo when you have a chance -- it should start with 'layers: '.  Thanks!

tobine:
thanks, @y-novikov . Just conferred w/ @cnorthrop and we determined that correct course of action is for ANGLE layer build to use -Bsymbolic as you had suggested in the ANGLE issue. The complete list of compiler options to be used for the layers on linux can be seen [here](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/CMakeLists.txt#L160).
Please double-check that everything matches up.
I'll close this PR.

PR #2318, opened 2018-01-04
build: Improve update ext srcs script for better CI

karl-lunarg:
Add options (--debug --release --32 --64) to the
update_external_sources.bat script so that it can
be used to build external libraries for only the
desired platform/configuration combinations.

This helps the AppVeyor CI builds by building only
the platform/configuration combination needed for the
current job.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -8,7 +8,7 @@
 # build version format
 version: "{build}"
 
-# Probably not useful for free accounts, but ask anyway.
+# Free accounts have a max of 1, but ask anyway.
 max_jobs: 4
 
 os:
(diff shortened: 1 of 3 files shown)


cnorthrop:
Excellent, thanks Karl!  On first inspection, looks good, I'll kick the tires on it this afternoon.  Should enable us to spin up some more Windows slaves and really bring down our presubmit test time.

karl-lunarg:
It seems we're down from about 58 mins to 36 mins, but that includes switching from 2015 to 2013.

cnorthrop:
I misread this PR.  It doesn't include changes to `build_windows_target.bat` so we can't parallelize yet.  Karl, could you also enhance that script to take the same parameter mix?

karl-lunarg:
I believe that `build_windows_targets.bat` already has parameters to build only 32-bit or only 64-bit.  But it builds both Debug and Release for each.   Is this good enough, or do you need it to build only Debug or Release as specified by parameters?

karl-lunarg:
OK, I modified `build_windows_targets.bat` to allow building only the debug or only the release configs.

cnorthrop:
This is giving some nice speed ups on a single system (laptop with 8 cores).  Should be even better if we can bring up some more Windows build systems:

```
Before
toolchain time:  ~10m30s
project time:  ~9m21s

After (running 4-wide 32/64/debug/release)
toolchain time:  ~4m0s
project time:  ~6m11s
```

I'm seeing the follow errors periodically on my system, even without running in parallel.  Most likely more missing Windows dependencies in cmake somewhere.  Doesn't need to hold up this PR, but worth looking into.  I don't know that we've seen it on the farm:
```
CUSTOMBUILD : CMake error : Cannot open file for write: C:/cygwin64/home/cody/code/sandbox/Vulkan-LoaderAndValidationLayers-A/bui
ld32/loader/cmake_install.cmake.tmp [C:\cygwin64\home\cody\code\sandbox\Vulkan-LoaderAndValidationLayers-A\build32\layers\VkLayer
_utils.vcxproj]
CUSTOMBUILD : CMake error : : System Error: Permission denied [C:\cygwin64\home\cody\code\sandbox\Vulkan-LoaderAndValidationLayer
s-A\build32\layers\VkLayer_utils.vcxproj]
```

karl-lunarg:
I didn't see any messages like those during my testing.  Perhaps you had Visual Studio open while running CMake?

PR #2325, opened 2018-01-08
Update LVL to version 1.0.67 of the Vulkan header

mark-lunarg:
- updated vulkan.h
- updated cgenerator.py
- updated generator.py
- updated reg.py
- updated vk.xml
- updated layer json files
- updated vulkan.hpp
- updated vuid_mapping.py
- updated vk_validation_error_database.txt for spec changes, new VUIDs
- updated vk_validation_error_messages.h


mark-lunarg:
The failure is due to running clang-format on Khronos-hosted files, and is OK to ignore for this PR.

PR #2326, opened 2018-01-08
Python Script Updates

mark-lunarg:
Removed an unused script and changed the default output mode for doc validator tests


tests/_vkvalidatelayerdoc.ps1
```diff
@@ -18,9 +18,7 @@ if ($args[0] -eq "-Debug") {
     $dPath = "Release"
 }
 
-if (($args[0] -eq "terse_mode") -Or ($args[1] -eq "terse_mode")) {
-    $output_mode = "terse_mode"
-}
+$output_mode = "terse_mode"
```
jzulauf-lunarg: Shouldn't we replace the arg 0/1 check with "verbose_mode" and set terse unless verbose is found?

tests/vkvalidatelayerdoc.sh
```diff
@@ -24,7 +24,7 @@ printf "$GREEN[ RUN      ]$NC $0\n"
 pushd ../../scripts
 
 # Validate that layer database matches source contents
-python3 vk_validation_stats.py $1
+python3 vk_validation_stats.py terse_mode
```
jzulauf-lunarg: same thing here.  look for -v or --vebose_mode as $a and pass on terse or not as appropriate.

mark-lunarg:
Changed logic over to use a 'verbose' arg.

tests/vkvalidatelayerdoc.sh
```diff
@@ -24,7 +24,7 @@ printf "$GREEN[ RUN      ]$NC $0\n"
 pushd ../../scripts
 
 # Validate that layer database matches source contents
-python3 vk_validation_stats.py $1
+python3 vk_validation_stats.py
```
jzulauf-lunarg: We've still lost the ability to control the output mode in this script.

tests/_vkvalidatelayerdoc.ps1
```diff
@@ -18,18 +18,14 @@ if ($args[0] -eq "-Debug") {
...
-
 write-host -background black -foreground green "[  RUN     ] " -nonewline
 write-host "vkvalidatelayerdoc.ps1: Validate layer documentation"
 
 # Run doc validation from project scripts dir
 push-location ..\..\scripts
 
 # Validate that layer documentation matches source contents
-python vk_validation_stats.py $output_mode
+python vk_validation_stats.py
```
jzulauf-lunarg: We used to be able to control the output mode, but we've lost that here.

mark-lunarg: True, and having the verbose output is an unnecessary use-case for these scripts.  The verbose output is still available by running the doc validator directly.

PR #2327, opened 2018-01-08
layers: Make DescriptorSetLayout robust for CTS

jzulauf-lunarg:
CTS sends an invalid descriptor set that the validation tracking was not
robust to -- two binding with identical binding ID values.  After
reporting the invalid input data, the test application could crash.
Validation internal data is now created in a way that invalid data
should less easily cause program termination.

Change-Id: Ib84a5234267f7bece9be377badd22696e3633b1b


chrisforbes:
Which CTS test is doing this?

Tony-LunarG:
dEQP-VK.api.descriptor_pool.out_of_pool_memory

jzulauf-lunarg:
@chrisforbes  -- The CTS should probably be fixed, but my recent change made LVL *less* robust in this regard.  It's a quick fix that doesn't hurt perf.

Issue #2241, opened 2017-11-30
Null pointer access in PreCallValidateBindBufferMemory


asuonpaa:
I'm seeing a null pointer access in core_validation.cpp line 3758 function PreCallValidateBindBufferMemory. There mem_info is being used without checking if it's null:
```
// Validate memory requirements size
        if (buffer_state->requirements.size > (mem_info->alloc_info.allocationSize - memoryOffset)) {
```

Earlier in the same function it is checked:
```
// Validate bound memory range information
        auto mem_info = GetMemObjInfo(dev_data, mem);
        if (mem_info) {
```

mark-lunarg:
@asuonpaa, were there any validation errors prior to reaching this point?  The Core Validation layer (to a large but not complete extent) relies on other layers to do these types of checks, and as you have noticed does not perform them pervasively.  If an invalid access was encountered with no other validation errors then we **would** need to make a change here -- let us know!

asuonpaa:
@mark-lunarg : Thanks for your help. I'm not seeing errors before this. Also if that object would be guaranteed to not be null by some other layer, then the existing checks of mem_info would be unnecessary. So IMO a consistent solution is to check the pointer at each location it's being used in that function.

mark-lunarg:
@asuonpaa, I see.  Is there any way you can get us a straightforward repro case, or some instructions on how to reproduce this issue -- this does sound like a layer bug, one way or the other.  As for the Null checks, for the most part they should be unnecessar, and the time it would take to scrub the layers of unnecessary checks is not near the top of the priority list.

mark-lunarg:
@asuonpaa, is there any way you can share a test case?

tobine:
Note that having a valid VkDeviceMemory param in vkBindBufferMemory() is VALIDATION_ERROR_1700c601 which is checked in object_tracker.

asuonpaa:
The root cause seems to be the fact I'm using multiple device and allocator objects. Most likely those are confusing the validator. I'm going to roll back to using only a single device object so the original validator issue is no longer bothering me.

jzulauf-lunarg:
There is a sparse memory CTS with the same issue... I have a fix pending in https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/2336 if you'd like to review/comment/try.

jzulauf-lunarg:
Regarding: 
[...]

PR #2336, opened 2018-01-12
layers: Make bind more robust to invalid memory

jzulauf-lunarg:
Addresses crashes in validation forcertain sparse memory CTS testing.

Added nullptr check to BindBufferMemory and BindImageMemory validation
to cope with invalid memory arguments (reported elsewhere).

Change-Id: I52266490ccfe9eff7d5f451cddc4d3bbf94b3c5d


jzulauf-lunarg:
Should also fix #2241

PR #2340, opened 2018-01-13
build: Use DevSim from a VulkanTools build in CI

karl-lunarg:
- Change Travis-CI to build VulkanTools and use the resulting
  DevSim binary instead of the one checked into the repo.
- Delete the DevSim binary and JSON file from layers/linux.
- Remove CMake rules related to DevSim binary.


karl-lunarg:
I wanted to check with @tobine and @cnorthrop to make sure that removing the DevSim binary isn't going to cause any unanticipated problems.  Our only view of its usage is in Travis-CI; we don't know if other CI systems depend on it.

Instead of using the binary from the repo, the Travis-CI job now builds VulkanTools and uses DevSim from there.   With extra parameters on VulkanTools' update_external_sources.sh and cmake, building DevSim only adds about 45 secs to the job time.

mikew-lunarg:
Cool, LGTM.  This will add additional valuable test coverage for DevSim.

cnorthrop:
I like it, good idea.  Didn't realize DevSim could be built so quickly.

Issue #2341, opened 2018-01-15
Loader leaks critical sections


scyganINTEL:
When running any Vulkan app under ['Application Verifier'](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/application-verifier) - with all basic checks selected, I am getting following error on `FreeLibrary(<vulkan-1.dll handle): 

```
=======================================
VERIFIER STOP 0000000000000201: pid 0x13CC: Unloading DLL containing an active critical section. 

	00007FFBAF88F7A0 : Critical section address. Run !cs -s <address> to get more information.
	00000151877DC650 : Critical section initialization stack trace. Run dps <address> to dump the stack trace.
	000001524081BFE0 : DLL name address.
	00007FFBAF790000 : DLL base address.
```

I have matched the address of the critical section to `loader_json_lock` mutex from https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/874fbdaad9733e71274906e65475d716ffb07c70/loader/loader.c#L1909: 
```
void loader_initialize(void) {
    // initialize mutexs
    loader_platform_thread_create_mutex(&loader_lock);
    loader_platform_thread_create_mutex(&loader_json_lock);

    // initialize logging
```

This looks like a bug - both mutex-es from above are not destroyed anywhere. 

However, in the absence of DllMain (?)  it is unclear to me where to put the cleanup code - hence this issue.

scyganINTEL:
I am seeing this on sdk-1.0.65 and latest master (6ce028589b360d98c1789762895aaf6462792ea4)

lenny-lunarg:
When I try running cube from master, I actually get a different error from the Application Verifier:

```
VERIFIER STOP 0000000000000900: pid 0x6230: A heap allocation was leaked. 

	000001B395F42550 : Address of the leaked allocation. Run !heap -p -a <address> to get additional information about the allocation.
	000001B38A482C80 : Address to the allocation stack trace. Run dps <address> to view the allocation stack.
	000001B394F76FD8 : Address of the owner dll name. Run du <address> to read the dll name.
	00007FFFA6FC0000 : Base of the owner dll. Run .reload <dll_name> = <address> to reload the owner dll. Use 'lm' to get more information about the loaded and unloaded modules.
```

The call stack shows that this is happening when calling `FreeLibrary()` on the Intel driver. Checking the name of the dll that's creating the error gives:

```
[...]

PR #2343, opened 2018-01-16
loader: Clean up mutexes on destruction

lenny-lunarg:
When unloading the vulkan loader dynamic library, a couple of mutexes were not getting cleaned up. This PR adds a mechanism to clean them up.

This PR addresses issue #2341.

### DIFF
--- a/loader/loader.c
+++ b/loader/loader.c
@@ -92,8 +92,6 @@ uint32_t g_loader_log_msgs = 0;
 loader_platform_thread_mutex loader_lock;
 loader_platform_thread_mutex loader_json_lock;
 
-LOADER_PLATFORM_THREAD_ONCE_DECLARATION(once_init);
-
 void *loader_instance_heap_alloc(const struct loader_instance *instance, size_t size, VkSystemAllocationScope alloc_scope) {
     void *pMemory = NULL;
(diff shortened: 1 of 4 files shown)


loader/loader.c
```diff
@@ -5997,3 +5998,26 @@ VKAPI_ATTR VkResult VKAPI_CALL terminator_EnumerateInstanceLayerProperties(const
...
+        default:
+            // Do nothing
+            break;
+    }
+    return TRUE;
+}
+#else
+__attribute__((constructor)) void loader_init_dll() { loader_initialize(); }
+
+__attribute__((destructor)) void loader_free_dll() { loader_release(); }
```
MarkY-LunarG: I don't understand how the above two items get triggered on Linux.  Plus the name _dll shouldn't be used for any other platform.

lenny-lunarg: The naming was a poor choice. I must have been thinking in terms of Windows when I wrote that. I'll fix it before I merge.

The two linux items get triggered by the `__attribute__` labels. For more info go [here](https://gcc.gnu.org/onlinedocs/gcc-4.3.0/gcc/Function-Attributes.html) and search for the constructor and destructor section.

PR #2349, opened 2018-01-17
Add VUs for param limit checks in CmdDispatch and CmdDispatchBaseKHX

### LINKED ISSUE #2296: vkCmdDispatch doesn't validate groupCount parameters against maxComputeWorkGroupCount limits
These VU statements on vkCmdDispatch don't appear to be implemented:

* `groupCountX` must be less than or equal to `VkPhysicalDeviceLimits::maxComputeWorkGroupCount[0]`
* `groupCountY` must be less than or equal to `VkPhysicalDeviceLimits::maxComputeWorkGroupCount[1]`
* `groupCountZ` must be less than or equal to `VkPhysicalDeviceLimits::maxComputeWorkGroupCount[2]`

See [this StackOverflow question](https://stackoverflow.com/questions/47927866/vulkan-buffer-workgroupid-not-returning-actual-value-when-large-number-of-elemen). A quick grep through the validation code didn't find anything that looked like an implementation of this check.

daveh-lunarg:
Adds valid usage checks for CmdDispatch & CmdDispatchBaseKHX parameters to validate count and base parameters against the device maxComputeWorkGroupCount limits.  Adds a new test to exercise each of the VUs.

Resolves #2296.


layers/parameter_validation_utils.cpp
```diff
@@ -2554,6 +2558,88 @@ bool pv_vkCreateDescriptorPool(VkDevice device, const VkDescriptorPoolCreateInfo
...
 
+bool pv_vkCmdDispatch(VkCommandBuffer commandBuffer, uint32_t groupCountX, uint32_t groupCountY, uint32_t groupCountZ) {
+    bool skip = false;
+    layer_data *device_data = GetLayerDataPtr(get_dispatch_key(commandBuffer), layer_data_map);
+
+    if (groupCountX > device_data->device_limits.maxComputeWorkGroupCount[0]) {
+        skip |=
+            log_msg(device_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
+                    HandleToUint64(commandBuffer), __LINE__, VALIDATION_ERROR_19c00304, LayerName,
+                    "vkCmdDispatch(): groupCountX (%1d) exceeds device limit maxComputeWorkGroupCount[0] (%1d). %s", groupCountX,
```
mikes-lunarg: %1d is odd format specifier for uint32_t data, %u makes more sense, PRIu32 is probably best

tests/layer_validation_tests.cpp
```diff
@@ -8408,6 +8408,118 @@ TEST_F(VkLayerTest, InvalidPipeline) {
...
 
+TEST_F(VkLayerTest, CmdDispatchExceedLimits) {
+    TEST_DESCRIPTION("Compute dispatch with dimensions that exceed device limits");
+
+    // Enable KHX device group extensions, if available
+    bool khx_dg_ext_available = false;
+    ASSERT_NO_FATAL_FAILURE(InitFramework(myDbgFunc, m_errorMonitor));
+    if (DeviceExtensionSupported(gpu(), nullptr, VK_KHX_DEVICE_GROUP_EXTENSION_NAME)) {
+        m_device_extension_names.push_back(VK_KHX_DEVICE_GROUP_EXTENSION_NAME);
+        if (DeviceExtensionSupported(gpu(), nullptr, VK_KHX_DEVICE_GROUP_CREATION_EXTENSION_NAME)) {
```
mikes-lunarg: VK_KHX_device_group_creation is an instance extension

tobine:
vkCmdDispatchIndirect() has similar VUs (VALIDATION_ERROR_06e00342, 4, 6). It's a bit trickier, though, as the x, y, z values are read from a buffer at execution time so validation probably makes sense in core_validation layer and would be delayed 'til QueueSubmit time. Ideally those checks could share code from vkCmdDispatch() checks which requires making code common between PV and CV layers.

tests/layer_validation_tests.cpp
daveh-lunarg: But it has 'device' right there in the name...  :)

You're right, of course.

layers/parameter_validation_utils.cpp
daveh-lunarg: Yes

layers/parameter_validation_utils.cpp
```diff
@@ -2554,6 +2554,88 @@ bool pv_vkCreateDescriptorPool(VkDevice device, const VkDescriptorPoolCreateInfo
...
+
+    return skip;
+}
+
+bool pv_vkCmdDispatchBaseKHX(VkCommandBuffer commandBuffer, uint32_t baseGroupX, uint32_t baseGroupY, uint32_t baseGroupZ,
+                             uint32_t groupCountX, uint32_t groupCountY, uint32_t groupCountZ) {
+    bool skip = false;
+    layer_data *device_data = GetLayerDataPtr(get_dispatch_key(commandBuffer), layer_data_map);
+
+    // Paired if {} else if {} tests used to avoid any possible uint underflow
```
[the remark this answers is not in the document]
daveh-lunarg: That would seem to require either genericizing the error message text to be less specific, or parameterizing the helper to the extent it's nearly as ugly as log_msg itself. I think it's generally better to avoid the proliferation of variations-on-a-theme helpers.  Especially so in this case, where as you mention the LOC savings would be minimal.

layers/parameter_validation_utils.cpp
```diff
@@ -2554,6 +2554,88 @@ bool pv_vkCreateDescriptorPool(VkDevice device, const VkDescriptorPoolCreateInfo
...
+    uint32_t limit = device_data->device_limits.maxComputeWorkGroupCount[0];
+    if (baseGroupX >= limit) {
+        skip |= log_msg(device_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
```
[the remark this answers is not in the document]
daveh-lunarg: Yep

Issue #2184, opened 2017-11-06
Break apart DSUsageBitsErrors test and use appropriate helpers


chrisforbes:

mark-lunarg:
See #2181 for context.

jzulauf-lunarg:
Cleaned.

PR #2359, opened 2018-01-19
Clean up/use framework in DSUsageBitsErrors

jzulauf-lunarg:
Cleaned up DSUsageBitsError, using the framework calls to reduce boilerplate code to improve maintainability and readability.  Added helper calls as needed.  Addresses (fixes?) #2184

### DIFF
--- a/tests/vktestbinding.h
+++ b/tests/vktestbinding.h
@@ -17,6 +17,7 @@
  *
  * Author: Courtney Goeltzenleuchter <courtney@LunarG.com>
  * Author: Cody Northrop <cody@lunarg.com>
+ * Author: John Zulauf <jzulauf@lunarg.com>
  */
 
 #ifndef VKTESTBINDING_H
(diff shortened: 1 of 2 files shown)


tests/vktestbinding.h
```diff
@@ -615,13 +630,30 @@ class DescriptorPool : public internal::NonDispHandle<VkDescriptorPool> {
     std::vector<DescriptorSet *> alloc_sets(const Device &dev, const DescriptorSetLayout &layout, uint32_t count);
     DescriptorSet *alloc_sets(const Device &dev, const DescriptorSetLayout &layout);
 
+    template <typename PoolSizes>
```
jzulauf-lunarg: The interface is written to be agnostic to vector or array (both of which support data and size), but if folks find it too odd, I can wind it back to just vector.

jzulauf-lunarg:
@chrisforbes I wasn't clear on the "break apart" element of #2184.  It's just a single loop over all the issues, once you clean the cruft.  The clean up saves about half the lines, but there's any number of other tests that could shed line count with a similar application of helpers.

jzulauf-lunarg:
Possible to get a quick review to merge and close?

chrisforbes:
@jzulauf-lunarg the 'break apart' was to make failures easy to interpret from just the test name. These complex tests with many subtests always require more digging.

Issue #2365, opened 2018-01-21
Validate push descriptor count in vkCreateDescriptorSetLayout


zeux:
vkCreateDescriptorSetLayout says that the number of push descriptors is limited by maxPushDescriptors that can be obtained by querying it (and has to be at least 32):

> If flags contains VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR, then the total number of elements of all bindings must be less than or equal to VkPhysicalDevicePushDescriptorPropertiesKHR::maxPushDescriptors

 There is no validation for this right now.

jzulauf-lunarg:
Added to vkCreateDescriptorSetLayout validation

Issue #2362, opened 2018-01-20
Validate descriptor types for vkCmdPushDescriptorSetKHR


zeux:
I'm not sure whether it makes sense to handle it in `vkCmdPushDescriptorSetKHR`, or when creating descriptors with `VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR` (which requires #2361...), but the spec says that push descriptor sets can not work with dynamic descriptor types, and validation layers don't check this at creation or `vkCmdPushDescriptorSetKHR` time.

jzulauf-lunarg:
This should be added to DescriptorSetLayout::ValidateCreateInfo
VUID-VkDescriptorSetLayoutCreateInfo-flags-00280

jzulauf-lunarg:
fixed in validation of vkCreateDescriptorSetLayout

PR #2369, opened 2018-01-23
Validate push descriptor/descriptor set layout create interactions

jzulauf-lunarg:
Add VUID checks and unit tests to address #2362, #2365.  Additionally add "required extension" check and unit test for vkCreateDescriptorSetLayout path re: VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR usage, and repaired unit test not enabling required extension for ..._BIT_KHR.

VALIDATION_ERROR_05000230  VkDescriptorSetLayoutCreateInfo-flags-00280
VALIDATION_ERROR_05000232  VkDescriptorSetLayoutCreateInfo-flags-00281


layers/core_validation.cpp
```diff
@@ -4704,7 +4718,9 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateSampler(VkDevice device, const VkSamplerCre
 
 static bool PreCallValidateCreateDescriptorSetLayout(layer_data *dev_data, const VkDescriptorSetLayoutCreateInfo *create_info) {
     if (dev_data->instance_data->disabled.create_descriptor_set_layout) return false;
-    return cvdescriptorset::DescriptorSetLayout::ValidateCreateInfo(dev_data->report_data, create_info);
+    return cvdescriptorset::DescriptorSetLayout::ValidateCreateInfo(dev_data->report_data, create_info,
```
jzulauf-lunarg: Note the layer_data is not defined outside core_validation.cpp -- so we pass the req'd data piece-wise.

layers/core_validation.cpp
```diff
@@ -2173,6 +2178,15 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDevice(VkPhysicalDevice gpu, const VkDevice
     // Store physical device properties and physical device mem limits into device layer_data structs
     instance_data->dispatch_table.GetPhysicalDeviceMemoryProperties(gpu, &device_data->phys_dev_mem_props);
     instance_data->dispatch_table.GetPhysicalDeviceProperties(gpu, &device_data->phys_dev_props);
+
+    if (device_data->extensions.vk_khr_push_descriptor) {
```
jzulauf-lunarg: This is the simplest way to get *one* extension property loaded.  Will need to refactor to capture all extension limits.

layers/descriptor_sets.cpp
```diff
@@ -107,17 +107,54 @@ cvdescriptorset::DescriptorSetLayout::DescriptorSetLayout(const VkDescriptorSetL
...
-                                                              const VkDescriptorSetLayoutCreateInfo *create_info) {
+bool cvdescriptorset::DescriptorSetLayout::ValidateCreateInfo(const debug_report_data *report_data,
+                                                              const VkDescriptorSetLayoutCreateInfo *create_info,
+                                                              const bool push_descriptor_ext, const uint32_t max_push_descriptors) {
     bool skip = false;
     std::unordered_set<uint32_t> bindings;
+    uint64_t total_descriptors = 0;
+
+    const bool push_descriptor_set = create_info->flags & VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR;
+    if (push_descriptor_set && !push_descriptor_ext) {
```
jzulauf-lunarg: There is no VUID for using extension defined types, enums etc, except for the body text 

> (31.2. Extensions) ... Enabling an extension does not change behavior of functionality exposed by the core Vulkan API or any other extension, other than making valid the use of the commands, enums and structures defined by that extension.

Which implies that using the define types, etc. is *invalid* if the extension is *not* enabled..

scripts/helper_file_generator.py
```diff
@@ -1142,22 +1143,35 @@ def GenerateTypeMapHelperHeader(self):
             '    }}',
             '    return found;',
             '}}',
+            '',
```
jzulauf-lunarg: helpers that allow for terse "correct by design" initialization of sType /pNext fields

scripts/helper_file_generator.py
```diff
@@ -1175,6 +1189,13 @@ def GenerateTypeMapHelperHeader(self):
             if item.ifdef_protect != None:
                 code.append('#endif // %s' % item.ifdef_protect)
 
+        # Generate utilites for all types
```
jzulauf-lunarg: as the utilities reference the traits, move the utilities after the traits

tests/layer_validation_tests.cpp
```diff
@@ -21472,6 +21472,91 @@ TEST_F(VkLayerTest, DuplicateDescriptorBinding) {
...
+    ds_layout_ci.bindingCount = 1;
+    ds_layout_ci.pBindings = &binding;
+
+    std::string error = "Attemped to use VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR in ";
+    error = error + "VkDescriptorSetLayoutCreateInfo::flags but its required extension ";
+    error = error + VK_KHR_PUSH_DESCRIPTOR_EXTENSION_NAME;
+    error = error + " has not been enabled.";
+
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, error.c_str());
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_05000232);
```
jzulauf-lunarg: This silences an unexpected, as the second error message is implied by the first.

scripts/helper_file_generator.py
mark-lunarg: three I's in utilities

jzulauf-lunarg:
@tobine -- Needed to add  VK_KHR_push_descriptor to the mock_icd ignored list, because a DevicePhysicalProperties2 -> pNext query was failing to return a valid value, and causing validation to create unexpected errors in Positive tests and in the required "positive" functionality to set up negative tests.

If you have a better short term solution, I'm game.  Long term it would nice to be able to enable the extension with proper Get*Property* support.

tobine:
@jzulauf-lunarg Yes, I'm ok with disabling VK_KHR_push_descriptor in mock for short-term. Can you open bug on me to update mock to improve GPDP2 support for extension in the mock?

PR #2372, opened 2018-01-24
Don't complain about unused vertex buffers being VK_NULL_HANDLE

### LINKED ISSUE #2371: Non consecutive vertex buffer bindings causes layer to report a warning
Create multiple vertex buffers with non-consecutive binding indices:
```
VkVertexInputBindingDescription binding1{0, stride1, VK_VERTEX_INPUT_RATE_VERTEX};
VkVertexInputBindingDescription binding2{2, stride2, VK_VERTEX_INPUT_RATE_VERTEX};
```
Bind vertex buffers to specified indices:
```
vkCmdBindVertexBuffers(commandBuffer, 0, 1, &buffer1, &offset);
vkCmdBindVertexBuffers(commandBuffer, 2, 1, &buffer2, &offset);
```
Model renders as expected but validation layer emits
`validation layer: Cannot submit cmd buffer using deleted buffer 0x0.`

chrisforbes:
Fixes #2371.

There are still some corner cases here -- there's no requirement that a vertex buffer NOT consumed by the pipeline be valid, see VU language:

```
All vertex input bindings accessed via vertex input variables declared in the vertex shader entry point’s interface must have valid buffers bound
```


chrisforbes:
There's a bigger chunk of cleanup work here-- we should get rid of the parallel tracking of vertex buffers and use the normal resource tracking for it instead. At draw time, we should take the actual set of buffer bindings that are needed for the pipeline, and add those to the CB's dependent resources map.

PR #2373, opened 2018-01-24
Demo fixes for submodule operation

mark-lunarg:
The VulkanTools repo is being reorganized, and some LVL-side cmake/structure fixes were necessary.


scripts/vk_validation_stats.py
```diff
@@ -193,6 +193,10 @@ def __init__(self, source_file_list, generated_source_file_list, generated_sourc
                     if os.path.isfile(filepath):
                         qualified_paths.append(filepath)
                         break
+                    filepath = '../../../%s/submodules/Vulkan-LoaderAndValidationLayers/layers/%s' % (build_dir, source)
```
karl-lunarg: This assumes that the parent repo is placing the LVL repo as a submodule in a specific location.  So, this won't work if a parent repo decides to put LVL someplace else.

Do we really expect this script to be run in repos where LVL is a submodule?

If so, is there a way to get the directory containing THIS script and then constructing filepath by appending ../build_dir/layers/source to it. (as in the code just above).

tests/_vkvalidatelayerdoc.ps1
```diff
@@ -4,7 +4,12 @@
 #    cd C:\src\Vulkan-LoaderAndValidationLayers\build\tests
 #    .\vkvalidatelayerdoc.ps1 [-Debug]
 
-if (-not (Test-Path -LiteralPath '..\..\scripts')) {
+# Run doc validation from project scripts dir
+if (Test-Path -LiteralPath '..\..\scripts') {
+    push-location ..\..\scripts
+} elseif (Test-Path -LiteralPath '..\..\..\..\submodules\Vulkan-LoaderAndValidationLayers\scripts') {
```
karl-lunarg: Same comment as above.

tests/vkvalidatelayerdoc.sh
```diff
@@ -10,8 +10,14 @@ else
     GREEN=''
     NC=''
 fi
+
+# Run doc validation from project scripts dir
 # If we can't find the source dir then skip
-if [ ! -d "../../scripts" ]; then
+if [ -d "../../scripts" ]; then
+    pushd ../../scripts
+elif [ -d "../../../../submodules/Vulkan-LoaderAndValidationLayers/scripts" ]; then
```
karl-lunarg: and ditto...

scripts/vk_validation_stats.py
mark-lunarg: I was thinking CI would include one of the 'run_all_tests' scripts, and so it'd need to work.  If not, I can just leave these scripts alone.  O/W, we're not giving anything up by having hard-coded paths as this script, if of less use in VT than in LVL, is unquestionably of little or no utility elsewhere.  Should I just remove the doc validator script changes?

mark-lunarg:
We decided to leave the doc validator scripts as-is, and only support them in the LVL repo.  Removed the commit containing those changes.

Issue #2361, opened 2018-01-20
Validate descriptor operations for VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR


zeux:
The spec says that

> VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR specifies that descriptor sets must not be allocated using this layout, and descriptors are instead pushed by vkCmdPushDescriptorSetKHR

Validation layers currently allow allocating descriptors with layouts created with this flag, this most likely needs to be an error.

jzulauf-lunarg:
VUID-VkDescriptorSetAllocateInfo-pSetLayouts-00308

jzulauf-lunarg:
Addressed in above commit.

PR #2375, opened 2018-01-26
Add Validation for push descriptor sets re allocation and push

jzulauf-lunarg:
Addresses #2361 (ensuring push descriptor set are not allocated) and adds validation for vkCmdPushDescriptorSetKHR VUID's


layers/core_validation.cpp
```diff
@@ -6045,11 +6046,73 @@ VKAPI_ATTR void VKAPI_CALL CmdBindDescriptorSets(VkCommandBuffer commandBuffer,
...
+static bool PreCallValidateCmdPushDescriptorSetKHR(layer_data *device_data, GLOBAL_CB_NODE *cb_state,
+                                                   const VkPipelineBindPoint bind_point, const VkPipelineLayout layout,
+                                                   const uint32_t set, const uint32_t descriptor_write_count,
+                                                   const VkWriteDescriptorSet *descriptor_writes, const char *func_name) {
+    bool skip = false;
+    skip |= ValidateCmd(device_data, cb_state, CMD_PUSHDESCRIPTORSET, func_name);
+    skip |= ValidateCmdQueueFlags(device_data, cb_state, func_name, (VK_QUEUE_GRAPHICS_BIT | VK_QUEUE_COMPUTE_BIT),
+                                  VALIDATION_ERROR_1be02415);
+    skip |= ValidatePipelineBindPoint(device_data, cb_state, bind_point, func_name,
+                                      {{VALIDATION_ERROR_1be002d6, VALIDATION_ERROR_1be002d6}});
```
jzulauf-lunarg: Note the {{. If the formal argument were a std::vector, this would be required, but for std::array only { *should* be required, but for portability reasons (clang until very recently warned unnecessarily), {{ needs to be used even for std::array.  Thankfully {{ is legal and valid for std::array, while the std::vector two-element single brace {} initialization would yield a {length, fill-value} vector 3.3GB in size... so using {{ is both the portable and safe choice.

layers/core_validation.cpp
```diff
@@ -6045,11 +6046,73 @@ VKAPI_ATTR void VKAPI_CALL CmdBindDescriptorSets(VkCommandBuffer commandBuffer,
...
+    }
+    return skip;
+}
+
+static bool PreCallValidateCmdPushDescriptorSetKHR(layer_data *device_data, GLOBAL_CB_NODE *cb_state,
+                                                   const VkPipelineBindPoint bind_point, const VkPipelineLayout layout,
+                                                   const uint32_t set, const uint32_t descriptor_write_count,
+                                                   const VkWriteDescriptorSet *descriptor_writes, const char *func_name) {
+    bool skip = false;
+    skip |= ValidateCmd(device_data, cb_state, CMD_PUSHDESCRIPTORSET, func_name);
```
jzulauf-lunarg: ValidateCmd needs some way to associate the CMD_ enum, the func_name with the appropriate VUID/VALIDATION_ERROR_xxxxxxxx value for the command re: "must be in recording state.  ValidateCmd does this check, and either can manually add the correct value as a command line arg in each of the 32 (and all subsequent) places, or find a way to create and code-gen'd mapping.

layers/descriptor_sets.cpp
```diff
@@ -1824,12 +1824,16 @@ bool cvdescriptorset::ValidateAllocateDescriptorSets(const core_validation::laye
 
     for (uint32_t i = 0; i < p_alloc_info->descriptorSetCount; i++) {
         auto layout = GetDescriptorSetLayout(dev_data, p_alloc_info->pSetLayouts[i]);
-        if (!layout) {
-            skip |=
```
jzulauf-lunarg: Took out the internal consistency check, as at this point it's impossible to distinguish between a missing map entry and a bogus VkDescriptorSetLayout handle.

layers/vk_validation_error_database.txt
```diff
@@ -360,7 +360,7 @@ VALIDATION_ERROR_04a30401~^~Y~^~Unknown~^~vkCreateDescriptorPool~^~VUID-VkDescri
...
 VALIDATION_ERROR_04c00266~^~Y~^~AllocDescriptorFromEmptyPool~^~vkAllocateDescriptorSets~^~VUID-VkDescriptorSetAllocateInfo-descriptorPool-00307~^~!(VK_KHR_maintenance1)~^~The spec valid usage text states 'descriptorPool must have enough free descriptor capacity remaining to allocate the descriptor sets of the specified layouts' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkDescriptorSetAllocateInfo-descriptorPool-00307)~^~
-VALIDATION_ERROR_04c00268~^~N~^~None~^~VkDescriptorSetAllocateInfo~^~VUID-VkDescriptorSetAllocateInfo-pSetLayouts-00308~^~(VK_KHR_push_descriptor)~^~The spec valid usage text states 'Each element of pSetLayouts must not have been created with VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR set' (https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VUID-VkDescriptorSetAllocateInfo-pSetLayouts-00308)~^~
```
jzulauf-lunarg: Note that each of this entries is changed in the "add VUID check" and "add unit test for VUID check" commits.

scripts/vk_validation_stats.py
```diff
@@ -213,14 +213,14 @@ def parse(self):
                         continue
                     # Find enums
                     #if 'VALIDATION_ERROR_' in line and True not in [ignore in line for ignore in ['[VALIDATION_ERROR_', 'UNIQUE_VALIDATION_ERROR_CODE']]:
-                    if ' VALIDATION_ERROR_' in line:
```
jzulauf-lunarg: when putting the error codes in an intializer-list, clang formats them with a leading and trailing { and } for first and last elements with no space.  This allows for finding those error codes.

scripts/vk_validation_stats.py
```diff
@@ -213,14 +213,14 @@ def parse(self):
...
+                    if 'VALIDATION_ERROR_' in line:
                         # Need to isolate the validation error enum
                         #print("Line has check:%s" % (line))
                         line_list = line.split()
                         enum_list = []
                         for str in line_list:
                             if 'VALIDATION_ERROR_' in str and True not in [ignore_str in str for ignore_str in ['[VALIDATION_ERROR_', 'VALIDATION_ERROR_UNDEFINED', 'UNIQUE_VALIDATION_ERROR_CODE']]:
-                                enum_list.append(str.strip(',);'))
+                                enum_list.append(str.strip(',);{}'))
```
jzulauf-lunarg: can't avoid adjoining {} so, add to strip list.

tests/vkrenderframework.cpp
```diff
@@ -508,10 +508,12 @@ VkDeviceObj::VkDeviceObj(uint32_t id, VkPhysicalDevice obj, std::vector<const ch
     queue_props = phy().queue_properties();
 }
 
-uint32_t VkDeviceObj::QueueFamilyWithoutCapabilities(VkQueueFlags capabilities) {
```
jzulauf-lunarg: Needed the "Match" variant for the new tests, and "Without" is a subset.

tests/layer_validation_tests.cpp
```diff
@@ -21470,6 +21470,120 @@ TEST_F(VkLayerTest, AllocatePushDescriptorSet) {
     vkDestroyDescriptorSetLayout(m_device->handle(), ds_layout, nullptr);
 }
 
+TEST_F(VkLayerTest, PushDescriptorSetCmdPushBadArgs) {
+    TEST_DESCRIPTION("Attempt to push a push descriptor set with incorrect arguments.");
+    if (InstanceExtensionSupported(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME)) {
+        m_instance_extension_names.push_back(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
+    } else {
+        printf("             Did not find VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME; skipped.\n");
```
tobine: Same comment as before, could report this similar to line 21486 below, which is a bit friendlier.

tests/layer_validation_tests.cpp
```diff
@@ -21470,6 +21470,120 @@ TEST_F(VkLayerTest, AllocatePushDescriptorSet) {
...
+    const VkPipelineLayoutObj pipeline_layout(m_device, {&push_ds_layout, &ds_layout});
+    ASSERT_TRUE(pipeline_layout.initialized());
+
+    // Create a descriptor to push
+    const uint32_t buffer_data[4] = {4, 5, 6, 7};
+    VkConstantBufferObj buffer_obj(m_device, sizeof(buffer_data), &buffer_data);
+    ASSERT_TRUE(buffer_obj.initialized());
+
+    // Create a "write" struct, noting that the buffer_info cannot be a temporary arg (the return from write_descriptor_set
+    // references its data), and the DescriptorSet() are can be temporary, because the value is ignored
```
tobine: Extraneous "are" in comment here, I believe

layers/core_validation.cpp
```diff
@@ -6045,11 +6046,73 @@ VKAPI_ATTR void VKAPI_CALL CmdBindDescriptorSets(VkCommandBuffer commandBuffer,
...
-static void PreCallRecordCmdPushDescriptorSetKHR(layer_data *device_data, VkCommandBuffer commandBuffer,
+// Validates that the supplied bind point is supported for the command buffer (vis. the command pool)
+// Takes array of error codes as some of the VUID's (e.g. vkCmdBindPipeline) are written per bindpoint
+// TODO add vkCmdBindPipeline bind_point validation using this call.
+bool ValidatePipelineBindPoint(layer_data *device_data, GLOBAL_CB_NODE *cb_state, VkPipelineBindPoint bind_point,
+                               const char *func_name,
+                               const std::array<UNIQUE_VALIDATION_ERROR_CODE, VK_PIPELINE_BIND_POINT_RANGE_SIZE> &bind_errors) {
+    bool skip = false;
+    auto pool = GetCommandPoolNode(device_data, cb_state->createInfo.commandPool);
+    if (pool) {  // The loss of a pool is a recording cmd is reported in DestroyCommandPool
```
mikes-lunarg: I think an "is" should be "in" in the comment

layers/core_validation.cpp
jzulauf-lunarg: @tobine I decided to do a one-off simple code gen to create a map (and update the enums), then leave and assert in-place for anything new to trip over...  (my next PR)

layers/core_validation.cpp
```diff
@@ -6065,12 +6128,17 @@ VKAPI_ATTR void VKAPI_CALL CmdPushDescriptorSetKHR(VkCommandBuffer commandBuffer
                                                    VkPipelineLayout layout, uint32_t set, uint32_t descriptorWriteCount,
                                                    const VkWriteDescriptorSet *pDescriptorWrites) {
     layer_data *device_data = GetLayerDataPtr(get_dispatch_key(commandBuffer), layer_data_map);
+    auto cb_state = GetCBNode(device_data, commandBuffer);
```
mikes-lunarg: should we get the cb_state while holding the global_lock?

PR #2376, opened 2018-01-29
Fix Issue 2338

### LINKED ISSUE #2338: cube: Calling vkCreateSwapchainKHR with zero width/height is not allowed
With the existing "cube" demos, when they are minimized on Windows the code receives a VK_ERROR_OUT_OF_DATE_KHR error return from vkQueuePresentKHR() or vkAcquireNextImageKHR(). It handles this by calling demo_resize() which tears everything down and recreates everything.

When an application is minimized vkGetPhysicalDeviceSurfaceCapabilitiesKHR() may return a currentExtent for the image with width=0,height=0. This is normal. However, the bug in cube is it then uses these width/height values as input to vkCreateSwapchainKHR().

The VU for vkCreateSwapchainKHR() in the latest Vulkan spec (https://www.khronos.org/registry/vulkan/specs/1.0-wsi_extensions/html/vkspec.html#vkCreateSwapchainKHR) says the following:
>>>
* imageExtent members width and height must both be non-zero
>>>

Cube should be updated to not create an invalid swapchain. Perhaps wait until the user un-minimizes the application.

jeremyk-lunarg:
Fixed crashes and validation errors in cube.c and added a validation check for VkSwapchainCreateInfoKHR.

Check added VALIDATION_ERROR_14600d32

Could not add validation layer test for this error at this time because there is currently limited testing framework for WSI enabled tests.

Fixes #2338.


jeremyk-lunarg:
@mark-lunarg, please review this.

jeremyk-lunarg:
@mark-lunarg Updated

jeremyk-lunarg:
@mark-lunarg, updated.  Looks like it's passing now.

PR #2385, opened 2018-01-31
Add VUID/Valdation Errror coverage for the "must be in the recording state" 

jzulauf-lunarg:
There is a VUID with exactly the same wording for all of the vkCmd* and vkEndCommandBuffer -- that the command buffer must be in the recording state.  This change expands coverage for this for all non-KHX, non-vendor extension vkCmd* and converts the existing centralized reporting to utilized the correct VUID for each specific command


layers/buffer_validation.cpp
```diff
@@ -3951,14 +3954,7 @@ bool PreCallValidateCmdCopyImageToBuffer(layer_data *device_data, VkImageLayout
     bool skip = ValidateBufferImageCopyData(report_data, regionCount, pRegions, src_image_state, "vkCmdCopyImageToBuffer");
 
     // Validate command buffer state
-    if (CB_RECORDING != cb_node->state) {
```
jzulauf-lunarg: This code duplicates the contents for ValidateCmd... so we convert to that (now that ValidateCmd will correctly output the VALIDATION_ERROR...  useful as this allows us to use the existing unit test for this (and the next) VU to function as a spot check for the refactor.

layers/buffer_validation.cpp
```diff
@@ -4027,14 +4023,7 @@ bool PreCallValidateCmdCopyBufferToImage(layer_data *device_data, VkImageLayout
     bool skip = ValidateBufferImageCopyData(report_data, regionCount, pRegions, dst_image_state, "vkCmdCopyBufferToImage");
 
     // Validate command buffer state
-    if (CB_RECORDING != cb_node->state) {
```
jzulauf-lunarg: Same comment as above.

layers/core_validation_types.h
```diff
@@ -371,62 +371,71 @@ struct RENDER_PASS_STATE : public BASE_NODE {
     RENDER_PASS_STATE(VkRenderPassCreateInfo const *pCreateInfo) : createInfo(pCreateInfo) {}
 };
 
-// Cmd Buffer Tracking
+// vkCmd tracking -- complete as of header 1.0.68
+// please keep in "none, then sorted" order
+// Note: grepping vulkan.h for VKAPI_CALL.*vkCmd will return all functions except vkEndCommandBuffer
+
```
jzulauf-lunarg: The updated list was generated by one-time scripts that fished for the "must be in the recording state" (from the validation database) and vkCmd* from vulkan.h -- and diff'd to ensure common result.  Note that the enum names are automated (upper case) based on the vkCmd<basename> and thus there are some changes to existing enum (mostly dropping "STATE").

layers/core_validation_types.h
```diff
@@ -839,8 +848,7 @@ bool rangesIntersect(layer_data const *dev_data, MEMORY_RANGE const *range1, VkD
 bool ValidateBufferMemoryIsValid(layer_data *dev_data, BUFFER_STATE *buffer_state, const char *functionName);
 void SetBufferMemoryValid(layer_data *dev_data, BUFFER_STATE *buffer_state, bool valid);
 bool ValidateCmdSubpassState(const layer_data *dev_data, const GLOBAL_CB_NODE *pCB, const CMD_TYPE cmd_type);
-
-
+bool ValidateCmd(layer_data *dev_data, const GLOBAL_CB_NODE *cb_state, const CMD_TYPE cmd, const char *caller_name);
```
jzulauf-lunarg: buffer_validation needed to see this.

layers/vk_validation_error_database.txt
```diff
@@ -1718,7 +1718,7 @@ VALIDATION_ERROR_1780064c~^~N~^~Unknown~^~vkCmdBeginQuery~^~VUID-vkCmdBeginQuery
...
 VALIDATION_ERROR_17802401~^~Y~^~None~^~vkCmdBeginQuery~^~VUID-vkCmdBeginQuery-commandBuffer-parameter~^~core~^~The spec valid usage text states 'commandBuffer must be a valid VkCommandBuffer handle' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkCmdBeginQuery-commandBuffer-parameter)~^~implicit
-VALIDATION_ERROR_17802413~^~N~^~Unknown~^~vkCmdBeginQuery~^~VUID-vkCmdBeginQuery-commandBuffer-recording~^~core~^~The spec valid usage text states 'commandBuffer must be in the recording state' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkCmdBeginQuery-commandBuffer-recording)~^~implicit
+VALIDATION_ERROR_17802413~^~Y~^~None~^~vkCmdBeginQuery~^~VUID-vkCmdBeginQuery-commandBuffer-recording~^~core~^~The spec valid usage text states 'commandBuffer must be in the recording state' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkCmdBeginQuery-commandBuffer-recording)~^~implicit
```
jzulauf-lunarg: Changed test to None for all of the new "Y" cases.  Using the preexisting two cases in ImageBufferCopyTests (which were converted to use ValidateCmd, as the spot check.  Can add more if reviewers think it necessary.

layers/core_validation.cpp
```diff
@@ -1670,6 +1670,73 @@ static bool ReportInvalidCommandBuffer(layer_data *dev_data, const GLOBAL_CB_NOD
...
+    {CMD_COPYBUFFER, VALIDATION_ERROR_18c02413},
+    {CMD_COPYBUFFERTOIMAGE, VALIDATION_ERROR_18e02413},
+    {CMD_COPYIMAGE, VALIDATION_ERROR_19002413},
+    {CMD_COPYIMAGETOBUFFER, VALIDATION_ERROR_19202413},
+    {CMD_COPYQUERYPOOLRESULTS, VALIDATION_ERROR_19402413},
+    {CMD_DEBUGMARKERBEGINEXT, VALIDATION_ERROR_19602413},
+    {CMD_DEBUGMARKERENDEXT, VALIDATION_ERROR_19802413},
+    {CMD_DEBUGMARKERINSERTEXT, VALIDATION_ERROR_19a02413},
+    {CMD_DISPATCH, VALIDATION_ERROR_19c02413},
+    // Exclude KHX (if not already present) { CMD_DISPATCHBASEKHX, VALIDATION_ERROR_19e02413 },
```
jzulauf-lunarg: This defensive commenting is to ensue the documentation validator doesn't think that the VALIDATION_ERRORS are present when they aren't (and thus spitting out errors about the corresponding "~N~" entries

layers/core_validation.cpp
```diff
@@ -11539,219 +11625,276 @@ VKAPI_ATTR VkResult VKAPI_CALL DebugMarkerSetObjectTagEXT(VkDevice device, VkDeb
...
-    {"vkFreeMemory", (void*)FreeMemory},
-    {"vkBindBufferMemory", (void*)BindBufferMemory},
-    {"vkGetBufferMemoryRequirements", (void*)GetBufferMemoryRequirements},
-    {"vkGetImageMemoryRequirements", (void*)GetImageMemoryRequirements},
-    {"vkGetQueryPoolResults", (void*)GetQueryPoolResults},
-    {"vkBindImageMemory", (void*)BindImageMemory},
-    {"vkQueueBindSparse", (void*)QueueBindSparse},
-    {"vkCreateSemaphore", (void*)CreateSemaphore},
-    {"vkCreateEvent", (void*)CreateEvent},
+    {"vkGetInstanceProcAddr", (void *)GetInstanceProcAddr},
```
jzulauf-lunarg: Clang format somehow decided all these lines were touch enough to add the " " before the "*"

mikes-lunarg: Same for #2384, last one to master will have a little merge conflict to deal with

layers/vk_validation_error_database.txt
tobine: Seems reasonable to me. ROI on adding this tests at this point not high.

PR #2390, opened 2018-02-01
Move LVL temporary & worker MSVC cmake targets into solution explorer subfolder

mark-lunarg:
This sets up cmake to locate the temporary or worker MSVC projects into a subfolder called lvl_cmake_targets. This simply unclutters tte Visual Studio Solution Explorer projects pane, and does not affect functionality.

For example, all of the solution explorer projects such as *-def-file, *-json, and the file generator targets will show up in the lvl_cmake_targets folder.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,10 @@ if (USE_CCACHE)
     endif(CCACHE_FOUND)
 endif()
 
+# Enable cmake folders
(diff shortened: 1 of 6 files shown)


tobine:
@mark-lunarg looks like build errors on screenshot and api-dump layers. Ping for re-review when CI is green.

PR #2391, opened 2018-02-01
Add LAYER_EXPORT specifiers to vk_layer_config

mark-lunarg:
Required for hassle-free linking to layer_utils lib.

### DIFF
--- a/layers/vk_layer_config.cpp
+++ b/layers/vk_layer_config.cpp
@@ -69,11 +69,11 @@ std::string getEnvironment(const char *variable) {
 #endif
 }
 
-const char *getLayerOption(const char *_option) { return g_configFileObj.getOption(_option); }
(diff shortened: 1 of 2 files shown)


cnorthrop:
The Android build is fine with these changes, not sure what's going on with building downstream VulkanTools for CI testing.

karl-lunarg:
This was needed because before VulkanTools changed to use LVL as a submodule, there was a layer_utilsvt library in VT that the VT layers linked against.  The layer_utilsvt library was accidentally built to NOT set symbol visibility to hidden, so the VT layers linking to layer_utilsvt were able to find all the symbols.

After we moved LVL to a submodule in VT we removed layer_utilsvt entirely and now the VT layers use layer_utils from LVL.   The layer_utils lib from LVL was linked properly to hide symbols, but some symbols were not explicitly exported.  This MR fixes that.

PR #2392, opened 2018-02-01
vkCreateImage valid usage checks / tests

daveh-lunarg:
Adds 18 valid usage checks to vkCreateImage(), with about 2/3 of those in core validation and the rest in parameter validation.  Includes a pretty much full-body refactoring of PreCallValidateCreateImage() in buffer_validation.cpp.

Fixed some existing tests that tripped the new VUs, and modified them to use VUID-based error matching instead of string matching.

Added a new test to exercise all the new VUs that weren't otherwise being tested.


layers/buffer_validation.cpp
```diff
@@ -634,141 +634,197 @@ bool PreCallValidateCreateImage(layer_data *device_data, const VkImageCreateInfo
...
+                          (uint64_t)pCreateInfo->extent.depth * (uint64_t)pCreateInfo->arrayLayers *
+                          (uint64_t)pCreateInfo->samples * (uint64_t)FormatSize(pCreateInfo->format);
+
+    // Round up to imageGranularity boundary
+    VkDeviceSize imageGranularity = GetPhysicalDeviceProperties(device_data)->limits.bufferImageGranularity;
+    uint64_t ig_mask = imageGranularity - 1;
+    total_size = (total_size + ig_mask) & ~ig_mask;
 
-    if (totalSize > ImageFormatProperties.maxResourceSize) {
+    if (total_size > format_limits.maxResourceSize) {
```
tobine: No VUID for this case. Does it make sense to file spec update to create one?

daveh-lunarg: Yeah, probably. I submitted as spec issue 1155. Good catch!

daveh-lunarg:
Appveyor is stuck, but PR is passing on Travis and local CI.  Pushing.

Issue #2346, opened 2018-01-16
external_revision_generator.py should not depend on Git


cdwfs:
The `external_revision_generator.py` script is used to generate a unique ID for a particular state of the validation code, to determine whether a blob of serialized validation cache data is compatible with the version of the validation code currently in use by an application. The script currently uses the repository's Git commit ID as the unique ID, but this creates an implicit build requirement that the layers code is being hosted in a Git repository. A different, more portable method for UUID generation should be used instead.

cnorthrop:
Hey Cort,

Pinging @chrisforbes too.

Any update on this?  Could we just use a random number instead of something tied to git?  This is affecting the Android NDK downstream.

-C

cdwfs:
No concrete progress yet; just ideas.

- The ideal ID would correspond to a particular state of the spirv-tools shader validation code.

- The next best thing would be a build number from the spirv-tools library; I see a build-version.inc in the spirv-tools build directory, but its contents also seems to be derived from the Git commit ID.

- Worst-case, any old random number for a given build of the core_validation layer would _probably_ be an acceptable ID; it would limit the ability to share validation cache data between different builds of the exact same layers code, but that's unlikely to a significant problem in practice.

- Another approach would be to add an optional flag to `external_revision_generator.py` that lets users override the default Git-derived ID with any arbitrary value of their choosing (ideally, something semantically equivalent to a spirv-tools commit ID or build number). That would be the most flexible solution, but it's really just passing the problem on to developers who build layers in non-Git environments.

cnorthrop:
A couple more ideas:

- Use a timestamp.  This is akin to the random number, but trackable.

- It looks like `build-version.inc` is created by `update_build_version.py`, which can use the `git rev-parse`, but doesn't require it.  So you could get a string out to identify the approximate SPIR-V version, which would hopefully be compatible between hashes until `CHANGES` were updated? i.e.:
  - `"v2017.1-dev", "SPIRV-Tools v2017.1-dev unknown hash, 2018-02-01"`
  vs
  - `"v2017.1-dev", "SPIRV-Tools v2017.1-dev spirv-1.1-rev1-463-g7e2d26c"`

cdwfs:
If the spirv-tools build-version.inc works with or without Git, that's suddenly a much more compelling solution! I wish the timestamp it included in that case were a bit higher-resolution; would it be reasonable to add a HH:MM:SS suffix?

PR #2393, opened 2018-02-02
Fix issue #2346

cdwfs:
Chris/Cody: I haven't run these through the build farm yet (will try that tomorrow), but please take a look anyway to ensure you're okay with the general direction.

One potential improvement that occurs to me as I type this just now: the validation cache code ultimately wants a UUID for its cache ID. `external_revision_generator.py` currently takes either a SHA1 Git commit ID or the contents of an arbitrary revision file (converted to a SHA1 hash if necessary), and writes the hash to the header file (which is converted back to a UUID at runtime by the layers). Perhaps the script should just convert its input to a UUID immediately, and write _that_ to the header file (so validation cache runtime code can ingest it directly). What do you think -- useful simplification, or overthinking?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -318,7 +318,10 @@ endmacro()
 # Define macro used for generating header files containing commit IDs for external dependencies
 macro(run_external_revision_generate source_dir symbol_name output)
     add_custom_command(OUTPUT ${output}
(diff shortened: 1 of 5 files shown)


cnorthrop:
Thanks, Cort.  I like the change, but it will be Monday before I get a chance to test it.

cnorthrop:
Okay, I added a commit that allows the android-generate.* scripts to continue working as is, allowing downstream toolchains to work without modification.  I'll merge this to master soon.

cnorthrop:
cherry-picked to master

PR #2394, opened 2018-02-02
layers: fix unexpected errors in layer tests

jeremyk-lunarg:
Fixes various layer validation tests that were producing unexpected error messages.


krOoze:
There's some work duplication with #2222...

tests/layer_validation_tests.cpp
```diff
@@ -7244,6 +7244,10 @@ TEST_F(VkLayerTest, RenderPassInUseDestroyedSignaled) {
 
     // Wait for queue to complete so we can safely destroy rp
     vkQueueWaitIdle(m_device->m_queue);
+    // Because ObjectTracker is ahead of core_validation in the layer chain this proper delete is considered by the validation
+    // layers to be a double delete.
```
jzulauf-lunarg: This happens because core_validation (as configured for the layer_validation_tests) "skips" calling the lower layers upon finding the error condition above, but ObjectTracker doesn't (can't) know that it did.  ObjectTracker succeeds (as thus deletes it's record of the handle) while the object still lives in the driver.  

Given the "skip" configuration of the ObjectTracker, the lower layers aren't called for any subsequent calls to vkDestroyRenderPass with the same handle, making the handle impossible to actually delete.

krOoze: Why is the `rp` being destroyed here in the first place?

mikes-lunarg: @jzulauf-lunarg Adding the expected object tracker error will caused debug report callback to return false, ie not skip the call.  The second destroy makes it past object tracker.

@krOoze  It's good practice to vkDestroy anything we vkCreate within a test.  In this case vkDestroyRenderPass is called twice because core_validation stops the first one from getting to the driver.  In reality it probably doesn't make a difference, you would hope that the driver cleans up when the device gets destroyed.

krOoze: @mikes-lunarg Certainly, but only for objects that are not direct subject of the test, and are therefore in undefined state.

jeremyk-lunarg:
@krOoze @jzulauf-lunarg  updated

tests/layer_validation_tests.cpp
chrisforbes: @krOoze this isn't just tidiness -- the spec /requires/ apps to clean up all their objects before destroying the device. "You would hope" the driver can cope with you not doing that, but at that point you're in undefined behavior.

krOoze: @chrisforbes That's nice, but as @jzulauf-lunarg pointed out, the handle is indeleteble.

chrisforbes: @jzulauf-lunarg any thoughts on dealing with this layer composition problem? We'd previously discussed either merging all the layers together (not a real solution if anyone else provides a layer), or adding a side channel for results of skippable functions that don't return a result.

tests/layer_validation_tests.cpp
```diff
@@ -7244,6 +7244,8 @@ TEST_F(VkLayerTest, RenderPassInUseDestroyedSignaled) {
 
     // Wait for queue to complete so we can safely destroy rp
     vkQueueWaitIdle(m_device->m_queue);
+    m_errorMonitor->SetUnexpectedError("If renderPass is not VK_NULL_HANDLE, renderPass must be a valid VkRenderPass handle");
```
jzulauf-lunarg: Matches MkeS best practices comment.

PR #2401, opened 2018-02-07
Clang-format level-set

daveh-lunarg:
We now have enforcement of clang-format on GitHub, so once all source files are compliant they should remain so.  This is a whole-file clang-format of all .h and .cpp files in the test and layers directories.

I also took the opportunity to re-flow string literals that had been fragmented by previous line width settings of clang-format. These are the bulk of the diffs, overall.

It's intended, and AFAICT is true that this PR is entirely whitespace and comment changes only.

Assigning to a large group as a heads-up - this PR touches a lot of code.


daveh-lunarg:
Travis CI is currently verifying clang-format using version 5.0.0. I used the latest version (5.0.1) to format these changes. This PR generated some local discussion about whether there was a preferred c-f version that was < 5.0.x, but without a definitive answer. Accepting this PR will mean accepting c-f 5.0+ as the preferred version.

CONTRIBUTING.md discussed the c-f requirement without specifying a version, so I've added a paragraph to recommend version 5.0+.

layers/core_validation.cpp
```diff
@@ -9142,26 +9089,25 @@ VKAPI_ATTR void VKAPI_CALL CmdExecuteCommands(VkCommandBuffer commandBuffer, uin
...
-                            ") executed within render pass (0x%" PRIx64
-                            ") must have had vkBeginCommandBuffer() called w/ VK_COMMAND_BUFFER_USAGE_RENDER_PASS_CONTINUE_BIT "
-                            "set. %s",
-                            HandleToUint64(pCommandBuffers[i]), HandleToUint64(pCB->activeRenderPass->renderPass),
-                            validation_error_map[VALIDATION_ERROR_1b2000c0]);
+                        skip |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                                        VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT, HandleToUint64(pCommandBuffers[i]),
+                                        __LINE__, VALIDATION_ERROR_1b2000c0, "DS",
+                                        "vkCmdExecuteCommands(): Secondary Command Buffer (0x%" PRIx64
+                                        ") executed within render pass (0x%" PRIx64
```
krOoze: Too bad, it still does not handle the macros well...

layers/core_validation.cpp
```diff
@@ -10044,15 +9982,15 @@ static bool PreCallValidateCreateSwapchainKHR(layer_data *dev_data, const char *
             // TODO: should restrict search only to queue families of VkDeviceQueueCreateInfos, not whole phys. device
             return (qs.first.gpu == dev_data->physical_device) && qs.second;
         };
-        const auto& support = surface_state->gpu_queue_support;
+        const auto &support = surface_state->gpu_queue_support;
```
krOoze: Any chance to switch to type-stickiness for pointers and references?
The google style allows it. And the only reason for name-stickyness is that you can write `int a, &b, *c;`, which is strictly forbidden by the style anyway.

krOoze:
`clang-format` defaults to 3.8 for me. But 5.0 is available so I guess it is OK... but maybe worth a thought.

layers/core_validation_types.h
```diff
@@ -267,7 +267,7 @@ class IMAGE_STATE : public BINDABLE {
 
     ~IMAGE_STATE() {
         if ((createInfo.sharingMode == VK_SHARING_MODE_CONCURRENT) && (createInfo.queueFamilyIndexCount > 0)) {
-            delete [] createInfo.pQueueFamilyIndices;
+            delete[] createInfo.pQueueFamilyIndices;
```
krOoze: huh. didn't even know `delete_[]` was valid syntax...

layers/object_tracker_utils.cpp
```diff
@@ -210,31 +210,29 @@ bool ValidateDescriptorSet(VkDevice device, VkDescriptorPool descriptor_pool, Vk
...
-                                   false, VALIDATION_ERROR_15c00286, VALIDATION_ERROR_15c00009);
+            skip |= ValidateObject(disp, desc->pTexelBufferView[idx2], kVulkanObjectTypeBufferView, false,
+                                   VALIDATION_ERROR_15c00286, VALIDATION_ERROR_15c00009);
         }
     }
 
     if ((desc->descriptorType == VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER) ||
-        (desc->descriptorType == VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE) ||
-        (desc->descriptorType == VK_DESCRIPTOR_TYPE_STORAGE_IMAGE) ||
+        (desc->descriptorType == VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE) || (desc->descriptorType == VK_DESCRIPTOR_TYPE_STORAGE_IMAGE) ||
```
krOoze: That's kinda worse... oh, the tyranny of a dumb styler script.

layers/parameter_validation.h
```diff
@@ -887,15 +884,15 @@ static std::string get_result_description(VkResult result) {
...
-*/
+ * Validate return code.
+ *
+ * Print a message describing the reason for failure when an error code is returned.
+ *
+ * @param report_data debug_report_data object for routing validation messages.
+ * @param apiName Name of API call being validated.
+ * @param ignore vector of VkResult return codes to be ignored
+ * @param value VkResult value to validate.
+ */
```
krOoze: While turning stones: These comments seem like for some documentation software (are we using one?). Maybe should be deleted (eventually)?

layers/shader_validation.h
```diff
@@ -120,37 +120,32 @@ class ValidationCache {
...
+        if (data[0] != size) return;
+        if (data[1] != VK_VALIDATION_CACHE_HEADER_VERSION_ONE_EXT) return;
         uint8_t expected_uuid[VK_UUID_SIZE];
         Sha1ToVkUuid(SPIRV_TOOLS_COMMIT_ID, expected_uuid);
-        if (memcmp(&data[2], expected_uuid, VK_UUID_SIZE) != 0)
-            return;  // different version
+        if (memcmp(&data[2], expected_uuid, VK_UUID_SIZE) != 0) return;  // different version
 
-        data = (uint32_t const*)(reinterpret_cast<uint8_t const *>(data) + headerSize);
+        data = (uint32_t const *)(reinterpret_cast<uint8_t const *>(data) + headerSize);
```
krOoze: The style recommends putting `const` first. Apparently not enforced...

tests/layer_validation_tests.cpp
```diff
@@ -5923,9 +5891,9 @@ TEST_F(VkLayerTest, InvalidDescriptorSetLayout) {
 TEST_F(VkLayerTest, WriteDescriptorSetIntegrityCheck) {
     TEST_DESCRIPTION(
         "This test verifies some requirements of chapter 13.2.3 of the Vulkan Spec "
-        "1) A uniform buffer update must have a valid buffer index."
-        "2) When using an array of descriptors in a single WriteDescriptor,"
-        "     the descriptor types and stageflags must all be the same."
+        "1) A uniform buffer update must have a valid buffer index. "
+        "2) When using an array of descriptors in a single WriteDescriptor, the descriptor types and stageflags "
+        "must all be the same. "
```
krOoze: Why the new spaces?

tests/layer_validation_tests.cpp
```diff
@@ -11926,6 +11847,7 @@ TEST_F(VkLayerTest, ImageBarrierWithBadRange) {
 
     m_commandBuffer->begin();
 
+    // clang-format off
```
krOoze: Why?

I think I seen exceptions elsewhere too. Should they be re-evaluated?

tests/layer_validation_tests.cpp
```diff
@@ -13291,9 +13207,9 @@ TEST_F(VkLayerTest, Maint1BindingSliceOf3DImage) {
     // Meat of the test.
     m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_046002ae);
 
-    VkDescriptorImageInfo dii = { VK_NULL_HANDLE, view, VK_IMAGE_LAYOUT_GENERAL };
-    VkWriteDescriptorSet write = { VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET, nullptr,
-        set.set_, 0, 0, 1, VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE, &dii, nullptr, nullptr };
+    VkDescriptorImageInfo dii = {VK_NULL_HANDLE, view, VK_IMAGE_LAYOUT_GENERAL};
+    VkWriteDescriptorSet write = {VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET, nullptr, set.set_, 0,      0, 1,
+                                  VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE,       &dii,    nullptr,  nullptr};
```
krOoze: Shows to me:
obrazek

What's going on with the spaces??

tests/layer_validation_tests.cpp
```diff
@@ -15663,8 +15552,7 @@ TEST_F(VkPositiveLayerTest, CreatePipelineComplexTypes) {
         "void main() {}";
     char const *tcsSource =
         "#version 450\n"
-        "layout(vertices=3) out;"
-        "struct S { int x; };\n"
+        "layout(vertices=3) out;struct S { int x; };\n"
```
krOoze: Probably missing `\n`. This string should not be merged like that.

tests/layer_validation_tests.cpp
```diff
@@ -16021,8 +15912,7 @@ TEST_F(VkLayerTest, CreateShaderModuleCheckBadCapability) {
     char const *vsSource =
         "#version 450\n"
         "\n"
-        "layout(xfb_buffer = 1) out;"
-        "void main(){\n"
+        "layout(xfb_buffer = 1) out;void main(){\n"
```
krOoze: ditto

layers/core_validation.cpp
daveh-lunarg: This PR is just about getting the existing codebase aligned to the existing .clang-format (which is Google plus line-width and include-sorting overrides).  Tweaking the .clang-format file is a separate issue. (But FYI, I agree with your preference here).

layers/shader_validation.h
daveh-lunarg: I don't think const placement is within clang-format's ability.

tests/layer_validation_tests.cpp
daveh-lunarg: It's a single string after concatenation, and the spaces make sense in that case.  (Well, except the last one).

layers/core_validation.cpp
krOoze: Yea, On the other hand it would be nice to do it together (while this already will force people to rebase their work)...

tests/layer_validation_tests.cpp
daveh-lunarg: In this case c-f just couldn't make sense of the nested scopes, and just completely messed up the alignment.  After a couple failed attempts to coerce it, I just decided to exclude that chunk.

tests/layer_validation_tests.cpp
daveh-lunarg: Yeah, that is a bit strange. Seems to be aligning elements vertically. It's definitely intentional - when I take them out c-f puts them back.

tests/layer_validation_tests.cpp
krOoze: Actually, it probably prevents the c-f from merging the literals for some weird reasons...
Might as well go for `\n` (it obviously is a numbered list)?

layers/shader_validation.h
krOoze: Yea, apparently needs to be done manually. At least I did not see it in the options scanning quickly over them....

tests/layer_validation_tests.cpp
daveh-lunarg: Agreed, but it's a harmless artifact in the original code.  Don't want to correct it in this PR.

daveh-lunarg: Caused by the string-merge prior to clang-format. I'll correct it.

tests/layer_validation_tests.cpp
krOoze: Odd is it does not render them when copied...
Maybe make sure there are no Unicode characters there.

daveh-lunarg:
Thanks @krOoze for the thorough review!

daveh-lunarg:
With no further discussion I'm going to assume [tacit approval](https://en.wikipedia.org/wiki/Qui_tacet_consentire_videtur#qui_tacet_consentire_videtur) and push this in the next day or two.

CONTRIBUTING.md
```diff
@@ -57,6 +57,8 @@ a good reason is "This violates the style guide, but it improves type safety."
 * Run **clang-format** on your changes to maintain consistent formatting
     * There are `.clang-format files` present in the repository to define clang-format settings
       which are found and used automatically by clang-format.
+	* **clang-format** binaries are available from the LLVM orginization, here: https://clang.llvm.org/. The GitHub CI system currently verifies
```
karl-lunarg: Technically, GitHub doesn't have CI.  It is probably better to say:

> Our CI system (Travis-CI) currently uses clang-format version 5.0.0 to check that the lines of code you have changed are formatted properly.  It is recommended that you use the same version to format your code prior to submission.

karl-lunarg:
>With no further discussion I'm going to assume tacit approval and push this in the next day or two.

Yes, let's set this time for the afternoon of the 14th. (MST)

CONTRIBUTING.md
daveh-lunarg: Ok.

daveh-lunarg:
Per today's discussion, I'll rebase Thu near end of day and plan to push this on Friday 16th AM.

PR #2405, opened 2018-02-13
loader: Don't add duplicate JSON files from Khronos registry path.

NorbertGarnysAMD:
To make sure we avoid enumerating old JSON files/drivers that might be present
in the non-device specific area of the registry when a newer device specific
JSON file is present, do a check before adding.
Find the file name, without path, of the JSON file found in the non-device
specific registry location. If the same JSON file name is already found in
the list, don't add it again.


loader/loader.c
```diff
@@ -408,6 +408,57 @@ VKAPI_ATTR VkResult VKAPI_CALL vkSetDeviceDispatch(VkDevice device, void *object
 }
 
 #if defined(_WIN32)
+
+// Append the JSON path data to the list and allocate/grow the list if it's not large enough.
+// Function returns true if filename was appended to reg_data list.
+// Caller should free reg_data.
+static bool loaderAddJsonEntry(const struct loader_instance *inst,
```
lenny-lunarg: You split some of the code into this function, but then only used it in one place. It looks to me like you could add the check to avoid duplicating filenames in here, then call it from `loaderGetRegistryFiles` where you added the check.

loader/loader.c
```diff
@@ -732,11 +746,37 @@ VkResult loaderGetRegistryFiles(const struct loader_instance *inst, char *locati
                         inst, VK_DEBUG_REPORT_INFORMATION_BIT_EXT, 0, "Located json file \"%s\" from registry \"%s\\%s\"", name,
                         hive == DEFAULT_VK_REGISTRY_HIVE ? DEFAULT_VK_REGISTRY_HIVE_STR : SECONDARY_VK_REGISTRY_HIVE_STR, location);
                     if (strlen(*reg_data) == 0) {
+                        // The list is emtpy. Add the first entry.
                         (void)snprintf(*reg_data, name_size + 1, "%s", name);
+                        found = true;
                     } else {
-                        (void)snprintf(*reg_data + strlen(*reg_data), name_size + 2, "%c%s", PATH_SEPARATOR, name);
+                        // At this point the reg_data variable contains other JSON paths, likely from the PNP/device section
```
lenny-lunarg: Here you do the check to avoid repeating filenames, but it seems like it would be better to add the check to the `loaderAddJsonEntry` function that you added and then call that from here. That way we could reuse the same code to check for repeated files and add them to the list otherwise.

loader/loader.c
ChuckM2: I work with Norbert and worked on this change.  (I am very new to GitHub, so bear with me.)  I can comment on some of the questions/feedback.
-It's true that the new function, loaderAddJsonEntry(), is currently only used once.  I have additional changes that also use this function, but I am not allowed to share them yet.  Hopefully I can share them soon.  I separated the code into the function now to avoid any future merge conflicts (since I already had one recently). 
-There is a good reason that we only check for duplicates in the Khronos registry location and not in the PnP registry location(s).   In the previous Windows driver install "model", drivers were installed to Windows\System32 (and SysWow64) and registry keys were written to the HKLM Khronos location.  Only one vendor driver with file name Foo and JSON file with file name Bar could be present in System32 at a time.  In new Windows 10 versions, there is a new driver install "model", both because of Windows Driver Store use and Universal Windows Drivers.  In this new model, it is valid for one vendor to have two or more drivers installed on the system at a time.  Each of the N active adapters will have a VulkanDriverName registry key in the PnP location for that device and a different driver Foo and JSON file Bar in the driver store folder for that driver.  So it's possible and valid for there to be multiple JSON files named Bar, but in different driver store folders.  One unfortunately restriction of Universal Windows Driver installs is that it cannot delete old HKLM Khoronos registry keys or old driver files in Windows\System32.  Therefore, when we detect a new PnP registry location in use, we should avoid any JSON files with identical names in the legacy Khronos registry key location.  
Let's look at an extreme example.  Let's say we detect the following registry keys:  A. PNP VulkanDriverName key from adapter A pointing to C:\some_driver_store_path\XX\Bar.JSON.   B. PNP VulkanDriverName key from adapter B pointing to C:\some_driver_store_path\YY\Bar.JSON.  C. HKLM Khronos key C:\Windows\System32\Bar.JSON.  In this case A and B are valid, but C should be ignored.  This new change will cause only C to be ignored because 1. it's in the Khronos location and 2. Bar.JSON is already in the list.
Let's look at another extreme example.  Let's say we detect the following registry keys:  A. PNP VulkanDriverName key from adapter A pointing to C:\some_driver_store_path\XX\Bar.JSON.   B. PNP VulkanDriverName key from adapter B pointing to C:\some_driver_store_path\XX\Bar.JSON.  In this case A and B are exact duplicates because the exact path matches.  There already exists code to remove exact duplicates in loader_get_manifest_files().   Search for "// Check if the file is already present".  This code will handle skipping JSON files that have already been loaded once.
-I don't know that it occurred to me to use the loaderAddJsonEntry() in the loaderGetRegistryFiles() function (which is used for the Khronos registry location).  If this is important, than I can look into doing this.  Currently the loaderAddJsonEntry() is slightly PnP registry location specific since the debug print outputs "%s: Located json file \"%s\" from PnP registry: %".

loader/loader.c
ChuckM2: See previous comment.

loader/loader.c
lenny-lunarg: Thanks for the reply. I see why you don't remove redundant entries from the PnP registries now. My only concern is that when you have one driver in the Khronos registries, it's currently possible for that one driver to pick up 2 devices. If the drivers are the same in the PnP registries, then picking up two drivers would mean that each driver would pick up both devices, and a Vulkan app would see 4 devices, when there's really only 2 on the system. To avoid this, drivers would need to make sure that they only load the device they're associated with in the PnP registry. In any case, that's a concern that this PR neither introduces nor aggravates, so it's not worth holding up this PR over that.

My other comments are also minor and can always be changed later if there's a need.

ChuckM2: I admit that I have done very limited Vulkan testing with multiple physical devices/adapters.  This is how I think it's supposed to work.  Say the system has 2 physical devices, each with a PnP registry location, and both are using the same driver.  So the PnP registry search will find two VulkanDriverName keys (one for each device) that point to the same full JSON path (to the driver store).  The code in loader_get_manifest_files() should load the JSON file the first time it's encountered and skip it the second time, because it's already loaded (with the same full path+name).  Therefore the app should see one driver and 2 devices (just like it used to work when using the Khronos registry key and drivers in System32).

PR #2409, opened 2018-02-13
demos: ensure DISPLAY envar is valid

mikew-lunarg:
If the DISPLAY environment variable is unset or null, the cube demo will
emit "Cannot find a compatible Vulkan installable client driver (ICD)."
That message is not helpful when attempting to resolve the issue, since
the problem is with the envar, not the ICD.

Change-Id: Idf145cf0da906a4a8352b5bb136f88d9c1bca90d


demos/cube.cpp
```diff
@@ -2438,6 +2445,13 @@ Demo::Demo()
 #elif defined(VK_USE_PLATFORM_XLIB_KHR)
 
     void Demo::create_xlib_window() {
+        const char *display = getenv("DISPLAY");
```
karl-lunarg: If you compile this for VK_USE_PLATFORM_XLIB_KHR, this won't work (or compile) because display is a member variable in the Demo structure.  Should probably pick a different name for this local `display` variable.

The Travis-CI build passes because it probably only builds the XCB variant, in which case, there is no display member in Demo.

One of our local CI machines failed because it builds the XLIB flavor.

mikew-lunarg:
variables have been renamed to avoid collision.  Thanks Karl.

Issue #2213, opened 2017-11-16
VL reports too early when recording secondary command buffer


mark-lunarg:
This issue was submitted by Dean Sekulic on [LunarXchange, #654](https://vulkan.lunarg.com/issue/view/58b4083666315130fe3c2d78):

I got "Vulkan internal report (code 1115): vkCmdClearAttachments(): The area defined by pRects[0] is not contained in the area of the current render pass instance."

but I'm only recording secondary command buffer, primary command buffer didn't yet got vkCmdBeginRenderPass! Like in this case: vkBeginCommandBuffer( vkSecondaryCmdBuffer, &vkCmdBufferBeginInfo);

`VkClearAttachment vkClear; vkClear.colorAttachment = 0; vkClear.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT; vkClear.clearValue.color.float32[0] = 0.0; vkClear.clearValue.color.float32[1] = 0.0; vkClear.clearValue.color.float32[2] = 0.0; vkClear.clearValue.color.float32[3] = 0.0; const VkClearRect vkRect = { { { 0,0 },{ 128,128 }, }, 0,1 }; pvkCmdClearAttachments( vkCmdBuffer, 1, &vkClear, 1, &vkRect);`

And right here VL reports! It's true that at that time primary command buffer didn't record begin render pass yet, but I think this report should be moved to execution of secondary command buffer.

This basically will involve completing this TODO in PreCallValidateCmdClearAttachments:

> // TODO: This check should be moved to CmdExecuteCommands or QueueSubmit to cover secondary CB cases

jeremyk-lunarg:
Issue should to be fixed by PR #2411

PR #2411, opened 2018-02-14
tests: Add test to verify issue 2213 fixed

jeremyk-lunarg:
Adds a VkPositiveLayerTest to check that vkCmdClearAttachment does not cause a validation error when used in a secondary command buffer.  The validation layers should wait until the secondary command buffer is run by a primary command buffer before checking if vkCmdClearAttachment is called within a render pass.

Verifies that issue #2213 is fixed.


mark-lunarg:
This LGTM if CI passes.  Be nice if the commit message was improved.

jeremyk-lunarg:
CI is passing.  Will merge this tomorrow morning if there are no other comments.

PR #2412, opened 2018-02-15
Handle multiple GetSwapchainImageKHR calls

### LINKED ISSUE #2410: vkGetSwapchainImagesKHR called more than once, returns another handle (ID)s
When enabled validation layer and 'vkGetSwapchainImagesKHR' is called twice, 
I think that invalid handle (ID) returned.

I tried to vkGetSwapchainImagesKHR without validation layer, so value was same.
I expect the values to be the same when enabled validation layer.

example:
```
int count = 0; 
vkGetSwapchainImagesKHR(device, swapchain, &count, nullptr); // (e.g.) return count = 3.

VkImage images[3];
vkGetSwapchainImagesKHR(device, swapchain, &count, images); // (e.g.) return ID = 4,5,6

// again
vkGetSwapchainImagesKHR(device, swapchain, &count, images); // (e.g.) return ID = 7,8,9 
```

thanks

mark-lunarg:
Unique Objects was fixed to prevent uniquifying duplicate image objects for duplicate GetSwapchainImages calls.

Fixes #2410.


demos/cube.c
```diff
@@ -1302,6 +1302,11 @@ static void demo_prepare_buffers(struct demo *demo) {
     err = demo->fpGetSwapchainImagesKHR(demo->device, demo->swapchain,
                                         &demo->swapchainImageCount,
                                         swapchainImages);
+
```
tobine: Is this intended to be landed or is it accidentally included?
If you intend to land it then should include a comment here to indicate that this is not recommended style, but is being done here as a means to exercise validation layers

mark-lunarg: Nope, part of my local testing.  I'll remove it before pushing.

layers/unique_objects.cpp
```diff
@@ -502,7 +502,14 @@ VKAPI_ATTR VkResult VKAPI_CALL GetSwapchainImagesKHR(VkDevice device, VkSwapchai
         if ((*pSwapchainImageCount > 0) && pSwapchainImages) {
             std::lock_guard<std::mutex> lock(global_lock);
             for (uint32_t i = 0; i < *pSwapchainImageCount; ++i) {
-                pSwapchainImages[i] = WrapNew(my_device_data, pSwapchainImages[i]);
+                uint64_t unwrapped_handle = HandleToUint64(pSwapchainImages[i]);
+                if (my_device_data->swapchain_image_handle_map.find(unwrapped_handle) ==
+                    my_device_data->swapchain_image_handle_map.end()) {
+                    pSwapchainImages[i] = WrapNew(my_device_data, pSwapchainImages[i]);
+                    my_device_data->swapchain_image_handle_map[unwrapped_handle] = HandleToUint64(pSwapchainImages[i]);
```
krOoze: Is this technically correct?
`VkImage` is a non-dispatchable (i.e. non-unique) handle, right? Theoretically the driver could return several identical handles in `pSwapchainImages` (or the same handles for different swapchain), so this approach would not be right.

The map should perhaps associante to the (wrapped, unique) swapchain handle, and not the `VkImage` returned.

layers/unique_objects.h
```diff
@@ -79,6 +79,9 @@ struct layer_data {
     // uses unwrapped handles
     std::unordered_map<VkRenderPass, SubpassesUsageStates> renderpasses_states;
 
+    // Map of swapchain image handles to wrapped IDs -- we don't want to re-wrap these if they exist
+    std::unordered_map<uint64_t, uint64_t> swapchain_image_handle_map;
```
krOoze: Seems premature to convert to `uint64_t` here. Might be more readable to have a specialized `std::hash`. Or these potentially small maps should perhaps be only `std::map` (I would assume comparable performance and less memory usage for small N).

layers/unique_objects.cpp
mark-lunarg: If a driver used non-unique object handles in a swapchain, things would go probably go sideways regardless.  That said, having a per-swapchain map instead of a per-device map leaves less wiggle room here, and would cover destroySwapchain well.  Thanks for the suggestion.

krOoze: Now that you brought that up, how does it handle `destroySwapchain`. If I read it correctly it just leaks the handles and `swapchain_image_handle_map` would just grow forewer with each new swapchain.

PR #2414, opened 2018-02-16
Handle multiple GetSwapchainImageKHR calls

### LINKED ISSUE #2410: vkGetSwapchainImagesKHR called more than once, returns another handle (ID)s

mark-lunarg:
Unique Objects was fixed to prevent uniquifying duplicate image objects for multiple GetSwapchainImageKHR calls.  As the GetXxx results are immutable here, saved off the wrapped handles for each swapchain and return them for subsequent calls.  Data is destroyed when swapchain destroyed.

Fixes #2410.


layers/unique_objects.cpp
```diff
@@ -491,24 +491,53 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateSharedSwapchainsKHR(VkDevice device, uint32
...
+    if ((VK_SUCCESS == result) || (VK_INCOMPLETE == result)) {
         if ((*pSwapchainImageCount > 0) && pSwapchainImages) {
             std::lock_guard<std::mutex> lock(global_lock);
-            for (uint32_t i = 0; i < *pSwapchainImageCount; ++i) {
-                pSwapchainImages[i] = WrapNew(my_device_data, pSwapchainImages[i]);
+            std::unordered_map<VkSwapchainKHR, std::vector<VkImage>>::iterator item =
+                my_device_data->swapchain_wrapped_image_handle_map.find(wrapped_swapchain_handle);
+            if (item == my_device_data->swapchain_wrapped_image_handle_map.end()) {
+                // This swapchain is new, add it to the map
+                auto &indices_vector = my_device_data->swapchain_wrapped_image_handle_map[wrapped_swapchain_handle];
```
krOoze: Why is it called `indices_vector`? You seem to store handles.

layers/unique_objects.cpp
```diff
@@ -491,24 +491,53 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateSharedSwapchainsKHR(VkDevice device, uint32
...
+                auto &wrapped_handles_vector = item->second;
+                uint32_t num_existing_indices = static_cast<uint32_t>(wrapped_handles_vector.size());
+                for (uint32_t i = 0; i < num_existing_indices; i++) {
+                    pSwapchainImages[i] = wrapped_handles_vector[i];
+                }
+                // If we have new indices (i.e., on a VK_INCOMPLETE) wrap, save, and return them
+                for (uint32_t i = num_existing_indices; i < *pSwapchainImageCount; i++) {
+                    pSwapchainImages[i] = WrapNew(my_device_data, pSwapchainImages[i]);
+                    wrapped_handles_vector.push_back(pSwapchainImages[i]);
+                }
```
krOoze: All of this seems unnecessarily complicated to me. Why not just:

```
auto &swapchain_images = my_device_data->swapchain_wrapped_image_handle_map[wrapped_swapchain_handle];
for( uint32_t i = swapchain_images.size(); i < *pSwapchainImageCount; ++i )
    swapchain_images.push_back( WrapNew(pSwapchainImages[i]) );
for( uint32_t i = 0; i < *pSwapchainImageCount; ++i ) pSwapchainImages[i] = swapchain_images[i];
```
that should get rid of most of the `if`s, iterators, `find`s and whatnot

layers/unique_objects.cpp
```diff
@@ -491,24 +491,53 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateSharedSwapchainsKHR(VkDevice device, uint32
...
             }
         }
     }
     return result;
 }
 
+VKAPI_ATTR void VKAPI_CALL DestroySwapchainKHR(VkDevice device, VkSwapchainKHR swapchain, const VkAllocationCallbacks *pAllocator) {
+    layer_data *dev_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
+    std::unique_lock<std::mutex> lock(global_lock);
+    dev_data->swapchain_wrapped_image_handle_map.erase(swapchain);
```
krOoze: We should perhaps remove the images from `unique_id_mapping` first too?

layers/unique_objects.cpp
```diff
@@ -491,24 +491,53 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateSharedSwapchainsKHR(VkDevice device, uint32
...
+                    indices_vector.push_back(pSwapchainImages[i]);
+                }
+            } else {  // Existing swapchain, return existing wrapped handles
+                auto &wrapped_handles_vector = item->second;
+                uint32_t num_existing_indices = static_cast<uint32_t>(wrapped_handles_vector.size());
+                for (uint32_t i = 0; i < num_existing_indices; i++) {
+                    pSwapchainImages[i] = wrapped_handles_vector[i];
+                }
+                // If we have new indices (i.e., on a VK_INCOMPLETE) wrap, save, and return them
+                for (uint32_t i = num_existing_indices; i < *pSwapchainImageCount; i++) {
```
jzulauf-lunarg: Could reserve(*pSwapchainImageCount) on the vector to avoid reallocation in the for.

layers/unique_objects.h
```diff
@@ -79,6 +79,10 @@ struct layer_data {
     // uses unwrapped handles
     std::unordered_map<VkRenderPass, SubpassesUsageStates> renderpasses_states;
 
+    // Map of wrapped swapchain handles to arrays of wrapped swapchain image IDs
+    // Each swapchain has an immutable list of wrapped swapchain image IDs -- always return these IDs if they exist
+    std::unordered_map<VkSwapchainKHR, std::vector<VkImage>> swapchain_wrapped_image_handle_map;
```
krOoze: PS: personally I hate comments. They lie and are unmaintainable. Is there anything in them here that cannot be potentially expressed in code? Most of the comments seems to repeat what the variable type and name already contains.

layers/unique_objects.cpp
```diff
@@ -491,24 +491,53 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateSharedSwapchainsKHR(VkDevice device, uint32
...
+                // This swapchain is new, add it to the map
+                auto &indices_vector = my_device_data->swapchain_wrapped_image_handle_map[wrapped_swapchain_handle];
+                for (uint32_t i = 0; i < *pSwapchainImageCount; i++) {
+                    pSwapchainImages[i] = WrapNew(my_device_data, pSwapchainImages[i]);
+                    indices_vector.push_back(pSwapchainImages[i]);
+                }
+            } else {  // Existing swapchain, return existing wrapped handles
+                auto &wrapped_handles_vector = item->second;
+                uint32_t num_existing_indices = static_cast<uint32_t>(wrapped_handles_vector.size());
+                for (uint32_t i = 0; i < num_existing_indices; i++) {
```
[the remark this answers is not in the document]
krOoze: +1, good eye. Should have been `*pSwapchainImageCount` it seems.

layers/unique_objects.cpp
jzulauf-lunarg: modulo a reserve for swap_chain_images to minimize realloc/copy I don't see any downsides to this form.  It avoids the two corner cases noted.

layers/unique_objects.cpp
```diff
@@ -491,24 +491,53 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateSharedSwapchainsKHR(VkDevice device, uint32
...
+            std::unordered_map<VkSwapchainKHR, std::vector<VkImage>>::iterator item =
+                my_device_data->swapchain_wrapped_image_handle_map.find(wrapped_swapchain_handle);
+            if (item == my_device_data->swapchain_wrapped_image_handle_map.end()) {
+                // This swapchain is new, add it to the map
+                auto &indices_vector = my_device_data->swapchain_wrapped_image_handle_map[wrapped_swapchain_handle];
+                for (uint32_t i = 0; i < *pSwapchainImageCount; i++) {
+                    pSwapchainImages[i] = WrapNew(my_device_data, pSwapchainImages[i]);
+                    indices_vector.push_back(pSwapchainImages[i]);
+                }
+            } else {  // Existing swapchain, return existing wrapped handles
```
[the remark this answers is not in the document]
krOoze: I believe that is done on [L502](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/2414/files#diff-662926c228ed6b13586841c311fdf4c8R502), or you meant something else?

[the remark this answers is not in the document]
jzulauf-lunarg: Right, I think I swapped the L506 and L502 tests in my head...  thinking 502 only guard the "found" case.

mark-lunarg:
Excellent suggestions Gentlemen -- reworked PR with the good stuff.  Thanks!

layers/unique_objects.h
mark-lunarg: Comments can be a matter of taste, for sure.  In general, we like comments for stuff that'll help folks new to the codebase, and for things that are non-obvious.  That line is different for everybody.  I think we're pretty good about catching lies and unmaintainability in code reviews, so keep up the good work!

layers/unique_objects.cpp
```diff
@@ -491,24 +491,46 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateSharedSwapchainsKHR(VkDevice device, uint32
...
     }
     return result;
 }
 
+VKAPI_ATTR void VKAPI_CALL DestroySwapchainKHR(VkDevice device, VkSwapchainKHR swapchain, const VkAllocationCallbacks *pAllocator) {
+    layer_data *dev_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
+    std::unique_lock<std::mutex> lock(global_lock);
+
+    auto &image_array = dev_data->swapchain_wrapped_image_handle_map[swapchain];
+    for (auto &image_handle : image_array) {
```
krOoze: Could spam some `const`s here...

layers/unique_objects.cpp
```diff
@@ -491,24 +491,46 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateSharedSwapchainsKHR(VkDevice device, uint32
...
+        dev_data->unique_id_mapping.erase(HandleToUint64(image_handle));
+    }
+    image_array.clear();
```
krOoze: That's unnecessary. `vector` is able to clean after itself.

jzulauf-lunarg: The clear is redundant, erase calls the std::vector destructor.

PR #2418, opened 2018-02-19
Fixes compilation error from non-qualified std:: function invocation

jzulauf-lunarg:
Added std:: to isnan invocation.

The code this fixes passed GitHub CI.  Need to follow up and understand.


jzulauf-lunarg:
Current theory is that Travis CI (g++ 4.8.4) is less picky than more up-to-date versions (g++ 5.4.0) about namespace resolution for std::isnan.

PR #2421, opened 2018-02-20
Add validation to catch barriers with unbound memory

### LINKED ISSUE #2289: Missing validation: Transitioning an image before binding memory to it is illegal
As described in the title.

When initializing a texture image, I use vkCmdPipelineBarrier to transition it to from UNDEFINED to TRANSFER_DST_OPTIMAL before vkCmdCopyBufferToImage, which should be a normal way to go about things. However, I accidentally did this *before* binding memory to the image. Turns out most implementations don't care, as presumably any transition from UNDEFINED is generally a no-op apart from possibly updating a bit of metadata. Intel's ANV Linux driver did care though, and promptly crashed.

The validation layers didn't catch this, but probably should - seems like an easy case to detect.

jzulauf-lunarg:
Fixes #2289 . Added validation checks for image and buffer memory barriers s.t. non-sparse images and buffers without bound memory will cause validation errors.  Cleaned up buffer size test which was triggering and "unexpected". Additionally cleaned up 2 adjacent validation checks that *have* VUID's but weren't using them.  

There is not a valid usage for 2289 (et. al.) but appears to be needed on the same style as 

> If buffer is non-sparse then it must be bound completely and contiguously to a single VkDeviceMemory object

VUID-VkBufferViewCreateInfo-buffer-00935

> If image is non-sparse then it must be bound completely and contiguously to a single VkDeviceMemory object

VUID-VkImageViewCreateInfo-image-01020

Given the general requirements of 11.6 ... 

> Non-sparse resources must be bound completely and contiguously to a single VkDeviceMemory object before the resource is... (used in) ... recording commands in a command buffer


layers/core_validation.cpp
```diff
@@ -7629,22 +7636,29 @@ static bool ValidateBarriers(layer_data *device_data, const char *funcName, GLOB
...
 
         if (buffer_state) {
-            auto buffer_size = buffer_state->requirements.size;
+            // There is no VUID for this, but there is blanket text:
+            //     "Non-sparse resources must be bound completely and contiguously to a single VkDeviceMemory object before
+            //     recording commands in a command buffer"
+            // TODO: Update this when VUID is defined
+            skip |= ValidateMemoryIsBoundToBuffer(device_data, buffer_state, funcName, VALIDATION_ERROR_UNDEFINED);
+
+            auto buffer_size = buffer_state->createInfo.size;
```
jzulauf-lunarg: This seems more correct (if more restrictive) than checking for the requirements size, as the spec specifically speaks of the "buffer size" not the "size of the allocation bound to the buffer".

tests/layer_validation_tests.cpp
```diff
@@ -11377,18 +11404,18 @@ TEST_F(VkLayerTest, InvalidBarriers) {
     buf_barrier.offset = 0;
     buf_barrier.size = VK_WHOLE_SIZE;
 
-    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "which is not less than total size");
-    // Exceed the internal memory size
-    buf_barrier.offset = buffer.memory_requirements().size + 1;
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_01800946);
+    // Exceed the buffer size
+    buf_barrier.offset = buffer.create_info().size + 1;
```
jzulauf-lunarg: matches the new (potentially) more restrictive limits in the validation layers.

tests/vktestbinding.h
```diff
@@ -406,6 +406,7 @@ class Buffer : public internal::NonDispHandle<VkBuffer> {
     // vkBindObjectMemory()
     void bind_memory(const DeviceMemory &mem, VkDeviceSize mem_offset);
 
+    const VkBufferCreateInfo &create_info() const { return create_info_; }
```
jzulauf-lunarg: Needed to get the boundaries for the new size tests.

layers/core_validation.cpp
```diff
@@ -7594,11 +7594,18 @@ static bool ValidateBarriers(layer_data *device_data, const char *funcName, GLOB
...
+                            "%s: Image Layout cannot be transitioned to UNDEFINED or PREINITIALIZED. %s", funcName,
+                            validation_error_map[VALIDATION_ERROR_0a00095c]);
         }
 
         if (image_data) {
+            // There is no VUID for this, but there is blanket text:
+            //     "Non-sparse resources must be bound completely and contiguously to a single VkDeviceMemory object before
+            //     recording commands in a command buffer."
+            // TODO: Update this when VUID is defined
+            skip |= ValidateMemoryIsBoundToImage(device_data, image_data, funcName, VALIDATION_ERROR_UNDEFINED);
```
jzulauf-lunarg: This validation is done is several other calls -- reusing.

layers/core_validation.cpp
```diff
@@ -7629,22 +7636,29 @@ static bool ValidateBarriers(layer_data *device_data, const char *funcName, GLOB
         skip |= ValidateBarrierQueueFamilies(device_data, funcName, cb_state, mem_barrier, buffer_state);
 
         if (buffer_state) {
-            auto buffer_size = buffer_state->requirements.size;
+            // There is no VUID for this, but there is blanket text:
+            //     "Non-sparse resources must be bound completely and contiguously to a single VkDeviceMemory object before
+            //     recording commands in a command buffer"
+            // TODO: Update this when VUID is defined
+            skip |= ValidateMemoryIsBoundToBuffer(device_data, buffer_state, funcName, VALIDATION_ERROR_UNDEFINED);
```
jzulauf-lunarg: No wheels reinvented here.

PR #2422, opened 2018-02-20
Vulkaninfo json generation for use with DevSim

shannon-lunarg:
Two additional vulkaninfo output modes.

Possible output modes:

1. Std console (human readable) only
2. Html only
3. **Json only**
4. **Html AND json**

Json specified with `-j`, `--json`, or `--json=1` (to specify which gpu to target on a multi-gpu system).

### DIFF
--- a/demos/vulkaninfo.c
+++ b/demos/vulkaninfo.c
@@ -96,23 +96,13 @@ static int ConsoleIsExclusive(void) {
 #define APP_SHORT_NAME "vulkaninfo"
 
 static bool html_output = false;
+static bool human_readable_output = true;
 static bool json_output = false;
 static uint32_t selected_gpu = 0;
 
 struct VkStructureHeader {
     VkStructureType sType;
     void *pNext;
-};
-
-struct AppGpu;
-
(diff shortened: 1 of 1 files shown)


demos/vulkaninfo.c
```diff
@@ -477,7 +477,7 @@ static const char *VkPresentModeString(VkPresentModeKHR mode) {
 #endif
 
 static bool CheckExtensionEnabled(const char *extension_to_check, const char **extension_list, uint32_t extension_count) {
-    for (uint32_t i = 0; i < extension_count; i++) {
+    for (uint32_t i = 0; i < extension_count; ++i) {
```
mark-lunarg: Not sure if your IDE is doing this automatically, or if you just couldn't stand it anymore, but the pre/post-increment changes should not go in this commit.  And if they're in their own commit, then be prepared to defend yourself in a flame-war.  Just-sayin'.

shannon-lunarg: Alrighty, I put these post-increment/pre-increment changes in their own commit.

PR #2423, opened 2018-02-20
Adds push constant range and "compatible for set N" logic to pipeline layout compatibility validation and recording

### LINKED ISSUE #2214: Core validation does not check push constant ranges to determine set compatibility between pipeline layouts
This is from [LunarXchange, #648](https://vulkan.lunarg.com/issue/view/588c4e776631512b4c4b1f50):

The Vulkan specification states:

> Two pipeline layouts are defined to be “compatible for set N” if they were created with matching (the same, or identically defined) descriptor set layouts for sets zero through N, and if they were created with identical push constant ranges.

The core validation layers however do not seem to check the push constant ranges when determining pipeline layout compatibility regarding descriptor sets. In my project I use two different pipeline layouts. The first one only defines a descriptor set with one uniform buffer (meant to contain view and projection matrices); no push constants. The second one defines two descriptor sets, the first with one uniform buffer (in order to be compatible with the first layout) and the second with additional resources like a sampler and sampled images. The second set also defines a push constant range (used to set the model matrix). When rendering I first bind the first descriptor set with the first layout and then the second descriptor set with the second layout. Why? Because it happens in two different objects.

According to the quoted paragraph above this should not work because the push constant ranges of both layouts are different. The core validation layer however does not report anything (no error, warn, perf, info). Is there a reason for this lack of errors/warnings?

jzulauf-lunarg:
Fixes #2214 .  Upgrades the pipeline layout compatibility and recording to (IIAC) match the compatibility text of of the spec.  Push constant and 1-N comparisons for compatibility are added for each set in the current layout.  Set update/disturb logic is now shared with push descriptors. 

As compatibility validation is deep nests (for each Set N { foreach set 1 through N } {for each set { for each binding in set }, where the inner three loops are comparing invariant quantities, the compatibility check has been reduced to a comparison of unique ID's which integrate each inner loop in  "canonical" definitions stored in dictionaries.  The dictionaries share a common container type and used the standard overload mechanism for std::equal_to and std::hash.

Specific layer state data (where appropriate) has been segregated into handle-specific and non-handle specific storage. Where possible, comparisons of unique ID's are used for "trivial accept" validation.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -1196,7 +1196,7 @@ static bool ValidateDrawState(layer_data *dev_data, GLOBAL_CB_NODE *cb_node, CMD
         result = validate_draw_state_flags(dev_data, cb_node, pPipe, indexed, msg_code);
 
     // Now complete other state checks
-    if (VK_NULL_HANDLE != state.pipeline_layout.layout) {
+    if (VK_NULL_HANDLE != state.pipeline_layout) {
         string errorString;
         auto pipeline_layout = pPipe->pipeline_layout;
 
@@ -1256,7 +1256,7 @@ static bool ValidateDrawState(layer_data *dev_data, GLOBAL_CB_NODE *cb_node, CMD
 static void UpdateDrawState(layer_data *dev_data, GLOBAL_CB_NODE *cb_state, const VkPipelineBindPoint bind_point) {
     auto const &state = cb_state->lastBound[bind_point];
     PIPELINE_STATE *pPipe = state.pipeline_state;
-    if (VK_NULL_HANDLE != state.pipeline_layout.layout) {
+    if (VK_NULL_HANDLE != state.pipeline_layout) {
         for (const auto &set_binding_pair : pPipe->active_slots) {
             uint32_t setIndex = set_binding_pair.first;
             // Pull the set node
@@ -6138,60 +6138,125 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
     if (!skip) dev_data->dispatch_table.CmdSetStencilReference(commandBuffer, faceMask, reference);
 }
 
+// Update pipeline_layout bind points applying the "Pipeline Layout Compatibility" rules
+static void UpdateLastBoundDescriptorSets(layer_data *device_data, GLOBAL_CB_NODE *cb_state,
+                                          VkPipelineBindPoint pipeline_bind_point, const PIPELINE_LAYOUT_NODE *pipeline_layout,
+                                          uint32_t first_set, uint32_t set_count,
+                                          const std::vector<cvdescriptorset::DescriptorSet *> descriptor_sets,
+                                          uint32_t dynamic_offset_count, const uint32_t *p_dynamic_offsets) {
+    // Defensive
+    assert(set_count);
+    if (0 == set_count) return;
+    assert(pipeline_layout);
+    if (!pipeline_layout) return;
+
+    uint32_t required_size = first_set + set_count;
+    const uint32_t last_binding_index = required_size - 1;
+    assert(last_binding_index < pipeline_layout->compat_for_set.size());
+
+    // Some useful shorthand
+    auto &last_bound = cb_state->lastBound[pipeline_bind_point];
+
+    auto &bound_sets = last_bound.boundDescriptorSets;
+    auto &dynamic_offsets = last_bound.dynamicOffsets;
+    auto &bound_compat_ids = last_bound.compat_id_for_set;
+    auto &pipe_compat_ids = pipeline_layout->compat_for_set;
+
+    const uint32_t current_size = static_cast<uint32_t>(bound_sets.size());
+    assert(current_size == dynamic_offsets.size());
+    assert(current_size == bound_compat_ids.size());
+
+    // We need this three times in this function, but nowhere else
+    auto push_descriptor_cleanup = [&last_bound](const cvdescriptorset::DescriptorSet *ds) -> bool {
+        if (ds && ds->IsPushDescriptor()) {
+            assert(ds == last_bound.push_descriptor_set.get());
+            last_bound.push_descriptor_set = nullptr;
+            return true;
+        }
+        return false;
+    };
+
+    // Clean up the "disturbed" before and after the range to be set
+    if (required_size < current_size) {
+        if (bound_compat_ids[last_binding_index] != pipe_compat_ids[last_binding_index]) {
+            // We're disturbing those after last, we'll shrink below, but first need to check for and cleanup the push_descriptor
+            for (auto set_idx = required_size; set_idx < current_size; ++set_idx) {
+                if (push_descriptor_cleanup(bound_sets[set_idx])) break;
+            }
+        } else {
+            // We're not disturbing past last, so leave the upper binding data alone.
+            required_size = current_size;
+        }
+    }
+
+    // We resize if we need more set entries or if those past "last" are disturbed
+    if (required_size != current_size) {
+        // TODO: put these size tied things in a struct (touches many lines)
+        bound_sets.resize(required_size);
+        dynamic_offsets.resize(required_size);
+        bound_compat_ids.resize(required_size);
+    }
+
+    // For any previously bound sets, need to set them to "invalid" if they were disturbed by this update
+    for (uint32_t set_idx = 0; set_idx < first_set; ++set_idx) {
+        if (bound_compat_ids[set_idx] != pipe_compat_ids[set_idx]) {
+            push_descriptor_cleanup(bound_sets[set_idx]);
+            bound_sets[set_idx] = nullptr;
+            dynamic_offsets[set_idx].clear();
+            bound_compat_ids[set_idx] = pipe_compat_ids[set_idx];
+        }
+    }
+
+    // Now update the bound sets with the input sets
+    const uint32_t *input_dynamic_offsets = p_dynamic_offsets;  // "read" pointer for dynamic offset data
+    for (uint32_t input_idx = 0; input_idx < set_count; input_idx++) {
+        auto set_idx = input_idx + first_set;  // set_idx is index within layout, input_idx is index within input descriptor sets
+        cvdescriptorset::DescriptorSet *descriptor_set = descriptor_sets[input_idx];
+
+        // Record binding (or push)
+        push_descriptor_cleanup(bound_sets[set_idx]);
+        bound_sets[set_idx] = descriptor_set;
+        bound_compat_ids[set_idx] = pipe_compat_ids[set_idx];  // compat ids are canonical *per* set index
+
+        if (descriptor_set) {
+            auto set_dynamic_descriptor_count = descriptor_set->GetDynamicDescriptorCount();
+            // TODO: Add logic for tracking push_descriptor offsets (here or in caller)
+            if (set_dynamic_descriptor_count && input_dynamic_offsets) {
+                const uint32_t *end_offset = input_dynamic_offsets + set_dynamic_descriptor_count;
+                dynamic_offsets[set_idx] = std::vector<uint32_t>(input_dynamic_offsets, end_offset);
+                input_dynamic_offsets = end_offset;
+                assert(input_dynamic_offsets <= (p_dynamic_offsets + dynamic_offset_count));
+            } else {
+                dynamic_offsets[set_idx].clear();
+            }
+            if (!descriptor_set->IsPushDescriptor()) {
+                // Can't cache validation of push_descriptors
+                cb_state->validated_descriptor_sets.insert(descriptor_set);
+            }
+        }
+    }
+}
+
+// Update the bound state for the bind point, including the effects of incompatible pipeline layouts
 static void PreCallRecordCmdBindDescriptorSets(layer_data *device_data, GLOBAL_CB_NODE *cb_state,
                                                VkPipelineBindPoint pipelineBindPoint, VkPipelineLayout layout, uint32_t firstSet,
                                                uint32_t setCount, const VkDescriptorSet *pDescriptorSets,
                                                uint32_t dynamicOffsetCount, const uint32_t *pDynamicOffsets) {
-    uint32_t total_dynamic_descriptors = 0;
-    string error_string = "";
-    uint32_t last_set_index = firstSet + setCount - 1;
-    auto last_bound = &cb_state->lastBound[pipelineBindPoint];
-
-    if (last_set_index >= last_bound->boundDescriptorSets.size()) {
-        last_bound->boundDescriptorSets.resize(last_set_index + 1);
-        last_bound->dynamicOffsets.resize(last_set_index + 1);
-    }
-    auto old_final_bound_set = last_bound->boundDescriptorSets[last_set_index];
     auto pipeline_layout = getPipelineLayout(device_data, layout);
-    for (uint32_t set_idx = 0; set_idx < setCount; set_idx++) {
-        cvdescriptorset::DescriptorSet *descriptor_set = GetSetNode(device_data, pDescriptorSets[set_idx]);
-        if (descriptor_set) {
-            last_bound->pipeline_layout = *pipeline_layout;
+    std::vector<cvdescriptorset::DescriptorSet *> descriptor_sets;
+    descriptor_sets.reserve(setCount);
 
-            if ((last_bound->boundDescriptorSets[set_idx + firstSet] != nullptr) &&
-                last_bound->boundDescriptorSets[set_idx + firstSet]->IsPushDescriptor()) {
-                last_bound->push_descriptor_set = nullptr;
-                last_bound->boundDescriptorSets[set_idx + firstSet] = nullptr;
-            }
-
-            last_bound->boundDescriptorSets[set_idx + firstSet] = descriptor_set;
-
-            auto set_dynamic_descriptor_count = descriptor_set->GetDynamicDescriptorCount();
-            last_bound->dynamicOffsets[firstSet + set_idx].clear();
-            if (set_dynamic_descriptor_count) {
-                last_bound->dynamicOffsets[firstSet + set_idx] =
-                    std::vector<uint32_t>(pDynamicOffsets + total_dynamic_descriptors,
-                                          pDynamicOffsets + total_dynamic_descriptors + set_dynamic_descriptor_count);
-                total_dynamic_descriptors += set_dynamic_descriptor_count;
-            }
-            cb_state->validated_descriptor_sets.insert(descriptor_set);
-        }
-        // For any previously bound sets, need to set them to "invalid" if they were disturbed by this update
-        if (firstSet > 0) {
-            for (uint32_t i = 0; i < firstSet; ++i) {
-                if (last_bound->boundDescriptorSets[i] &&
-                    !verify_set_layout_compatibility(last_bound->boundDescriptorSets[i], pipeline_layout, i, error_string)) {
-                    last_bound->boundDescriptorSets[i] = VK_NULL_HANDLE;
-                }
-            }
-        }
-        // Check if newly last bound set invalidates any remaining bound sets
-        if ((last_bound->boundDescriptorSets.size() - 1) > (last_set_index)) {
-            if (old_final_bound_set &&
-                !verify_set_layout_compatibility(old_final_bound_set, pipeline_layout, last_set_index, error_string)) {
-                last_bound->boundDescriptorSets.resize(last_set_index + 1);
-            }
-        }
+    // Construct a list of the descriptors
+    bool found_non_null = false;
+    for (uint32_t i = 0; i < setCount; i++) {
+        cvdescriptorset::DescriptorSet *descriptor_set = GetSetNode(device_data, pDescriptorSets[i]);
+        descriptor_sets.emplace_back(descriptor_set);
+        found_non_null |= descriptor_set != nullptr;
+    }
+    if (found_non_null) {  // which implies setCount > 0
+        UpdateLastBoundDescriptorSets(device_data, cb_state, pipelineBindPoint, pipeline_layout, firstSet, setCount,
+                                      descriptor_sets, dynamicOffsetCount, pDynamicOffsets);
+        cb_state->lastBound[pipelineBindPoint].pipeline_layout = layout;
     }
 }
 
@@ -6211,6 +6276,7 @@ static bool PreCallValidateCmdBindDescriptorSets(layer_data *device_data, GLOBAL
     if (last_set_index >= cb_state->lastBound[pipelineBindPoint].boundDescriptorSets.size()) {
         cb_state->lastBound[pipelineBindPoint].boundDescriptorSets.resize(last_set_index + 1);
         cb_state->lastBound[pipelineBindPoint].dynamicOffsets.resize(last_set_index + 1);
+        cb_state->lastBound[pipelineBindPoint].compat_id_for_set.resize(last_set_index + 1);
     }
     auto pipeline_layout = getPipelineLayout(device_data, layout);
     for (uint32_t set_idx = 0; set_idx < setCount; set_idx++) {
@@ -6391,15 +6457,15 @@ static bool PreCallValidateCmdPushDescriptorSetKHR(layer_data *device_data, GLOB
 static void PreCallRecordCmdPushDescriptorSetKHR(layer_data *device_data, GLOBAL_CB_NODE *cb_state,
                                                  VkPipelineBindPoint pipelineBindPoint, VkPipelineLayout layout, uint32_t set,
                                                  uint32_t descriptorWriteCount, const VkWriteDescriptorSet *pDescriptorWrites) {
-    if (set >= cb_state->lastBound[pipelineBindPoint].boundDescriptorSets.size()) {
-        cb_state->lastBound[pipelineBindPoint].boundDescriptorSets.resize(set + 1);
-        cb_state->lastBound[pipelineBindPoint].dynamicOffsets.resize(set + 1);
-    }
-    const auto &layout_state = getPipelineLayout(device_data, layout);
+    const auto &pipeline_layout = getPipelineLayout(device_data, layout);
+    if (!pipeline_layout) return;
     std::unique_ptr<cvdescriptorset::DescriptorSet> new_desc{
-        new cvdescriptorset::DescriptorSet(0, 0, layout_state->set_layouts[set], device_data)};
-    cb_state->lastBound[pipelineBindPoint].boundDescriptorSets[set] = new_desc.get();
+        new cvdescriptorset::DescriptorSet(0, 0, pipeline_layout->set_layouts[set], device_data)};
+
+    std::vector<cvdescriptorset::DescriptorSet *> descriptor_sets = {new_desc.get()};
+    UpdateLastBoundDescriptorSets(device_data, cb_state, pipelineBindPoint, pipeline_layout, set, 1, descriptor_sets, 0, nullptr);
     cb_state->lastBound[pipelineBindPoint].push_descriptor_set = std::move(new_desc);
+    cb_state->lastBound[pipelineBindPoint].pipeline_layout = layout;
 }
 
 VKAPI_ATTR void VKAPI_CALL CmdPushDescriptorSetKHR(VkCommandBuffer commandBuffer, VkPipelineBindPoint pipelineBindPoint,
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -95,9 +95,9 @@ size_t PipelineLayoutCompatDef::hash() const {
     hash_util::HashCombiner hc;
     // The set number is integral to the CompatDef's distinctiveness
     hc << set << push_constant_ranges.get();
-    const auto &layout_layout = *layout_id.get();
+    const auto &descriptor_set_layouts = *set_layouts_id.get();
     for (uint32_t i = 0; i <= set; i++) {
-        hc << layout_layout[i].get();
+        hc << descriptor_set_layouts[i].get();
     }
     return hc.Value();
 }
@@ -107,12 +107,18 @@ bool PipelineLayoutCompatDef::operator==(const PipelineLayoutCompatDef &other) c
         return false;
     }
 
-    const auto &layout_layout = *layout_id.get();
-    assert(set < layout_layout.size());
-    const auto &other_ll = *other.layout_id.get();
-    assert(set < other_ll.size());
+    if (set_layouts_id == other.set_layouts_id) {
+        // if it's the same set_layouts_id, then *any* subset will match
+        return true;
+    }
+
+    // They aren't exactly the same PipelineLayoutSetLayouts, so we need to check if the required subsets match
+    const auto &descriptor_set_layouts = *set_layouts_id.get();
+    assert(set < descriptor_set_layouts.size());
+    const auto &other_ds_layouts = *other.set_layouts_id.get();
+    assert(set < other_ds_layouts.size());
     for (uint32_t i = 0; i <= set; i++) {
-        if (layout_layout[i] != other_ll[i]) {
+        if (descriptor_set_layouts[i] != other_ds_layouts[i]) {
             return false;
         }
     }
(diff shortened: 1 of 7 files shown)


layers/core_validation_types.h
```diff
@@ -645,17 +680,18 @@ class PIPELINE_STATE : public BASE_NODE {
 // Track last states that are bound per pipeline bind point (Gfx & Compute)
 struct LAST_BOUND_STATE {
     PIPELINE_STATE *pipeline_state;
-    PIPELINE_LAYOUT_NODE pipeline_layout;
+    VkPipelineLayout pipeline_layout;
```
jzulauf-lunarg: Formerly we were copying by value the whole of the pipe_layout but only using the "layout" member, elminating needless copying.

layers/descriptor_sets.cpp
```diff
@@ -23,28 +23,34 @@
...
-cvdescriptorset::DescriptorSetLayout::DescriptorSetLayout(const VkDescriptorSetLayoutCreateInfo *p_create_info,
-                                                          const VkDescriptorSetLayout layout)
-    : layout_(layout),
-      layout_destroyed_(false),
-      flags_(p_create_info->flags),
-      binding_count_(0),
-      descriptor_count_(0),
-      dynamic_descriptor_count_(0) {
+cvdescriptorset::DescriptorSetLayoutDef::DescriptorSetLayoutDef(const VkDescriptorSetLayoutCreateInfo *p_create_info)
+    : flags_(p_create_info->flags), binding_count_(0), descriptor_count_(0), dynamic_descriptor_count_(0) {
```
jzulauf-lunarg: This is where we split the handle variant and invariant chunks. Ensuring that the interface remains the same to contain the number of lines changed by the refactor

layers/descriptor_sets.cpp
```diff
@@ -106,91 +112,47 @@ cvdescriptorset::DescriptorSetLayout::DescriptorSetLayout(const VkDescriptorSetL
...
-        skip |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0, __LINE__,
-                        VALIDATION_ERROR_05000232, "DS",
-                        "for push descriptor, total descriptor count in layout (%" PRIu64
-                        ") must not be greater than VkPhysicalDevicePushDescriptorPropertiesKHR::maxPushDescriptors (%" PRIu32
-                        "%s). %s",
-                        total_descriptors, max_push_descriptors, undefined, validation_error_map[VALIDATION_ERROR_05000232]);
-    }
-
-    return skip;
+size_t cvdescriptorset::DescriptorSetLayoutDef::hash() const {
```
jzulauf-lunarg: See hash_utils.h for how we are composing hashes on non-simple types.

layers/descriptor_sets.cpp
```diff
@@ -247,50 +209,61 @@ bool cvdescriptorset::DescriptorSetLayout::IsCompatible(DescriptorSetLayout cons
                                                         std::string *error_msg) const {
     // Trivial case
     if (layout_ == rh_ds_layout->GetDescriptorSetLayout()) return true;
-    if (descriptor_count_ != rh_ds_layout->descriptor_count_) {
+    if (get_layout_def() == rh_ds_layout->get_layout_def()) return true;
```
jzulauf-lunarg: Here we compare the canonical for for trivially valid, by-passing the detailed check, unless it isn't compatible.

layers/descriptor_sets.cpp
```diff
@@ -247,50 +209,61 @@ bool cvdescriptorset::DescriptorSetLayout::IsCompatible(DescriptorSetLayout cons
...
+    bool detailed_compat_check =
+        get_layout_def()->IsCompatible(layout_, rh_ds_layout->GetDescriptorSetLayout(), rh_ds_layout->get_layout_def(), error_msg);
+    // The detailed check should never tell us mismatching DSL are compatible
+    assert(!detailed_compat_check);
+    return detailed_compat_check;
+}
+
+bool cvdescriptorset::DescriptorSetLayoutDef::IsCompatible(VkDescriptorSetLayout ds_layout, VkDescriptorSetLayout rh_ds_layout,
```
jzulauf-lunarg: The DSL Def doesn't know a thing about the handles, so we have to supply them.

layers/descriptor_sets.cpp
```diff
@@ -346,6 +319,64 @@ bool cvdescriptorset::DescriptorSetLayout::VerifyUpdateConsistency(uint32_t curr
     return true;
 }
 
+// The DescriptorSetLayout stores the per handle data for a descriptor set layout, and references the common defintion for the
+// handle invariant portion
+cvdescriptorset::DescriptorSetLayout::DescriptorSetLayout(const VkDescriptorSetLayoutCreateInfo *p_create_info,
+                                                          const VkDescriptorSetLayout layout)
+    : layout_(layout), layout_destroyed_(false), layout_id_(get_canonical_id(p_create_info)) {}
```
jzulauf-lunarg: get_canonical turns the DSL's *definition* (i.e. the create info) into a unique ID using the dictionary of DSL def's.

layers/descriptor_sets.h
```diff
@@ -157,23 +161,115 @@ class DescriptorSetLayout {
...
     uint32_t dynamic_descriptor_count_;
     BindingTypeStats binding_type_stats_;
 };
 
+static bool operator==(const DescriptorSetLayoutDef &lhs, const DescriptorSetLayoutDef &rhs) {
+    return (lhs.GetCreateFlags() == rhs.GetCreateFlags()) && (lhs.GetBindings() == rhs.GetBindings());
+}
+
+// Canonical dictionary of DSL definitions -- independent of device or handle
+using DescriptorSetLayoutDict = hash_util::Dictionary<DescriptorSetLayoutDef, hash_util::HasHashMember<DescriptorSetLayoutDef>>;
```
jzulauf-lunarg: hash_util has utility wrappers for structs with hash methods and for ordered containers... that match the interface of std::hash (and use it under the covers by default)

layers/descriptor_sets.h
```diff
@@ -157,23 +161,115 @@ class DescriptorSetLayout {
...
+using DescriptorSetLayoutId = DescriptorSetLayoutDict::Id;
```
jzulauf-lunarg: This is the same approach that std:: containers use, allowing the implementation details to change, but the interface to remain the same.

layers/descriptor_sets.h
```diff
@@ -157,23 +161,115 @@ class DescriptorSetLayout {
...
+
+class DescriptorSetLayout {
+   public:
+    // Constructors and destructor
+    DescriptorSetLayout(const VkDescriptorSetLayoutCreateInfo *p_create_info, const VkDescriptorSetLayout layout);
+    // Validate create info - should be called prior to creation
+    static bool ValidateCreateInfo(const debug_report_data *, const VkDescriptorSetLayoutCreateInfo *, const bool, const uint32_t);
```
jzulauf-lunarg: Wrapping the DSL Def interface s.t. current code that uses the DSL interface can remain unchanged.

layers/hash_util.h
```diff
@@ -0,0 +1,164 @@
...
+// Hash and equality utilities for supporting hashing containers (e.g. unordered_set, unordered_map)
+namespace hash_util {
+
+// True iff both pointers are null or both are non-null
+template <typename T>
+bool similar_for_nullity(const T *const lhs, const T *const rhs) {
+    return ((lhs != nullptr) && (rhs != nullptr)) || ((lhs == nullptr) && (rhs == nullptr));
+}
+
+// Wrap std hash to avoid manual casts for the holes in std::hash (in C++11)
```
jzulauf-lunarg: These two SFINAE variants automatically reduce (or don't) enum's into their undelying types (or not).  The cannot be members of the HashCombiner class below as (AFAICT) SFINAE does not allow for SF within member templates.

SFINAE experts feel free to chime in.

layers/hash_util.h
```diff
@@ -0,0 +1,164 @@
...
+    return std::hash<Value>()(value);
+}
+
+template <typename Value>
+size_t HashWithUnderlying(Value value, typename std::enable_if<std::is_enum<Value>::value, void *>::type = nullptr) {
+    using Underlying = typename std::underlying_type<Value>::type;
+    return std::hash<Underlying>()(static_cast<const Underlying &>(value));
+}
+
+class HashCombiner {
```
jzulauf-lunarg: A class to compose hashes from various pieces.  Conceptually similar to boost functionality.

layers/hash_util.h
```diff
@@ -0,0 +1,164 @@
...
+    template <typename Value>
+    struct WrappedHash {
+        size_t operator()(const Value &value) const { return HashWithUnderlying(value); }
+    };
+
+    HashCombiner(Key combined = 0) : combined_(combined) {}
+    // magic and combination algorithm based on boost::hash_combine
+    // http://www.boost.org/doc/libs/1_43_0/doc/html/hash/reference.html#boost.hash_combine
+    // Magic value is 2^size / ((1-sqrt(5)/2)
+    static const uint64_t kMagic = sizeof(Key) > 4 ? Key(0x9e3779b97f4a7c16UL) : Key(0x9e3779b9U);
```
jzulauf-lunarg: Note the the magic value needs > 64 bit double to compute.  Gcc on linux/x64 does 80bit long doubles.

layers/hash_util.h
```diff
@@ -0,0 +1,164 @@
...
+struct HasHashMember {
+    size_t operator()(const T &value) const { return value.hash(); }
+};
+
+// A template to inherit std::hash overloads from when is an *ordered* constainer
+template <typename T>
+struct IsOrderedContainer {
+    size_t operator()(const T &value) const { return HashCombiner().Combine(value.cbegin(), value.cend()).Value(); }
+};
+
```
jzulauf-lunarg: Thar be the UID dragons below.  The key things are that the definitions are always stored as a unique shared_ptr, the hash and equality check are on the *value* referenced by the shared_ptr, and the built-in ordered_set::insert logic is used to keep the entries unique.

layers/hash_vk_types.h
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Author: John Zulauf <jzulauf@lunarg.com>
+ */
+#ifndef HASH_VK_TYPES_H_
```
jzulauf-lunarg: these are in there own include file to keep from having an include loops if defined in core_validation_type or descriptor_set etc.

layers/core_validation.cpp
```diff
@@ -5375,20 +5405,77 @@ static bool PreCallValiateCreatePipelineLayout(const layer_data *dev_data, const
     return skip;
 }
 
+
+// For repeatble sorting, not useful very useful for "memory in range" search
```
mark-lunarg: sp

layers/hash_vk_types.h
```diff
@@ -0,0 +1,86 @@
...
+template <>
+struct hash<VkPushConstantRange> {
+    size_t operator()(const VkPushConstantRange &value) const {
+        hash_util::HashCombiner hc;
+        return (hc << value.stageFlags << value.offset << value.size).Value();
+    }
+};
+}  // namespace std
+
+using PushConstantRanges = std::vector<VkPushConstantRange>;
```
jzulauf-lunarg: There isn't a built-in Vk* type for a collection of PCR's so we invent one.

layers/shader_validation.cpp
```diff
@@ -1329,7 +1329,7 @@ static bool validate_pipeline_shader_stage(layer_data *dev_data, VkPipelineShade
     auto descriptor_uses = collect_interface_by_descriptor_slot(report_data, module, accessible_ids);
 
     skip |= validate_specialization_offsets(report_data, pStage);
-    skip |= validate_push_constant_usage(report_data, &pipeline->pipeline_layout.push_constant_ranges, module, accessible_ids,
+    skip |= validate_push_constant_usage(report_data, pipeline->pipeline_layout.push_constant_ranges.get(), module, accessible_ids,
```
jzulauf-lunarg: one of the few place  the changed PCR storage definition is visible.

layers/core_validation.cpp
```diff
@@ -6048,63 +6135,108 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
 
-static void PreCallRecordCmdBindDescriptorSets(layer_data *device_data, GLOBAL_CB_NODE *cb_state,
-                                               VkPipelineBindPoint pipelineBindPoint, VkPipelineLayout layout, uint32_t firstSet,
-                                               uint32_t setCount, const VkDescriptorSet *pDescriptorSets,
-                                               uint32_t dynamicOffsetCount, const uint32_t *pDynamicOffsets) {
-    uint32_t total_dynamic_descriptors = 0;
-    string error_string = "";
-    uint32_t last_set_index = firstSet + setCount - 1;
-    auto last_bound = &cb_state->lastBound[pipelineBindPoint];
+// Update pipeline_layout bind points applying the "Pipeline Layout Compatibility" rules
```
jzulauf-lunarg: Reusing this logic for Bind and Push of descriptor sets, so refactor it out.

layers/core_validation.cpp
```diff
@@ -6048,63 +6135,108 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
+static void UpdateLastBoundDescriptorSets(layer_data *device_data, GLOBAL_CB_NODE *cb_state,
```
jzulauf-lunarg: Completely different from the previous logic and approach.

layers/core_validation.cpp
```diff
@@ -6048,63 +6135,108 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
+    // Some useful shorthand
+    auto &last_bound = cb_state->lastBound[pipeline_bind_point];
+
+    auto &bound_sets = last_bound.boundDescriptorSets;
+    auto &dynamic_offsets = last_bound.dynamicOffsets;
+    auto &bound_compat_ids = last_bound.compat_id_for_set;
+    auto &pipe_compat_ids = pipeline_layout->compat_for_set;
+
+    // We resize if we need more set entries or if those past "last" are disturbed
+    if (required_size > bound_sets.size() ||
```
jzulauf-lunarg: Here we use resizing both constructively and destructively.  Oh dear, I think I may need to clean up the push_descriptor set if it gets disturbed and shrunk away...  expect a small revision.

layers/core_validation.cpp
```diff
@@ -6048,63 +6135,108 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
+
+        // Clean up push descriptor
+        if (bound_sets[set_idx] && bound_sets[set_idx]->IsPushDescriptor()) {
+            assert(descriptor_set == last_bound.push_descriptor_set.get());
+            last_bound.push_descriptor_set = nullptr;
+        }
+
+        // Record binding (or push)
+        bound_sets[set_idx] = descriptor_set;
+        bound_compat_ids[set_idx] = pipe_compat_ids[set_idx];  // compat ids are canonical *per* set index
```
jzulauf-lunarg: This is where we track what the compatibility for set N was when we wrote this set.

layers/core_validation.cpp
```diff
@@ -6048,63 +6135,108 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
+            if (set_dynamic_descriptor_count && input_dynamic_offsets) {
+                const uint32_t *end_offset = input_dynamic_offsets + set_dynamic_descriptor_count;
+                dynamic_offsets[set_idx] = std::vector<uint32_t>(input_dynamic_offsets, end_offset);
+                input_dynamic_offsets = end_offset;
+                assert(input_dynamic_offsets <= (p_dynamic_offsets + dynamic_offset_count));
+            } else {
+                dynamic_offsets[set_idx].clear();
+            }
+            if (!descriptor_set->IsPushDescriptor()) {
+                // Can't cache validation of push_descriptors
```
jzulauf-lunarg: This is part of the Record path for BindDescriptors, so we also record that the validation path looked at this descriptor_set.

layers/core_validation.cpp
```diff
@@ -6048,63 +6135,108 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
+                                               VkPipelineBindPoint pipelineBindPoint, VkPipelineLayout layout, uint32_t firstSet,
+                                               uint32_t setCount, const VkDescriptorSet *pDescriptorSets,
+                                               uint32_t dynamicOffsetCount, const uint32_t *pDynamicOffsets) {
+    auto pipeline_layout = getPipelineLayout(device_data, layout);
+    std::vector<cvdescriptorset::DescriptorSet *> descriptor_sets;
+    descriptor_sets.reserve(setCount);
+
+    // Construct a list of the descriptors
+    bool found_non_null = false;
+    for (uint32_t i = 0; i < setCount; i++) {
```
jzulauf-lunarg: map from handles to pointers to use the refactored version above.

layers/core_validation.cpp
```diff
@@ -6301,15 +6434,15 @@ static bool PreCallValidateCmdPushDescriptorSetKHR(layer_data *device_data, GLOB
...
-    const auto &layout_state = getPipelineLayout(device_data, layout);
+    const auto &pipeline_layout = getPipelineLayout(device_data, layout);
+    if (!pipeline_layout) return;
     std::unique_ptr<cvdescriptorset::DescriptorSet> new_desc{
-        new cvdescriptorset::DescriptorSet(0, 0, layout_state->set_layouts[set], device_data)};
-    cb_state->lastBound[pipelineBindPoint].boundDescriptorSets[set] = new_desc.get();
+        new cvdescriptorset::DescriptorSet(0, 0, pipeline_layout->set_layouts[set], device_data)};
+
+    std::vector<cvdescriptorset::DescriptorSet *> descriptor_sets = {new_desc.get()};
+    UpdateLastBoundDescriptorSets(device_data, cb_state, pipelineBindPoint, pipeline_layout, set, 1, descriptor_sets, 0, nullptr);
```
jzulauf-lunarg: While the spec doesn't explicitly state that Pushing can disturb, it's hard to imagine any logical model that would look like the pipeline compatibility rules for which this would not be true.

Those that know better, please inform :)

jzulauf-lunarg:
I've addressed @mark-lunarg issue, the clang-format, and the push-descriptor issues from the first push.  Ready for detailed review.

layers/hash_util.h
daveh-lunarg: What happens if your doubles are not 80-bit?  Works (or degrades gracefully) on VS+Win10?  Android/ARM?

jzulauf-lunarg: To be more clear.  In order to compute the 64 bit magic number hard coded above, 64 bits of mantissa are needed and available with Intel extended precision.  If you tried to recreate the 64 bit magic number with IEEE 64 bit doubles you'd get 0x9e3779b97f4a7800 -- the same value with the last eleven bits zero'd -- roughly as you'd expect with a 52 bit mantissa.  (from the code I used to compute)
```
typedef long double L;
L phi64 = (L(1.) + std::sqrt(L(5.))) / L(2.);
L big64 = std::pow(L(2.), L(64.));
uint64_t d64 = double(big64) / double(phi64);
uint64_t g64 = big64 / phi64; // if long double == double, same as d64 above 
```

However, with the hard coded value present, no extended precision is needed to use the hash combiner.

layers/hash_util.h
```diff
@@ -0,0 +1,164 @@
+/* Copyright (c) 2015-2016 The Khronos Group Inc.
```
mark-lunarg: shader_validation already has a hash library, xxhash -- any way to leverage that here?  If this is impractical, what about removing that and using this one for shader validation?

jzulauf-lunarg: The only overlap is the "hash combination inmost function" -- the goals of the two interfaces are somewhat different.  The xxhash most certainly faster than the HashCombine/std::hash implementation, but is design around hashing contiguous *buffers*, not complex data types.  Having been bit in the recent past with pad-byte noise in hash digests, the "cast the struct to a char * the sizeof(struct ... )" approach isn't workable.

The xxhash utilization in shader validation could be replaced with the UID Dictionary, but I'm not sure the value of the change.

chrisforbes:
I'm wondering whether you need to also canonicalize the order of the bindings -- the spec says "identically defined", but does it mean identical creation structs, or an identical resulting layout?

Not finding much else to complain about in here; looks like a solid improvement.

layers/core_validation.cpp
```diff
@@ -5378,32 +5404,37 @@ struct PushConstantRangeCompare {
...
     PushConstantRanges ranges(sorted.size());
     for (const auto range : sorted) {
         ranges.emplace_back(*range);
     }
-    PushConstantRangesId from_input = std::make_shared<PushConstantRanges>(std::move(ranges));
+    return push_constant_ranges_dict.look_up(std::move(ranges));
+}
+
+// Dictionary of canoncial form of the pipeline set layout of descriptor set layouts
+static DescriptorSetLayoutLayoutDict layout_layout_dict;
```
tobine: This naming is difficult. These are the ds layouts that make up a pipeline layout, yes? Perhaps pipeline_ds_layouts_dict? Or pipe_layout_ds_layout_dict? Or layout_ds_layouts?

layers/descriptor_sets.h
```diff
@@ -53,11 +53,17 @@ struct IndexRange {
 typedef std::map<uint32_t, descriptor_req> BindingReqMap;
 
 /*
- * DescriptorSetLayout class
+ * DescriptorSetLayoutDef/DescriptorSetLayout classes
```
tobine: ok, cool, looks like you redid the comment here so ignore my earlier comment.

layers/core_validation.cpp
```diff
@@ -5413,10 +5444,22 @@ static void PostCallRecordCreatePipelineLayout(layer_data *dev_data, const VkPip
     PIPELINE_LAYOUT_NODE &plNode = dev_data->pipelineLayoutMap[*pPipelineLayout];
     plNode.layout = *pPipelineLayout;
     plNode.set_layouts.resize(pCreateInfo->setLayoutCount);
+    DescriptorSetLayoutLayoutDef layout_layout(pCreateInfo->setLayoutCount);
```
tobine: Wouldn't mind seeing all the "layout_layout" vars renamed as I find it a bit unclear.

layers/core_validation.cpp
```diff
@@ -5413,10 +5444,22 @@ static void PostCallRecordCreatePipelineLayout(layer_data *dev_data, const VkPip
...
+        layout_layout[i] = plNode.set_layouts[i]->get_layout_id();
     }
-    plNode.push_constant_ranges = get_definition_id(pCreateInfo);
+
+    // Get canonical form IDs for the "compatible for set" contents
+    plNode.push_constant_ranges = get_canonical_id(pCreateInfo);
+    auto ll_id = layout_layout_dict.look_up(layout_layout);
+    plNode.compat_for_set.reserve(pCreateInfo->setLayoutCount);
+
+    // Create table of "compatible for set N" cannonical forms for trivial accept validation
```
tobine: cannonical -> canonical

layers/core_validation.cpp
jzulauf-lunarg: Yeah, it's not great. I'll try to find something that is clearer but not of Faulkneresque verbosity.

chrisforbes:
@jzulauf-lunarg if we're short on CTS coverage for this area, can you file bugs so we get that fixed?

layers/core_validation.cpp
```diff
@@ -6119,63 +6119,128 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
+    // Some useful shorthand
+    auto &last_bound = cb_state->lastBound[pipeline_bind_point];
+
+    auto &bound_sets = last_bound.boundDescriptorSets;
+    auto &dynamic_offsets = last_bound.dynamicOffsets;
+    auto &bound_compat_ids = last_bound.compat_id_for_set;
+    auto &pipe_compat_ids = pipeline_layout->compat_for_set;
+
+    const uint32_t current_size = static_cast<uint32_t>(bound_sets.size());
+    assert(current_size == dynamic_offsets.size());
```
tobine: Do we set a dynamicOffset per bound descriptor? I believe we initially only had dynamicOffset per *dynamic* descriptor.
If this changed then need to also update the comment for dynamicOffsets in LAST_BOUND_STATE definition in core_validation_types.h which says:

> // one dynamic offset per dynamic descriptor bound to this CB

layers/core_validation.cpp
```diff
@@ -6119,63 +6119,128 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
+    // We resize if we need more set entries or if those past "last" are disturbed
+    if (required_size != current_size) {
+        // TODO: put these size tied things in a struct (touches many lines)
+        bound_sets.resize(required_size);
+        dynamic_offsets.resize(required_size);
+        bound_compat_ids.resize(required_size);
+    }
+
+    // For any previously bound sets, need to set them to "invalid" if they were disturbed by this update
+    for (uint32_t set_idx = 0; set_idx < first_set; ++set_idx) {
```
tobine: I need you to double-check this code. Relevant spec section:

`When binding a descriptor set (see Descriptor Set Binding) to set number N, if the previously bound descriptor sets for sets zero through N-1 were all bound using compatible pipeline layouts, then performing this binding does not disturb any of the lower numbered sets.`

I believe the if check below should be held constant for pipe_compat_ids[0]. If that matches all of the previous bound_compat_ids from 0 to right before first_set, then no need to disturb, otherwise all need to be disturbed.

Entirely possible I'm misunderstanding something here so lemme know what you think.

layers/core_validation.cpp
```diff
@@ -6119,63 +6119,128 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
-            if ((last_bound->boundDescriptorSets[set_idx + firstSet] != nullptr) &&
-                last_bound->boundDescriptorSets[set_idx + firstSet]->IsPushDescriptor()) {
-                last_bound->push_descriptor_set = nullptr;
-                last_bound->boundDescriptorSets[set_idx + firstSet] = nullptr;
-            }
+    // We resize if we need more set entries or if those past "last" are disturbed
+    if (required_size != current_size) {
+        // TODO: put these size tied things in a struct (touches many lines)
+        bound_sets.resize(required_size);
+        dynamic_offsets.resize(required_size);
```
tobine: I think this is new behavior from before in that dynamic_offsets is now as big as bound_sets regardless of type of descriptor. Please update comment referenced earlier.

https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/master/layers/core_validation_types.h#L653

layers/core_validation.cpp
```diff
@@ -6119,63 +6119,128 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
+            assert(ds == last_bound.push_descriptor_set.get());
+            last_bound.push_descriptor_set = nullptr;
+            return true;
+        }
+        return false;
+    };
 
-    if (last_set_index >= last_bound->boundDescriptorSets.size()) {
-        last_bound->boundDescriptorSets.resize(last_set_index + 1);
-        last_bound->dynamicOffsets.resize(last_set_index + 1);
```
jzulauf-lunarg: @tobine -- this is the matching dynamic offset array with the exact same resize.

layers/core_validation.cpp
```diff
@@ -6119,63 +6119,128 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
-            last_bound->boundDescriptorSets[set_idx + firstSet] = descriptor_set;
+        // Record binding (or push)
+        push_descriptor_cleanup(bound_sets[set_idx]);
+        bound_sets[set_idx] = descriptor_set;
+        bound_compat_ids[set_idx] = pipe_compat_ids[set_idx];  // compat ids are canonical *per* set index
 
+        if (descriptor_set) {
             auto set_dynamic_descriptor_count = descriptor_set->GetDynamicDescriptorCount();
-            last_bound->dynamicOffsets[firstSet + set_idx].clear();
-            if (set_dynamic_descriptor_count) {
```
jzulauf-lunarg: @tobine Here's where the per descriptor vector is created in the old code.  A std vector, empty if there's no dynamic descriptors, and filled if there are.

layers/core_validation.cpp
```diff
@@ -6119,63 +6119,128 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
 
-            last_bound->boundDescriptorSets[set_idx + firstSet] = descriptor_set;
+        // Record binding (or push)
+        push_descriptor_cleanup(bound_sets[set_idx]);
+        bound_sets[set_idx] = descriptor_set;
+        bound_compat_ids[set_idx] = pipe_compat_ids[set_idx];  // compat ids are canonical *per* set index
 
+        if (descriptor_set) {
             auto set_dynamic_descriptor_count = descriptor_set->GetDynamicDescriptorCount();
-            last_bound->dynamicOffsets[firstSet + set_idx].clear();
```
jzulauf-lunarg: there's always an entry for dynamicOffset for each set, it's just empty if the set isn't dynamic.

layers/core_validation.cpp
```diff
@@ -6119,63 +6119,128 @@ VKAPI_ATTR void VKAPI_CALL CmdSetStencilReference(VkCommandBuffer commandBuffer,
...
-        // For any previously bound sets, need to set them to "invalid" if they were disturbed by this update
-        if (firstSet > 0) {
-            for (uint32_t i = 0; i < firstSet; ++i) {
-                if (last_bound->boundDescriptorSets[i] &&
-                    !verify_set_layout_compatibility(last_bound->boundDescriptorSets[i], pipeline_layout, i, error_string)) {
-                    last_bound->boundDescriptorSets[i] = VK_NULL_HANDLE;
-                }
+            // TODO: Add logic for tracking push_descriptor offsets (here or in caller)
+            if (set_dynamic_descriptor_count && input_dynamic_offsets) {
+                const uint32_t *end_offset = input_dynamic_offsets + set_dynamic_descriptor_count;
```
jzulauf-lunarg: @tobine Here's the parallel logic for populating  the dynamic_offsets[set_idx] iff the set is dynamic, else clear it.

layers/core_validation.cpp
jzulauf-lunarg: ^^^ (to include in the set of "review comments)

layers/core_validation.cpp
jzulauf-lunarg: What's confusing is that while we do have an offset per dynamic descriptor, those offsets are organized in a vector of vectors -- the outer vector is indexed per set, the inner per dynamic descriptor within a set (or empty if none are present.)

layers/core_validation.cpp
jzulauf-lunarg: First one point I believe must be assert.  If for two pipeline layouts i and j, if compat(pipe_i, k) == compat(pipe_j, k), then compat(pipe_i, l) == compat(pipe_k, l) for all values l < k. Compatibility is cumulative.

The question in my mind is what state last bound states  compat_id's could be in at any given change.  When I change set[n], if it's incompatible, all sets above are cleared, (effectively set to null compat), and all sets < n are (by this logic) forced to be compatible with the binding pipeline_layout (and cleared if not).

Given that we start with a set vector that is effective an infinite null compat vector, the first bind leaves all entries up to N compat with p1 (the pipeline of the first bind).

Now, say we set N - m with p2 which is compat for (N-m) (and thus for all j < N-m), Sets < N-m are undisturbed and sets > N-M are undisturbed -- thus the last_bound compat ids are still the set for p1 (up to N) (since none where disturbed).

Now bind at k, N > k > N-m with p3 s.t. with is not compat for k but is compat for N-m, to me this means that [0..N-m] should be undisturbed, and for j > N-m, the should be disturbed only starting at the boundary where compat(p3, j) != compat(p1, j), and also everything above k is disturbed.

To me this says that an entry by entry comparison should work for the low set, though is perhaps overkill.  I could look entry by entry for the first incompat, and then blanket clear from there, but I can't see how I could end up with a different result.

If the "look for first incompat", then clear makes you more comfortable and seems clearer, I'm more than happy to change it.

jzulauf-lunarg: Or... I could go the other way.  If the last index is compatible, then we know all indices below *must* be compatible (since we always update on incompatibility) and not compare at all.  If the last index is incompatible, then do a bisection search to find the breakage and clear from there.

tobine: I agree with all of your analysis and statements. I think there was a typo in your second sentence where "compat(pipe_i, l) == compat(pipe_k, l)" is actually "compat(pipe_i, l) == compat(pipe_j, l)" but it made sense to me.

After going over the spec some more and conferring with @chrisforbes I think this check is good as it stands.

PR #2425, opened 2018-02-21
Reorganize and update BUILD.md

mark-lunarg:
Second hack at simplifying and organizing the BUILD.md file.  

Attempted to incorporate the excellent feedback from @karl-lunarg, @mikes-lunarg, @cnorthrop, and @krOoze.  Also did a reformatting/spelling pass.  Please have another quick look gentlemen.

Also, neat trick, while looking at the diff, click the *View* button and you can see the process markdown.


BUILD.md
```diff
@@ -1,79 +1,186 @@
...
 
-This repository does not contain a Vulkan-capable driver.
-Before proceeding, it is strongly recommended that you obtain a Vulkan driver from your graphics hardware vendor
-and install it.
+# Index
+1. [Contributing](#contributing)
+2. [Repository Set-Up](#repo-set-up)
+3. [Windows Build](#windows-build)
+4. [Linux Build](#linux-build)
+5. [Android Build](#android-build)
```
cnorthrop: Can this point to #building-on-android instead?  That's where the Android steps start.  Alternatively, you can rename the links below for uniformity.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+  - Select to install the optional sub-package to add Python to the system PATH environment variable.
+  - Ensure the `pip` module is installed (it should be by default)
+  - Python3.3 or later is necessary for the Windows py.exe launcher that is used to select python3 rather than python2 if both are
+installed
+- [Git](http://git-scm.com/download/win)
+  - Tell the installer to allow it to be used for "Developer Prompt" as well as "Git Bash".
+  - Tell the installer to treat line endings "as is" (i.e. both DOS and Unix-style line endings).
+  - Install each a 32-bit and a 64-bit version, as the 64-bit installer does not install the 32-bit libraries and tools.
+- Notes for using [Cygwin](https://www.cygwin.com)
+  - The tools specified above **must** precede Cygwin in your path
```
mikes-lunarg: Does this make sense?   I thought cygwin had its own shell and path

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+For example, for VS 2017 (Generators for other versions are [specified here](#win-cmake-generators)):
+```
+  cmake -G "Visual Studio 15 Win64" ..
+```
+   This will create a Windows solution file named 'VULKAN.sln' in the build directory.
+
+Launch Visual Studio and open the "VULKAN.sln" solution file in the build folder.  You may select "Debug" or "Release" from the
+Solution Configurations drop-down list. Start a build by selecting the Build->Build Solution menu item.
+
+Vulkan programs must be able to find and use the vulkan-1.dll library. Make sure it is either installed in the C:\Windows\System32
```
karl-lunarg: This section is talking about working with this VULKAN.sln package.  I believe that this solution contains post-build steps to copy the vulkan loader into the same directory as the various demos (cube, vulkaninfo) to get around the need to have the loader in System32 or on the PATH.  So the reader may think that they have to take care of this themselves and may be surprised to see that they don't have to.  In addition, many IHV driver packages install the vulkan loader to System32, so the reader may be surprised to find it there already.

It would probably be better to replace this text with something that explains what is actually happening in the solution.  If the user runs the debugger on one of the programs and sees the list of loaded modules, they'll see something different from what this current text is describing.  Could say something like "This solution copies the loader it built to each program's build directory to ensure that the program uses the loader built from this solution."

This current paragraph is still useful and could be moved to another section "Notes for developing your own apps" and mention that a (slightly older) loader may already be installed on the system with the driver and that putting their own loader in the PATH or the same dir as the app are alternatives.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+folder, or the PATH environment variable includes the folder that it is located in.
+
+To run Vulkan programs you must tell the Vulkan Loader where to find the libraries. This is described in a
```
karl-lunarg: Again, we're talking specifically about the VULKAN.sln here and I believe that some projects are configured to set VK_LAYER_PATH to ../layers or somesuch so that you don't have to worry about this.  It might be better to state that the solution is set up this way, rather than making the user guess at what they need to do.

This current paragraph can also be moved to a new "Notes" section.  It isn't really about the VULKAN.sln.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+
+| Build Platform               | 64-bit Generator | 32-bit Generator |
+|------------------------------|--------------------------|--------------------|
+| Microsoft Visual Studio 2013 | "Visual Studio 12 Win64" | "Visual Studio 12" |
+| Microsoft Visual Studio 2015 | "Visual Studio 14 Win64" | "Visual Studio 14" |
+| Microsoft Visual Studio 2017 | "Visual Studio 15 Win64" | "Visual Studio 15" |
 
-The build process uses CMake to generate makefiles for this project.
-The build generates the loader, layers, and tests.
+
```
karl-lunarg: This would be a good place to move the paragraphs I mentioned above.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
 This repo has been built and tested on the two most recent Ubuntu LTS versions. Currently, the oldest supported
 version is Ubuntu 14.04, meaning that the minimum supported compiler versions are GCC 4.8.2 and Clang 3.4, although
 earlier versions may work.
 It should be straightforward to adapt this repository to other Linux distributions.
 
-
-These packages are needed to build this repository:
+**Required Package List:**
 ```
 sudo apt-get install git cmake build-essential bison libx11-xcb-dev libxkbcommon-dev libmirclient-dev libwayland-dev libxrandr-dev
```
karl-lunarg: I'd be surprised if bison is still needed.  If you have time, maybe try uninstalling bison and see if you can still build.  I think we needed it when we had the old ILO driver in the repos.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+Example debug build (Note that the update\_external\_sources script used below builds external tools into predefined locations.
+See **Loader and Validation Layer Dependencies** for more information and other options):
+
+1. In a Linux terminal, cd Vulkan-LoaderAndValidationLayers -- the root of the cloned git repository
+2. Execute './update_external_sources.sh' -- this will download and build external components
+3. Create a 'build' directory, change into that directory, and run cmake:
 ```
-cd Vulkan-LoaderAndValidationLayers  # cd to the root of the cloned git repository
-./update_external_sources.sh
 cmake -H. -Bdbuild -DCMAKE_BUILD_TYPE=Debug
```
karl-lunarg: People have complained that the -H option is deprecated in cmake.  Could you try the 
```
mkdir dbuild
cd dbuild
cmake -DCMAKE_BUILD_TYPE=Debug ..
```
pattern to see if it works and change it if so?  This would be more consistent with the Windows directions.

BUILD.md
```diff
@@ -88,165 +195,114 @@ This command installs files to:
 
 You may need to run `ldconfig` in order to refresh the system loader search cache on some Linux systems.
 
-The list of installed files appears in the build directory in a file named `install_manifest.txt`.
-You can easily remove the installed files with:
+The list of installed files appears in the build directory in a file named `install_manifest.txt`. You can easily remove the
```
karl-lunarg: We added support for `sudo make uninstall`.  It is probably better to mention that here instead of the following command.   (Probably my bad for not changing this.)

Please start new sentences on their own line (Markdown style).

BUILD.md
```diff
@@ -88,165 +195,114 @@ This command installs files to:
...
+After making any changes to the repository, you should perform some quick sanity tests, including the run_all_tests shell script
+and the cube demo with validation enabled.
 
+To run the **validation test script**, in a terminal change to the build/tests directory and run:
 ```
-export ASFLAGS=--32
-export CFLAGS=-m32
-export CXXFLAGS=-m32
-export PKG_CONFIG_LIBDIR=/usr/lib/i386-linux-gnu
+    VK_LAYER_PATH=/home/user/Vulkan-LoaderAndValidationLayers/build/layers ./run_all_tests.sh
```
karl-lunarg: "/home/user" has not been explained yet, even though it might be obvious.
Since you are telling the user to switch to the build/tests dir, it may be more compact tell them to set it to `../layers`.  (make sure it works first).
Same comment for the Cube demo below.

BUILD.md
mark-lunarg: @cnorthrop, I intentionally indexed all of these entries to the beginning of their respective sections.  Are you suggesting changing the names of the index table from [Linux Build] to [Linux Instructions] or [Linux Section] for example?

cnorthrop: The spot you are jumping to skips some important steps for the Android build.  If you click it, you have to scroll up to pick up the exports.  Maybe those can just be moved to the section you've linked to.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+  - Tell the installer to treat line endings "as is" (i.e. both DOS and Unix-style line endings).
+  - Install each a 32-bit and a 64-bit version, as the 64-bit installer does not install the 32-bit libraries and tools.
+- Notes for using [Cygwin](https://www.cygwin.com)
+  - The tools specified above **must** precede Cygwin in your path
+
+### Windows Build - Microsoft Visual Studio
+
+1. Open a Developer Command Prompt for VS201x
+2. Cd Vulkan-LoaderAndValidationLayers -- the root of the cloned git repository
+3. Run 'update_external_sources.bat' -- this will download and build external components
```
krOoze: `&mdash;` should result in pretty-printed em-dash.

This can take a long time and takes GBs. Might be worth to mention the `--32` and `--64` switches.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+4. Create a 'build' directory, change into that directory, and run cmake
```
krOoze: Wrong tick marks, if you intended to format it as `build`.
"cmake" could also be formatted.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+
+For example, for VS 2017 (Generators for other versions are [specified here](#win-cmake-generators)):
```
krOoze: lowercase G?
"VS2017" is usually written without space.

Might be easier to instruct to use `cmake -h` to get available generators?

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+```
+  cmake -G "Visual Studio 15 Win64" ..
```
krOoze: The proper string acording to `cmake -h` is actually "Visual Studio 15 2017 Win64".

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+```
+   This will create a Windows solution file named 'VULKAN.sln' in the build directory.
+
+Launch Visual Studio and open the "VULKAN.sln" solution file in the build folder.  You may select "Debug" or "Release" from the
```
krOoze: Might want to pretty print `VULKAN.sln`, `build`, `Debug` and `Release`.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+Solution Configurations drop-down list. Start a build by selecting the Build->Build Solution menu item.
```
krOoze: Might want to pretty-print arrows with `&rarr;`.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+
+Vulkan programs must be able to find and use the vulkan-1.dll library. Make sure it is either installed in the C:\Windows\System32
+folder, or the PATH environment variable includes the folder that it is located in.
+
+To run Vulkan programs you must tell the Vulkan Loader where to find the libraries. This is described in a
+`LoaderAndLayerInterface` document in the `loader` folder in this repository. This describes both how ICDs and layers should
+be properly packaged, and how developers can point to ICDs and layers within their builds.
```
krOoze: I am not sure what these two paragraphs try to tell. Is it about how to run arbitrary programs against built loader or layers?

Interface doc is bit tl;dr. Might want to link to specific section, and perhaps tease the specific environment variables (e.g. `VK_LAYER_PATH`)

Might want to pretty-print `vulkan-1.dll`, `C:\Windows\System32` and `PATH`.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+
+### Windows Tests and Demos
+
+After making any changes to the repository, you should perform some quick sanity tests, including the run_all_tests powershell
+script and the cube demo with validation enabled.
```
krOoze: Might want to: `run_all_tests.ps1` and `cube`.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+        .\run_all_tests.ps1
+```
+For Debug builds:
+```
+        .\run_all_tests.ps1 -debug
+```
+
+This script will run the following tests:
+- vk\_loader\_validation\_tests: Tests Vulkan Loader handle wrapping
+- vk\_layer\_validation\_tests: Test Vulkan validation layers
```
krOoze: Might want to `vk_loader_validation_tests`. Should not need to escape inside the \` \`.

It runs only loader tests for me, and **not** layers. Might want to remove the exit code check from the script (replace by keypress between tests perhaps)...

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+```
+        .\run_all_tests.ps1
+```
+For Debug builds:
+```
+        .\run_all_tests.ps1 -debug
+```
+
+This script will run the following tests:
+- vk\_loader\_validation\_tests: Tests Vulkan Loader handle wrapping
```
krOoze: Wait, what handle wrapping?

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+- vk\_layer\_validation\_tests: Test Vulkan validation layers
+- vkvalidatelayerdoc: Tests that validation database is in up-to-date and in synchronization with the validation source code
```
krOoze: "in up-to-date"
"in synchronization"

-> up-to-date and synchronized

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+
+To run the Cube demo with validation in a Debug build configuration:
+- In the MSVC solution explorer, right-click on the 'cube' project and select 'Set As Startup Project'
+- Right click on cube again, select properties->Debugging->Command Arguments, change to '--validate', and save
```
krOoze: Should perhaps set `--validate` as default for Debug configuration.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+- From the main menu, select Debug->Start Debugging, or from the toolbar click 'Local Windows Debugger'
+
+Other demos that can be found in the dbuild/demos directory are:
```
krOoze: Just "build" was used above.
"dbuild" is perhaps specific for linux.

BUILD.md
```diff
@@ -1,79 +1,186 @@
...
+
+
+### Windows Notes
+
+<a name="win-cmake-generators"/>
+
+Cmake Visual Studio Generators:
+
+The above example used Visual Studio 2017, and specified its generator as "Visual Studio 15 Win64".
+The chosen generator shoudl match your Visual Studio Version. Popular Visual Studio generators include:
```
krOoze: Version -> version

It is not as much "popular", but more like those three being the only supported compilers by this repo.

BUILD.md
mark-lunarg: No, it doesn't make sense!  But @jeffbolznv is helping with his instructions.

BUILD.md
mark-lunarg: The loader has to wrap dispatchable handles.  The loader exercises that functionality.

BUILD.md
```diff
@@ -1,84 +1,205 @@
...
+### Windows Build - Microsoft Visual Studio
+
+1. Open a Developer Command Prompt for VS201x
+2. Change directory to `Vulkan-LoaderAndValidationLayers` -- the root of the cloned git repository
+3. Run `update_external_sources.bat` -- this will download and build external components
+4. Create a `build` directory, change into that directory, and run cmake
+
+For example, for VS2017 (generators for other versions are [specified here](#win-cmake-generators)):
+```
+  cmake -G "Visual Studio 15 Win64" ..
```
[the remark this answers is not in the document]
mark-lunarg: This works, dunno why, but you're right and I'll change them to match.

BUILD.md
```diff
@@ -1,84 +1,205 @@
...
+```
+        .\run_all_tests.ps1
+```
+For Debug builds:
+```
+        .\run_all_tests.ps1 -debug
+```
+
+This script will run the following tests:
+- `vk\_loader_validation_tests`:  Tests Vulkan Loader handle wrapping
```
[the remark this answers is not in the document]
mark-lunarg: Nice! And, I think so, but it's A Big Thing.

BUILD.md
```diff
@@ -1,84 +1,205 @@
...
 
-This repository does not contain a Vulkan-capable driver.
-Before proceeding, it is strongly recommended that you obtain a Vulkan driver from your graphics hardware vendor
-and install it.
+# Index
+1. [Contributing](#contributing)
+2. [Repository Set-Up](#repo-set-up)
+3. [Windows Build](#windows-build)
+4. [Linux Build](#linux-build)
+5. [Android Build](#android-build)
```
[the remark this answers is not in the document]
mark-lunarg: Nah, but I'll take another look at where it falls in the doc.

BUILD.md
```diff
@@ -1,84 +1,205 @@
...
+
+To run the validation test script, open a Powershell Console, change to the build/tests directory, and run:
+
+For Release builds:
+```
+        .\run_all_tests.ps1
+```
+For Debug builds:
+```
+        .\run_all_tests.ps1 -debug
```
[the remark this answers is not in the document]
mark-lunarg: No, but I'll make it match.  Thanks!

BUILD.md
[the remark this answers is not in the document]
krOoze: E.g. there are some alloc tests. Or layer loading tests. Could they be considered tests of "handle wrapping"?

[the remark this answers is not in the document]
mark-lunarg: Ah!  I see, I misunderstood your comments.  Will update the description.  Thanks a ton for reviewing this...

karl-lunarg:
I fixed a couplahundred markdown link gripes, reflowed text to remove insanely long lines, fixed a table.  Hopefully didn't change much actual content.

PR #2427, opened 2018-02-21
demos: Change logo absolute path to relative path

### LINKED ISSUE #2424: vulkaninfo: For HTML output, there is an image ref to a local filesystem.
```c
fprintf(out, "\t\t\t<img src='C:/Git/VulkanTools/layersvt/images/lunarg.png' />\n");
```

shannon-lunarg:
Fixes #2424

### DIFF
--- a/demos/vulkaninfo.c
+++ b/demos/vulkaninfo.c
@@ -674,7 +674,6 @@ void PrintHtmlHeader(FILE *out) {
     fprintf(out, "\t</head>\n");
     fprintf(out, "\t<body>\n");
     fprintf(out, "\t\t<div id='header'>\n");
-    fprintf(out, "\t\t\t<img src='C:/Git/VulkanTools/layersvt/images/lunarg.png' />\n");
     fprintf(out, "\t\t\t<h1>Vulkan Info</h1>\n");
     fprintf(out, "\t\t</div>\n");
     fprintf(out, "\t\t<div id='wrapper'>\n");


mark-lunarg:
VIA has a couple of examples of this, but I'd agree with Karl and just remove it, at least for now.

shannon-lunarg:
@karl-lunarg My thought process was that if the user put both repos in the recommended places (some overarching development directory) the logo would appear and if the configuration varied or if  VulkanTools didn't exist on the system the logo simply wouldn't show up. This is the case for Chrome but now that I've switched to viewing this on Firefox, I can see the problem..

I'd prefer a permanent URL to the logo (it looks nicer) but I can just remove the line for now.

Note: This line was copied and pasted from the api dump html template which suffers from the same erroneous absolute file path.

karl-lunarg:
Yes, but many users are going to be using vulkaninfo out of an SDK where there are no GitHub repos checked out.   In addition, it is possible that a user might run vulkaninfo to generate an HTML file and then send it via email to someone to look at.  The recipient isn't likely to have the repos and isn't likely to download the attachment into the exact location in the repo that would make this relative path work.

I think that removing the reference is the best solution.  If we decide to host the image file on a server someplace, then we could put it back.  But I don't think it is worth it.

I'll open an issue on VulkanTools to address the problem there.

Thanks for fixing this!

shannon-lunarg:
@karl-lunarg You make a lot of great points. Thanks for all the info!

shannon-lunarg:
Pushed to master

PR #2432, opened 2018-02-23
layers: Fix miscounting of descriptors for multiple stages

### LINKED ISSUE #2431: Validation counting descriptors too often against limits
At the moment in `core_validation.cpp`, the `maxDescriptorSet*` limits are counted in such a way that all shader stages which are enabled on the device and visible count for 1 against the limit, on each descriptor. So in the worst case a single descriptor can count 6 times against the limit. This was [added here](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/commit/44bf63f14c4a033f02bbaf7228d956e39e4d0f21#diff-617ec4ceb169e2ba63e3b5ba46ef3647R4913).

Chatting to @TobiasHector this is not the intention of the limit, instead it's about counting the number of descriptors total in a set, with each descriptor counting once, regardless of shader visibility - in contrast to the `maxPerStageDescriptor*` limits which are about how many descriptors can be visible to any one stage and so again each descriptor counts once.

chrisforbes:
Fixes #2431 -- We hit this issue independently yesterday. 

Descriptors are not intended to count multiple times against the
maxDescriptorSet* limits if they are accessible from multiple stages.

Required adjusting various subtests in CreatePipelineLayout*.

Also fixed some related issues in those tests:
- Input attachments are ONLY accessible to the fragment stage.
- Various assumptions about divisibility of the maxDescriptorSet* limits

There are still many robustness issues in these tests -- it appears we
don't hit them with real implementations, but still wrong.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,13 +96,11 @@ script:
       export VK_LAYER_PATH=external/VulkanTools/build/layersvt:dbuild/layers
       export VK_ICD_FILENAMES=dbuild/icd/VkICD_mock_icd.json
       dbuild/tests/vk_layer_validation_tests
-      VK_DEVSIM_FILENAME=tests/device_profiles/adreno_540.json dbuild/tests/vk_layer_validation_tests --devsim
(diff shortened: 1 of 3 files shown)


chrisforbes:
Sigh, CI is unhappy.

PR #2434, opened 2018-02-23
Fix 2408: record 3D image-slice layouts as subresources

### LINKED ISSUE #1650: core_validation: Possible bug when validation image layout state when using 2d view of a 3d image
Hi,

as with the maintenance1 extension, it is possible to use a slice of a 3d image as a 2d (array) view.
And if you try to use as 2d image view of a 3d image slice, the the lookup for slice 1 >= always returns VK_IMAGE_LAYOUT_UNDEFINED and the layout validation reports an error even if the complete 3d image is in the needed layout.

### LINKED ISSUE #2210: Validation layer reports image layout error when rendering into a slice of a 3D image
If I create a 3D image with `VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT_KHR` and transition it from `VK_IMAGE_LAYOUT_UNDEFINED` to `VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL` setting `VkImageSubresourceRange::layerCount = 1` in `VkImageMemoryBarrier` the validation layer will complain when rendering into a view targetting any other slice than 0 of the 3D image:
```
Cannot submit cmd buffer using image (0x33b) [sub-resource: aspectMask 0x1 array layer 1, mip level 0], with layout VK_IMAGE_LAYOUT_UNDEFINED when first use is VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL.
```
I suspect that this might be a false positive in the validation layer.

Note that I can get rid of this error message by setting `VkImageSubresourceRange::layerCount = depth` in `VkImageMemoryBarrier` but then the validation layer complains with:
```
vkCmdPipelineBarrier(): pImageMemoryBarriers[0].subresourceRange.baseArrayLayer + .layerCount (= 0 + 32 = 32) is greater than the arrayLayers of the image when it was created (i.e. greater than 1). The spec valid usage text states 'If subresourceRange::layerCount is not VK_REMAINING_ARRAY_LAYERS, subresourceRange::layerCount must be non-zero and subresourceRange::baseArrayLayer + subresourceRange::layerCount must be less than or equal to the arrayLayers specified in VkImageCreateInfo when image was created' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkImageMemoryBarrier-subresourceRange-01489)
```
But AFAIK the image barrier should always set `VkImageSubresourceRange::layerCount = 1` for 3D images.

### LINKED ISSUE #2408: Potential bug with 3d images/subresource layouts
I'm getting what I'm pretty sure are buggy validation errors when using a 3D texture aliased as 2d arrays. Attached is the API dump file, and in the companion log file, you can see the validation errors. This is with regards to 3d textures 0x961, 0x96c, 0x966, 0x96a, 0x963, etc at the bottom of the log file 
_(eg *** [ERROR:DS6] Obj 0x00000123314B0C00 Loc 2733 Cannot submit cmd buffer using image (0x966) [sub-resource: aspectMask 0x1 array layer 38, mip level 0], with layout VK_IMAGE_LAYOUT_UNDEFINED when first use is VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL.._

Basically there's no way to change the layout on subslices of a 3d texture, and I had existing layout changes for the whole texture, so why is the validation complaining about those?

[vk_apidump2.zip](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/files/1721433/vk_apidump2.zip)

daveh-lunarg:
Fixes a bug in image layout state handling that appeared with
khr_maintenance1. When a 3D image which was created with the
VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT_KHR flag is transitioned
between layouts, we now also record a transition for each depth
slice, treating it as an array layer subresource.

The actual fix is just two lines within TransitionImageLayouts().
Also included are some debug-enabling refactoring to distinguish
layout state updates vs additions, and support for new aspect
mask bits.

Fixes #2408 
Fixes #2210 
Fixes #1650 

Change-Id: Ibb5131a6dba8fcdeafe5e6d94af2decf751842ec

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -34,8 +34,9 @@
 #include "buffer_validation.h"
 
 void SetLayout(layer_data *device_data, GLOBAL_CB_NODE *pCB, ImageSubresourcePair imgpair, const VkImageLayout &layout) {
-    if (pCB->imageLayoutMap.find(imgpair) != pCB->imageLayoutMap.end()) {
-        pCB->imageLayoutMap[imgpair].layout = layout;
+    auto it = pCB->imageLayoutMap.find(imgpair);
+    if (it != pCB->imageLayoutMap.end()) {
+        it->second.layout = layout;
     } else {
         assert(imgpair.hasSubresource);
         IMAGE_CMD_BUF_LAYOUT_NODE node;
@@ -52,6 +53,11 @@ void SetLayout(layer_data *device_data, OBJECT *pObject, VkImage image, VkImageS
     SetLayout(device_data, pObject, imgpair, layout, VK_IMAGE_ASPECT_DEPTH_BIT);
     SetLayout(device_data, pObject, imgpair, layout, VK_IMAGE_ASPECT_STENCIL_BIT);
     SetLayout(device_data, pObject, imgpair, layout, VK_IMAGE_ASPECT_METADATA_BIT);
+    if (GetDeviceExtensions(device_data)->vk_khr_sampler_ycbcr_conversion) {
+        SetLayout(device_data, pObject, imgpair, layout, VK_IMAGE_ASPECT_PLANE_0_BIT_KHR);
+        SetLayout(device_data, pObject, imgpair, layout, VK_IMAGE_ASPECT_PLANE_1_BIT_KHR);
+        SetLayout(device_data, pObject, imgpair, layout, VK_IMAGE_ASPECT_PLANE_2_BIT_KHR);
+    }
 }
 
 template <class OBJECT, class LAYOUT>
@@ -66,7 +72,12 @@ void SetLayout(layer_data *device_data, OBJECT *pObject, ImageSubresourcePair im
 // Set the layout in supplied map
 void SetLayout(std::unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE> &imageLayoutMap, ImageSubresourcePair imgpair,
                VkImageLayout layout) {
-    imageLayoutMap[imgpair].layout = layout;
+    auto it = imageLayoutMap.find(imgpair);
+    if (it != imageLayoutMap.end()) {
+        it->second.layout = layout;  // Update
+    } else {
+        imageLayoutMap[imgpair].layout = layout;  // Insert
+    }
 }
 
 bool FindLayoutVerifyNode(layer_data const *device_data, GLOBAL_CB_NODE const *pCB, ImageSubresourcePair imgpair,
@@ -133,6 +144,11 @@ bool FindCmdBufLayout(layer_data const *device_data, GLOBAL_CB_NODE const *pCB, 
     FindLayoutVerifyNode(device_data, pCB, imgpair, node, VK_IMAGE_ASPECT_DEPTH_BIT);
     FindLayoutVerifyNode(device_data, pCB, imgpair, node, VK_IMAGE_ASPECT_STENCIL_BIT);
     FindLayoutVerifyNode(device_data, pCB, imgpair, node, VK_IMAGE_ASPECT_METADATA_BIT);
+    if (GetDeviceExtensions(device_data)->vk_khr_sampler_ycbcr_conversion) {
+        FindLayoutVerifyNode(device_data, pCB, imgpair, node, VK_IMAGE_ASPECT_PLANE_0_BIT_KHR);
+        FindLayoutVerifyNode(device_data, pCB, imgpair, node, VK_IMAGE_ASPECT_PLANE_1_BIT_KHR);
+        FindLayoutVerifyNode(device_data, pCB, imgpair, node, VK_IMAGE_ASPECT_PLANE_2_BIT_KHR);
+    }
     if (node.layout == VK_IMAGE_LAYOUT_MAX_ENUM) {
         imgpair = {image, false, VkImageSubresource()};
         auto imgsubIt = pCB->imageLayoutMap.find(imgpair);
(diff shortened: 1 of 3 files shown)


layers/buffer_validation.cpp
```diff
@@ -66,7 +72,11 @@ void SetLayout(layer_data *device_data, OBJECT *pObject, ImageSubresourcePair im
 // Set the layout in supplied map
 void SetLayout(std::unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE> &imageLayoutMap, ImageSubresourcePair imgpair,
                VkImageLayout layout) {
-    imageLayoutMap[imgpair].layout = layout;
+    auto it = imageLayoutMap.find(imgpair);
+    if (it != imageLayoutMap.end())
```
mark-lunarg: Can you afford a few braces?  I have some to spare.

layers/buffer_validation.cpp
```diff
@@ -222,7 +253,11 @@ void SetGlobalLayout(layer_data *device_data, ImageSubresourcePair imgpair, cons
 
 // Set the layout on the cmdbuf level
 void SetLayout(layer_data *device_data, GLOBAL_CB_NODE *pCB, ImageSubresourcePair imgpair, const IMAGE_CMD_BUF_LAYOUT_NODE &node) {
-    pCB->imageLayoutMap[imgpair] = node;
+    auto it = pCB->imageLayoutMap.find(imgpair);
```
mark-lunarg: Again, por favor?

layers/buffer_validation.cpp
daveh-lunarg: I thought clang-format would take care of it ;)

layers/buffer_validation.cpp
```diff
@@ -66,7 +72,12 @@ void SetLayout(layer_data *device_data, OBJECT *pObject, ImageSubresourcePair im
 // Set the layout in supplied map
 void SetLayout(std::unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE> &imageLayoutMap, ImageSubresourcePair imgpair,
                VkImageLayout layout) {
-    imageLayoutMap[imgpair].layout = layout;
```
tobine: Is the new code changing behavior? Seems like the ultimate effect is the same. What am I missing?

layers/core_validation.cpp
```diff
@@ -2815,6 +2815,7 @@ static void PostCallRecordQueueSubmit(layer_data *dev_data, VkQueue queue, uint3
                 cbs.push_back(submit->pCommandBuffers[i]);
                 for (auto secondaryCmdBuffer : cb_node->linkedCommandBuffers) {
                     cbs.push_back(secondaryCmdBuffer->commandBuffer);
+                    UpdateCmdBufImageLayouts(dev_data, secondaryCmdBuffer);
```
tobine: Good. We should also be able to hoist the "incrementResources()" call in the loop below into this loop instead and then just kill the second loop.

layers/buffer_validation.cpp
```diff
@@ -222,7 +255,12 @@ void SetGlobalLayout(layer_data *device_data, ImageSubresourcePair imgpair, cons
 
 // Set the layout on the cmdbuf level
 void SetLayout(layer_data *device_data, GLOBAL_CB_NODE *pCB, ImageSubresourcePair imgpair, const IMAGE_CMD_BUF_LAYOUT_NODE &node) {
-    pCB->imageLayoutMap[imgpair] = node;
```
tobine: Same ? as above. Is behavior changing here?

layers/buffer_validation.cpp
```diff
@@ -212,7 +239,13 @@ bool FindLayout(const std::unordered_map<ImageSubresourcePair, IMAGE_LAYOUT_NODE
 // Set the layout on the global level
 void SetGlobalLayout(layer_data *device_data, ImageSubresourcePair imgpair, const VkImageLayout &layout) {
     VkImage &image = imgpair.image;
-    (*core_validation::GetImageLayoutMap(device_data))[imgpair].layout = layout;
```
tobine: Is this changing behavior?

layers/buffer_validation.cpp
daveh-lunarg: No change in behavior, this is an example of _"Also included are some debug-enabling refactoring to distinguish layout state updates vs additions"_, where for debugging purposes it's necessary to distinguish the update/insert cases. 

It adds a couple lines of code but shouldn't be any meaningful performance impact.

layers/core_validation.cpp
daveh-lunarg: Maybe, as long as incrementResources() has no side effects or interaction with UpdateCmdBufImageLayouts(), but I'd tend to assume it was segregated like that intentionally. This stuff is complex enough I don't want to make that assumption (but happy to review your PR ;)).

layers/buffer_validation.cpp
daveh-lunarg: Same

layers/buffer_validation.cpp
daveh-lunarg: Same

daveh-lunarg: I missed the VerifyAspectsPresent() useage - I'll update that.

And re:  _"layout code overall should be re-done. Too complicated"_ Amen!

daveh-lunarg: @tobine , re: _"Checking spec seems that there are various new VUs related to those bits. If we don't have a follow-on issue to implement those checks can you add one for tracking?"_ 

We sort of have standing task to implement VUs not yet implemented - do you think we really need a GitHub issue to track this set?

daveh-lunarg:
Updated to include VerifyAspectsPresent() check of multi-plane bits.

Note that this PR is primarily to close the listed issues, plus some opportunistic changes to aid debugging and head off problems with aspect bits that now include multi-plane image support. There are still unresolved corner cases with depth slice vs array layer, which I'll work on resolving. I agree with Tobin that a full overhaul could be beneficial, but that probably won't percolate to the top of my list soon.

layers/buffer_validation.cpp
tobine: Yes, my thought was give it a specific issue to hoist priority. My general thought is that its better to hoist priority of newer VUs over past VUs that haven't yet been done and don't have specific VUs. This is because older stuff w/o specific issues may not be causing many problems for users, but newer stuff is unknown so better to get checks in ASAP in order to help early-adopters.

PR #2446, opened 2018-02-27
Add validusage.json to LVL repo

mark-lunarg:
In the future, this file will be updated as part of the Vulkan header update process, and will be used as part of managing VUID identifiers and lifetimes.

@mikes-lunarg, I'm thinking we should modify spec.py to use this file **by default**?


mikes-lunarg:
@mark-lunarg That would be my preference.  Right now spec.py goes out and gets the latest validusage.json from the web which is fine for public header updates but becomes awkward in other scenarios.

- [ ] TODO for updating the wiki when this gets merged

PR #2451, opened 2018-02-28
Guard new-ish formats in an #ifdef to not break older vulkan.h such as Android uses.

davidhubbard:
Android ships VK_HEADER_VERSION as old as 28, which chokes on
`VK_FORMAT_*_2PLANE_*` and `VK_FORMAT_*_3PLANE_*`

### DIFF
--- a/layers/vk_format_utils.cpp
+++ b/layers/vk_format_utils.cpp
@@ -866,6 +866,8 @@ VK_LAYER_EXPORT VkExtent3D FormatCompressedTexelBlockExtent(VkFormat format) {
 
 VK_LAYER_EXPORT uint32_t FormatPlaneCount(VkFormat format) {
     switch (format) {
+// Android's version of vulkan.h is outdated and does not know these.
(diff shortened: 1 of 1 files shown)


krOoze:
I would like to note this repo already has (updated) [https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/tree/master/include/vulkan](Vulkan header). I assume those are used during build?

tobine:
Closing this PR. We generally keep the NDK vulkan header reasonably up-to-date but if it's lagging you can use the latest header from this repo. No need to customize core repo based on old header versions. If you're specifically limited to an older header then you'll have to get older versions of the layers or customize them locally for your particular header.

volcanoauthors:
@tobine I'm not sure how one would use a newer header with new format values, then pass them to the Android Vulkan entrypoint and it would work.

krOoze:
@volcanoauthors It would work normally...? (Assuming you enable the required extension)
I am not exactly sure what you are asking.

PR #2455, opened 2018-03-01
macOS: Add macOS support

karl-lunarg:


loader/LoaderAndLayerInterface.md
```diff
@@ -937,6 +942,33 @@ environment variables are only used for non-suid programs.  See
...
+    /etc/vulkan/explicit_layer.d
+    /etc/vulkan/implicit_layer.d
+    /usr/local/share/vulkan/explicit_layer.d
+    /usr/local/share/vulkan/implicit_layer.d
+    /usr/share/vulkan/explicit_layer.d
+    /usr/share/vulkan/implicit_layer.d
+    $HOME/.local/share/vulkan/explicit_layer.d
+    $HOME/.local/share/vulkan/implicit_layer.d
+
+1. &lt;bundle&gt; is the directory containing a bundled application.  It is scanned first.
```
MarkY-LunarG: Should this be < and >?  Seems like an editor put the wrong items in.

MarkY-LunarG:
Nothing major jumps out at me.  A few small nits on the MD file.

loader/LoaderAndLayerInterface.md
karl-lunarg: Those are escapes for those symbols.  Otherwise, they look like inline html.

layers/macos/VkLayer_core_validation.json
```diff
@@ -0,0 +1,29 @@
+{
+    "file_format_version" : "1.1.0",
+    "layer" : {
+        "name": "VK_LAYER_LUNARG_core_validation",
+        "type": "GLOBAL",
+        "library_path": "./libVkLayer_core_validation.dylib",
+        "api_version": "1.0.65",
```
mark-lunarg: Do you care that these are not 1.1.70 or 1.0.69?

karl-lunarg: I'll fix those.

loader/CMakeLists.txt
```diff
@@ -219,27 +227,70 @@ else()
...
+        target_link_libraries(${API_LOWERCASE} "-framework CoreFoundation")
+
+        # Build vulkan.framework
+        set(FRAMEWORK_HEADERS
+            ${CMAKE_SOURCE_DIR}/include/vulkan/vk_icd.h
+            ${CMAKE_SOURCE_DIR}/include/vulkan/vk_layer.h
+            ${CMAKE_SOURCE_DIR}/include/vulkan/vk_platform.h
+            ${CMAKE_SOURCE_DIR}/include/vulkan/vk_sdk_platform.h
+            ${CMAKE_SOURCE_DIR}/include/vulkan/vulkan.h
+            ${CMAKE_SOURCE_DIR}/include/vulkan/vulkan.hpp
```
lenny-lunarg: I would be suspicious of these lines when using this repo as a submodule. I encountered a problem recently where using` CMAKE_BINARY_DIR` broke things when including the repo in VulkanTools because the value of that variable was different. I suspect that the same thing applies here and that it would be better to use:

    ${CMAKE_CURRENT_SOURCE_DIR}/../include/vulkan/whatever.h

It might be helpful to try updating the submodule locally to catch anything like that.

layers/macos/VkLayer_core_validation.json
lenny-lunarg: All of the mac layer json files have an old version. We'll just need to make sure they get updated with the linux/windows files.

loader/CMakeLists.txt
karl-lunarg: I was thinking about deferring this to when we work on VulkanTools, but I'll look into it now.

I also would want to put some thought into using `${PROJECT_SOURCE_DIR}`, which is the dir where the `CMakeLists.txt` file containing `project(VULKAN)` resides.  This may be a little more elegant than the pathrobatics involved with relative paths.

layers/macos/VkLayer_core_validation.json
karl-lunarg: yes, will fix.

loader/CMakeLists.txt
lenny-lunarg: Actually, `PROJECT_SOURCE_DIR` is probably a better solution. I'd need to look up how it works to be sure, but the main thing is that it needs to be based off a directory that won't change.

mark-lunarg:
THE CMAKE REVOLUTION CANNOT COME SOON ENOUGH.

karl-lunarg:
discussed changes applied and all commits in this PR squashed to one commit.

PR #2456, opened 2018-03-01
build: Fix -Wmicrosoft-cast warnings with clang-cl

nico:
MSVC allows implicit conversion from function pointers to void*.
The C++ standard doesn't allow this, and clang-cl emits a warning
when this is done.  This change fixes these warnings in
Vulkan-LoaderAndValidationLayers.

Arguably, loader_platform_get_proc_address() should return a
FARPROC so that the compiler can help checking calling conventions,
but this is the simplest fix that unblocks us.

This is needed for https://crbug.com/550065

### DIFF
--- a/loader/vk_loader_platform.h
+++ b/loader/vk_loader_platform.h
@@ -296,7 +296,7 @@ static void loader_platform_close_library(loader_platform_dl_handle library) { F
 static void *loader_platform_get_proc_address(loader_platform_dl_handle library, const char *name) {
     assert(library);
     assert(name);
-    return GetProcAddress(library, name);
+    return (void *)GetProcAddress(library, name);
 }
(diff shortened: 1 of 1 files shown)


mark-lunarg:
@lenny-lunarg, please don't forget to fix up the commit message before taking this.

nico:
What's wrong with the commit message? I can change it.

nico:
Done.

nico:
How long do pull requests usually take to get merged? After this lands, I need to update the version of vulkan that ANGLE uses to include this change, and then I need to update the version of ANGLE that Chromium uses, which will take a while -- so it'd be cool if this could make it in soon.

nico:
Thank you!

PR #2460, opened 2018-03-02
Implement checks for dedicated allocation extensions (c.f. 2444)

### LINKED ISSUE #2444: Implement VU for dedicated allocation extensions
We've recently seen a game bind a dedicated allocation created for one image to a different image, and the layers didn't catch it. The relevant VUs are:

  * [[VUID-vkBindImageMemory-memory-01509]]
    If the sname:VkmemoryAllocateInfo provided when pname:memory was
    allocated included an instance of slink:VkMemoryDedicatedAllocateInfoKHR
    in its pname:pNext chain, and
    slink:VkMemoryDedicatedAllocateInfoKHR::pname:image was not
    ename:VK_NULL_HANDLE, then pname:image must: equal
    slink:VkMemoryDedicatedAllocateInfoKHR::pname:image and
    pname:memoryOffset must: be zero.

  * [[VUID-vkBindBufferMemory-memory-01508]]
    If the sname:VkmemoryAllocateInfo provided when pname:memory was
    allocated included an instance of slink:VkMemoryDedicatedAllocateInfoKHR
    in its pname:pNext chain, and
    slink:VkMemoryDedicatedAllocateInfoKHR::pname:buffer was not
    ename:VK_NULL_HANDLE, then pname:buffer must: equal
    slink:VkMemoryDedicatedAllocateInfoKHR::pname:buffer and
    pname:memoryOffset must: be zero.

and similarly for BindImage/BufferMemory2, though it looks like the corresponding VUs may be missing?

jzulauf-lunarg:
Added validation checks and unit tests for the following valid usage  w.r.t. binding dedicated allocations to buffers and images.

    VALIDATION_ERROR_17000bc8 VUID-vkBindBufferMemory-memory-01508
    VALIDATION_ERROR_17400bca VUID-vkBindImageMemory-memory-01509

The same code path also is used to validate the BindImageMemory2/BindBufferMemory2, but has no matching valid usage (in process).  Unit testing is added only for the VUID cases above.

Fixes #2444 .


layers/core_validation.h
```diff
@@ -68,11 +68,6 @@
 // TODO : Is there a way to track when Cmd Buffer finishes & remove mem references at that point?
 // TODO : Could potentially store a list of freed mem allocs to flag when they're incorrectly used
 
-struct GENERIC_HEADER {
```
jzulauf-lunarg: I removed the last usage of the cut and paste pNext walking loop -- killing GENERIC_HEADER to prevent recurrence.  Use LvlGenericHeader and lvl_find_in_chain instead.

PR #2461, opened 2018-03-04
tests: Improve Image creation tests

krOoze:
- should fix platform-specific Unexpected messages
  - relies on mandatory support formats now
- add extra tests for features vs usage checks
  - and make them non-conditional (no longer silently skipped)
- add tests for all other missing things
- buncha esthetics


tests/layer_validation_tests.cpp
```diff
@@ -19101,214 +19101,299 @@ TEST_F(VkLayerTest, ImageFormatLimits) {
...
-    image_ci.extent.width = std::max(img_limits.maxExtent.width, dev_limits.maxImageDimension1D) + 1;
-    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_09e0076e);
-    vkCreateImage(m_device->handle(), &image_ci, NULL, &image);
-    m_errorMonitor->VerifyFound();
+    {
+        VkImageCreateInfo image_ci = safe_image_ci;
+        image_ci.imageType = VK_IMAGE_TYPE_1D;
+        VkImageFormatProperties img_limits;
+        ASSERT_VK_SUCCESS(GPDIFPHelper(gpu(), &image_ci, &img_limits));
+        image_ci.extent = {img_limits.maxExtent.width + 1, 1, 1};
```
krOoze: Trips the dev_sim and mock_icd. The json profiles do not contain `VkImageFormatProperties`, so it may happen that `maxExtent < maxImageDimensionX` (which is invalid impl).

Quick workaround would be to re-introduce `max` of the two values.

Correct fix should be:
- In the cookie-cutter template of `mock_icd` make the limit UINT32_MAX
- In the specific device profile of `dev_sim` make sure to return larger or equal values with `vkGetPhysicalDeviceImageFormatProperties` than the `VkPhysicalDeviceLimits`

tobine: This looks like a good update and would be nice to land. I'm fine with the quick workaround to get this landed. We can also modify the built-in values of the mock_icd to work towards correct solution. For DevSim change, can you submit a feature request in VulkanTools repo?

krOoze: @tobine As am I. Already have that locally. As well as fix for CI errors.
But discovered there are some things that should be tested here and are not.
Also get bit stuck on testing `VALIDATION_ERROR_09e00758` which is bit tricky to do **correctly**.

krOoze:
Ok, this is ready for review (assuming CI passes).

PR #2463, opened 2018-03-05
Update CmdPushConstant Valid Usages

### LINKED ISSUE #2442: Validating push constant ranges with overlapping shader stages
Valid usage statements for vkCmdPushConstants [were recently rephrased](https://github.com/KhronosGroup/Vulkan-Docs/commit/ab08f0951ef1ad9b84db93f971e113c1d9d55609) to allow overlapping ranges in different shader stages. The core validation layer hasn't been updated to reflect these changes, thus causing false validation errors.

Old VU:
>  stageFlags must match exactly the shader stages used in layout for the range specified by offset and size

New VUs:

> For each byte in the range specified by offset and size and for each shader stage in stageFlags, there must be a push constant range in layout that includes that byte and that stage
> 
> For each byte in the range specified by offset and size and for each push constant range that overlaps that byte, stageFlags must include all stages in that push constant range’s VkPushConstantRange::stageFlags

jzulauf-lunarg:
Updated the core validation layer and matching units to reflect changes
to the valid usage statements in the Vulkan specification to better
support the use of overlapping push constant ranges.  The following
valid usage was removed from the specification and the core validation
layer.

VUID-vkCmdPushConstants-stageFlags-00367 VALIDATION_ERROR_1bc002de

Two new valid usages have been added to the core validation layer.

VALIDATION_ERROR_1bc00e06 VUID-vkCmdPushConstants-offset-01795
VALIDATION_ERROR_1bc00e08 VUID-vkCmdPushConstants-offset-01796

The individual unit tests within the larger InvalidPushConstants test
case have been updated to reflect the changes to the valid usage checks.

Fixes #2442 .


layers/vk_validation_error_database.txt
```diff
@@ -2382,13 +2382,13 @@ VALIDATION_ERROR_1ba02413~^~N~^~Unknown~^~vkCmdProcessCommandsNVX~^~VUID-vkCmdPr
...
 VALIDATION_ERROR_1bc00009~^~Y~^~Unknown~^~vkCmdPushConstants~^~VUID-vkCmdPushConstants-commonparent~^~core~^~The spec valid usage text states 'Both of commandBuffer, and layout must have been created, allocated, or retrieved from the same VkDevice' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkCmdPushConstants-commonparent)~^~implicit
-VALIDATION_ERROR_1bc002de~^~Y~^~InvalidPushConstants~^~vkCmdPushConstants~^~VUID-vkCmdPushConstants-stageFlags-00367~^~core~^~The spec valid usage text states 'stageFlags must match exactly the shader stages used in layout for the range specified by offset and size' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkCmdPushConstants-stageFlags-00367)~^~
+VALIDATION_ERROR_1bc002de~^~N~^~None~^~vkCmdPushConstants~^~VUID-vkCmdPushConstants-stageFlags-00367~^~core~^~The spec valid usage text states 'stageFlags must match exactly the shader stages used in layout for the range specified by offset and size' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkCmdPushConstants-stageFlags-00367)~^~
```
jeremyk-lunarg: If the documentation has changed, should we eventually remove this?

mark-lunarg: Taking them out hasn't been part of the process, but this'll change in the next few weeks.

jzulauf-lunarg: Somebody removed the obsolete one from main.

jzulauf-lunarg:
Can I get a review (one way of the other?)  Thanks.

tests/layer_validation_tests.cpp
```diff
@@ -9304,26 +9304,41 @@ TEST_F(VkLayerTest, InvalidPushConstants) {
     vkCmdPushConstants(m_commandBuffer->handle(), pipeline_layout_obj.handle(), 0, 0, 16, dummy_values);
     m_errorMonitor->VerifyFound();
 
+    // Positive tests for the overlapping ranges
```
polarina: Would it be practical to move this test to VkPositiveLayerTests?

polarina: It is my understanding that overlapping push constant ranges are a new feature in Vulkan 1.1. Do we want to check the current Vulkan version before proceeding?

jzulauf-lunarg: @polarina -- when there are a small number of positive cases which share the same setup with a set of negative cases, it's fairly common to have them in the same test to avoid test setup duplication -- in this case they are as much validating the test *setup* as the validation check.

The overlapping range issue is considered a 1.0 spec clarification, as opposed to a new feature, first found in 1.0.69. ( https://github.com/KhronosGroup/Vulkan-Docs/commit/ab08f0951ef1ad9b84db93f971e113c1d9d55609 )

Thanks for the detailed look.

PR #2465, opened 2018-03-06
Update Contributing.md for validation message output information

### LINKED ISSUE #2140: Scrub error messages and add suggestions where appropriate
Tracking issue based on survey feedback. In general make sure that when error messages can give suggestions about *how* to fix the problem they should do so to assist the user.

mark-lunarg:
Fixes #2140.  Add guidelines for error message output, including specifying object handles, values, and user assistance on HOW to fix the error.

Will also update our internal docs to match, and this doc receive an overhaul and expansion in the next few weeks.


krOoze:
To add suggestions is well-meaning, but oftentimes values dump + VU quote is self-explanatory (and sometimes even less confusing).

Also giving really good suggestions may need more code and several slightly different messages per-check.

Would it make sense to add suggestions to only the most obvious cases. At least for the time being; I think layers full coverage should take priority over this.

PR #2467, opened 2018-03-06
layers/vk_layer_config.cpp: fix win32 include case

lachs0r:
Fixes mingw-w64 compilation on case-sensitive file systems.


krOoze:
I wonder why is it there in the first place. Why not `getenv` on Windows as well?

lachs0r:
I don’t know and didn’t check what this is actually used for, but the usual MSVCRT non-Unicode bullshit applies. So you can’t use straight getenv on Windows either way.

krOoze:
Why not? It is a standard function.
What specific bullshit? If it is non-unicode, then it should be identical to `GetEnvironmentVariable`, no? (Besides this code would not even work with UNICODE as that would return `wchar`s)

PR #2472, opened 2018-03-08
Merge 1.1.70

mikes-lunarg:


mikes-lunarg:
@cnorthrop Can you look at f1b7e57 specifically to see if I missed anything?

layers/core_validation.cpp
```diff
@@ -11746,94 +11762,171 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumerateDeviceExtensionProperties(VkPhysicalDevi
...
+                                                                                  pPhysicalDeviceGroupProperties);
+    if (result == VK_SUCCESS) {
+        PostCallRecordEnumeratePhysicalDeviceGroups(instance_data, pPhysicalDeviceGroupCount, pPhysicalDeviceGroupProperties);
+    }
+    return result;
+}
+
+VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDeviceGroupsKHR(
+    VkInstance instance, uint32_t *pPhysicalDeviceGroupCount, VkPhysicalDeviceGroupPropertiesKHR *pPhysicalDeviceGroupProperties) {
+    bool skip = false;
```
tobine: It looks like all of the code in this function body is identical to the function body of EnumeratePhysicalDeviceGroups() above. Can you verify and if so just call one from the other? I'd suggest making KHR ver call non-KHR version.

layers/unique_objects.cpp
```diff
@@ -590,14 +623,28 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDescriptorUpdateTemplateKHR(VkDevice device
...
+    std::unique_lock<std::mutex> lock(global_lock);
+    uint64_t descriptor_update_template_id = reinterpret_cast<uint64_t &>(descriptorUpdateTemplate);
+    dev_data->desc_template_map.erase(descriptor_update_template_id);
+    descriptorUpdateTemplate = (VkDescriptorUpdateTemplate)dev_data->unique_id_mapping[descriptor_update_template_id];
+    dev_data->unique_id_mapping.erase(descriptor_update_template_id);
+    lock.unlock();
+    dev_data->dispatch_table.DestroyDescriptorUpdateTemplate(device, descriptorUpdateTemplate, pAllocator);
+}
+
+// This is the extension version of this routine.  The core version is above.
```
tobine: Can this version call into core and kill function body?

layers/unique_objects.cpp
```diff
@@ -590,14 +623,28 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDescriptorUpdateTemplateKHR(VkDevice device
...
 VKAPI_ATTR void VKAPI_CALL DestroyDescriptorUpdateTemplateKHR(VkDevice device,
                                                               VkDescriptorUpdateTemplateKHR descriptorUpdateTemplate,
                                                               const VkAllocationCallbacks *pAllocator) {
     layer_data *dev_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
     std::unique_lock<std::mutex> lock(global_lock);
     uint64_t descriptor_update_template_id = reinterpret_cast<uint64_t &>(descriptorUpdateTemplate);
     dev_data->desc_template_map.erase(descriptor_update_template_id);
-    descriptorUpdateTemplate = (VkDescriptorUpdateTemplateKHR)unique_id_mapping[descriptor_update_template_id];
+    descriptorUpdateTemplate = (VkDescriptorUpdateTemplate)unique_id_mapping[descriptor_update_template_id];
```
tobine: should this be dev_data->unique_id_mapping...?

layers/unique_objects.cpp
```diff
@@ -685,6 +732,23 @@ void *BuildUnwrappedUpdateTemplateBuffer(layer_data *dev_data, uint64_t descript
...
+        std::lock_guard<std::mutex> lock(global_lock);
+        descriptorSet = Unwrap(dev_data, descriptorSet);
+        descriptorUpdateTemplate = (VkDescriptorUpdateTemplate)dev_data->unique_id_mapping[template_handle];
+    }
+    void *unwrapped_buffer = BuildUnwrappedUpdateTemplateBuffer(dev_data, template_handle, pData);
+    dev_data->dispatch_table.UpdateDescriptorSetWithTemplate(device, descriptorSet, descriptorUpdateTemplate, unwrapped_buffer);
+    free(unwrapped_buffer);
+}
+
+// This is the extension version of this routine.  The core version is above.
```
tobine: Can this call into core version?

layers/unique_objects.cpp
```diff
@@ -694,7 +758,7 @@ VKAPI_ATTR void VKAPI_CALL UpdateDescriptorSetWithTemplateKHR(VkDevice device, V
     {
         std::lock_guard<std::mutex> lock(global_lock);
         descriptorSet = Unwrap(descriptorSet);
-        descriptorUpdateTemplate = (VkDescriptorUpdateTemplateKHR)unique_id_mapping[template_handle];
+        descriptorUpdateTemplate = (VkDescriptorUpdateTemplate)unique_id_mapping[template_handle];
```
tobine: Should this be dev_data->unique_id_mapping?

layers/object_tracker_utils.cpp
```diff
@@ -1006,6 +1006,34 @@ VKAPI_ATTR void VKAPI_CALL DestroyCommandPool(VkDevice device, VkCommandPool com
...
+        if (instance_data->queue_family_properties.size() < *pQueueFamilyPropertyCount) {
+            instance_data->queue_family_properties.resize(*pQueueFamilyPropertyCount);
+        }
+        for (uint32_t i = 0; i < *pQueueFamilyPropertyCount; i++) {
+            instance_data->queue_family_properties[i] = pQueueFamilyProperties[i].queueFamilyProperties;
+        }
+    }
+}
+
+// Note: This is the extension version of this routine.  The core version is above.
```
tobine: Can it call into core version above?

scripts/helper_file_generator.py
```diff
@@ -185,7 +185,9 @@ def genGroup(self, groupinfo, groupName):
             for elem in groupElem.findall('enum'):
                 if elem.get('supported') != 'disabled':
                     item_name = elem.get('name')
-                    value_list.append(item_name)
+                    # Avoid duplicates
```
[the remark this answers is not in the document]
mikes-lunarg: fixed e48f549

scripts/lvl_genvk.py
```diff
@@ -517,18 +548,20 @@ def genTarget(args):
...
-                directory = args.directory)
+    makeGenOpts(args)
 
     if (args.target in genOpts.keys()):
         createGenerator = genOpts[args.target][0]
         options = genOpts[args.target][1]
 
         if not args.quiet:
             write('* Building', options.filename, file=sys.stderr)
+            write('* options.versions          =', options.versions, file=sys.stderr)
```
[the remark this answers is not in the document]
mikes-lunarg: Those lines come from the Vulkan-Docs genvk.py so I would prefer to keep them the same.  I did merge in the latest genvk.py to cut down on the default verbosity when building: e377dac

layers/core_validation.cpp
mikes-lunarg: The downchain call has to be different between the two, but we are reusing the Pre/Post functions for both KHR and non-KHR versions.

layers/unique_objects.cpp
mikes-lunarg: We moved the map global recently https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/2453 because of issues with surface unwrapping in vkGetDeviceGroupSurfacePresentModes

loader/extension_manual.h
```diff
@@ -25,13 +25,13 @@
 // These functions, for whatever reason, require more complex changes than
 // can easily be automatically generated.
 
-VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDeviceGroupsKHR(
+/*VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDeviceGroupsKHR(
```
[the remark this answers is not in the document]
lenny-lunarg: It should be killed. I just pushed a commit to remove it.

layers/unique_objects.cpp
mikes-lunarg: Still needs to call KHR version from dispatch table, but agree that this duplicate code needs to be cleaned up

layers/unique_objects.cpp
mikes-lunarg: #2453

mikes-lunarg:
Opened #2474 to deal with the broader KHR/core code reuse issue.  I would like to get this merged soon because it is blocking VulkanTools and VulkanSamples from merging 1.1 content

tobine:
I've convinced myself that this is in sufficient state to land. Note that Travis-CI fails. Note clear exactly why. Gets the message "No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself." seemingly while it's doing clang-format checks so maybe volume of code is taking too long to clang-format?
If that's the case then after this lands follow-on commits should clear up in CI again.

tobine:
Thanks for breaking this up into so many focused CLs, which made review as nice as it could be for such a massive PR.

mikes-lunarg:
Yep, clang-format generates huge diffs on header updates,  I’ll look into excluding them from the format rules in future

PR #2475, opened 2018-03-09
WIP: Logger Strawman

jzulauf-lunarg:
TODO: Rebase this on top of MikeS recent changes.
    
For your review/critique is a quick and dirty working prototype of a new message logger for the validation layers. 

The goals were:

* No repetition of known data (i.e. types/VUIDs)
* No casts when constructing messages. 
* Flexible formatting options.
* No formatting if message will not be emitted.
* A starting place for more brainstorming

I've given 3 examples of what the new message create would look like.  The programming flow is:
 
* Create logger object with the fixed data
* Add to object (noop'd if flags are not enabled)
   * sprintf style
   * add* methods (including handle and pointer automated handling)
   * <<  for types where built-in stream out is acceptable
* Either manually post the the report (if no skip variable is present) --or--
* Allow the logger object to go out of scope ("report" is called in the destructor if not done already)

    
Missing:
* even better automatic casting/overload.  (given further code generation)
* detailed interface design (naming, and method vs. operator)
    
Messy:
* Use of MACRO for msg generation (to compress flags and __LINE__)
* add_x and add_h for hex and handle handling...
* transitions between built-in type support (<< operations) and .add* method operations
* Using << style formatting. (maybe not familiar to maintainers)
    
Pretty Cool:
* Automatic object type detection
* No-op output functions if messages are not enabled
* Built-in sprintf conversions
* Automatic update of "skip", and post to debug_log_msg in report/destructor
* Automatic appending of validation_error_message (when avail)
* No more PRI ...


layers/vk_layer_logging.h
```diff
@@ -439,4 +455,131 @@ uint64_t HandleToUint64(HANDLE_T h);
...
+    bool reported_;
+};
+
+// stream out to log msg using the public interface
+template <typename OutType>
+LogMsg &operator<<(LogMsg &logger, const OutType &out) {
+    return logger.add(out);
+}
+
+#define LOG_ERR_MSG(skip, msg, report_data, object, code, prefix) \
```
jzulauf-lunarg: if the `__LINE__` and prefix are going away, we can avoid the macro by subclassing LogMsg for the two common cases without loss of brevity.

scripts/helper_file_generator.py
```diff
@@ -716,32 +717,49 @@ def GenerateObjectTypesHeader(self):
         object_types_header += '// Helper array to get Vulkan VK_EXT_debug_report object type enum from the internal layers version\n'
         object_types_header += 'const VkDebugReportObjectTypeEXT get_debug_report_enum[] = {\n'
         object_types_header += '    VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, // No Match\n'
+
```
jzulauf-lunarg: For the templates below, I want to have a cross-reference map... and since I had it... thought I'd use it.

scripts/helper_file_generator.py
```diff
@@ -716,32 +717,49 @@ def GenerateObjectTypesHeader(self):
...
+        vko_map = {to_key(vko_re, vko) : vko for vko in self.core_object_types}
         for object_type in type_list:
-            search_type = object_type.replace("kVulkanObjectType", "").lower()
-            for vk_object_type in self.core_object_types:
-                target_type = vk_object_type.replace("VK_OBJECT_TYPE_", "").lower()
-                target_type = target_type.replace("_", "")
-                if search_type == target_type:
-                    object_types_header += '    %s,   // %s\n' % (vk_object_type, object_type)
-                    break
+            vk_object_type = vko_map[object_type.replace("kVulkanObjectType", "").lower()]
```
jzulauf-lunarg: since I already have the map, why not use it.

scripts/helper_file_generator.py
```diff
@@ -716,32 +717,49 @@ def GenerateObjectTypesHeader(self):
...
+                                '    static const VkDebugReportObjectTypeEXT kDebugObjectType = {};\n' +
+                                '    static const VkObjectType kCoreObjectType = {};\n')
+        unknown_enums = ( "kVulkanObjectTypeUnknown", "VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT", "VK_OBJECT_TYPE_UNKNOWN")
+        object_types_header += vko_template_body_fmt.format(*unknown_enums)
+        object_types_header += '};\n\n'
+
+        for object_type in type_list:
+            ot_key = object_type.replace("kVulkanObjectType", "").lower()
+            vk_type = object_type.replace("kVulkanObjectType", "Vk")
+            object_types_header += '// Object type enums for %s\n' % (vk_type)
```
jzulauf-lunarg: You could get away with just the kVulkanObjectType enum (and look up the others) but as the compiler is likely just constexpr the whole thing, it's free so have them all.

layers/vk_layer_logging.h
```diff
@@ -439,4 +455,131 @@ uint64_t HandleToUint64(HANDLE_T h);
...
+    template <typename ObjectType>
+    LogMsg(bool *status, const debug_report_data *debug_data, VkFlags msg_flags, ObjectType src_object, size_t location,
+           UNIQUE_VALIDATION_ERROR_CODE msg_code, const char *layer_prefix)
+        : LogMsg(status, debug_data, msg_flags, src_object, location, static_cast<int32_t>(msg_code), layer_prefix) {
+        msg_code_is_vu_ = true;
+    }
+
+    template <typename IntType>
+    LogMsg &add_x(IntType out) {
+        if (will_report_) {
```
jzulauf-lunarg: note the noop'ing controlled by will_report_

layers/core_validation.cpp
```diff
@@ -2681,12 +2677,10 @@ static bool validateQueueFamilyIndices(layer_data *dev_data, GLOBAL_CB_NODE *pCB
...
     if (pPool && queue_state) {
         if (pPool->queueFamilyIndex != queue_state->queueFamilyIndex) {
-            skip |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_COMMAND_BUFFER_EXT,
-                            HandleToUint64(pCB->commandBuffer), __LINE__, VALIDATION_ERROR_31a00094, "DS",
-                            "vkQueueSubmit: Primary command buffer 0x%" PRIx64
-                            " created in queue family %d is being submitted on queue 0x%" PRIx64 " from queue family %d. %s",
-                            HandleToUint64(pCB->commandBuffer), pPool->queueFamilyIndex, HandleToUint64(queue),
-                            queue_state->queueFamilyIndex, validation_error_map[VALIDATION_ERROR_31a00094]);
+            LOG_ERR_MSG(&skip, msg, dev_data->report_data, pCB->commandBuffer, VALIDATION_ERROR_31a00094, "DS");
+            msg.add("vkQueueSubmit: Primary command buffer ").add_h(pCB->commandBuffer);
```
jzulauf-lunarg: with more code gen, the add_h could be removed in favor of "overloads per handle type", s.t. all handles are known for what they are.

layers/core_validation.cpp
```diff
@@ -412,11 +412,10 @@ static bool ValidateMemoryIsValid(layer_data *dev_data, VkDeviceMemory mem, uint
     DEVICE_MEM_INFO *mem_info = GetMemObjInfo(dev_data, mem);
     if (mem_info) {
         if (!mem_info->bound_ranges[bound_object_handle].valid) {
-            return log_msg(dev_data->report_data, VK_DEBUG_REPORT_WARNING_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_MEMORY_EXT,
-                           HandleToUint64(mem), __LINE__, MEMTRACK_INVALID_MEM_REGION, "MEM",
-                           "%s: Cannot read invalid region of memory allocation 0x%" PRIx64 " for bound %s object 0x%" PRIx64
-                           ", please fill the memory before using.",
-                           functionName, HandleToUint64(mem), object_string[type], bound_object_handle);
+            LOG_WARN_MSG(nullptr, msg, dev_data->report_data, mem, MEMTRACK_INVALID_MEM_REGION, "MEM");
```
jzulauf-lunarg: In looking for good examples I stumbled across  VerifyBoundMemoryIsValid , which is using the *wrong* object types (hard coded image for things that can be buffers)... implementing the auto detection is going to take changing some interfaces (like that one) to not convert handles (which will probably fix some latent bugs) and templatizing things like  VerifyBoundMemoryIsValid... that or adding LogMsg constructors that take uint64_t/object_type pairs.

layers/core_validation.cpp
krOoze: `skip`, `report_data`, `objType` + `object`, layer name, and function name are usually constants for the whole function.
Would be nice not having to repeat them. Maybe `class Logger`? Then `Logger log(all the above)` and then per case `log.err(veid, msg)` or `log.warn()`?

jzulauf-lunarg: Perhaps a LoggerContext that could be used in a LogMsg constructor

Issue #2476, opened 2018-03-10
trampoline.c is missing core entrypoints


mchock-nv:
Specifically, 

vkCreateDescriptorUpdateTemplate
vkDestroyDescriptorUpdateTemplate
vkUpdateDescriptorSetWithTemplate

lenny-lunarg:
Resolved via 27157624a7efc98de08c563f5b2c4f2732a68a46 and 8984ee77864cc9f4d8b235262a97d0622a5c9883. See the linked PR for more details.

PR #2479, opened 2018-03-10
Add missing core entrypoints to trampoline.c

mchock-nv:
The entrypoints associated with VK_KHR_descriptor_update_template didn't
make it into the trampoline file when the extension was promoted to
core. This change adds them.

See Issue #2476


lenny-lunarg:
Thanks for the PR and sorry that I didn't get around to this sooner. The PR mostly looks good, but there's a couple of extra things that need to be done to set up the entrypoints. I pushed your PR to master as commit 27157624a7efc98de08c563f5b2c4f2732a68a46 and I also added commit 8984ee77864cc9f4d8b235262a97d0622a5c9883 to finish things up. My commit adds the functions to the def file and the loader's internal GetProcAddr function.

This PR will show as closed instead of merged because I tweaked your commit before I pushed it. The only thing I did was to change the commit message. We write our commits in the form "component: Commit message", so I just added "loader:" to start of the message.

Issue #2486, opened 2018-03-14
Potentially false validation error if image size > maxResourceSize


pppessi:
Running `dEQP-VK.pipeline.render_to_image.core.2d_array.huge.width_height_layers.r8g8b8a8_unorm` produces a validation error:

> ERROR: Object: 0x0 | CreateImage resource size exceeds allowable maximum Image resource size = 0x400000000, maximum resource size = 0x80000000 (code 0x00000000 from Image at Image:867)

when trying to create an image using the following info:
```
VkImageCreateInfo
{
    VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO,
    0,
    0,
    VK_IMAGE_TYPE_2D,
    VK_FORMAT_R8G8B8A8_UNORM,
    { 4096, 4096, 1 },
    1,
    256,
    VK_SAMPLE_COUNT_1_BIT,
    VK_IMAGE_TILING_OPTIMAL,
    VK_IMAGE_USAGE_TRANSFER_SRC_BIT | VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT,
    VK_SHARING_MODE_EXCLUSIVE,
    0,
    0,
    VK_IMAGE_LAYOUT_UNDEFINED
}
```
I don't think this is quite right. Specification says:

> There is no mechanism to query the size of an image before creating it, to compare that size against maxResourceSize. If an application attempts to create an image that exceeds this limit, the creation will fail and vkCreateImage will return VK_ERROR_OUT_OF_DEVICE_MEMORY.

> Even if vkGetPhysicalDeviceImageFormatProperties2 returns success and the parameters to vkCreateImage are all within the returned limits, vkCreateImage must fail and return VK_ERROR_OUT_OF_DEVICE_MEMORY if the resulting size of the image would be larger than maxResourceSize.

On this device imageGranularity is 1 so it should not affect this in any way.

krOoze:
`4096 x 4096 x 256 x 4 = 0x4 0000 0000`, which **is** greater than 0x8000 0000(=2<sup>31</sup>),  so that checks out.

It should be a `WARNING`, not `ERROR`, oterwise I don't see what the problem is?

tobine:
Spec quote from @pppessi is out of date, potentially. Current language is:

[...]

Issue #1612, opened 2017-03-22
Best Practices/Assistant/Performance Layer


mark-lunarg:
This issue is a place holder for a utility layer which provides application feedback for performance-related checks.  For details and to add items, please see/add to the [LunarG Github VulkanTools project Wiki](https://github.com/LunarG/VulkanTools/wiki) or add an issue to this repo or VulkanTools and we'll add to the master list.

The following Github issues have been created and are now tracked by this issue.

- #1401 Performance warning: VK_PIPELINE_STAGE_ALL_COMMANDS_BIT or VK_PIPELINE_STAGE_ALL_GRAPHICS_BIT
- #1253 Performance warning: Pipeline barriers used for layout transitions instead of Render Passes
- #1155 Performance warning: High number of memory objects

karl-lunarg:
#1628 "VK_LAYER_LUNARG_core_validation queue submissions list grows forever and causes crash" might be another candidate condition to check for in this layer.

RippeR37:
Idea: Maybe this layer could also check if same state is set multiple times in a row which can lead to lower performance?

mark-lunarg:
From @Tony-LunarG: Assuming that the 99% common workflow is [BeginCommandBuffer; write commands to command buffer; EndCommandBuffer], it would be nice to warn developer if the code is recording two command buffers simultaneously. Would also need to validate assumption about workflow. Note that it needs to be thread smart since we expect multiple threads to be recording multiple command buffers simultaneously

MarkY-LunarG:
Idea: Detect case of LVL issue #1917 and warn.

karl-lunarg:
Idea: #2061 - Add a layer that writes a unique ID to a CPU-visible buffer whenever a marker is encountered.  After a GPU crash, the buffer can be examined to see the last ID written.

HansKristian-ARM:
Is this supposed to be a generic layer for best practices on Vulkan which collects best practices from a spec POV, or would there be room for IHV specifics as well (which are probably where the most useful/interesting checks will be)?

mark-lunarg:
@HansKristian-ARM, the Assistant Layer effort is currently focused on the former.  However, it is a simple matter to create related assistant-type layers specific to a particular IHV.  The aim of the layer framework is to make the creation of these types of layers much more straightforward.

hrydgard:
[...]

PR #2487, opened 2018-03-14
layers: Downgrade maxResourceSize check to warning

krOoze:
maybe addresses #2486


krOoze:
PS: Ideally, it should trigger iff it actually returns the `VK_ERROR_OUT_OF_DEVICE_MEMORY`. Is there some obscure case where the image could actually use less resources than its parameters would suggest?

mark-lunarg:
@krOoze, I'll to move this check to the assistant layer (#1612).  I think we also talked about adding an interceptor to warn on any 'bad' return codes from the display driver, so I'll add that to the list as well (I think most of these have been removed from the layers).

PR #2488, opened 2018-03-15
build: Don't require glslang if not building layers

chewi:
This makes distribution packaging easier.


mark-lunarg:
Thanks @chewi!

PR #2490, opened 2018-03-16
Fix queue bits for CmdWriteTimeStamp

mark-lunarg:
Transfer queue is called out but was omitted. Verified all other Cmd queue flag cases as well.


mark-lunarg:
Thanks @jzulauf-lunarg!

PR #2491, opened 2018-03-16
Fix kVulkanObjectType generation issues w.r.t. 1.1 changes

jzulauf-lunarg:
With 1.1 the VkDebugReportObjectTypeEXT enums are no longer be updated (in favor of the VkObjectType), which meant there is no long a 1-1 relationship with the kVulkanObjectTypes.  Addtionally, handle types that were promoted in 1.1 from extension to core needed special handling given aliasing in the registry.

### DIFF
--- a/scripts/helper_file_generator.py
+++ b/scripts/helper_file_generator.py
@@ -19,6 +19,7 @@
 #
 # Author: Mark Lobodzinski <mark@lunarg.com>
 # Author: Tobin Ehlis <tobine@google.com>
+# Author: John Zulauf <jzulauf@lunarg.com>
 
 import os,re,sys
 import xml.etree.ElementTree as etree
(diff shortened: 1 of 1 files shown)


scripts/helper_file_generator.py
```diff
@@ -771,35 +771,36 @@ def GenerateObjectTypesHeader(self):
             object_types_header += '    "%s",\n' % fixup_name
         object_types_header += '};\n'
 
+        # Key creation for map comprehensions to convert from k<Name> to VK<Name> helper
+        to_key = lambda regex, raw_key: re.search(regex, raw_key).group(1).lower().replace("_","")
```
mikes-lunarg: I think if you are going to give it a name, you might as well use def syntax

PR #2500, opened 2018-03-19
Fix Superfluous and Erroneous inttypes.h Includes

liam-middlebrook:
The files that explicitly include <inttypes.h> are C++ files. The
convention for including C header files in C++ programs is to use a
special C++ header file that wraps the C header file. For example
<inttypes.h> would be included as <cinttypes>.

In vk_layer_logging.h:34 <cinttypes> is included.

In buffer_validation.cpp and parameter_validation_utils.cpp
"vk_layer_logging.h" is directly included.

In core_validation.cpp "vk_layer_logging.h" is indirectly included via
core_validation.h

This change removes the invalid and superfluous use of C header files from the
aforementioned C++ files.

Reviewed-by: dkoch
Reviewed-by: ddadap

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -22,7 +22,6 @@
 // Allow use of STL min and max functions in Windows
 #define NOMINMAX
 
-#include <inttypes.h>
 #include <sstream>
 #include <string>
 
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -51,7 +51,6 @@
 #include <string.h>
 #include <string>
 #include <valarray>
-#include <inttypes.h>
 
 #include "vk_loader_platform.h"
 #include "vk_dispatch_table_helper.h"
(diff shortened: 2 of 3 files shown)


krOoze:
Including C standard headers in C++ program is not erroneous. (The diference is only guarantees about the `std` namespace &mdash; C style headers guarantees global visibility, and C++ style guarantees `std::` namespaced stuff &mdash; but both is valid thing to include).

Since we are (probably) using only the macros (which are always guaranteed to have global visibility), we should indeed rather use `<cinttypes>`.

You also claim it is superfluous. How so?

liam-middlebrook:
It's erroneous for a range of older versions of glibc, I should have clarified that as opposed to erroneous altogether. I've tested against 2.12 and 2.17, but was not able to reproduce the issue with 2.25.

It's superfluous because <cinttypes> was already being included in the `vk_layer_logging.h` file.

mark-lunarg:
Fixed commit message for format (contributing.md) and pushed to Master.  Thanks!

PR #2508, opened 2018-03-23
Determine need for line width dynamic state based on topology reaching rasterizer, not input topology

chrisforbes:


chrisforbes:
Squashed down so we won't preserve my missteps forever.

PR #2510, opened 2018-03-23
Various fixes for memory leaks

polarina:


layers/object_tracker_utils.cpp
```diff
@@ -312,13 +312,22 @@ void CreateSwapchainImageObject(VkDevice dispatchable_object, VkImage swapchain_
 
 void DeviceReportUndestroyedObjects(VkDevice device, VulkanObjectType object_type, enum UNIQUE_VALIDATION_ERROR_CODE error_code) {
     layer_data *device_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
-    for (auto item = device_data->object_map[object_type].begin(); item != device_data->object_map[object_type].end();) {
-        ObjTrackState *object_info = item->second;
+    for (auto &item : device_data->object_map[object_type]) {
```
tobine: I think item and ObjTrackState below can now both be "const", yes?

polarina: Correct. Fixed.

layers/object_tracker.h
```diff
@@ -223,6 +225,28 @@ void CreateObject(T1 dispatchable_object, T2 object, VulkanObjectType object_typ
     }
 }
 
+template <typename T1, typename T2>
+void DestroyObjectSilently(T1 dispatchable_object, T2 object, VulkanObjectType object_type) {
+    layer_data *device_data = GetLayerDataPtr(get_dispatch_key(dispatchable_object), layer_data_map);
+
+    auto object_handle = HandleToUint64(object);
+    assert(object_handle != VK_NULL_HANDLE);
```
daveh-lunarg: Is it necessary to assert() here?  My expectation for a function named Destroy*Silently() is to destroy something if it exists, or exit without complaint if it doesn't exist.

daveh-lunarg: I not overly bothered by the name, just wanted to be clear on the intention that the calling code has to ensure that it's passing a valid handle.

LGTM also.  Barring any other discussion I'll push this by EOD.

PR #2514, opened 2018-03-27
Fix CopyImage VUIDs broken by recent spec update

daveh-lunarg:
Changes to core spec between 1.0.68 and .71 modified two existing VUIDs (VUID-VkImageCopy-srcImage-01785 & 01786) by splitting them each into two, one new and _one that re-uses the existing VUID text_. This is a fix to update the existing checks and add the newly created ones, and update the tests to match.

_This points out a problem with specification changes that re-use VUIDs.  The existing checks/tests for the previous VUID language remain consistent with each other (though both wrong according to the updated spec), and there is no automated way to notice they are now broken. The database records them as implemented and the existing test continues to pass. (The newly-created VUIDs are not a problem because they show up in the database as new and unimplemented.)_


mikes-lunarg:
The spec style guide does say that semantic VU changes should remove the VUID so that a new one can be assigned.  A gentle reminder to spec editors might be necessary if this continues to happen.

https://www.khronos.org/registry/vulkan/specs/1.1/styleguide.html#_updating_vuid_tags_when_valid_usage_statements_change

Issue #2517, opened 2018-03-27
ABA issue with freeing and allocation of command buffers and destruction of command pools in core validation layer


fduranleau-gl:
We have stumbled recently on a bug in core validation layer. The bug lies in the layer's DestroyCommandPool API entry function. Here is the function:

```
VKAPI_ATTR void VKAPI_CALL DestroyCommandPool(VkDevice device, VkCommandPool commandPool, const VkAllocationCallbacks *pAllocator) {
    layer_data *dev_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
    unique_lock_t lock(global_lock);
    bool skip = PreCallValidateDestroyCommandPool(dev_data, commandPool);
    if (!skip) {
        lock.unlock();
        dev_data->dispatch_table.DestroyCommandPool(device, commandPool, pAllocator);
        lock.lock();
        if (commandPool != VK_NULL_HANDLE) {
            PostCallRecordDestroyCommandPool(dev_data, commandPool);
        }
    }
}
```

The problem is that the command pool's state is erased after destroying the command pool in the next layer while unlocked. After this line executed
```
        dev_data->dispatch_table.DestroyCommandPool(device, commandPool, pAllocator);
```
any other thread can allocate any number of command buffers, and it might (and will, we did get it) happen that memory freed by the above line is reused, and thus it's state will reuse the existing entry in the layer's command buffer map . So when we get to
```
            PostCallRecordDestroyCommandPool(dev_data, commandPool, cp_state);
```
it will destroy the command buffer's state, leaving the created command buffer with no state, and later on, we hit an assert (or a crash) about the command buffer's state being null.

I do not know why clearing the layer's command pool data must be done after calling the next layer's function. Looking in what PostCallRecordDestroyCommandPool does, it does not seem to require that order. Then again I might miss something. On our end, I replaced the code inside `if (!skip)` for
```
    //...
    if (!skip) {
        if (commandPool != VK_NULL_HANDLE) {
            PostCallRecordDestroyCommandPool(dev_data, commandPool, cp_state);
        }
        lock.unlock();
        dev_data->dispatch_table.DestroyCommandPool(device, commandPool, pAllocator);
    }
    //...
```
and aside from the fact that the function now does not bare the right name (rename it as `PreCallRecordDestroyCommandPool`?), it solved our problem. It also follows a normal reverse-order destruction sequence.

[...]

PR #2518, opened 2018-03-27
layers: Fix command buffer allocation ABA bug (issue #2517)

fduranleau-gl:
In core validation layer's DestroyCommandPool API entry function, the
ext layer's API call is executed outside of the lock, and before the
command pool's command buffers are erased from the layer's command
buffer map. That means that between next layer's call and the time the
lock is regained, any other thread can allocate other command buffers.
In particular, it may reuse the memory freed the command, and then the
core validation layer will not insert a new entry in the command buffer
map. So after the lock in regained and the command pool's command
buffers are erased from the map, leaving those allocated with no state,
leading to either crashes or assertions later on. The fix is to move the
release of command buffers before releasing the lock the first, before
calling the next layer's function, such that they will be removed from
the map before any other thread might reuse those addresses.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -4519,9 +4519,7 @@ VKAPI_ATTR void VKAPI_CALL DestroyCommandPool(VkDevice device, VkCommandPool com
     unique_lock_t lock(global_lock);
     bool skip = PreCallValidateDestroyCommandPool(dev_data, commandPool);
     if (!skip) {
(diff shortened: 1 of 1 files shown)


mark-lunarg:
Looking at this now.  The CI failure seems spurious.  Also, if this is a real problem I'll need to update DestroyDescriptorPool similarly.

tobine:
Fixes #2517

Issue #1986, opened 2017-08-06
Changing image layouts in release/acquire image barriers triggers false positives.


Themaister:
When using two queues with different families, it is possible to trigger a misleading error about image layouts.

E.g. in a TRANSFER queue
```
srcQueue = TRANSFER
dstQueue = GRAPHICS
oldLayout = TRANSFER_DST_OPTIMAL
newLayout = SHADER_READ_ONLY_OPTIMAL
srcAccess = TRANSFER_WRITE
dstAccess = 0 // semaphore takes care of it
```
this signals a semaphore which graphics waits on, and needs to do the same layout transition. The transition only actually happen once according to spec.

Graphics:
```
srcQueue = TRANSFER
dstQueue = GRAPHICS
oldLayout = TRANSFER_DST_OPTIMAL
newLayout = SHADER_READ_ONLY_OPTIMAL
srcAccess = 0 // semaphore takes care of it
dstAccess = SHADER_READ_BIT
```
The acquire barrier triggers an error where the command buffer expects TRANSFER_DST_OPTIMAL (because it's transitioning from it), but the transition actually hasn't happened yet.

```
[Vulkan]: Error: DS: Cannot submit cmd buffer using image (0xb6) [sub-resource: aspectMask 0x1 array layer 0, mip level 0], with layout VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL when first use is VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL.
```
The release operation probably should not change the layout for the purposes of the tracker.

chrisforbes:
Looking into this as already investigating some other bad behavior in the same area.

mark-lunarg:
Duplicate of #865? Will leave them both open for now since @chrisforbes bit.

jzulauf-lunarg:
@chrisforbes -- is this still on your active TODO list?  I'm starting to look at #1550 and I don't want to duplicate efforts.

chrisforbes:
@jzulauf-lunarg I don't have a lot of time to look at validation at the moment, but #2017 is pretty close to what you want for this.

jzulauf-lunarg:
@chrisforbes  I'll pull and rebase from #2017, and move forward with a new PR.

jzulauf-lunarg:
Should be fixed by 2521 above

Issue #1550, opened 2017-03-09
Cross-queue memory barriers not validated properly


phaulos:
Seems that validation layers have some limitations in validating cross-queue barriers. For example `dEQP-VK.synchronization.op.multi_queue.fence.write_update_buffer_read_ubo_vertex.buffer_16384_exclusive` is failing with:

```
ERROR: vkCmdPipelineBarrier(): dstStageMask flag VK_PIPELINE_STAGE_VERTEX_SHADER_BIT is not compatible with the queue family properties of this command buffer. For more information refer to Vulkan Spec Section '6.5. Pipeline Barriers' which states 'Any pipeline stage included in srcStageMask or dstStageMask must be supported by the capabilities of the queue family specified by the queueFamilyIndex member of the VkCommandPoolCreateInfo structure that was used to create the VkCommandPool that commandBuffer was allocated from, as specified in the table of supported pipeline stages.' (https://www.khronos.org/registry/vulkan/specs/1.0-extensions/xhtml/vkspec.html#VkDependencyFlagBits) (code 0x000009d1 from DL at CommandBuffer:8286)
```

This command is targeted at command buffer for queue index 1 which is transfer-only queue (and thus doesn't support VK_PIPELINE_STAGE_VERTEX_SHADER_BIT). However `VkBufferMemoryBarrier::dstQueueFamilyIndex` points to queue index 0 which is graphics-capable queue. Seems that `dstStageMask` is incorrectly checked against queue associated with the command buffer, not the `dstQueueFamilyIndex`.

krOoze:
Actually `dstStageMask` or `srcStageMask` should be completely ignored on transfer barriers.
I think this is esentially another duplicate of #865 and #1986.

jzulauf-lunarg:
Should be fixed by 2521 above.

Issue #865, opened 2016-08-21
Error when  moving exclusive image ownership between queue families with simultaneous image transition


ParticlePeter:
The issue is discussed in detail [here](https://forums.khronos.org/showthread.php/13153-Issue-moving-exclusive-image-ownership-between-queue-families-and-synchronization).
Summarizing the scenario:
1.) write into an `VK_SHARING_MODE_EXCLUSIVE` image from graphics queue family
2.) transition image to another queue family with compute capability for reading

[As per spec](https://www.khronos.org/registry/vulkan/specs/1.0-wsi_extensions/xhtml/vkspec.html#resources-sharing) two similar pipeline image barriers are required, one in the source queue for releasing the ownership and one in the destination queue for acquiring the ownership. The spec does not mention that source and destination layout must not differ, hence this might be an spec issue rather then a validation issue.
However, if the layouts do differ (in my case layout transition from `VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL` to `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL`) validation reports this error (layout transition applied twice?):

`Cannot submit cmd buffer using image (0xa) [sub-resource: aspectMask 0x1 array layer 0, mip level 0], with layout VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL when first use is VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL.`

No error is reported when the layout transition happens elsewhere (render pass) and old and new layout are both and destination layout `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL` in both of the pipeline barriers.

lentinem:
I'm a bit confused. Is the layout transiton applied twice? If that is the case then the error is correct as you cannot transition an image from COLOR_ATTACHMENT if it's in layout SHADER_READ_ONLY. Otherwise it would help to have a repo case so we can track down what is happening. Thanks!

ParticlePeter:
My example is inside a D abstraction layer and I doubt you want to read it. Let's take a look into the spec first and try to clarify before I create an example.
[Per Spec: Resource Sharing Mode](https://www.khronos.org/registry/vulkan/specs/1.0-wsi_extensions/xhtml/vkspec.html#resources-sharing), in particular this sentence:

[...]

PR #2521, opened 2018-03-28
Eliminate false positives in queue family ownership transitions / add positive tests

jzulauf-lunarg:
Two classes of false positives are address

1) Incorrectly validated src(dst)StageMask on acquire(release) operations
2) doubly accounting image layout transitions during release/acquire barrier pairs (from @chrisforbes)

This is done by detecting the type of barrier operations and gating validation based on they operation type.  For 1) this is at command granularity, for 2) at barrier granularity

In order to add tests, the vk_testing and Vk*Obj framework was enhanced w.r.t. queue and command buffer, while maintaining (with one, small, addressed exception) interface compatibility.

Addresses issues: #1986, #1550, #865, #2522

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -8051,7 +8051,7 @@ VKAPI_ATTR void VKAPI_CALL CmdWaitEvents(VkCommandBuffer commandBuffer, uint32_t
             }
             cb_state->eventUpdates.emplace_back(
                 [=](VkQueue q) { return validateEventStageMask(q, cb_state, eventCount, first_event_index, sourceStageMask); });
-            TransitionImageLayouts(dev_data, commandBuffer, imageMemoryBarrierCount, pImageMemoryBarriers);
+            TransitionImageLayouts(dev_data, cb_state, imageMemoryBarrierCount, pImageMemoryBarriers);
         }
     }
     lock.unlock();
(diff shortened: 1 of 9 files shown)


tests/vktestbinding.cpp
```diff
@@ -395,17 +400,25 @@ void Queue::submit(const std::vector<const CommandBuffer *> &cmds, Fence &fence)
...
+    return result;
 }
 
-void Queue::submit(const CommandBuffer &cmd, Fence &fence) { submit(std::vector<const CommandBuffer *>(1, &cmd), fence); }
+VkResult Queue::submit(const CommandBuffer &cmd, const Fence &fence, bool expect_success) {
+    return submit(std::vector<const CommandBuffer *>(1, &cmd), fence, expect_success);
+}
 
-void Queue::submit(const CommandBuffer &cmd) {
+VkResult Queue::submit(const CommandBuffer &cmd, bool expect_success) {
```
jzulauf-lunarg: Adding result information as the new callers are interested in it.

tests/vktestbinding.h
```diff
@@ -213,7 +214,11 @@ class Device : public internal::Handle<VkDevice> {
     const std::vector<Queue *> &graphics_queues() const { return queues_[GRAPHICS]; }
     const std::vector<Queue *> &compute_queues() { return queues_[COMPUTE]; }
     const std::vector<Queue *> &dma_queues() { return queues_[DMA]; }
-    uint32_t queue_family_without_capabilities(VkQueueFlags capabilities);
+
+    typedef std::vector<std::unique_ptr<Queue>> QueueFamilyQueues;
+    typedef std::vector<QueueFamilyQueues> QueueFamilies;
+    const QueueFamilyQueues &queue_family_queues(uint32_t queue_family) const;
```
jzulauf-lunarg: the style doesn't match Google C++, but the vk_testing namespace has been kept consistent in it's non-conformance until now (while Vk*Obj is an amalgam) -- so staying consistent as well.

layers/core_validation.cpp
```diff
@@ -8067,8 +8106,10 @@ static bool PreCallValidateCmdPipelineBarrier(layer_data *device_data, GLOBAL_CB
                                               uint32_t bufferMemoryBarrierCount, const VkBufferMemoryBarrier *pBufferMemoryBarriers,
                                               uint32_t imageMemoryBarrierCount, const VkImageMemoryBarrier *pImageMemoryBarriers) {
     bool skip = false;
-    skip |= ValidateStageMasksAgainstQueueCapabilities(device_data, cb_state, srcStageMask, dstStageMask, "vkCmdPipelineBarrier",
-                                                       VALIDATION_ERROR_1b80093e);
+    auto barrier_op_type = ComputeBarrierOperationsType(device_data, cb_state, bufferMemoryBarrierCount, pBufferMemoryBarriers,
```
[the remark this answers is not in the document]
jzulauf-lunarg: It think it *would* be safe as the only checks that are skipped are for the barriers with src and dst queue families, but I'll clean this up to be less fragile.

jzulauf-lunarg:
@mikes-lunarg -- addressed your comments.

PR #2484, opened 2018-03-13
Build fails on Linux/x32 due to 32-bit pointers but 64-bit registers.

jduerstock:
On Linux/x32 systems, it's not enough to test for __x86_64__.
Pointers are 32-bit, as represented by __ILP32__, so it must be
tested as well.

PR #2523, opened 2018-03-28
layers: Correct non-dispatchable type handling

jzulauf-lunarg:
Update threading layer non-dispatchable type handling to reflect updates
to the vulkan header for Linux/x32 memory model support.  Includes new
static asserts to detect future mismatches.

This in response to [Pull Request 2484](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/2484) -- as the original poster does not have a signed CLA.  The updated #define logic is copied *not* from 2484, but from vulkan_core.h, and then clang formatted.

As I don't have access to x32 systems pinging the author of 2484.

Tagging @cnorthrop to ensure no Android issues. (e.g. does Android support Arm/ILP32 or do some other VK_DEFINE_NON_DISPATCHABLE_HANDLE magic that we'd need to be aware of?)

### DIFF
--- a/layers/threading.h
+++ b/layers/threading.h
@@ -26,11 +26,20 @@
 #include "vk_layer_config.h"
 #include "vk_layer_logging.h"
 
-#if defined(__LP64__) || defined(_WIN64) || defined(__x86_64__) || defined(_M_X64) || defined(__ia64) || defined(_M_IA64) || \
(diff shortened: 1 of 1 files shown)


jduerstock:
This fixes my build problems in my x32 chroot.  Thank you.

PR #2531, opened 2018-03-29
Add new YCbCr formats and CmdCopyImage checks

daveh-lunarg:
The VK_KHR_sampler_ycbcr_conversion extension added new formats, which then became core with Vulkan 1.1.  This PR adds support for the new single- and multi-plane formats within the layer format utility fxns, and adds checks for VUIDs added or modified within CmdCopyImage.  Includes tests for new VUIDs, and adds guards to existing tests for modified VUIDs.

This is the first chunk of what will be several PRs related to this extension, dealing with ~20 of the 128 VUIDs added or modified.


tobine:
Couple of test fails

`[ RUN      ] VkLayerTest.CopyImageSinglePlane422Alignment
/home/travis/build/KhronosGroup/Vulkan-LoaderAndValidationLayers/tests/vkrenderframework.cpp:968: Failure
Failed
VkImageObj::init() error: unsupported tiling configuration. Usage: 0x3, supported optimal features: 0
/home/travis/build/KhronosGroup/Vulkan-LoaderAndValidationLayers/tests/layer_validation_tests.cpp:20328: Failure
Value of: image_422.initialized()
  Actual: false
Expected: true
[  FAILED  ] VkLayerTest.CopyImageSinglePlane422Alignment (31 ms)
[ RUN      ] VkLayerTest.CopyImageMultiplaneAspectBits
/home/travis/build/KhronosGroup/Vulkan-LoaderAndValidationLayers/tests/vkrenderframework.cpp:968: Failure
Failed
VkImageObj::init() error: unsupported tiling configuration. Usage: 0x3, supported optimal features: 0
/home/travis/build/KhronosGroup/Vulkan-LoaderAndValidationLayers/tests/layer_validation_tests.cpp:20455: Failure
Value of: mp3_image.initialized()
  Actual: false
Expected: true
[  FAILED  ] VkLayerTest.CopyImageMultiplaneAspectBits (31 ms)`

daveh-lunarg:
There's some guesswork involved, but I think Travis (mock) is providing conflicting answers about whether a format is supported, depending on how you ask.  i.e. vkGetPhysicalDeviceFormatProperties() provides no feature bits, although vkGetPhysicalDeviceImageFormatProperties() returns VK_SUCCESS.

I've updated the tests to use a more comprehensive format check anyway, actually a good incentive to consolidate what had become a handful of one-offs.

tests/layer_validation_tests.cpp
```diff
@@ -20803,7 +21069,10 @@ TEST_F(VkLayerTest, CopyImageAspectMismatch) {
 
     // Src and dest aspect masks don't match
     copyRegion.dstSubresource.aspectMask = VK_IMAGE_ASPECT_STENCIL_BIT;
-    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_09c00112);
+    bool ycbcr = (DeviceExtensionEnabled(VK_KHR_SAMPLER_YCBCR_CONVERSION_EXTENSION_NAME) ||
+                  (m_device->props.apiVersion >= VK_API_VERSION_1_1));
+    UNIQUE_VALIDATION_ERROR_CODE vuid = (ycbcr ? VALIDATION_ERROR_09c00c1e : VALIDATION_ERROR_09c00112);
```
krOoze: I have bit of an issue with tests being written this way. It tests completely different things on different machines.

On 1.1 `VALIDATION_ERROR_09c00112` test will get unnecessarily skipped , and on 1.0 impl the `VALIDATION_ERROR_09c00c1e` will get silently skipped.

In the long-term, should tests be written like this?

daveh-lunarg: 09c00112 is _replaced_ by 09c00c1e in the sampler_YCbCr_conversion extension, which then moved to core in Vulkan 1.1.  So, based on the driver version, only one of these checks may be run, and the other is undefined.

The valid usage statements are:

```09c00112: "The aspectMask member of srcSubresource and dstSubresource must match" ```

```09c00c1e: "If neither the calling command's srcImage nor the calling command's dstImage has a multi-planar image format then the aspectMask member of srcSubresource and dstSubresource must match" ```

krOoze: > So, based on the driver version, only one of these checks may be run, and the other is undefined.

Not true for 1.1 impl which can run both.
Besides, historically if a test of a check cannot be performed, it writes something like "test skipped because". Which informs person messing with that check, that he needs to test his changes on a different machine (or on mock_icd with relevant profile).

That's why am I asking if tests should be written like this going forward. There is obvious downside to doing so.

> The valid usage statements are:

Completelly different, with likely different check implementation. Testing one does not imply the second is covered by tests as well. It is irrelevant if it feels like one "replaces" the other. By definition. different VUID means different VU semantics.

jzulauf-lunarg: @krOoze  -- Dave's right... looking at the documentation source 

VUID-VkImageCopy-aspectMask-00137 (aka 09c00112)
VUID-VkImageCopy-srcImage-01551 (aka 09c00c1e) 

are guarded by complementary [copies.txt](ifdefs https://github.com/KhronosGroup/Vulkan-Docs/blob/master/doc/specs/vulkan/chapters/copies.txt#L539)

krOoze: @jzulauf-lunarg I never said there are not. That is **not** the point.
The point is 1.1 implementation can run both branches.

It is really quite similar to extensions. We test with both extension on, and with it off (to test both branches of the ifdef). By your logic, we would only need to run with all extensions enabled, because the ifdefs are complementary. That is not the case.

daveh-lunarg: The '112 VUID is _replaced by_ the '1ce VUID with the sampler_YCbCr_conversion extension, which became core with V1.1. They have an identical condition, with the later one adding only a caveat that the images not be multi-plane. Since prior to the sampler_YCbCr extension it was also true that the image must not be multi-plane, it's clear that these two VUIDs are exactly identical semantically. It's true they are checked in different places within buffer_validation, but the implementation is functionally identical. 

For any given combination of instance + device + extension, _exactly one_ of these two VUIDs will be in effect.  __Never both.__ In this case it is acceptable to use the same test for differing VUIDs that check an identical condition, and this will be a common pattern going forward, post-V1.1.

krOoze: > It's true they are checked in different places within buffer_validation, but the implementation is functionally identical.

Meaning if one of the places have a bug, the tests will potentially pretend it passes anyway.

> For any given combination of instance + device + extension, exactly one of these two VUIDs will be in effect. Never both.

Now you mention it, we do test extensions separately. Only one of two VUIDs is in effect too, **yet** we test with core, and with extension enabled to test **both** effects.

At this point I am just repeating myself.
If you say it is tolerated, and how things are to be done in future regarding minor versions, I take your word for it.

daveh-lunarg: @krOoze, perhaps you're missing some historical context wrt layer test requirements. The only hard rule is that every VUID must have a test, and the test must run correctly on whatever combination of device, API version, extension, etc. if finds itself on.  If the test cannot succeed in a given environment, because of a missing format or extension or whatever, it must detect so and exit gracefully.

Beyond that, there is no _requirement_ to force different api level or extension scenarios for a given VUID, although it's fine if a test author decides such would be beneficial. Such a requirement would seem to lead to geometrically increasing test code with every api version and extension.

It's also valid to combine, sub-class, or otherwise optimize tests in any way that doesn't violate the primary rule. It seems to me you'd prefer this specific test to be two, one that runs in one environment and one that runs in the other. That would not be incorrect, but given the either/or nature of these two VUIDs it seems natural to combine them, to me at least. I believe there is absolutely no loss of test coverage in doing so. If you do think so, can you specify a combination of api version and extension that would be missed, or report incorrectly?

krOoze: > it must detect so and exit gracefully.

Is skipping VUID test silently considered graceful exit?

krOoze: > If you do think so, can you specify a combination of api version and extension that would be missed, or report incorrectly?

Yes,
1.1 impl\testmachine -> would pass even if there was bug in `VALIDATION_ERROR_09c00112`, because it tests `VALIDATION_ERROR_09c00c1e` only, yet VUDB would claim both cases are tested

1.0 impl\testmachine -> would pass while even if there was bug in `VALIDATION_ERROR_09c00c1e` and not inform the check is skipped because of lack of support, yet VUDB would claim both cases are tested

mark-lunarg: Seems like, as long as there is test coverage for both platforms, it can be said that both cases *are* tested. I'm cool with this one.

krOoze: @mark-lunarg In all likelihood all testmachines (usually desktop) will be 1.1. So one of the platforms will practically never be tested (while it easily can, as 1.1 impl must support 1.0 apps).
I suppose dev_sim will eventually catch any such problems on the CI server...

Now core tests are already done and separate and only 1.1 specific stuff needs to be done. If 1.1 (and 1.2 and 1.3) stuff gets mixed within core tests that will be hard to undo later if so decided.

daveh-lunarg: >Yes,
>1.1 impl\testmachine -> would pass even if there was bug in VALIDATION_ERROR_09c00112, because it tests VALIDATION_ERROR_09c00c1e only, yet VUDB would claim both cases are tested

Which would be fine, because in this scenario 09c00112 does not exist.  It is not defined in the spec, and cannot and must not be checked.

>1.0 impl\testmachine -> would pass even if there was bug in VALIDATION_ERROR_09c00c1e and not inform the check is skipped because of lack of support, yet VUDB would claim both cases are tested

Similarly.  Regarless of whether you had a 1.1 device, or a 1.1-capable device but requested 1.0, then VALIDATION_ERROR_09c00c1e __does not exist__.

>1.0 impl\testmachine + VK_KHR_SAMPLER_YCBCR_CONVERSION extension -> not tested at all

I can't understand why you say that. The test runs, unconditionally. In this case it would provoke VALIDATION_ERROR_09c00c1e.

daveh-lunarg: >Now impl of core tests are already done and separate from everything else and only 1.1 specific stuff needs to be done.

This is a pleasant thought, but far from true.

>If 1.1 (and 1.2 and 1.3) stuff gets mixed this way within core tests that will be hard to undo and separate later, if so decided.

Yes. This is absolutely a concern, and not at all unique to this project. In our discussions locally we've come to agreement that the idea of multiple, eventually many, simultaneous repos of validation code all requiring support and updating is more problematic than a single repo with increasingly complex validation logic. But neither option is very appealing.

krOoze:
> Which would be fine, because in this scenario 09c00112 does not exist. It is not defined in the spec, and cannot and must not be checked.

Except in the Vulkan 1.0 spec. Which applies to apps that claim they are Vulkan 1.0 apps. Since we are validating apps not impls, we have to use the spec that is relevant to the given app. And tests would/should just mirror that.

I should just make an Issue. I did not expect this to be controversial and it is not specific to this PR.
For now it is important to silence the error, so you should merge this anyway.

tests/layer_validation_tests.cpp
krOoze: > Which would be fine, because in this scenario 09c00112 does not exist. It is not defined in the spec, and cannot and must not be checked.

Except in the Vulkan 1.0 spec. Which applies to apps that claim they are Vulkan 1.0 apps. Since we are validating apps not impls, we have to use the spec that is relevant to the given app. And tests would/should just mirror that.

I should just make an Issue. I did not expect this to be controversial and it is not specific to this PR.
For now it is important to silence the error, so you should merge this anyway.

PR #2536, opened 2018-03-30
Fix multi-plane aspect bit check

### LINKED ISSUE #2527: Multiplanar aspect flags are still reporting an error
This is a duplicate of #2378 but because I can't reopen it I'm creating a new one.

The issue is that whenever the aspect flags are checked, the validation layer does not yet recognize the multiplanae aspect flags (VK_IMAGE_ASPECT_PLANE_0_BIT and so on) as valid and prints the following error:

> vkCreateImageView(): Color image formats must have the VK_IMAGE_ASPECT_COLOR_BIT set. The spec valid usage text states 'aspectMask must be a valid combination of VkImageAspectFlagBits values' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkImageSubresource-aspectMask-parameter)

>vkCmdPipelineBarrier(): Color image formats must have the VK_IMAGE_ASPECT_COLOR_BIT set. The spec valid usage text states 'aspectMask must be a valid combination of VkImageAspectFlagBits values' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkImageSubresource-aspectMask-parameter)

The issue occurs with current git master (37beb71e6fdb90d1bf40089abf98bb3c).

daveh-lunarg:
The FormatIsColor() util fxn classified anything not depth or stencil as 'color', and applied the aspect bit checks correspondingly. This PR segregates multi-plane formats from the catch-all color classification and adds multi-plane specific aspect bit checks for them.

Resolves #2527.

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -3147,6 +3147,18 @@ bool ValidateImageAspectMask(layer_data *device_data, VkImage image, VkFormat fo
                             HandleToUint64(image), VALIDATION_ERROR_0a400c01,
                             "%s: Stencil-only image formats can have only the VK_IMAGE_ASPECT_STENCIL_BIT set.", func_name);
         }
+    } else if (FormatIsMultiplane(format)) {
+        VkImageAspectFlags valid_flags = VK_IMAGE_ASPECT_COLOR_BIT | VK_IMAGE_ASPECT_PLANE_0_BIT | VK_IMAGE_ASPECT_PLANE_1_BIT;
+        if (3 == FormatPlaneCount(format)) {
+            valid_flags = valid_flags | VK_IMAGE_ASPECT_PLANE_2_BIT;
+        }
+        if ((aspect_mask & valid_flags) != aspect_mask) {
+            skip |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT,
+                            HandleToUint64(image), VALIDATION_ERROR_0a400c01,
+                            "%s: Multi-plane image formats may have only VK_IMAGE_ASPECT_COLOR_BIT or VK_IMAGE_ASPECT_PLANE_n_BITs "
+                            "set, where n = [0, 1, 2].",
+                            func_name);
+        }
     }
     return skip;
 }
--- a/layers/vk_format_utils.h
+++ b/layers/vk_format_utils.h
@@ -119,10 +119,12 @@ VK_LAYER_EXPORT VkFormatCompatibilityClass FormatCompatibilityClass(VkFormat for
 VK_LAYER_EXPORT VkDeviceSize SafeModulo(VkDeviceSize dividend, VkDeviceSize divisor);
 
 static inline bool FormatIsUndef(VkFormat format) { return (format == VK_FORMAT_UNDEFINED); }
-static inline bool FormatIsColor(VkFormat format) { return !(FormatIsUndef(format) || FormatIsDepthOrStencil(format)); }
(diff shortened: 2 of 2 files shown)


layers/buffer_validation.cpp
```diff
@@ -3238,6 +3238,25 @@ bool ValidateImageAspectMask(layer_data *device_data, VkImage image, VkFormat fo
                             "%s: Stencil-only image formats can have only the VK_IMAGE_ASPECT_STENCIL_BIT set. %s", func_name,
                             validation_error_map[VALIDATION_ERROR_0a400c01]);
         }
+    } else if (FormatIsMultiplane(format)) {
```
jzulauf-lunarg: This handling is somewhat different (at least for barriers) if the image is disjoint, where more than one can be specified... https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VUID-VkImageMemoryBarrier-image-01672

daveh-lunarg: Y'know, back in my day a man'd do one kind of aspectMask validation on an image, whether he was settin' up a memory barrier or creatin' an imageView.  Didn't matter, we just had one kind and it was good enough for whatever you needed. Then these kids come along with their multi-plane images and YCbCr conversions.....   bah! 

Anyway, good catch @jzulauf-lunarg! I'll take this one back to the drawing board...

daveh-lunarg:
For purpose of resolving this issue, it's sufficient just to exclude the multi-plane formats from categorization by the FormatIsColor() util function.  I've loosened the multi-plane checks in ValidateImageAspectMask() to just verify that all aspectMask bits are within the overall allowed subset.  There are specific VUIDs that capture the different requirements for barriers and image views that (will, soon) provide the domain-specific validation of the aspect mask.

layers/buffer_validation.cpp
```diff
@@ -3147,6 +3147,18 @@ bool ValidateImageAspectMask(layer_data *device_data, VkImage image, VkFormat fo
                             HandleToUint64(image), VALIDATION_ERROR_0a400c01,
                             "%s: Stencil-only image formats can have only the VK_IMAGE_ASPECT_STENCIL_BIT set.", func_name);
         }
+    } else if (FormatIsMultiplane(format)) {
```
jzulauf-lunarg: This version has the benefit of brevity -- however, when you flesh out the rest of the domain specific checks, it's going to get bigger.  I think this is the right balance of providing initial coverage while avoiding "false positives".

layers/buffer_validation.cpp
```diff
@@ -3147,6 +3147,18 @@ bool ValidateImageAspectMask(layer_data *device_data, VkImage image, VkFormat fo
...
+        if (3 == FormatPlaneCount(format)) {
+            valid_flags = valid_flags | VK_IMAGE_ASPECT_PLANE_2_BIT;
+        }
+        if ((aspect_mask & valid_flags) != aspect_mask) {
+            skip |= log_msg(report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_EXT,
+                            HandleToUint64(image), VALIDATION_ERROR_0a400c01,
+                            "%s: Multi-plane image formats may have only VK_IMAGE_ASPECT_COLOR_BIT or VK_IMAGE_ASPECT_PLANE_n_BITs "
+                            "set, where n = [0, 1, 2].",
+                            func_name);
+        }
```
jzulauf-lunarg: the [0... n] range could be set with a const char assignment overwritten in the '(3 ==' clause to get a more accurate message, but I don't know if it's worth re-spinning the PR for.

PR #2539, opened 2018-04-02
Correct apiVersion validation checks

### LINKED ISSUE #2515: CreateInstance warning for apiVersion 0
Calling vkCreateInstance with apiVersion 0 will trigger a warning in the validation layers. 

> Warning: Unrecognized CreateInstance->pCreateInfo->pApplicationInfo.apiVersion number -- (0x00000000) assuming VK_API_VERSION_1_1.

(source: https://github.com/gfx-rs/gfx/pull/1890#discussion_r177464526)

The Vulkan specification states that:
> Providing    a    NULL VkInstanceCreateInfo::pApplicationInfo or    providing    an apiVersion  of      0     is      equivalent      to      providing      an      apiVersion  of VK_MAKE_VERSION(1,0,0)

Therefore, it should be fine to pass 0 here.

jzulauf-lunarg:
Fixes #2515.

Error reporting for apiVersion validation checks were incorrect w.r.t. 1.1 specification, these have been corrected. In support of this:

* Unit tests were added for apiVersion zero , ~~warning, and error conditions~~  ( _removed the later two tests due to configuration sensistivity_ )
* Unit test framework added temp debug callback support during CreateInstance
* Duplicate error messages from desktop loader were eliminated (to allow unit test)
* Desktop loader apiVersion "fixup" code was made more robust (to support the additional unit test  abuse)

The error and warning unit test cases are disabled for Android. (Per @lenny-lunarg, Need @cnorthrop for the best valid (if any) way to enable this testing)


jzulauf-lunarg:
There are two stray files... will push version without strays.

cnorthrop:
Pinging @ianelliottus

ianelliottus:
Regarding the Android loader, it also protects 1.0 ICDs, so that they only see an apiVersion of 1.0.

layers/parameter_validation_utils.cpp
```diff
@@ -263,18 +263,29 @@ VKAPI_ATTR VkResult VKAPI_CALL vkCreateInstance(const VkInstanceCreateInfo *pCre
...
-            uint32_t specified_api_version = pCreateInfo->pApplicationInfo->apiVersion & ~VK_VERSION_PATCH(~0);
-            if (!(specified_api_version == VK_API_VERSION_1_0) && !(specified_api_version == VK_API_VERSION_1_1)) {
-                LOGCONSOLE(
-                    "Warning: Unrecognized CreateInstance->pCreateInfo->pApplicationInfo.apiVersion number -- (0x%08x) assuming "
-                    "%s.\n",
-                    pCreateInfo->pApplicationInfo->apiVersion,
-                    (api_version == VK_API_VERSION_1_0) ? "VK_API_VERSION_1_0" : "VK_API_VERSION_1_1");
+        // Note: From the spec--
+        //  Providing a NULL VkInstanceCreateInfo::pApplicationInfo or providing an apiVersion of 0 is equivalent to providing
+        //  an apiVersion of VK_MAKE_VERSION(1, 0, 0).  (a.k.a. VK_API_VERSION_1_0)
```
ianelliottus: I'm not fully sure of the thinking for the new code (at least after looking at InitFromInstanceCreateInfo()).  This code seems to be locking into there only being 1.0 and 1.1 versions.  Seems like this code and InitFromInstanceCreateInfo() will have to change again when we do 1.2.  

Beyond that, I'm not sure why there's so much code before line 276.  I realize I must be missing something (I've not looked at the validation code in a while), but I don't see how effective_api_version and api_version_nopatch will ever be different (unless the app requests 1.2 or greater, since InitFromInstanceCreateInfo clamps the version to a max of 1.1).

tests/layer_validation_tests.cpp
```diff
@@ -28254,6 +28254,48 @@ TEST_F(VkPositiveLayerTest, MultiplaneImageTests) {
...
+         "loader_add_to_layer_list: Explicit layer VK_LAYER_LUNARG_core_validation is using an old API version",
+         "loader_add_to_layer_list: Explicit layer VK_LAYER_LUNARG_device_profile_api is using an old API version",
+         "loader_add_to_layer_list: Explicit layer VK_LAYER_LUNARG_object_tracker is using an old API version",
+         "loader_add_to_layer_list: Explicit layer VK_LAYER_LUNARG_parameter_validation is using an old API version"}};
+    for (auto warning : expected) {
+        m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_WARNING_BIT_EXT, warning);
+    }
+
+    app_info.apiVersion = VK_MAKE_VERSION(999U, 999U, 999U);
+    ASSERT_NO_FATAL_FAILURE(InitFramework(myDbgFunc, m_errorMonitor));
```
jzulauf-lunarg: @cnorthrop @ianelliottus  -- this line was failing in our Android CI environment... would skipping this test (or expecting failure) if  vkEnumerateInstanceVersion (on Android) returning 1.0 be sufficient to avoid an unexpected failure?

cnorthrop: Confirmed this test is failing here even on more recent Android versions.  Should this be a loader test instead of layer, given that it is driven by loader functionality?

tests/layer_validation_tests.cpp
```diff
@@ -28254,6 +28254,48 @@ TEST_F(VkPositiveLayerTest, MultiplaneImageTests) {
...
+    m_errorMonitor->VerifyFound();
+}
+
+TEST_F(VkLayerTest, ApiVersionError) {
```
cnorthrop: This test is passing for me locally.  Since it is ifdef'd out, does that mean it is failing for you?

loader/loader.c
```diff
@@ -5235,14 +5242,15 @@ VKAPI_ATTR VkResult VKAPI_CALL terminator_CreateInstance(const VkInstanceCreateI
 
         // Create an instance, substituting the version to 1.0 if necessary
         VkApplicationInfo icd_app_info;
+        uint32_t icd_version_nopatch = VK_MAKE_VERSION(VK_VERSION_MAJOR(icd_version), VK_VERSION_MINOR(icd_version), 0);
         uint32_t requested_version = pCreateInfo == NULL || pCreateInfo->pApplicationInfo == NULL ? VK_API_VERSION_1_0 : pCreateInfo->pApplicationInfo->apiVersion;
-        if (requested_version > icd_version) {
+        if ((requested_version != 0) && (icd_version_nopatch == VK_API_VERSION_1_0)) {
```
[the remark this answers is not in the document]
ianelliottus: Sorry, I just looked at the code history.  You're right.  Older versions of the Android loader just passed on apiVersion (as did the desktop loader--at least I recall reviewing a @MarkY-LunarG change; but your loader gets updated on user systems and ours doesn't).  Newer versions errored in the loader before calling the driver.  The newest (not-yet-released) Android loader downgrades the version to 1.0 for a 1.0 driver.  This newest version is what I was referring to.

tests/layer_validation_tests.cpp
ianelliottus: Hmm.  I should think about this some more.  This is an area where 1.1 changed semantics, which is unfortunate.  

Here's a quick thought.  There is a way of testing what version of loader you have (i.e. 1.0.* vs. 1.1 and beyond).  This is done by calling GIPA of the new vkEnumerateInstanceVersion global function.  If NULL, you have a 1.0 loader (which should return an error for >1.0.*).  If non-NULL, call the function and observe what version the loader supports (i.e. loader version == instance version).  You can hide this check around that logic.

tests/layer_validation_tests.cpp
jzulauf-lunarg: It is failing.  Likely our Android CI device is a 1.0 device.  I've also discovered that there are other ways in which the loader/icd/layers can interact that also result in failures/unexpected (the devsim layer for one).  

Given that sensitivity to components outside my control, I'm just going to kill these latter two tests, as they are more trouble than they are worth.  @cnorthrop / @ianelliottus  thanks for looking at this.

tests/layer_validation_tests.cpp
```diff
@@ -28254,6 +28254,14 @@ TEST_F(VkPositiveLayerTest, MultiplaneImageTests) {
     vkDestroyImage(device(), image, NULL);
 }
 
+TEST_F(VkPositiveLayerTest, ApiVersionZero) {
+    TEST_DESCRIPTION("Check that apiVersion = 0 is valid.");
+    m_errorMonitor->ExpectSuccess();
+    app_info.apiVersion = 0U;
+    ASSERT_NO_FATAL_FAILURE(InitFramework(myDbgFunc, m_errorMonitor));
```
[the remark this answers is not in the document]
daveh-lunarg: ```app_info``` is a framework member variable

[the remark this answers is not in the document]
krOoze: @daveh-lunarg Don't take me so literally.
The `pApplicationInfo == nullptr` case, which is also supposed to have equal semantics to `VK_API_VERSION_1_0`.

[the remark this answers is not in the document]
jzulauf-lunarg: @krOoze  -- Didn't want to fiddle that much with the framework just for this case.  Will gladly review any PR that wants to improve coverage here :).  Framework changes should be in their own commit within the PR.

PR #2544, opened 2018-04-03
layers:Don't overwrite pColorBlendState ptr

### LINKED ISSUE #2543: Feature Request: disable removing of blend state info on pipeline create
Hi,

the situation is as follows.

We create a new graphics pipeline, the fragment shader writes to one or more color outputs. The used pass marks all color attachments as unused via VK_ATTACHMENT_UNUSED. Leaving only depth/stencil to be used.

We always provide the blend state info, even in this case. This works fine, as expected. But if we enable in this case the debug layer, it filters in this case the blend state info of the pipeline create struct - setting the pointer to null.

This causes a crash inside some drivers (we already contacted the vendor).

Is there a possibility to disable this at runtime, currently we avoid this by using a modified version of the debug layer. But this is not a great solution to avoid this.

tobine:
Fixes #2543

When creating PSO, keep any pColorBlendState ptr value even if it
should be ignored. Only create safe_struct wrapper in case where
value shouldn't be ignored.


tobine:
Running this as a test PR to see if keeping this ignored ptr causes any issues with testing. We really shouldn't be over-writing these ptrs in validation so alternative is to just keep original value in ignored cases instead of replacing with NULL.
There are a few more such cases so if this works I'll test those out as well.

PR #2549, opened 2018-04-04
Add log_msg overload that will accept text VUIDs

mark-lunarg:
This ability is necessary to support private extension development, where numeric VUIDs are unavailable, and even text VUIDs will not appear in the VUID database.

Added an overload of log_msg that will take text VUIDs 
(e.g., **"VUID-VkDeviceQueueInfo2-queueFamilyIndex-01842"**)
in place of the usual integer VUID (e.g., **VALIDATION_ERROR_42200e64**).

A new map of the VUID strings to the old integer VUIDs has been added to vk_validation_error_messages.h.

vk_layer_logging.h was modified to handle outputting the text VUIDs _without_ changing the callbacks.  This is important since many ISVs are likely using the current callback structure.

If a text VUID is used with the DEBUG_REPORT extension, the message will look like:

```[ VUID-VkGraphicsPipelineCreateInfo-attachmentCount-00746 ] Object: VK_NULL_HANDLE (Type = 19) | vkCreateGraphicsPipelines(): Render pass (0x6) subpass 0 has colorAttachmentCount of 1 which doesn't match the pColorBlendState->attachmentCount of 0. The spec valid usage text states 'If rasterization is not disabled and the subpass uses color attachments, the attachmentCount member of pColorBlendState must be equal to the colorAttachmentCount used to create subpass' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkGraphicsPipelineCreateInfo-attachmentCount-00746)```

The DEBUG_UTILS extension handles text VUIDs natively.

If a text VUID is specified and does NOT exist in the error_message database, the message will appear as before but without the appended spec text.

### CODE REVIEW

File: scripts/spec.py [hunk 1]
```diff
@@ -188,27 +188,32 @@ def genHeader(self, header_file):
...
+        file_contents.append('\n#include <string>')
+        file_contents.append('#include <unordered_map>')
         file_contents.append('\n// enum values for unique validation error codes')
         file_contents.append('//  Corresponding validation error message for each enum is given in the mapping table below')
         file_contents.append('//  When a given error occurs, these enum values should be passed to the as the messageCode')
         file_contents.append('//  parameter to the PFN_vkDebugReportCallbackEXT function')
         enum_decl = ['enum UNIQUE_VALIDATION_ERROR_CODE {\n    VALIDATION_ERROR_UNDEFINED = -1,']
-        error_string_map = ['static std::unordered_map<int, char const *const> validation_error_map{']
+        alias_error_string_map = ['static std::unordered_map<int, char const *const> validation_error_map{']
+        vuid_error_string_map = ['static std::unordered_map<std::string, int> validation_error_text_map{']
```
tobine: These var names are confusing as they seem backwards to me. I would think "vuid_error_string_map" is a map of VUID number to error string, but it appears that's not the case.
Should we call it vuid_int_to_error_map and vuid_string_to_error_map? I'd prefer that. The map names in the c++ layer code are more clear as default name remains for int->error while "text" is inserted for string->error.

Semantically I think if we're going to refer to the unique string values as "VUID" then we should just make it VUID_int and VUID_string to avoid confusion, at least while both exist. If there's a point in the future where VUID_int no longer exists then VUID_string can become the "VUID" by default.

mark-lunarg: Sure, fixed up the names as suggested.

mark-lunarg:
Incorporated feedback and pushed to Master.

PR #2555, opened 2018-04-06
scripts: Add spec.py check for ext changes

mikes-lunarg:
When the surrounding asciidoctor conditionals of a VUID change it may
impact validation even if the VUID text does not change.  Detect this in
spec.py and flag for manual intervention.

Change-Id: Idf802955ba9bbfe911edd4114fb28bdddf5fcc9b

### DIFF
--- a/scripts/spec.py
+++ b/scripts/spec.py
@@ -160,8 +160,15 @@ class Specification:
                 self.error_db_dict[error_enum] = {'check_implemented': 'N',
                                                   'testname': 'None',
                                                   'note': ''}
-            # update database entry with data from json file
             vuid_db_data = self.error_db_dict[error_enum]
+            # check if vuid is implemented but changed extension scope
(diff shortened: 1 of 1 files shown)


tobine:
The change is fine but the hard out w/ manual intervention is tough if this happens much. Is the intended manual action to update the validation code immediately and then change DB file to have new extension text so that error is gone for re-run?

mikes-lunarg:
The idea with the hard exit is to break the update process similar to what currently happens when a VUID changes and the build or doc validator or vuid_mapping fails.  The remedy is exactly as you described.  I don't have any data for how often this happens, but my hunch is that it is rare and many of the cases will be semantically equivalent DB updates with no code changes.

If it becomes too frequent we can print a warning and NOT update the ext column.  Then there would be a periodic task to work through the warnings in batches but it wouldn't block the header update.

PR #2559, opened 2018-04-06
Remove wrap-objects test, files

mark-lunarg:
In anticipation of repo split, removing wrap objects tests that rely on running the layer_validation_tests.  There remains a wrap objects test case in the loader_validation_tests.


mark-lunarg:
We could probably change the existing LayerPresent test (and keep more of the run_loader_tests script) by using the 'test' layer that is already built instead of the validation layers.  Suggestions welcome.

mark-lunarg:
Anyone? Anyone?

cnorthrop:
I don't think there's any Android impact here.  Although we do run the loader tests on desktop to make sure we don't break anything.  May not need to do that anymore after repo split.

PR #2564, opened 2018-04-10
layers: add level/layer to ValidateImageAspectLayout log msg

cdwfs:
This helps clarify the case where a transition fails due to a specific subresource being in an unexpected layout.


cdwfs:
arrgh, self-test failures; will fix momentarily.

PR #2565, opened 2018-04-10
Add automated travis check for commit message format

mark-lunarg:
This change should check commit message formats against CONTRIBUTING.md.  PRs containing commit messages that don't follow the spec will fail the new test, but Travis is set up to allow these tests to fail -- this check is only run on PRs.  

Failing them should NOT result in a failed PR, though results can be checked by examining the Travis details for a submitted PR. For example on this PR, click on 'Show all checks' next to "All Checks Have Passed" below, then select "continuous-integration/travis-ci/pr — The Travis CI build passed ->Details".

### DIFF
--- a/scripts/check_commit_message_format.sh
+++ b/scripts/check_commit_message_format.sh
@@ -1,0 +1,102 @@
+#!/bin/bash
+# Copyright (c) 2018 Valve Corporation
+# Copyright (c) 2018 LunarG, Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Checks commit messages against project standards in CONTRIBUTING.md document
+# Script to determine if commit messages in Pull Request are properly formatted.
+# Exits with non 0 exit code if reformatting is needed.
+
+# Disable subshells
+shopt -s lastpipe
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m' # No Color
+
+# TRAVIS_COMMIT_RANGE contains range of commits for this PR
+
+# Get user-supplied commit message text for applicable commits and insert
+# a unique separator string identifier. The git command returns ONLY the
+# subject line and body for each of the commits.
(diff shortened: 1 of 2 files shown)


scripts/check_commit_message_format.sh
```diff
@@ -0,0 +1,102 @@
...
+# TRAVIS_COMMIT_RANGE contains range of commits for this PR
+
+# Get user-supplied commit message text for applicable commits and insert
+# a unique separator string identifier. The git command returns ONLY the
+# subject line and body for each of the commits.
+COMMIT_TEXT=$(git log ${TRAVIS_COMMIT_RANGE} --pretty=format:"XXXNEWLINEXXX"%n%B)
+
+# Bail if there are none
+if [ -z "${COMMIT_TEXT}" ]; then
+  echo -e "${GREEN}No source code to check for formatting.${NC}"
```
jzulauf-lunarg: Should the message say "commit message" instead of "source code"

scripts/check_commit_message_format.sh
```diff
@@ -0,0 +1,102 @@
...
+  exit 0
+elif ! echo $TRAVIS_COMMIT_RANGE | grep -q "\.\.\."; then
+  echo -e "${GREEN}No source code to check for formatting.${NC}"
```
jzulauf-lunarg: same comment.

daveh-lunarg:
What's the net result here?  PRs with poorly-formatted commit messages will fail Travis checks, but still be able to push? If so, I worry about causing confusion about whether a CI check failed "for real" or for poor commit character counting.

I'd prefer to stick to public shaming rather than using CI for this.

mark-lunarg:
Public shaming is working like this:  out of the last 250 commits, nearly 50 don't fit the guidelines.  Having/needing to shame people is not something everyone enjoys, though it'd be fantastic if a shame-meister stepped forward to manage things.  Another option is to give it up entirely, but the middle is exhausting.

karl-lunarg:
The check is executed as an additional job tagged with "allow_failures" which means that if this new job fails, it does not change the status of the overall check/build result. 

What I don't know is how this appears in the "All checks have passed" panel on this page.  If only the commit message check does not pass, I don't know if it causes the green checkmark to be something other than a green checkmark.  Further, I don't know if "Show all checks" panel exposes the failure.  But certainly clicking through to the Travis job status page would show a failing commit message check.

The first comment on this PR suggests that a failing commit message check would only be visible to someone that clicked all the way through to the Travis job status page.  If this is true, then this new check really shouldn't bother most people and certainly won't block merging.  But the reviewers and PR author should probably click through to the Travis job status page to check.  Otherwise, this all won't do much good.

And BTW, this repo is NOT configured to block committing a change if all the checks do not pass.

karl-lunarg:
Anyway, I think that this is OK to try and see how it works over time.  We can always disable it if it causes more trouble than it solves.
:+1: after requested changes are made.

mark-lunarg:
Updated for feedback and pushed to Master.

Issue #2563, opened 2018-04-10
Missing checks for VALIDATION_ERROR_32a00272


eskilsund:
Resetting descriptor pools with in-flight cmdbuffers seems to lack validation and I couldn't find any open issues on this. So here's my attempt at addressing this. (Feel free to reject if this is already tracked somewhere else.)

As seen in layers/core_validation.cpp
```
VKAPI_ATTR VkResult VKAPI_CALL ResetDescriptorPool(VkDevice device, VkDescriptorPool descriptorPool,
                                                    VkDescriptorPoolResetFlags flags) {
    // TODO : Add checks for VALIDATION_ERROR_32a00272
```
Failing to fence properly before resetting a pool seems like a common application issue so it would be good to get into the core validation layer.

I've already forked master and made a suggested fix here https://github.com/eskilsund/Vulkan-LoaderAndValidationLayers/commit/9fea1df674f9fa4f373dd1db6817510997e0b374

I can create a pull request and link to this issue if wanted.

mark-lunarg:
@daveh-lunarg, can you take a look at the fix @eskilsund has pointed to?  It looks like it's ready for a PR to me, but you're the expert!

chrisforbes:
I'd say go ahead with creating the pull request, and have the discussion of the details there.

daveh-lunarg:
PR #2566 is merged so I guess this issue should be closed.

PR #2566, opened 2018-04-11
layers: Add VALIDATION_ERROR_32a00272

eskilsund:
This fixes issue #2563 adding VALIDATION_ERROR_32a00272 to core
validation layer to catch reset of descriptor pool in-flight. Also
adding an accompanying test case and an update of the validation layer
database to reflect new validation and test.


layers/vk_validation_error_database.txt
```diff
@@ -3386,7 +3386,7 @@ VALIDATION_ERROR_32802801~^~Y~^~None~^~vkResetCommandPool~^~VUID-vkResetCommandP
...
 VALIDATION_ERROR_32809001~^~Y~^~Unknown~^~vkResetCommandPool~^~VUID-vkResetCommandPool-flags-parameter~^~core~^~The spec valid usage text states 'flags must be a valid combination of VkCommandPoolResetFlagBits values' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkResetCommandPool-flags-parameter)~^~implicit, TBD in parameter validation layer.
-VALIDATION_ERROR_32a00272~^~N~^~Unknown~^~vkResetDescriptorPool~^~VUID-vkResetDescriptorPool-descriptorPool-00313~^~core~^~The spec valid usage text states 'All uses of descriptorPool (via any allocated descriptor sets) must have completed execution' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkResetDescriptorPool-descriptorPool-00313)~^~
+VALIDATION_ERROR_32a00272~^~Y~^~Unknown~^~vkResetDescriptorPool~^~VUID-vkResetDescriptorPool-descriptorPool-00313~^~core~^~The spec valid usage text states 'All uses of descriptorPool (via any allocated descriptor sets) must have completed execution' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkResetDescriptorPool-descriptorPool-00313)~^~
```
mark-lunarg: The 'Unknown' field here should be changed to 'DescriptorPoolInUseResetSignaled' -- it's the name of the layer validation test that verifies the check.

layers/core_validation.cpp
```diff
@@ -5534,11 +5560,15 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDescriptorPool(VkDevice device, const VkDes
...
                                                    VkDescriptorPoolResetFlags flags) {
-    // TODO : Add checks for VALIDATION_ERROR_32a00272
+    lock_guard_t lock(global_lock);
     layer_data *dev_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
+
+    // Make sure sets being destroyed are not currently in-use
+    bool skip = validateIdleDescriptorSetForPoolReset(dev_data, descriptorPool);
+    if (skip) return VK_ERROR_VALIDATION_FAILED_EXT;
+
     VkResult result = dev_data->dispatch_table.ResetDescriptorPool(device, descriptorPool, flags);
```
mark-lunarg: Need to unlock before any down-chain call.

layers/core_validation.cpp
```diff
@@ -5534,11 +5560,15 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDescriptorPool(VkDevice device, const VkDes
...
     if (VK_SUCCESS == result) {
-        lock_guard_t lock(global_lock);
```
mark-lunarg: This needs to be retained -- clearDescriptorPool must be guarded.

layers/core_validation.cpp
```diff
@@ -1582,6 +1582,32 @@ static bool validateIdleDescriptorSet(const layer_data *dev_data, VkDescriptorSe
...
+// Validate that given pool does not store any descriptor sets used by an in-flight CmdBuffer
+// pool stores the descriptor sets to be validated
+// Return false if no errors occur
+// Return true if validation error occurs and callback returns true (to skip upcoming API call down the chain)
+static bool validateIdleDescriptorSetForPoolReset(const layer_data *dev_data, const VkDescriptorPool pool) {
+    if (dev_data->instance_data->disabled.idle_descriptor_set) return false;
+    bool skip = false;
+    DESCRIPTOR_POOL_STATE *pPool = GetDescriptorPoolState(dev_data, pool);
+    for (auto ds : pPool->sets) {
+        if (ds != VK_NULL_HANDLE) {
```
jzulauf-lunarg: GetDescriptorPoolState returns a pointer which should be compare against null or (safely) implicitly converted to boolean.
```
if ( ds != nullptr)
```
or
```
if (ds)
```

layers/core_validation.cpp
```diff
@@ -1582,6 +1582,32 @@ static bool validateIdleDescriptorSet(const layer_data *dev_data, VkDescriptorSe
...
+            auto set_node = dev_data->setMap.find(ds->GetSet());
```
jzulauf-lunarg: This is a wasted round trip. ds is a cvdescriptorset::DescriptorSet *, which correlates to the VkDescriptorSet returned by GetSet, which is mapped *back* to ds in the setMap by the find (the logic of which duplicates an existing utility GetSetNode).

layers/core_validation.cpp
```diff
@@ -1582,6 +1582,32 @@ static bool validateIdleDescriptorSet(const layer_data *dev_data, VkDescriptorSe
...
+            if (set_node != dev_data->setMap.end()) {
+                if (set_node->second->in_use.load()) {
```
jzulauf-lunarg: Replacing 1594-1597 with
```
if (ds && and ds->in_use.load()) {
```
should be equivalent.

layers/core_validation.cpp
```diff
@@ -5534,11 +5560,15 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDescriptorPool(VkDevice device, const VkDes
 
 VKAPI_ATTR VkResult VKAPI_CALL ResetDescriptorPool(VkDevice device, VkDescriptorPool descriptorPool,
                                                    VkDescriptorPoolResetFlags flags) {
-    // TODO : Add checks for VALIDATION_ERROR_32a00272
+    lock_guard_t lock(global_lock);
```
jzulauf-lunarg: The pattern elsewhere in the code is to lock *after* acquiring the layer data pointer... also use unique_lock_t to allow unlocking and relocking.

layers/core_validation.cpp
```diff
@@ -1582,6 +1582,32 @@ static bool validateIdleDescriptorSet(const layer_data *dev_data, VkDescriptorSe
...
+        if (ds != VK_NULL_HANDLE) {
+            auto set_node = dev_data->setMap.find(ds->GetSet());
+            if (set_node != dev_data->setMap.end()) {
+                if (set_node->second->in_use.load()) {
+                    skip |=
+                        log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT,
+                                VK_DEBUG_REPORT_OBJECT_TYPE_DESCRIPTOR_POOL_EXT, HandleToUint64(pool), VALIDATION_ERROR_32a00272,
+                                "It is invalid to call vkResetDescriptorPool() with descriptor sets in use by a command buffer. %s",
+                                validation_error_map[VALIDATION_ERROR_32a00272]);
+                    if (skip) break;
```
jzulauf-lunarg: It might be better to return all descriptor set handles that are inuse rather than only the first.  @mark-lunarg Thoughts?

jzulauf-lunarg:
LGTM -- waiting on checks to +1.

layers/core_validation.cpp
eskilsund: I left this for you to decide. From my experience this generates quite a few validation errors for  complex apps.

eskilsund:
Seems like the AppVeyor build failed but I am unsure why? Anyone able to interpret the build error for me?

karl-lunarg:
I think that there is a dependency specification problem that causes race conditions between some projects when built in parallel.  It doesn't need to block this submission.

karl-lunarg:
I've submitted #2599 that hopefully addresses the race condition.

tests/layer_validation_tests.cpp
```diff
@@ -8376,6 +8376,131 @@ TEST_F(VkLayerTest, DescriptorPoolInUseDestroyedSignaled) {
     // TODO : It seems Validation layers think ds_pool was already destroyed, even though it wasn't?
 }
 
+TEST_F(VkLayerTest, DescriptorPoolInUseResetSignaled) {
```
jzulauf-lunarg: It seems like this test case could be much smaller...

PR #2577, opened 2018-04-13
layers:Update comments to avoid compiler warnings

tobine:
The simple "// fall through" comment causes the compiler to omit a
warning.

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -1234,10 +1234,10 @@ static bool RegionIntersects(const VkImageCopy *rgn0, const VkImageCopy *rgn1, V
         switch (type) {
             case VK_IMAGE_TYPE_3D:
                 result &= RangesIntersect(rgn0->srcOffset.z, rgn0->extent.depth, rgn1->dstOffset.z, rgn1->extent.depth);
-            // Intentionally fall through to 2D case
+                // fall through
             case VK_IMAGE_TYPE_2D:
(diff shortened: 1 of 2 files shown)


layers/buffer_validation.cpp
```diff
@@ -1234,10 +1234,10 @@ static bool RegionIntersects(const VkImageCopy *rgn0, const VkImageCopy *rgn1, V
         switch (type) {
             case VK_IMAGE_TYPE_3D:
                 result &= RangesIntersect(rgn0->srcOffset.z, rgn0->extent.depth, rgn1->dstOffset.z, rgn1->extent.depth);
-            // Intentionally fall through to 2D case
+                // fall through
```
jzulauf-lunarg: why shortening the extant comments?

chrisforbes: This is one of the specific comment forms that gcc7 recognizes as marking a fall through.

jzulauf-lunarg: The comment method isn't flexible at all.  "The whole body of the C or C++ style comment should match the given regular expressions listed below."

How about defining 
```
#ifdef ... // Compiler filtered if checks
...
#elif ... // some GCC condition
#define FALLTHROUGH [[gnu::fallthrough]];
#elif ... // C++17 check
   [[fallthrough]]; // C++17 and above 
```

and then doing a 
```
FALLTHROUGH // Some actually useful comment
```

in the code.

daveh-lunarg: Comment language could also be preserved by breaking into multiple lines? e.g.

```// fall through```
```// for this specific reason```

mark-lunarg:
👍 @daveh-lunarg

Issue #2578, opened 2018-04-14
loader: Pass the device dispatch table handle to the ICD before CreateDevice


jdrouan-goog:
Our ICD implementation requires that the future address of the dispatch table be in scope during CreateDevice.

polarina:
For what purpose does your ICD require this?

jdrouan-goog:
Our platform requires this the way it is implemented. We think this is a relatively small change that should be compatible with the current ecosystem of ICDs.

cc: @jfroy

lenny-lunarg:
You can see the linked PR for an explanation of why I don't want to pass the entire dispatch table to the ICD, but is there any reason why you actually need the dispatch table? Generally, it's much easier to just pass function pointers to the `vkGet*ProcAddr` functions. Would this still address your needs or is there some reason that you actually need the dispatch table.

I still find it odd that you would need these pointers at all, but I would be far more receptive to sharing them through that method, rather than just passing the dispatch table.

ianelliottus:
I just started looking at this.  I will try to get some additional info to see if I can help.  I will follow-up in the PR.

jfroy:
I am helping @jdrouan-goog on this PR. We think the change we're making here will be transparent to all existing ICDs because of the established rules for scanning and parsing extension structures, e.g. ignore and skip what you do not recognize.

Without going into all the technical details and motivations, we are building a virtual ICD that forwards a portion of its work to a "delegate" ICD. As part of this, we want to data to dispatchable objects, much like layers do. Unfortunately, with the current loader implementation, when vkCreateDevice executes in an ICD, the final dispatch table pointer has not been set by the loader. This is our attempt at solving this problem.

It occurs to me an alternative may be possible. Looking at https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/5de1ff1ff66aa41e271b87dcd0b47701711f6e17/loader/loader.c#L5577, could we move that line above the fpCreateDevice call?

lenny-lunarg:
> We think the change we're making here will be transparent to all existing ICDs because of the established rules for scanning and parsing extension structures, e.g. ignore and skip what you do not recognize.

[...]

PR #2579, opened 2018-04-14
loader: Send device dispatch table address to ICDs

jdrouan-goog:
Send ICDs the device dispatch table address using
VK_STRUCTURE_TYPE_LOADER_DEVICE_CREATE_INFO. This is to support
implementations that need the handle to be in scope during the execution
of CreateDevice.

See issue #2578

### DIFF
--- a/include/vulkan/vk_layer.h
+++ b/include/vulkan/vk_layer.h
@@ -82,7 +82,8 @@ typedef VkResult(VKAPI_PTR *PFN_PhysDevExt)(VkPhysicalDevice phys_device);
  */
 typedef enum VkLayerFunction_ {
     VK_LAYER_LINK_INFO = 0,
-    VK_LOADER_DATA_CALLBACK = 1
+    VK_LOADER_DATA_CALLBACK = 1,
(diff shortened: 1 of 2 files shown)


ianelliottus:
Two generic points while I try to learn more about the goals for this change:

@lenny-lunarg 's first point reminds me that such a change will need to be formally changing the documented loader-ICD interface.  It will probably need broader review and buy-in from others (including the desktop IHVs and potentially the Android loader and IHV folks).

@lenny-lunarg's 2nd point is a good one ... the desktop loader is shared by all of the ICDs on a system.  Laptops often have an integrated and discrete ICD, which come from 2 different IHVs.  If one driver is updated, the non-updated driver still needs to work with the loader that came along with the updated driver.

jfroy:
See some more details in the related issue.

mark-lunarg:
@jdrouan-goog, @ianelliottus, @lenny-lunarg, @jfroy, this repository will close for write access on Sunday, 5/13/2018.  If it is pushed before that time it will be present in the follow-on Vulkan-Loader repository on Monday, otherwise a new PR will be required in the new repo.

PR #2583, opened 2018-04-16
layers: Validate subpass dependency access vs stage masks

### LINKED ISSUE #2570: No validation warning for barriers with redundant access flags
The spec says this:

> Certain access types are only performed by a subset of pipeline stages. Any synchronization command that takes both stage masks and access masks uses both to define the access scopes - only the specified access types performed by the specified stages are included in the access scope. An application must not specify an access flag in a synchronization command if it does not include a pipeline stage in the corresponding stage mask that is able to perform accesses of that type. The following table lists, for each access flag, which pipeline stages can perform that type of access.

To me this suggests that, for example, the following barrier:

	dependencies[1].dstStageMask = VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT;
	dependencies[1].dstAccessMask = VK_ACCESS_COLOR_ATTACHMENT_READ_BIT | VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT | VK_ACCESS_SHADER_READ_BIT | VK_ACCESS_TRANSFER_READ_BIT;

Is something an application "must not specify" since there are no stages in the mask that can issue transfer reads. It would be nice to get a validation warning/error for this.

This seems to be the case for both external pass dependencies and vkCmdPipelineBarrier.

chrisforbes:
Fixes #2570


layers/core_validation.cpp
```diff
@@ -8954,6 +8954,24 @@ static bool ValidateDependencies(const layer_data *dev_data, FRAMEBUFFER_STATE c
...
         }
     }
+    // Validate all the subpass dependencies themselves
+    for (uint32_t i = 0; i < pCreateInfo->dependencyCount; ++i) {
+        VkSubpassDependency const &dependency = pCreateInfo->pDependencies[i];
+
+        if (!ValidateAccessMaskPipelineStage(dependency.srcAccessMask, dependency.srcStageMask)) {
+            skip |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0,
+                            VALIDATION_ERROR_13e006c8,
+                            "CreateRenderPass: pDependencies[%u].srcAccessMask (0x%X) is not supported by srcStageMask (0x%X).", i,
```
tobine: Is this down CmdBeginRenderPass()?

layers/core_validation.cpp
```diff
@@ -8954,6 +8954,24 @@ static bool ValidateDependencies(const layer_data *dev_data, FRAMEBUFFER_STATE c
             CheckPreserved(dev_data, pCreateInfo, i, subpass.pInputAttachments[j].attachment, subpass_to_node, 0, skip);
         }
     }
+    // Validate all the subpass dependencies themselves
+    for (uint32_t i = 0; i < pCreateInfo->dependencyCount; ++i) {
+        VkSubpassDependency const &dependency = pCreateInfo->pDependencies[i];
+
+        if (!ValidateAccessMaskPipelineStage(dependency.srcAccessMask, dependency.srcStageMask)) {
+            skip |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0,
```
tobine: Could include the renderPass object as dbg obj here and in call below as well.

layers/core_validation.cpp
chrisforbes: So it is. I'm moving these new checks to CreateRenderPass-time as I originally intended.

layers/core_validation.cpp
```diff
@@ -9167,10 +9167,25 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateRenderPass(VkDevice device, const VkRenderP
...
-                                             VALIDATION_ERROR_13e006b8, VALIDATION_ERROR_13e006bc);
-        skip |= ValidateStageMaskGsTsEnables(dev_data, pCreateInfo->pDependencies[i].dstStageMask, "vkCreateRenderPass()",
-                                             VALIDATION_ERROR_13e006ba, VALIDATION_ERROR_13e006be);
+        auto const &dependency = pCreateInfo->pDependencies[i];
+        skip |= ValidateStageMaskGsTsEnables(dev_data, dependency.srcStageMask, "vkCreateRenderPass()", VALIDATION_ERROR_13e006b8,
+                                             VALIDATION_ERROR_13e006bc);
+        skip |= ValidateStageMaskGsTsEnables(dev_data, dependency.dstStageMask, "vkCreateRenderPass()", VALIDATION_ERROR_13e006ba,
+                                             VALIDATION_ERROR_13e006be);
+
+        if (!ValidateAccessMaskPipelineStage(dependency.srcAccessMask, dependency.srcStageMask)) {
```
jzulauf-lunarg: Most "Validate" calls return a skip value ... Validate<Foo> == false implies "Foo is valid" (or at least logging doesn't care enough about it to generate a skip) -- and also encapsulate message logging.  For validation check returning "true" for valid... we should likely have some *other* naming convention... or imply fold in the logging to this Validate call and return skip.

chrisforbes: @jzulauf-lunarg shall we tidy that up when we fix `CreateRenderPass` validation in general? There's plenty of problems with this function.

PR #2586, opened 2018-04-16
Add validation for inter-extension dependencies

jzulauf-lunarg:
Valid usages now exist requiring that dependencies of give extension must be specified if the extension is specified.  The existing vk_extension_helper is (and its generator are refactored to support this, missing dependencies in the layer_validation_tests are corrected, and validation is added to CreateInstance and CreateDevice in the parameter_validation layer.

Please note that the various transformations of the code generator are present in separate commits, each (until the last) producing the same effective output as the initial form.


layers/parameter_validation_utils.cpp
```diff
@@ -191,6 +191,68 @@ static bool ValidateQueueFamilies(layer_data *device_data, uint32_t queue_family
     return skip;
 }
 
+static bool validate_api_version(const instance_layer_data *instance_data, uint32_t api_version, uint32_t effective_api_version) {
```
jzulauf-lunarg: This is encapsulation of an existing check.

layers/parameter_validation_utils.cpp
```diff
@@ -191,6 +191,68 @@ static bool ValidateQueueFamilies(layer_data *device_data, uint32_t queue_family
...
+                            HandleToUint64(instance_data->instance), VALIDATION_ERROR_UNDEFINED,
+                            "Unrecognized CreateInstance->pCreateInfo->pApplicationInfo.apiVersion number (0x%08x). Assuming %s.\n",
+                            api_version, effective_api_string);
+        }
+    }
+    return skip;
+}
+
+template <typename ExtensionState>
+static bool validate_extension_reqs(const instance_layer_data *instance_data, const ExtensionState &extensions,
```
jzulauf-lunarg: Device and Instance creation use identical logic just on different types.

layers/parameter_validation_utils.cpp
```diff
@@ -191,6 +191,68 @@ static bool ValidateQueueFamilies(layer_data *device_data, uint32_t queue_family
...
+    auto info = ExtensionState::get_info(extension_name);
+
+    if (!info.state) {
+        return skip;  // Unknown extensions cannot be checked so report OK
+    }
+
+    // Check agains the reqs list in the info
+    std::vector<const char *> missing;
+    for (const auto &req : info.requires) {
+        if (!(extensions.*(req.enabled))) {
```
jzulauf-lunarg: This strange syntax is the "pointer to data member", an approach already used in the *Extensions helper classes, and just expanded in its use here.

layers/parameter_validation_utils.cpp
```diff
@@ -191,6 +191,68 @@ static bool ValidateQueueFamilies(layer_data *device_data, uint32_t queue_family
...
+    }
+    auto info = ExtensionState::get_info(extension_name);
+
+    if (!info.state) {
+        return skip;  // Unknown extensions cannot be checked so report OK
+    }
+
+    // Check agains the reqs list in the info
+    std::vector<const char *> missing;
+    for (const auto &req : info.requires) {
```
jzulauf-lunarg: The we loop over the dependency requirement here so we can log only a single error below for *all* missing requirements.

layers/parameter_validation_utils.cpp
```diff
@@ -265,25 +327,11 @@ VKAPI_ATTR VkResult VKAPI_CALL vkCreateInstance(const VkInstanceCreateInfo *pCre
 
         uint32_t effective_api_version = my_instance_data->extensions.InitFromInstanceCreateInfo(api_version, pCreateInfo);
 
-        uint32_t api_version_nopatch = VK_MAKE_VERSION(VK_VERSION_MAJOR(api_version), VK_VERSION_MINOR(api_version), 0);
```
jzulauf-lunarg: moved to validate_api_version, above

layers/parameter_validation_utils.cpp
```diff
@@ -431,9 +479,19 @@ VKAPI_ATTR void VKAPI_CALL vkDestroyDebugUtilsMessengerEXT(VkInstance instance,
     }
 }
 
+template <typename ExtensionState>
+static bool extension_state_by_name(const ExtensionState &extensions, const char *extension_name) {
```
jzulauf-lunarg: the *Info structs are a little odd to use, so we just encapsulate it.

layers/parameter_validation_utils.cpp
```diff
@@ -442,16 +500,15 @@ static bool ValidateDeviceCreateInfo(instance_layer_data *instance_data, VkPhysi
         }
     }
 
-    bool maint1 = false;
-    bool negative_viewport = false;
-
     if ((pCreateInfo->enabledExtensionCount > 0) && (pCreateInfo->ppEnabledExtensionNames != NULL)) {
+        maint1 = extension_state_by_name(extensions, VK_KHR_MAINTENANCE1_EXTENSION_NAME);
```
jzulauf-lunarg: In the refactor, the extension name to extension field is now a hash, so we take advantage of it here.

layers/parameter_validation_utils.cpp
```diff
@@ -526,11 +583,21 @@ VKAPI_ATTR VkResult VKAPI_CALL vkCreateDevice(VkPhysicalDevice physicalDevice, c
...
     auto my_instance_data = GetLayerDataPtr(get_dispatch_key(physicalDevice), instance_layer_data_map);
     assert(my_instance_data != nullptr);
+
+    // Query and save physical device limits for this device, needed for validation
+    VkPhysicalDeviceProperties device_properties = {};
+    my_instance_data->dispatch_table.GetPhysicalDeviceProperties(physicalDevice, &device_properties);
+
+    // Set up the extension structure also for validation.
+    DeviceExtensions extensions;
+    uint32_t api_version =
```
jzulauf-lunarg: We need to know the extensions we think we have before we validate, so we'll create the struct of it before there's any device_data to store it in.

layers/parameter_validation_utils.cpp
```diff
@@ -442,16 +500,15 @@ static bool ValidateDeviceCreateInfo(instance_layer_data *instance_data, VkPhysi
...
+        maint1 = extension_state_by_name(extensions, VK_KHR_MAINTENANCE1_EXTENSION_NAME);
+        negative_viewport = extension_state_by_name(extensions, VK_AMD_NEGATIVE_VIEWPORT_HEIGHT_EXTENSION_NAME);
+
         for (size_t i = 0; i < pCreateInfo->enabledExtensionCount; i++) {
             skip |= validate_string(instance_data->report_data, "vkCreateDevice", "pCreateInfo->ppEnabledExtensionNames",
                                     pCreateInfo->ppEnabledExtensionNames[i]);
-            if (strcmp(pCreateInfo->ppEnabledExtensionNames[i], VK_KHR_MAINTENANCE1_EXTENSION_NAME) == 0) maint1 = true;
-            if (strcmp(pCreateInfo->ppEnabledExtensionNames[i], VK_AMD_NEGATIVE_VIEWPORT_HEIGHT_EXTENSION_NAME) == 0)
-                negative_viewport = true;
+            skip |= validate_extension_reqs(instance_data, extensions, VALIDATION_ERROR_1fc00ad6, "device",
```
jzulauf-lunarg: since we're already looping over the array of extensions, add the requirement check here.

layers/parameter_validation_utils.cpp
```diff
@@ -560,21 +627,8 @@ VKAPI_ATTR VkResult VKAPI_CALL vkCreateDevice(VkPhysicalDevice physicalDevice, c
...
-
-            uint32_t specified_api_version = device_properties.apiVersion & ~VK_VERSION_PATCH(~0);
-            if (!(specified_api_version == VK_API_VERSION_1_0) && !(specified_api_version == VK_API_VERSION_1_1)) {
-                LOGCONSOLE(
-                    "Warning: Unrecognized CreateInstance->pCreateInfo->pApplicationInfo.apiVersion number -- (0x%8x) assuming "
-                    "%s.\n",
-                    device_properties.apiVersion,
-                    (my_device_data->api_version == VK_API_VERSION_1_0) ? "VK_API_VERSION_1_0" : "VK_API_VERSION_1_1");
-            }
+            my_device_data->api_version = api_version;
```
jzulauf-lunarg: Now that we have a device_data to store the information, copy it over.

layers/vk_layer_utils.h
```diff
@@ -32,6 +33,58 @@
 #endif
 
 #ifdef __cplusplus
+// Traits objects to allow string_join to operate on collections of const char *
```
jzulauf-lunarg: This started off smaller (at least in my mind), and "join" really is a useful thing to have.

loader/debug_utils.c
```diff
@@ -923,6 +923,7 @@ bool debug_utils_ReportFlagsToAnnotFlags(VkDebugReportFlagsEXT dr_flags, bool de
         return false;
     }
     *da_type = 0;
+    *da_severity = 0;
```
jzulauf-lunarg: This kills a compile time warning... it's in its own commit if one wanted to stick it in a new PR...

scripts/helper_file_generator.py
```diff
@@ -162,14 +162,22 @@ def beginFeature(self, interface, emit):
 
         if self.featureName == 'VK_VERSION_1_0' or self.featureName == 'VK_VERSION_1_1':
             return
+        name = self.featureName
```
jzulauf-lunarg: See the separate commits to see how this got built up.

It turned out to be useful to index (for sorting) by the actual extension name...

scripts/helper_file_generator.py
```diff
@@ -162,14 +162,22 @@ def beginFeature(self, interface, emit):
...
-        if 'EXTENSION_NAME' not in name:
+        name_define = nameElem.get('name')
+        if 'EXTENSION_NAME' not in name_define:
             print("Error in vk.xml file -- extension name is not available")
+        requires = interface.get('requires')
+        if requires is not None:
+            required_extensions = requires.split(',')
+        else:
+            required_extensions = list()
+        info = { 'define': name_define, 'ifdef':self.featureExtraProtect, 'reqs':required_extensions }
```
jzulauf-lunarg: I'm using a dict here to avoid defining a class...  could be done, but the usage here is limited in scope and function.  Essentially, instance_extension_info and device_extension_info need to hold more data while avoiding adding members to the class.

scripts/helper_file_generator.py
```diff
@@ -487,104 +495,150 @@ def GenerateExtensionHelperHeader(self):
             'vk_khr_variable_pointers',
             'vk_khr_dedicated_allocation',
             ]
+        output = [
```
jzulauf-lunarg: According to various source, incrementally appending to strings is not guaranteed to execute in linear time, while appending to list and then joining them is.  so this is the preferred "pythonic" best practice.  Note the absence of \n and the presence of , at the EOL.

scripts/helper_file_generator.py
```diff
@@ -487,104 +495,150 @@ def GenerateExtensionHelperHeader(self):
...
+            struct  = [struct_decl]
+            struct.extend([ '    bool %s{false};' % field_name[ext_name] for ext_name, info in extension_items])
+
+            # Construct the extension information map -- mapping name to data member (field), and required extensions
+            # The map is contained within a static function member for portability reasons.
+            info_type = '%sInfo' % type
+            info_map_type = '%sMap' % info_type
+            req_type = '%sReq' % type
+            req_vec_type = '%sVec' % req_type
+            struct.extend([
```
jzulauf-lunarg: The *Extension structs now have helper structs within to store the additional information that the prior "std::pair" was ill suited to.

scripts/helper_file_generator.py
```diff
@@ -487,104 +495,150 @@ def GenerateExtensionHelperHeader(self):
...
+                '',
+                '    struct %s {' % req_type,
```
jzulauf-lunarg: the req_type struct encapsulates the extension dependency as a member pointer to avoid per dependency hash lookups, but stores the string for convenience (and avoiding another unordered_map to going from member pointer back to name)

scripts/helper_file_generator.py
```diff
@@ -487,104 +495,150 @@ def GenerateExtensionHelperHeader(self):
...
+                '    };',
+                '    typedef std::vector<%s> %s;' % (req_type, req_vec_type),
+                '    struct %s {' % info_type,
+                '       %s(bool %s::* state_, const %s requires_): state(state_), requires(requires_) {}' % ( info_type, struct_type, req_vec_type),
+                '       bool %s::* state;' % struct_type,
+                '       %s requires;' % req_vec_type,
+                '    };',
+                '',
+                '    typedef std::unordered_map<std::string,%s> %s;' % (info_type, info_map_type),
+                '    static const %s &get_info(const char *name) {' %info_type,
```
jzulauf-lunarg: The *Extensions get the name indexed info about each extension from a static function, as not all supported compilers allow for static object member initialization of non integral types.

scripts/helper_file_generator.py
```diff
@@ -487,104 +495,150 @@ def GenerateExtensionHelperHeader(self):
...
-            struct += '        // Initialize struct data\n'
+            struct.extend([
+                '',
+                '        static const std::vector<const char *> V_1_0_promoted_%s_extensions = {' % type.lower() ])
+            struct.extend(['            %s_EXTENSION_NAME,' % ext_name.upper() for ext_name in promoted_ext_list])
+            struct.extend([
+                '        };',
+                '',
+                '        // Initialize struct data, robust to invalid pCreateInfo',
+                '        if (pCreateInfo->ppEnabledExtensionNames) {',
```
jzulauf-lunarg: As this init is happening *before* validation (i.e. as data used during validation) it's made more robust to bad data.

scripts/helper_file_generator.py
```diff
@@ -487,104 +495,150 @@ def GenerateExtensionHelperHeader(self):
...
+                '            for (uint32_t i = 0; i < pCreateInfo->enabledExtensionCount; i++) {',
+                '                if (!pCreateInfo->ppEnabledExtensionNames[i]) continue;',
+                '                auto info = get_info(pCreateInfo->ppEnabledExtensionNames[i]);',
+                '                if(info.state) this->*(info.state) = true;',
```
jzulauf-lunarg: it's valid to have a unknown extension so, we intentional don't assert here.

scripts/helper_file_generator.py
```diff
@@ -487,104 +495,150 @@ def GenerateExtensionHelperHeader(self):
...
+                '            }',
+                '        }',
+                '        uint32_t api_version = NormalizeApiVersion(requested_api_version);',
+                '        if (api_version >= VK_API_VERSION_1_1) {',
+                '            for (auto promoted_ext : V_1_0_promoted_%s_extensions) {' % type.lower(),
+                '                auto info = get_info(promoted_ext);',
+                '                assert(info.state);',
```
jzulauf-lunarg: It's certainly *not* valid to no have info on a promoted extension... so we assert

scripts/helper_file_generator.py
```diff
@@ -487,104 +495,150 @@ def GenerateExtensionHelperHeader(self):
...
+                '                if (info.state) this->*(info.state) = true;',
+                '            }',
+                '        }',
+                '        return api_version;',
```
jzulauf-lunarg: This return is for historic reasons, frankly, the logic of NormalizeApiVersion is in the wrong place -- but as it was here, leaving it.

scripts/helper_file_generator.py
```diff
@@ -487,104 +495,150 @@ def GenerateExtensionHelperHeader(self):
...
+            '#ifndef VK_EXTENSION_HELPER_H_',
+            '#define VK_EXTENSION_HELPER_H_',
+            '#include <string>',
+            '#include <unordered_map>',
+            '#include <utility>',
+            '',
+            '#include <vulkan/vulkan.h>',
+            '']
+
+        def guarded(ifdef, value):
```
jzulauf-lunarg: This was a repeated pattern within this function. Possibly a file-scope utility function would be useful if(as) we refactor more code to the list oriented implementations, potentially returning a list to pass to "extend" directly, that or a class that wraps extend/append logic with whatever convenience functions built-in... "but not today!"

layers/vk_layer_utils.h
daveh-lunarg: Speaking as a curmudgeon, there's quite a bit of code in this PR dedicated to linear-time string concatenation, which I don't feel is a significant performance issue in validation code. I'm not against it since the work is all done, but I'm unsure of the ROI.

scripts/helper_file_generator.py
```diff
@@ -487,104 +495,150 @@ def GenerateExtensionHelperHeader(self):
...
+                '            return info->second;',
+                '        }',
+                '        return empty_info;',
+                '    }',
+                ''])
+
+            if type == 'Instance':
+                struct.extend([
+                    '    uint32_t NormalizeApiVersion(uint32_t specified_version) {',
+                    '        uint32_t api_version = (specified_version < VK_API_VERSION_1_1) ? VK_API_VERSION_1_0 : VK_API_VERSION_1_1;',
```
daveh-lunarg: Future-proof for api > 1.1?

layers/parameter_validation_utils.cpp
jzulauf-lunarg: The pointer to data member syntax is quite an ugly syntactic corner of C++, but this is the same approach that was used by the previous implementation.

scripts/helper_file_generator.py
jzulauf-lunarg: The concept behind normalized API is that it return the type for the highest API version the validation layers recognize.  The function probably shouldn't even be in the code-gen'd section (is inherited), but as it reflects the state of the validation code, I'm unsure of a way to future proof it that reflects the meaning.  

Is simply the fact the the validation layers were compiled against a header with a new minor sufficient to automatically up this version, or should be part of the deliberate set of changes that implement support for the new features of a minor release?

@daveh-lunarg , @mark-lunarg , @mikes-lunarg  -- thoughts on this?

layers/parameter_validation_utils.cpp
```diff
@@ -191,6 +191,68 @@ static bool ValidateQueueFamilies(layer_data *device_data, uint32_t queue_family
     return skip;
 }
 
+static bool validate_api_version(const instance_layer_data *instance_data, uint32_t api_version, uint32_t effective_api_version) {
+    bool skip = false;
+    uint32_t api_version_nopatch = VK_MAKE_VERSION(VK_VERSION_MAJOR(api_version), VK_VERSION_MINOR(api_version), 0);
+    if (api_version_nopatch != effective_api_version) {
+        const char *effective_api_string =
+            (effective_api_version == VK_API_VERSION_1_0) ? "VK_API_VERSION_1_0" : "VK_API_VERSION_1_1";
```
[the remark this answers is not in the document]
jzulauf-lunarg: Makes sense.

layers/vk_layer_utils.h
jzulauf-lunarg: Part of the motivation was to model Pythonic best practice for the new python scripting we're planning, and the other part to simplify the effective rewrite of the generator for the new meta data supported.  The list oriented approach allowed greater brevity with the list comprehensions. (which is also considered more Pythonic than the for this in that multiline loops.

The string_join function was a little more effort than expected (mostly because of the traits to allow const char * seps and container contents) 😀

PR #2588, opened 2018-04-17
Tobin build warning fixes

tobine:
Fix/prevent various loader and layer build warnings.

### DIFF
--- a/demos/vulkaninfo.c
+++ b/demos/vulkaninfo.c
@@ -2951,7 +2951,7 @@ static void AppGroupDump(const VkPhysicalDeviceGroupProperties *group, const uin
     }
 
     // Build create info for logical device made from all physical devices in this group.
-    char *extensions_list = VK_KHR_DEVICE_GROUP_EXTENSION_NAME;
+    const char *extensions_list = VK_KHR_DEVICE_GROUP_EXTENSION_NAME;
 
     VkDeviceGroupDeviceCreateInfoKHR dg_ci = {.sType = VK_STRUCTURE_TYPE_DEVICE_GROUP_DEVICE_CREATE_INFO_KHR,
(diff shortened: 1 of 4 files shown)


demos/vulkaninfo.c
```diff
@@ -2966,7 +2966,7 @@ static void AppGroupDump(const VkPhysicalDeviceGroupProperties *group, const uin
                                     .queueCreateInfoCount = 1,
                                     .pQueueCreateInfos = &q_ci,
                                     .enabledExtensionCount = 1,
-                                    .ppEnabledExtensionNames = &extensions_list};
+                                    .ppEnabledExtensionNames = (const char* const*)&extensions_list};
```
mikes-lunarg: There is no reason for extensions_list to point at non-const data, it could be defined as "const char *extensions_list = VK_KHR_DEVICE_GROUP_EXTENSION_NAME;" above and avoid the cast here

scripts/reg.py
```diff
@@ -96,11 +96,11 @@ def compareElem(self, info):
         if selfKeys != infoKeys:
             return False
 
-        # Ignore value of 'extname', as this will inherently be different
-        # when redefining the same interface in different feature and/or
-        # extension blocks.
+        # Ignore value of 'extname' and 'extnumber', as these will inherently
```
mikes-lunarg: Should this be submitted upstream to Vulkan-Docs?  Current [process](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/wiki#copy-new-vulkan-docs-file-changes-into-vulkan-loaderandvalidationlayers) will overwrite these changes next time the header is updated:

tobine: Ah, ok. Yes I'll do that.

PR #2593, opened 2018-04-18
loader: Clarify loader warning

MarkY-LunarG:
Clarify several loader warnings about layers so that they include
the name of the layer that is causing an issue.

Change-Id: Ie48d73f6cc8466c3114597831fef451b3ab32bf7


MarkY-LunarG:
Now using the "printf"-style functionality of the logging as suggested by Lenny instead of sprintf.

PR #2595, opened 2018-04-19
Validate against nonCoherentAtomSize

Tony-LunarG:
This was removed and Gitlab issue 611 was filed to verify the spec.  That issue has been resolved, and this commit add the checks and tests back in, taking into account the slight spec modification resulting from 611.

Github issues 329 & 2584
Gitlab Issue 611

Change-Id: I4cbc6c21f6fe4b6187b381b2565a90c61417bee5


tests/layer_validation_tests.cpp
```diff
@@ -2218,6 +2218,29 @@ TEST_F(VkLayerTest, InvalidMemoryMapping) {
     vkFlushMappedMemoryRanges(m_device->device(), 1, &mmr);
     m_errorMonitor->VerifyFound();
 
+    // Some platforms have an atomsize of 1 which makes the test meaningless
+    if (atom_size > 3) {
```
jzulauf-lunarg: why not > 1?

Issue #2605, opened 2018-04-23
LunarXchange issue #761 - vkGetImageSparseMemoryRequirements2 crashes when querying the count


jeremyk-lunarg:
Copied from LunarXchange [Issue-761](https://vulkan.lunarg.com/issue/view/5ade136de46ffe75aa6ffe23):

If i call vkGetImageSparseMemoryRequirements2(device, &info, &cnt, nullptr) with cnt = 0 to get the count of items returned, the validation layers crash in core_validation.cpp:4135 in PostCallRecordGetImageSparseMemoryRequirements2. The reqs parameter is my nullptr, but the req_count parameter is the overwritten result (in my case 1) returned by the driver. By the looks of it the loop that copies the data should only be executed if reqs is not null.

PR #2606, opened 2018-04-23
layers: Add handling for NULL in core post call

jeremyk-lunarg:
Added handling in
PostCallRecordGetImageSparseMemoryRequirements2
when a nullptr is passed to 'reqs'

Fixes issue #2605


layers/core_validation.cpp
```diff
@@ -4128,6 +4128,9 @@ VKAPI_ATTR void VKAPI_CALL GetImageSparseMemoryRequirements(VkDevice device, VkI
 
 static void PostCallRecordGetImageSparseMemoryRequirements2(IMAGE_STATE *image_state, uint32_t req_count,
                                                             VkSparseImageMemoryRequirements2KHR *reqs) {
+    if (reqs == nullptr) {
```
jzulauf-lunarg: A comment that indicates that this is not just "defensive" coding, but represents the common usage...
```
// Nothing to record on query for count only...
```
or some such, would help the next person :smile:

PR #2610, opened 2018-04-24
cube: Fix corruption

MarkY-LunarG:
We were freeing strings before we used them.  I'm not sure how
this wasn't causing corruption else-where, but I encountered it
on my own testing cube.

Change-Id: I089c7a79666def357d33147a80625152c3f8fad7

### DIFF
--- a/demos/cube.c
+++ b/demos/cube.c
@@ -355,6 +355,7 @@ struct demo {
 
     VkInstance inst;
     VkPhysicalDevice gpu;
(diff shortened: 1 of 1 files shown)


cforfang:
I'm not seeing the use after free? Is there use aside from the `strcmp()`s I'm missing?

MarkY-LunarG:
Killing because I think this was all in my branch.

PR #2615, opened 2018-04-25
layers: check color blend factors when testing dualSrcBlend

### LINKED ISSUE #2614: dualSrcBlend device feature is only checked if alpha blend factors use DSB
in validate_dual_src_blend_feature() (core_validation.cpp), there's a to make sure the dualSrcBlend device feature is enabled if one of the VK_BLEND_FACTOR_SRC1_* factors are used. However, only the src/dst alpha factors are checked; color factors must be checked as well. I'll fix this momentarily.

cdwfs:
Fixes #2614


cdwfs:
Okay, I split the checks out into the four VUIDs and added a test. I also moved the checks out of CmdBindPipelines() and into CreateGraphicsPipelines(), since there's no need to check this every time the pipeline is bound, right?

tobine:
Yep, this all looks great, thanks @cdwfs !
I'll let the CI finish tonight and merge tomorrow if clean and no one else beats me to it.

Issue #2483, opened 2018-03-13
sort pieces of vulkaninfo output where sorting makes sense


jjulianoatnv:
I've been diffing output of vulkaninfo across different driver versions and across different GPUs.

There is a block of text near the top of the output that enumerates one instance extension per line. The order of lines in this block appears to be random. Maybe it is the iteration order through the keys of a hash map, or something like that.

The text itself is amenable to alphanumeric sort. The purpose of this issue is to propose that text blocks in vulkaninfo output be alphanumerically sorted in cases where doing so makes sense, and where the text is otherwise non-deterministically sorted.

Candidates for alphanumeric soft include:
- List of instance extensions
- List of device extensions

Things that need not be sorted alphanumerically include:
- Dump of the values of fields of a struct. Printing them in the order in which they appear in the struct definition is both sensible and deterministic.
- Dump of the bits that are set in a bitfield. Printing them in bit-position order is both sensible and deterministic.

For instance extensions, the order is up to the loader because the loader constructs a union-or-intersection of extensions from each physical device, and physical devices can themselves use different orders.

For device extensions, even when comparing from a single vendor, the order appears to be non-deterministic. This suggests that either the loader or the vulkaninfo app is already changing the order from that which is returned by the ICD. Even so, a vendor's ICD is not required to use a canonical sort order. It makes sense for vulkaninfo to ignore the ICD's own sort order and to output the extension list in a canonical sort order. Alphanumeric sort is probably the most sensible canonical sort order.

jjulianoatnv:
Oh, for device extensions vulkaninfo might be using the ICD's order. When I look at my dGPU output it seems to have a stable order, but my iGPU's output does not. That could just mean that the iGPU driver does have itself have a stable ordering.

Still, when I have only a dGPU, the instance extension order is not stable, implying that the loader is probably itself using a random order. (That's still a guess.) And IMO it still makes sense to ignore the ICD's order or the loader's order, and for vulkaninfo to use a stable order regardless for instance and device extension lists.

shannon-lunarg:
Fixed with PR #2617.

PR #2617, opened 2018-04-25
demos: Sort most vkinfo output alphanumerically

shannon-lunarg:
Changes consist of reordering print statements for all paths of output (human_readable, html, and json) and the addition of employing the qsort function (and new comparison functions) to reorder global layers and extension properties in the app instance.

Please let me know if you think I've sorted sections of output that maybe shouldn't be sorted.

Sections left unsorted because the current ordering seems to make the most sense: Presentable Surfaces, VkQueueFamilyProperties, and Format Properties

Change-Id: I075a6897c3070f70d7073c22e7100c49c19073a3

### DIFF
--- a/demos/vulkaninfo.c
+++ b/demos/vulkaninfo.c
@@ -1711,179 +1711,179 @@ static void AppGpuDumpFeatures(const struct AppGpu *gpu, FILE *out) {
 
     if (html_output) {
         fprintf(out, "\t\t\t\t\t<details><summary>VkPhysicalDeviceFeatures</summary>\n");
-        fprintf(out, "\t\t\t\t\t\t<details><summary>alphaToOne                              = <div class='val'>%u</div></summary></details>\n", features.alphaToOne                             );
-        fprintf(out, "\t\t\t\t\t\t<details><summary>depthBiasClamp                          = <div class='val'>%u</div></summary></details>\n", features.depthBiasClamp                         );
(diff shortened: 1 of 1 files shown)


shannon-lunarg:
Ah. I should mention this is in reference to LVL issue #2483.

@mikes-lunarg Some of the reordering of the larger structs, e.g., VkPhysicalDeviceLimits, was inspired by the alphanumerical ordering on gpuinfo.com.

jeremyk-lunarg:
The issue states:

> Candidates for alphanumeric soft include:
>
> * List of instance extensions
> * List of device extensions
>
> Things that need not be sorted alphanumerically include:
>
> * Dump of the values of fields of a struct. Printing them in the order in which they appear in the struct definition is both sensible and deterministic.
> * Dump of the bits that are set in a bitfield. Printing them in bit-position order is both sensible and deterministic.

Looks like some of these order changes violate the second part of this statement.

mikes-lunarg:
Totally makes sense for the extension lists to be sorted.

I'm still not convinced on struct fields, the issue even specifically calls them out to be printed in struct order.  I would argue that there are also places where sorting fields is worse, like splitting up min/max pairs or this:
````
features.sparseResidency16Samples
features.sparseResidency2Samples
features.sparseResidency4Samples
features.sparseResidency8Samples
````

shannon-lunarg:
@jeremyk-lunarg @mikes-lunarg There was already conflicting ordering (that I introduced many commits ago) so I think I just made an  unpopular decision to conform to that. I have since reordered to conform to the struct ordering and pushed changes to my branch.

shannon-lunarg:
Pushed to master.

Issue #2608, opened 2018-04-24
core_validation::PostCallRecordFreeMemory() crashes at vkFreeMemory() call time


DominikWitczakAMD:
This is only seen happening when multiple (4 in this case) threads (each using its own device created using the same shared Vulkan instance) do Vulkan stuff:

![image](https://user-images.githubusercontent.com/17761066/39183853-b1ee9ad6-47c1-11e8-9f87-730734d867b7.png)

![image](https://user-images.githubusercontent.com/17761066/39183917-ebe51ab2-47c1-11e8-82fe-3f0f8ddecdee.png)

No validation errors are reported prior to the crash.

This is seen happening with the latest version of the layers, checked out from this repo earlier today.

jzulauf-lunarg:
I can see a race for two threads freeing the same memory on the same device, where one thread would end up with a pointer to a DEVICE_MEM_INFO erased by another thread -- though that would violate the "Host Synchronization" for vkFreeMemory.  I'll patch that a push that for review and test.  I don't see anyway two threads accessing two different devices could do the same thing however.

DominikWitczakAMD:
Exactly, I'd assume that such condition would be prepended with a validation error, telling the same mem block is about to be released for the second time. I'm not getting any such error. In fact, I'm not getting any errors at all, apart from the crash.

Please ping me when the patch lands so I can give it a go. If the crashes are still there afterward, I can provide you with a repro so you can have something to work with.

jzulauf-lunarg:
Created WIP branch with two commits.  One adds defensive/logging code for detect races and not dying, and the other plugs a potential hole in the threading layers logging (at some performance cost for singly threaded applications).

https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/tree/zulauf_free_mem_crash_2608

DominikWitczakAMD:
Excellent. I'll get back to you later today with further feed-back.

DominikWitczakAMD:
OK, I think I know what's cracking here. This issue appears to be directly related to the other one I also reported yesterday (https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/issues/2607).

I've modified my framework to:

1) Printf TID + mem handle at mem alloc time
2) Printf TID + mem handle at mem free time

Here's what I'm getting right before the new checks you added yesterday kick in:

[...]

PR #2620, opened 2018-04-26
PreRecord Free/Destroy calls

### LINKED ISSUE #2607: Threading layer: Race condition in the layer causes incorrect "object simultaneously used in thread .. and thread .." errors to be reported
Consider the following scenario:

1. Thread A calls vkDestroyBuffer()
2. Thread B calls vkCreateBuffer()

It is legitimate for the driver to return the same handle if the calls are executed in a serialized manner, in order: A followed by B.

The validation layer chain does some additional chores after VK impl returns execution flow from vkDestroyBuffer(), before passing control back to the app. However, this additional housekeeping work does not seem to be protected by any sort of a critical section. This can lead to a situation where the threading layer notices the same handle is being used in two threads at the same time, even though that very handle has already had ownership released from the original object and passed over to a new object.

jzulauf-lunarg:
Fixes #2607 and #2608 

To avoid Destroy/Create and Free/Alloc race conditions (when a driver
returns a destroyed(freed) handle between the PreCall and PostCall
phases of a validation intercept routine, PostCallRecord functionality
has been moved to PreCall.  This affects the following calls

	FreeMemory
	DestroyFence
	DestroySemaphore
	DestroyEvent
	DestroyQueryPool
	DestroyBuffer
	DestroyBufferView
	DestroyImage
	DestroyImageView
	DestroyPipeline
	DestroySampler
	DestroyDescriptorSetLayout
	DestroyFramebuffer
	DestroyRenderPass
	FreeDescriptorSets

Did not alter the destroy functions for Instance, Device, or the Debug
or layer extensions.

@cnorthrop -- this change touches a lot of routines making sure the Android view is okay

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -887,7 +887,7 @@ bool PreCallValidateDestroyImage(layer_data *device_data, VkImage image, IMAGE_S
     return skip;
 }
 
-void PostCallRecordDestroyImage(layer_data *device_data, VkImage image, IMAGE_STATE *image_state, VK_OBJECT obj_struct) {
+void PreCallRecordDestroyImage(layer_data *device_data, VkImage image, IMAGE_STATE *image_state, VK_OBJECT obj_struct) {
     core_validation::invalidateCommandBuffers(device_data, image_state->cb_bindings, obj_struct);
     // Clean up memory mapping, bindings and range references for image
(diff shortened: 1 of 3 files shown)


DominikWitczakAMD:
While I can't comment on actual changes in the layer since I don't have any experience working with that code, I can definitely confirm I can't repro the reported issues with these changes in. Good work :-)

PR #2625, opened 2018-05-01
Remove Fill Memory warnings

### LINKED ISSUE #2616: "Cannot read invalid region of memory" when using external memory.
Hello. When I'm using a `VkDeviceMemory`  with `VkImportMemoryFdInfoKHR` I'm getting an error:

> Cannot read invalid region of memory allocation 0x17 for bound Buffer object 0x16, please fill the memory before using.

In case if memory created with a flag `VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT` I have to make a dummy call to `vkMapMemory()` and `vkUnmapMemory()` to make this warning disappear. I didn't try to solve when the flag is `VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT`, but obviously mapping/unmapping will not work here.

There's of course an external memory, created with another device, which is updated separately.

mark-lunarg:
Implementation was incomplete -- when broken it produced incorrect warnings and when working produced non-specified warnings.  If this has a place in the future, it's in the assistant layer.

Fixes #2616.

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -1078,11 +1078,6 @@ void PreCallRecordCmdClearImage(layer_data *dev_data, VkCommandBuffer commandBuf
     auto image_state = GetImageState(dev_data, image);
     if (cb_node && image_state) {
         AddCommandBufferBindingImage(dev_data, cb_node, image_state);
-        std::function<bool()> function = [=]() {
-            SetImageMemoryValid(dev_data, image_state, true);
-            return false;
-        };
-        cb_node->queue_submit_functions.push_back(function);
         for (uint32_t i = 0; i < rangeCount; ++i) {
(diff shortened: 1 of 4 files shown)


tobine:
clang-format error in CI

tests/layer_validation_tests.cpp
```diff
@@ -1359,7 +1359,7 @@ TEST_F(VkLayerTest, ReservedParameter) {
     vkCreateEvent(device(), &event_info, NULL, &event_handle);
     m_errorMonitor->VerifyFound();
 }
-
+/*
```
chrisforbes: Why?

tobine:
Killing all that code looks good. Just needs update to pass clang-format.

mark-lunarg:
@chrisforbes, that's what I get for trying to work in an airport.  Restored the test, which I'd temporarily removed for local testing.  Also fixed c-f issue.

Issue #2448, opened 2018-02-27
Missing indexed draw call validation


DominikWitczakAMD:
Consider the following scenario:

1. Index buffer is assigned a memory range that fits X elements, each index taking N bytes.
2. Application records an indexed draw command which uses the index buffer and wants to draw Y elements using start index Z, where Y+Z > X

Use case 2:

1. Index buffer is assigned a memory range that fits X elements, each index taking N bytes.
2. Application records a bind index buffer command which uses a non-zero start offset.
3. Application records an indexed draw command which uses the index buffer and wants to draw Y elements using start index Z, where Y+Z <= X *but* the memory range the specified index range maps to exceeds available buffer memory

Core validation layer does not inform the draw call may try to use out-of-bound indices, even though there is a corresponding VU defined for the function:

``
indexSize * (firstIndex + indexCount) + offset) must be less than or equal to the size of the currently bound index buffer, with indexSize being based on the type specified by indexType, where the index buffer, indexType, and offset are specified via vkCmdBindIndexBuffer
``

karl-lunarg:
I have a new validation layer test written for this.  Working on the validation layer checks now.

@DominikWitczakAMD, the layers don't currently analyze the Indirect flavors of Draw and DrawIndexed.  Are you concerned at all about these Indirect versions?

DominikWitczakAMD:
@karl-lunarg: I'm not exactly concerned about the indirect variants at this point *but* if we have a gap here, it wound be useful to have it patched up as well :)

DominikWitczakAMD:
Frankly, it's going to be a nightmare. :-( Even if the buffers used to host dispatch/draw call args used host-coherent mappable memory, the layers would still have a hard time with cases where a single cmd buffer does the following:

1. Stash cmd buffer args into a buffer using a dispatch/draw call.
2. Issue an indirect dispatch/draw call that consumes data generated in step 1.

The layer would need to transparently split the cmd buffer to two, inject an intermediary cmd buffer that fires a buffer barrier which lets you HOST_READ and verify the generated data on CPU. This can quickly get out of hand, especially if the submission chain gets complex (eg. the submission sets semaphores or fences)

Ugh.

PR #2633, opened 2018-05-03
Add checks and tests for indexed draw count and offset

karl-lunarg:
Addresses #2448 

I have separate commits for the checks and the tests for ease of review; can squash if that is desirable.


layers/vk_validation_error_database.txt
```diff
@@ -2099,7 +2099,7 @@ VALIDATION_ERROR_1a400396~^~N~^~Unknown~^~vkCmdDrawIndexed~^~VUID-vkCmdDrawIndex
...
-VALIDATION_ERROR_1a40039e~^~N~^~Unknown~^~vkCmdDrawIndexed~^~VUID-vkCmdDrawIndexed-indexSize-00463~^~core~^~The spec valid usage text states '(indexSize * (firstIndex + indexCount) + offset) must be less than or equal to the size of the bound index buffer, with indexSize being based on the type specified by indexType, where the index buffer, indexType, and offset are specified via vkCmdBindIndexBuffer' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkCmdDrawIndexed-indexSize-00463)~^~
+VALIDATION_ERROR_1a40039e~^~Y~^~IndexBufferBadSizeOffset~^~vkCmdDrawIndexed~^~VUID-vkCmdDrawIndexed-indexSize-00463~^~core~^~The spec valid usage text states '(indexSize * (firstIndex + indexCount) + offset) must be less than or equal to the size of the bound index buffer, with indexSize being based on the type specified by indexType, where the index buffer, indexType, and offset are specified via vkCmdBindIndexBuffer' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkCmdDrawIndexed-indexSize-00463)~^~
```
jzulauf-lunarg: Sorry I missed this, but for full bisectability the N -> Y change should be in the layers: ... commit and the Unknown -> _test_name_ change should be in the tests: ... commit.

karl-lunarg: Correct.  I caught that too before the last update.  I'm planning on squashing the commits, so I didn't fix it then, although I should have.   If you think that squashing is NOT a good idea, please say so and I will change it to Y in the layers commit.  Otherwise, I'll probably squash and merge at the end of the day today.

karl-lunarg: As discussed, I'll change it to Y instead of squashing.

PR #2636, opened 2018-05-03
layers: remove duplicate GetEnabledExtensions function

cdwfs:
GetDeviceExtensions() does the same thing & is more widely-used.


tobine:
bizarre CI error that I've never seen before and isn't clearly related to the CL... @cdwfs can you make a minor change in commit message or something and re-push to trigger CI again?

Issue #2602, opened 2018-04-21
Should "actual + rdur" be "desired + rdur" in cube.c ActualTimeLate()?


bradgrantham:
https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/826d9699aa66007e729c0649b283038a676d61d1/demos/cube.c#L572

mark-lunarg:
@mikes-lunarg, this related to your cube glitches maybe?

mikes-lunarg:
@mark-lunarg I don't think it is related.  ActualTimeLate() is part of the VK_GOOGLE_display_timing demo embedded within cube, which my desktop system does not support.

ianelliottus:
@bradgrantham , @mark-lunarg , and @mikes-lunarg , I'll take a look at this (sounds like code that I wrote).

My understanding is that Keith Packard wrote VK_GOOGLE_display_timing support for at least one Mesa Vulkan driver (paid for by Valve).  I think it's for AMD hardware on Linux, but I'm not certain.

ianelliottus:
@bradgrantham you are correct that there is a bug.  As @karl-lunarg described, #2637 is a fix that we should take.

PR #2637, opened 2018-05-03
demos: Fix cube ActualTimeLate function

liam-middlebrook:
The ActualTimeLate function in cube.c is supposed to check if a
VkPastPresentationTimingGOOGLE structure indicates that the
presentation happened within a refresh-cycle duration of the
desired presentation time.

However currently ActualTimeLate will always return false, because
it is calculating the presentation deadline as `actual + rdur`.
This change calculated the presentation deadline as
`desired + rdur`.

Reviewed-by: Andy Ritger <aritger@nvidia.com>


karl-lunarg:
Looks like a fix for #2602.  I'll ping @ianelliottus  for input.  But, @liam-middlebrook, can you verify that you observed that this change actually fixes the problem?  If so, then I think we should merge it with or without Ian's input.

ianelliottus:
@liam-middlebrook and @karl-lunarg, I'll take a look.

I also am interested whether a problem was actually observed (while running the code)?

ianelliottus:
I re-familiarized myself with the code, and yes, this is a good fix of a real bug.  Thanks!

PR #2643, opened 2018-05-04
layers: image transfer granularity check fixes

cdwfs:
- Use appropriate VUIDs for all checks
- Update tests to check for VUIDs instead of message substrings
- Image granularity tests should be also be run on compressed images.
  They're *mostly* redundant with existing compressed-format-only checks,
  but I can't convince myself that they're *always* redundant, and the
  spec doesn't say to skip them.
- Removed comments that the checks should be moved elsewhere; IMO they're
  fine where they are, but I welcome second opinions.
- Compressed-format alignment checks should also apply to _422 single-plane
  formats.


tobine:
@Whenning42 I believe this overlaps with your current work. Please review and update your work to build on these changes.

PR #2644, opened 2018-05-04
Improve check coverage for subpass attachment use conflicts

chrisforbes:


tobine:
clang-format issue in CI

layers/core_validation.cpp
```diff
@@ -8962,16 +8962,84 @@ static bool ValidateAttachmentIndex(layer_data *dev_data, uint32_t attachment, u
...
+            return "preserve";
+        case ATTACHMENT_RESOLVE:
+            return "resolve";
+        default:
+            return "(multiple)";
+    }
+}
+
+static bool AddAttachmentUse(layer_data *dev_data, uint32_t subpass, std::vector<uint8_t> &attachmentUses,
+                             std::vector<VkImageLayout> &attachmentLayouts, uint32_t attachment, uint8_t newUse,
```
tobine: kill cameCase in var naming: attachment_uses, attachment_layouts, new_use

layers/core_validation.cpp
```diff
@@ -8962,16 +8962,84 @@ static bool ValidateAttachmentIndex(layer_data *dev_data, uint32_t attachment, u
...
+                             VkImageLayout layout) {
+    if (attachment >= attachmentUses.size()) return false; /* out of range, but already reported */
+
+    bool skip = false;
+    auto &uses = attachmentUses[attachment];
+    if (uses & newUse) {
+        skip |= log_msg(dev_data->report_data, VK_DEBUG_REPORT_ERROR_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_UNKNOWN_EXT, 0,
+                        DRAWSTATE_INVALID_RENDERPASS, "CreateRenderPass: subpass %u already uses attachment %u as a %s attachment.",
```
tobine: We're certainly not 100% consistent on this but for function name prefix to error message I think the majority case and my preference is complete function name w/ "()" to clearly indicate function vs. struct. So in this case, and those error messages below, "vkCreateRenderPass()".

layers/core_validation.cpp
```diff
@@ -8962,16 +8962,84 @@ static bool ValidateAttachmentIndex(layer_data *dev_data, uint32_t attachment, u
...
+                        subpass, attachment, string_AttachmentType(newUse));
+    } else if (uses & ~ATTACHMENT_INPUT || (uses && (newUse == ATTACHMENT_RESOLVE || newUse == ATTACHMENT_RESOLVE))) {
```
tobine: This check is puzzling me.
Won't the first check, (uses & ~ATTACHMENT_INPUT), be hit on any non-attachment input use?
Also, the second part repeats the same check twice, is one of them supposed to be ATTACHMENT_PRESERVE instead of RESOLVE?

layers/core_validation.cpp
```diff
@@ -8980,72 +9048,78 @@ static bool ValidateRenderpassAttachmentUsage(layer_data *dev_data, const VkRend
...
-
         unsigned sample_count = 0;
+        bool subpass_performs_resolve = false;
 
         for (uint32_t j = 0; j < subpass.colorAttachmentCount; ++j) {
-            uint32_t attachment;
             if (subpass.pResolveAttachments) {
-                attachment = subpass.pResolveAttachments[j].attachment;
-                skip |= ValidateAttachmentIndex(dev_data, attachment, pCreateInfo->attachmentCount, "Resolve");
+                auto const &attachment = subpass.pResolveAttachments[j];
```
tobine: Can this be "attachment_ref" or something slightly different to avoid the "attachment.attachment" uses below?

tests/layer_validation_tests.cpp
```diff
@@ -4200,6 +4200,106 @@ TEST_F(VkLayerTest, RenderPassAttachmentIndexOutOfRange) {
...
+
+    VkRenderPassCreateInfo rpci = {VK_STRUCTURE_TYPE_RENDER_PASS_CREATE_INFO, nullptr, 0, 1, attach, 1, subpasses, 0, nullptr};
+    VkRenderPass rp;
+
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, VALIDATION_ERROR_140006ae);
+    vkCreateRenderPass(m_device->device(), &rpci, nullptr, &rp);
+    m_errorMonitor->VerifyFound();
+}
+
+TEST_F(VkLayerTest, RenderPassAttachmentUsedTwiceOK) {
```
tobine: Should be a VkPositiveLayerTest.

layers/core_validation.cpp
chrisforbes: One of these was indeed supposed to be `ATTACHMENT_PRESERVE`. This is what I get for writing software on no sleep ;)

tests/layer_validation_tests.cpp
chrisforbes: Yes, it should. Minor complaint that this makes it execute much later, far away from the related test cases.

layers/core_validation.cpp
chrisforbes: OK, I will fix all of these.

layers/core_validation.cpp
chrisforbes: As for the first check: `uses` is the existing uses of the attachment. `newUse` is the one we're adding.

chrisforbes: For the second check: `uses` is going to be `ATTACHMENT_INPUT` or zero. This implements the 'one is `ATTACHMENT_INPUT` and neither is `ATTACHMENT_RESOLVE` or `ATTACHMENT_PRESERVE`' requirement

layers/core_validation.cpp
chrisforbes: OK.

layers/core_validation.cpp
chrisforbes: Fixed, here and in the caller.

tests/layer_validation_tests.cpp
```diff
@@ -4241,6 +4241,107 @@ TEST_F(VkLayerTest, RenderPassAttachmentIndexOutOfRange) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(VkLayerTest, RenderPassAttachmentUsedTwiceColor) {
+    ASSERT_NO_FATAL_FAILURE(Init());
```
tobine: No TEST_DESCRIPTION() at the start of any of the new tests.

PR #2656, opened 2018-05-08
Fix an error with 3D image layout transitions

### LINKED ISSUE #2136: Core Validation: Renderpasses whose attachments point to non-overlapping layers of 2D array images generate validation errors
Let X be a renderpass with a total of 6 attachments. X consists of a total of 3 subpasses (S1, S2, S3), each rendering consecutively to a color attachment C1, C2 and C3, and resolving to resolve attachments R1, R2 and R3 (one per subpass).

Trick here is that R1, R2 and R3 all point to the same 2D array image, but - at the same time - it's a unique, nonoverlapping slice. In other words, R1 renders to slice 0, R2 renders to slice 1 and R3 gets slice 2.

Core validation layer doesn't seem to enjoy this. It appears that ``ValidateCmdBufImageLayouts()`` does not recognize the fact that intra-subpass image layout transitions occur at image subresource level, instead of image level. This leads to a weird validation error which claims that:

```
Cannot submit cmd buffer using image (0x5571ac0) [sub-resource: aspectMask 0x1 array layer 0, mip level 0], with layout VK_IMAGE_LAYOUT_UNDEFINED when first use is VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL.
```

FWIW, each resolve attachment uses an ``VkAttachmentDescription::initialLayout`` of ``VK_IMAGE_LAYOUT_UNDEFINED`` and ``VkAttachmentDescription::finalLayout`` of ``VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL``, as well as ``VkSubpassDescription::layout`` value of ``VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL``.

As usual, I have a repro available hidden in my cuff, if an executable would be of help.

daveh-lunarg:
When a 3D image subresource layout is transitioned via a 2D or 2D_ARRAY image view, all depth slices within the subresource mip levels transition simultaneously, even if the subresource specifies a non-complete set of array layers. Fixed layout recording to reflect this special case.

Another instance of (depth slices == array layers, _except when they don't_.)   ¯\\\_(ツ)\_/¯

Resolves #2136.

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -295,13 +295,23 @@ void SetImageLayout(layer_data *device_data, GLOBAL_CB_NODE *cb_node, const IMAG
     image_subresource_range.levelCount = 1;
     SetImageLayout(device_data, cb_node, image_state, image_subresource_range, layout);
 }
+
 // Set image layout for all slices of an image view
 void SetImageViewLayout(layer_data *device_data, GLOBAL_CB_NODE *cb_node, VkImageView imageView, const VkImageLayout &layout) {
     auto view_state = GetImageViewState(device_data, imageView);
     assert(view_state);
 
-    SetImageLayout(device_data, cb_node, GetImageState(device_data, view_state->create_info.image),
-                   view_state->create_info.subresourceRange, layout);
+    IMAGE_STATE *image_state = GetImageState(device_data, view_state->create_info.image);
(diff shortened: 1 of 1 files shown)


tobine:
Do we have a test for this case?

daveh-lunarg:
Pushing this PR now to get the bug fix in before the repo transition.  Will follow up with a test next week.

PR #2671, opened 2018-05-11
icd:Fix Windows build

tobine:
Make sure we generate vk_typemap_helper.h for mock icd.


tobine:
@karl-lunarg did you move appveyor to VS-project based build for speed-up or some other benefit? I vaguely recall update but not exact reason.

karl-lunarg:
Yes, I think that was the only way to parallelize the build, and it is also the more "natural" way to do things in AppVeyor.
I'm going to have to re-do all the CI configs anyway as part of the repo split.  And we're probably getting rid of all the "build*.bat" and "update" scripts anyway.

Since the fix for this was adding a dependency, it seems that it isn't really a "bat file vs AppVeyor" sort of issue, but one of just a missing and needed dependency.  The different results from the bat file driven build and the AppVeyor build may have just been caused by the different ordering of parallel build steps.

Hopefully, this is sufficient to explain the observations and we can continue with the current approach in the AppVeyor builds.

karl-lunarg:
> @karl-lunarg did you move appveyor to VS-project based build for speed-up or some other benefit? I vaguely recall update but not exact reason.

I recall the exact reason now.  AppVeyor had/has a time limit on jobs.  When we were building all 4 flavors of the repo in one job driven by a "before_build" script, we always hit the time limit.

Moving to the "project-based" build causes AppVeyor to start a new job for each item in the Platform-Configuration matrix and we avoid doing ALL the work in a single job.

PR #2673, opened 2018-05-11
Tobin gpdf2 support

tobine:
This is related to #2670 but not exactly a fix.
A couple of updates to prepare core validation for handling GPDF2 extended struct. Not checking any extension structs yet but getting things ready for that with some more changes.
Added mock GPDP2 support for VkPhysicalDeviceDescriptorIndexingPropertiesEXT struct.

### DIFF
--- a/scripts/mock_icd_generator.py
+++ b/scripts/mock_icd_generator.py
@@ -767,6 +767,33 @@ CUSTOM_C_INTERCEPTS = {
 ''',
 'vkGetPhysicalDeviceProperties2KHR': '''
     GetPhysicalDeviceProperties(physicalDevice, &pProperties->properties);
+    const auto *desc_idx_props = lvl_find_in_chain<VkPhysicalDeviceDescriptorIndexingPropertiesEXT>(pProperties->pNext);
+    if (desc_idx_props) {
+        VkPhysicalDeviceDescriptorIndexingPropertiesEXT* write_props = (VkPhysicalDeviceDescriptorIndexingPropertiesEXT*)desc_idx_props;
+        write_props->maxUpdateAfterBindDescriptorsInAllPools = 500000;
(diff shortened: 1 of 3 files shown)


scripts/mock_icd_generator.py
```diff
@@ -767,6 +767,33 @@
...
+    const auto *desc_idx_props = lvl_find_in_chain<VkPhysicalDeviceDescriptorIndexingPropertiesEXT>(pProperties->pNext);
+    if (desc_idx_props) {
+        VkPhysicalDeviceDescriptorIndexingPropertiesEXT* write_props = (VkPhysicalDeviceDescriptorIndexingPropertiesEXT*)desc_idx_props;
+        write_props->maxUpdateAfterBindDescriptorsInAllPools = 500000;
+        write_props->shaderUniformBufferArrayNonUniformIndexingNative = false;
+        write_props->shaderSampledImageArrayNonUniformIndexingNative = false;
+        write_props->shaderStorageBufferArrayNonUniformIndexingNative = false;
+        write_props->shaderStorageImageArrayNonUniformIndexingNative = false;
+        write_props->shaderInputAttachmentArrayNonUniformIndexingNative = false;
+        write_props->robustBufferAccessUpdateAfterBind = true;
```
jzulauf-lunarg: The rest of the values were copied from the spec.  Where did the default for these two come from?

tobine: I read the description and felt like it would allow more testing features to make them supported by default in the mock.
When in doubt the mock philosophy is to say it supports features for greater testing options. When DevSim has extension support it can turn them off if desired.

mark-lunarg:
@tobine, if anyone DOESN'T want this pushed, SN, OFHYP.

